PR #17, opened 2021-12-22
Update to use zap logger

dantingl:
1. use development mode with consoleEncoder and debug log level
2. zap set V(1) as debug log level, -zap-log-level can change log level


zhengxiexie:
test

PR #26, opened 2022-01-12
Add health checker for nsx operator

seanpang-vmware:


pkg/nsx/client.go
```diff
@@ -41,12 +51,21 @@ func GetClient(cf *config.NSXOperatorConfig) *Client {
 	groupClient := domains.NewGroupsClient(restConnector(cluster))
 	securityClient := domains.NewSecurityPoliciesClient(restConnector(cluster))
 	ruleClient := security_policies.NewRulesClient(restConnector(cluster))
+	nsxChecker := func(req *http.Request) error {
+		health := cluster.Health()
+		if GREEN == health {
```
dantingl, line 56: orange should also be fine?

pkg/nsx/client.go
```diff
@@ -12,18 +14,26 @@ import (
...
 
 	"github.com/vmware-tanzu/nsx-operator/pkg/config"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/ratelimiter"
 )
 
+var typeToString = map[ClusterHealth]string{
+	RED:    "RED",
+	ORANGE: "ORANGE",
+	GREEN:  "GREEN",
+}
```
dantingl, line 27: I think we could define constants together for cluster.go and client.go at the same place

pkg/nsx/client.go
```diff
@@ -41,12 +51,21 @@ func GetClient(cf *config.NSXOperatorConfig) *Client {
 	groupClient := domains.NewGroupsClient(restConnector(cluster))
 	securityClient := domains.NewSecurityPoliciesClient(restConnector(cluster))
 	ruleClient := security_policies.NewRulesClient(restConnector(cluster))
+	nsxChecker := func(req *http.Request) error {
```
dantingl, line 54: Could we define a separate function in case that we may add more logic for health check later

seanpang-vmware:
can not update this request due to repo change to public, will generate a new pull request.

PR #35, opened 2022-01-18
Add nsx status checker for liveness probe

seanpang-vmware:
Use localhost:8383/healthz to invoke api get request.

### DIFF
--- a/pkg/nsx/client.go
+++ b/pkg/nsx/client.go
@@ -4,6 +4,8 @@
 package nsx
 
 import (
+	"errors"
+	"net/http"
 	"strings"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


pkg/nsx/cluster.go
```diff
@@ -31,6 +31,12 @@ const (
 	GREEN
 )
 
+var typeToString = map[ClusterHealth]string{
+	RED:    "RED",
+	ORANGE: "ORANGE",
+	GREEN:  "GREEN",
+}
```
dantingl, line 38: Can we merge ClusterHealth and typeToString together? Let we just use string for health return value

pkg/nsx/client.go
```diff
@@ -41,12 +64,14 @@ func GetClient(cf *config.NSXOperatorConfig) *Client {
...
 	ruleClient := security_policies.NewRulesClient(restConnector(cluster))
+	nsxChecker := newNSXHealthChecker(cluster)
 	return &Client{
 		NsxConfig:      cf,
 		RestConnector:  restConnector(cluster),
 		QueryClient:    queryClient,
 		GroupClient:    groupClient,
 		SecurityClient: securityClient,
 		RuleClient:     ruleClient,
+		NSXChecker:     *nsxChecker,
```
TaoZou1, line 75: it seems no need to have a function newNSXHealthChecker(), it's only used in Line 67.
nsxChecker is not needed.
Line 75 -> NSXChecker:     NSXHealthChecker{cluster: c},

TaoZou1 (approved):

PR #38, opened 2022-01-19
Add MatchExpression support for SecurityPolicy

timdengyun:
1.Support MatchExpression Operator: 'NotIn', 'Exists',
'DoesNotExist' for VM/Pod/Namespace label selectors
in SecurityPolicy

2.Given NSX doesn't support MatchExpression Operator 'In',
and only five criteria are allowed currently,
which results into a gigantic group expression body to be passed to NSX.
So, only allow just one Operator 'In' MatchExpressions with at most
of five values in it.

3.Add NSX-T limitation and NSGroup Criteria check

### DIFF
--- a/pkg/nsx/services/firewall.go
+++ b/pkg/nsx/services/firewall.go
@@ -15,10 +15,23 @@ import (
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra/domains"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	MaxCriteriaExpressions      int = 5
+	MaxMixedCriteriaExpressions int = 15
+	MaxCriteria                 int = 5
+	MaxTotalCriteriaExpressions int = 35
+	MaxMatchExpressionInOp      int = 1
+	MaxMatchExpressionIn        int = 1
+	MaxMatchExpressionInValues  int = 5
+	ClusterTagCount             int = 1
+	ProjectTagCount             int = 1
 )
 
 type SecurityPolicyService struct {
@@ -82,7 +95,12 @@ func (service *SecurityPolicyService) buildSecurityPolicy(obj *v1alpha1.Security
 	policyPriority := int64(obj.Spec.Priority)
 	nsxSecurityPolicy.SequenceNumber = &policyPriority
 
-	policyGroup, policyGroupPath, _ := service.buildPolicyGroup(obj)
+	policyGroup, policyGroupPath, err := service.buildPolicyGroup(obj)
+	if err != nil {
+		log.Error(err, "failed to build policy group")
+		return nil, nil, err
+	}
+
 	nsxSecurityPolicy.Scope = []string{policyGroupPath}
 	if policyGroup != nil {
 		nsxGroups = append(nsxGroups, *policyGroup)
@@ -125,8 +143,31 @@ func (service *SecurityPolicyService) buildPolicyGroup(obj *v1alpha1.SecurityPol
 		return nil, "ANY", nil
 	}
 
+	targetGroupCount, targetGroupTotalExprCount := 0, 0
+	criteriaCount, totalExprCount := 0, 0
+	var err error = nil
+	var errorMsg string = ""
 	for i, target := range appliedTo {
-		service.updateTargetExpressions(obj, &target, &policyGroup, i)
+		criteriaCount, totalExprCount, err = service.updateTargetExpressions(obj, &target, &policyGroup, i)
+		if err == nil {
+			targetGroupCount += criteriaCount
+			targetGroupTotalExprCount += totalExprCount
+		} else {
+			return nil, "", err
+		}
+	}
+	log.V(1).Info("build policy target group criteria", "total criteria", targetGroupCount, "total expressions of criteria", targetGroupTotalExprCount)
+
+	if targetGroupCount > MaxCriteria {
+		errorMsg = fmt.Sprintf("total counts of policy target group criteria %d exceed NSX limit of %d", targetGroupCount, MaxCriteria)
+	} else if targetGroupTotalExprCount > MaxTotalCriteriaExpressions {
+		errorMsg = fmt.Sprintf("total expression counts in policy target group criteria %d exceed NSX limit of %d", targetGroupTotalExprCount, MaxTotalCriteriaExpressions)
+	}
+
+	if len(errorMsg) != 0 {
+		err = errors.New(errorMsg)
+		log.Error(err, "validate policy group criteria nsx limit failed")
+		return nil, "", err
 	}
 
 	policyGroupPath := service.buildPolicyGroupPath(obj)
@@ -201,45 +242,90 @@ func (service *SecurityPolicyService) buildBasicTags(obj *v1alpha1.SecurityPolic
 	return tags
 }
 
-func (service *SecurityPolicyService) updateTargetExpressions(obj *v1alpha1.SecurityPolicy, target *v1alpha1.SecurityPolicyTarget, group *model.Group, idx int) {
+func (service *SecurityPolicyService) updateTargetExpressions(obj *v1alpha1.SecurityPolicy, target *v1alpha1.SecurityPolicyTarget, group *model.Group, idx int) (int, int, error) {
+	var err error = nil
+	var tagValueExpression *data.StructValue = nil
+	var memberType string = "SegmentPort"
+	var matchLabels map[string]string
+	var matchExpressions *[]metav1.LabelSelectorRequirement = nil
+	var mergedMatchExpressions *[]metav1.LabelSelectorRequirement = nil
+	var opInValueCount, totalCriteriaCount, totalExprCount = 0, 0, 0
+	var matchLabelsCount, matchExpressionsCount = 0, 0
+
+	if target.PodSelector != nil && target.VMSelector != nil {
+		errorMsg := "PodSelector and VMSelector are not allowed to set in one group"
+		err = errors.New(errorMsg)
+		log.Error(err, "build selector expressions failed")
+		return 0, 0, err
+	}
+
 	service.appendOperatorIfNeeded(&group.Expression, "OR")
-	expressions := data.NewListValue()
-	expressionFrame := data.NewStructValue(
-		"",
-		map[string]data.DataValue{
-			"expressions":   expressions,
-			"resource_type": data.NewStringValue("NestedExpression"),
-		},
-	)
-	group.Expression = append(group.Expression, expressionFrame)
+	expressions := service.buildGroupExpression(&group.Expression)
 
 	clusterExpression := service.buildExpression(
-		"Condition", "SegmentPort",
+		"Condition", memberType,
 		fmt.Sprintf("%s|%s", util.TagScopeNCPCluster, service.getCluster()),
-		"Tag", "EQUALS",
+		"Tag", "EQUALS", "EQUALS",
 	)
 	expressions.Add(clusterExpression)
+
 	if target.PodSelector != nil {
 		service.addOperatorIfNeeded(expressions, "AND")
 		// TODO: consider to use project_uid instead of project
 		nsExpression := service.buildExpression(
-			"Condition", "SegmentPort",
+			"Condition", memberType,
 			fmt.Sprintf("%s|%s", util.TagScopeNCPProject, obj.ObjectMeta.Namespace),
-			"Tag", "EQUALS",
+			"Tag", "EQUALS", "EQUALS",
 		)
 		expressions.Add(nsExpression)
-		service.updatePortExpressions(target.PodSelector.MatchLabels, expressions)
+
+		tagValueExpression = nsExpression
+		matchLabels = target.PodSelector.MatchLabels
+		matchExpressions = &target.PodSelector.MatchExpressions
 	}
 	if target.VMSelector != nil {
 		service.addOperatorIfNeeded(expressions, "AND")
 		nsExpression := service.buildExpression(
 			"Condition", "SegmentPort",
 			fmt.Sprintf("%s|%s", util.TagScopeNCPVIFProject, obj.ObjectMeta.Namespace),
-			"Tag", "EQUALS",
+			"Tag", "EQUALS", "EQUALS",
 		)
 		expressions.Add(nsExpression)
-		service.updatePortExpressions(target.VMSelector.MatchLabels, expressions)
+
+		tagValueExpression = nsExpression
+		matchLabels = target.VMSelector.MatchLabels
+		matchExpressions = &target.VMSelector.MatchExpressions
 	}
+	if target.PodSelector != nil || target.VMSelector != nil {
+		service.updateExpressionsMatchLables(matchLabels, memberType, expressions)
+		matchLabelsCount = len(matchLabels)
+		// PodSelector or VMSelector has two more built-in labels
+		matchLabelsCount += ClusterTagCount + ProjectTagCount
+
+		if matchExpressions != nil {
+			mergedMatchExpressions = service.mergeSelectorMatchExpression(*matchExpressions)
+			matchExpressionsCount = len(*mergedMatchExpressions)
+			opInValueCount, err = service.validateSelectorOpIn(*mergedMatchExpressions)
+
+			if err != nil {
+				log.Error(err, "validate operator 'IN' in label selector matchExpressions failed")
+				return 0, 0, err
+			}
+			err = service.updateExpressionsMatchExpression(*mergedMatchExpressions, matchLabels,
+				&group.Expression, clusterExpression, tagValueExpression, memberType, expressions)
+			if err != nil {
+				log.Error(err, "build label selector matchExpressions failed")
+				return 0, 0, err
+			}
+		}
+
+		totalCriteriaCount, totalExprCount, err = service.validateSelectorExpressions(matchLabelsCount, matchExpressionsCount, opInValueCount, memberType)
+		if err != nil {
+			log.Error(err, "validate label selector matchExpressions failed")
+			return 0, 0, err
+		}
+	}
+	return totalCriteriaCount, totalExprCount, nil
 }
 
 func (service *SecurityPolicyService) appendOperatorIfNeeded(policyExpression *[]*data.StructValue, op string) {
@@ -260,17 +346,46 @@ func (service *SecurityPolicyService) buildConjOperator(op string) *data.StructV
 	return operator
 }
 
-func (service *SecurityPolicyService) buildExpression(resource_type, member_type, value, key, operator string) *data.StructValue {
-	expression := data.NewStructValue(
+func (service *SecurityPolicyService) buildGroupExpression(policyExpression *[]*data.StructValue) *data.ListValue {
+	expressions := data.NewListValue()
+	expressionFrame := data.NewStructValue(
 		"",
 		map[string]data.DataValue{
-			"resource_type": data.NewStringValue(resource_type),
-			"member_type":   data.NewStringValue(member_type),
-			"value":         data.NewStringValue(value),
-			"key":           data.NewStringValue(key),
-			"operator":      data.NewStringValue(operator),
+			"expressions":   expressions,
+			"resource_type": data.NewStringValue("NestedExpression"),
 		},
 	)
+	*policyExpression = append(*policyExpression, expressionFrame)
+	return expressions
+}
+
+func (service *SecurityPolicyService) buildExpression(resource_type, member_type, value, key, operator, scope_op string) *data.StructValue {
+	var expression *data.StructValue
+	if scope_op == "NOTEQUALS" {
+		// when scope_op is "NOTEQUALS", the tag operator and value field will not be used
+		expression = data.NewStructValue(
+			"",
+			map[string]data.DataValue{
+				"resource_type":  data.NewStringValue(resource_type),
+				"member_type":    data.NewStringValue(member_type),
+				"value":          data.NewStringValue(value),
+				"key":            data.NewStringValue(key),
+				"scope_operator": data.NewStringValue(scope_op),
+			},
+		)
+	} else {
+		expression = data.NewStructValue(
+			"",
+			map[string]data.DataValue{
+				"resource_type":  data.NewStringValue(resource_type),
+				"member_type":    data.NewStringValue(member_type),
+				"value":          data.NewStringValue(value),
+				"key":            data.NewStringValue(key),
+				"operator":       data.NewStringValue(operator),
+				"scope_operator": data.NewStringValue(scope_op),
+			},
+		)
+	}
 	return expression
 }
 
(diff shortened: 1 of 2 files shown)


timdengyun:
To add UT later.

pkg/nsx/services/firewall.go
```diff
@@ -201,45 +218,77 @@ func (service *SecurityPolicyService) buildBasicTags(obj *v1alpha1.SecurityPolic
...
 			"Condition", "SegmentPort",
 			fmt.Sprintf("%s|%s", util.TagScopeNCPProject, obj.ObjectMeta.Namespace),
-			"Tag", "EQUALS",
+			"Tag", "EQUALS", "EQUALS",
 		)
 		expressions.Add(nsExpression)
-		service.updatePortExpressions(target.PodSelector.MatchLabels, expressions)
+		service.updateExpressionsMatchLables(target.PodSelector.MatchLabels, "SegmentPort", expressions)
+		if len(target.PodSelector.MatchLabels) > 0 {
+			podSelectorCount = 1
```
dantingl, line 245: Do you mean `podSelectorCount += 1`?

pkg/nsx/services/firewall.go
```diff
@@ -201,45 +218,77 @@ func (service *SecurityPolicyService) buildBasicTags(obj *v1alpha1.SecurityPolic
...
 			"Condition", "SegmentPort",
 			fmt.Sprintf("%s|%s", util.TagScopeNCPVIFProject, obj.ObjectMeta.Namespace),
-			"Tag", "EQUALS",
+			"Tag", "EQUALS", "EQUALS",
 		)
 		expressions.Add(nsExpression)
-		service.updatePortExpressions(target.VMSelector.MatchLabels, expressions)
+		service.updateExpressionsMatchLables(target.VMSelector.MatchLabels, "SegmentPort", expressions)
+		if len(target.VMSelector.MatchLabels) > 0 {
+			vMSelectorCount = 1
```
dantingl, line 267: Same here

pkg/nsx/services/firewall.go
```diff
@@ -201,45 +218,77 @@ func (service *SecurityPolicyService) buildBasicTags(obj *v1alpha1.SecurityPolic
...
+	if podSelectorCount > MaxGroupExpressions {
+		errorMsg := fmt.Sprintf("Total number of PodSelector target expressions:%d exceed NSX limit of %d", podSelectorCount, MaxGroupExpressions)
+		err = errors.New(errorMsg)
+		log.Error(err, "updateTargetExpressions")
+	}
+	if vMSelectorCount > MaxGroupExpressions {
+		errorMsg := fmt.Sprintf("Total number of VMSelector target expressions:%d exceed NSX limit of %d", vMSelectorCount, MaxGroupExpressions)
+		err = errors.New(errorMsg)
+		log.Error(err, "updateTargetExpressions")
+	}
```
dantingl, line 290: I don't think we need check the count separately here, as you will check the total count outside

pkg/nsx/services/firewall.go
```diff
@@ -201,45 +218,77 @@ func (service *SecurityPolicyService) buildBasicTags(obj *v1alpha1.SecurityPolic
...
 		expressions.Add(nsExpression)
-		service.updatePortExpressions(target.VMSelector.MatchLabels, expressions)
+		service.updateExpressionsMatchLables(target.VMSelector.MatchLabels, "SegmentPort", expressions)
+		if len(target.VMSelector.MatchLabels) > 0 {
+			vMSelectorCount = 1
+		}
+
+		if target.VMSelector.MatchExpressions != nil {
+			mergedMatchExpressions := service.mergeSelectorMatchExpression(target.VMSelector.MatchExpressions)
+			vMSelectorCount, err = service.validateSelectorExpressions(target.VMSelector.MatchLabels, *mergedMatchExpressions, "SegmentPort")
```
dantingl, line 272: use `vm` or `VM`

timdengyun: Thanks, updated.

pkg/nsx/services/firewall.go
timdengyun: No, if only vmSelector used, all labelSelectors for vmSelector will produce one group. So VMSelectorCount should be one. for example, 
- vmSelector:
     matchLabels:
          user: internal
          op: role
          user: internal

the above vmSelector will only produce one group, like

image

pkg/nsx/services/firewall.go
timdengyun: No, if only podSelector used, all labelSelectors for podSelector will produce one group. So podSelectorCount should be one.

pkg/nsx/services/firewall.go
timdengyun: The reason why we checked here is for one group limit. In fact, we have two limits:
1. Total number of groups, i.e. total max criteria in one NSGroup: 5
2. Maximum 5 expressions are allowed inside a nested expression
Here，we are checking the second one, inside one group, only allow 5 expressions. Outside, we need to check total 
group number limit.

pkg/nsx/services/firewall.go
timdengyun: destinations:
        - podSelector:
            matchLabels:
              role: backend
            matchExpressions:
              - {key: k2, operator: NotIn, values: [b1,b2]}
              - {key: k2, operator: NotIn, values: [b1,b2,b3]}
              - 

      - podSelector:
            matchLabels:
              role: backend
            matchExpressions:
              - {key: k2, operator: NotIn, values: [b1,b2]}
              - {key: k2, operator: NotIn, values: [b1,b2,b3]}
   
The above podSelector will one produce one group criteria:
image

pkg/nsx/services/firewall.go
```diff
@@ -432,60 +713,108 @@ func (service *SecurityPolicyService) buildRuleSrcGroup(obj *v1alpha1.SecurityPo
...
-		service.updatePortExpressions(peer.PodSelector.MatchLabels, expressions)
+
+		service.updateExpressionsMatchLables(peer.PodSelector.MatchLabels, "SegmentPort", expressions)
+		if len(peer.PodSelector.MatchLabels) > 0 {
+			podSelectorCount = 1
+		}
+
+		if peer.PodSelector.MatchExpressions != nil {
+			mergedMatchExpressions := service.mergeSelectorMatchExpression(peer.PodSelector.MatchExpressions)
+			podSelectorCount, err = service.validateSelectorExpressions(peer.PodSelector.MatchLabels, *mergedMatchExpressions, "SegmentPort")
```
heypnus, line 758: I feel not comfortable with validating while constructing a policy because that will introduce more validating during the process, which makes the code complex. Essentially we only want to know if the final policy/group/rule is legal, we don't care about in which step the exceeding occurs, right? So do you think if we have any way to validate only in the last step to simplify the code?

pkg/nsx/services/firewall.go
```diff
@@ -432,60 +713,108 @@ func (service *SecurityPolicyService) buildRuleSrcGroup(obj *v1alpha1.SecurityPo
...
+			if err == nil {
+				service.updateExpressionsMatchExpression(*mergedMatchExpressions, peer.PodSelector.MatchLabels,
```
heypnus, line 760: If you validate the expressions, then it seems that if err != nil, something should happen? e.g.

if err == nil {
  ...
} **else {
  log.Error(...)
  return ...
}**

pkg/nsx/services/firewall.go
```diff
@@ -281,18 +363,209 @@ func (service *SecurityPolicyService) addOperatorIfNeeded(expressions *data.List
...
+	var mexprInOpcount = 0
+	var mexprInValueCount = 0
+	var err error = nil
+	var errorMsg string = ""
+
+	for _, expr := range matchExpressions {
+		if expr.Operator == metav1.LabelSelectorOpIn {
+			mexprInOpcount++
+			mexprInValueCount += len(expr.Values)
+			if mexprInOpcount > MaxMatchExpressionInOp {
```
dantingl, line 434: Why we only support 1 `In` operator?
I think we could also support below case, right?
    matchExpressions:
      - {key: k1, operator: In, values: [v1]}
      - {key: k2, operator: In, values: [v2]}

pkg/nsx/services/firewall.go
```diff
@@ -281,18 +363,209 @@ func (service *SecurityPolicyService) addOperatorIfNeeded(expressions *data.List
...
+				errorMsg = fmt.Sprintf("Values of Opeator IN expressions:%d exceed limit of %d",
+					mexprInValueCount, MaxMatchExpressionInValues)
+				break
+			}
+		}
+	}
+
+	if len(errorMsg) != 0 {
+		err = errors.New(errorMsg)
+		log.Error(err, "validateSelectorExpressions")
```
dantingl, line 449: validate label selector failed

pkg/nsx/services/firewall.go
```diff
@@ -281,18 +363,209 @@ func (service *SecurityPolicyService) addOperatorIfNeeded(expressions *data.List
...
+		return 0, err
+	}
+
+	// Check total expressions from LabelSelectors in one group
+	totalExprCount := exprInCount + len(matchExpressions)
+	if totalExprCount > MaxGroupExpressions {
+		errorMsg = fmt.Sprintf("Total number of LabelSelectors nested expressions: %d exceed NSX limit of %d in one Group",
+			totalExprCount, MaxGroupExpressions)
+		err = errors.New(errorMsg)
+		log.Error(err, "validateSelectorExpressions")
```
dantingl, line 459: validate label selector failed

pkg/nsx/services/firewall.go
```diff
@@ -409,10 +682,18 @@ func (service *SecurityPolicyService) buildRuleAppliedGroup(obj *v1alpha1.Securi
...
-		service.updateTargetExpressions(obj, &target, &ruleAppliedGroup, i)
-
+		ruleGroupCount += service.updateTargetExpressions(obj, &target, &ruleAppliedGroup, i)
 	}
+	if ruleGroupCount > MaxGroupExpressions {
+		errorMsg := fmt.Sprintf("Total number of rule group expressions:%d exceed NSX limit of %d", ruleGroupCount, MaxGroupExpressions)
+		err = errors.New(errorMsg)
+		log.Error(err, "buildRuleAppliedGroup")
+	}
+	log.V(1).Info(fmt.Sprintf("ruleGroupCount: %v in buildRuleAppliedGroup", ruleGroupCount))
```
dantingl, line 695: Please use key-value pair for logging

pkg/nsx/services/firewall.go
```diff
@@ -432,60 +713,108 @@ func (service *SecurityPolicyService) buildRuleSrcGroup(obj *v1alpha1.SecurityPo
...
+	ruleSrcGroupCount := 0
+	var err error = nil
 	for i, peer := range sources {
-		service.updatePeerExpressions(obj, &peer, &ruleSrcGroup, i)
+		ruleSrcGroupCount += service.updatePeerExpressions(obj, &peer, &ruleSrcGroup, i)
 	}
+	if ruleSrcGroupCount > MaxGroupExpressions {
+		errorMsg := fmt.Sprintf("Total number of rule source group expressions:%d exceed NSX limit of %d", ruleSrcGroupCount, MaxGroupExpressions)
+		err = errors.New(errorMsg)
+		log.Error(err, "buildRuleSrcGroup")
```
dantingl, line 725: `buildRuleSrcGroup`: do not just add function name, use readable message

pkg/nsx/services/firewall.go
```diff
@@ -432,60 +713,108 @@ func (service *SecurityPolicyService) buildRuleSrcGroup(obj *v1alpha1.SecurityPo
...
+	}
+	log.V(1).Info(fmt.Sprintf("ruleSrcGroupCount: %v in buildRuleSrcGroup", ruleSrcGroupCount))
```
dantingl, line 727: same here

pkg/nsx/services/firewall.go
```diff
@@ -201,45 +218,77 @@ func (service *SecurityPolicyService) buildBasicTags(obj *v1alpha1.SecurityPolic
...
 
 	clusterExpression := service.buildExpression(
 		"Condition", "SegmentPort",
 		fmt.Sprintf("%s|%s", util.TagScopeNCPCluster, service.getCluster()),
-		"Tag", "EQUALS",
+		"Tag", "EQUALS", "EQUALS",
 	)
 	expressions.Add(clusterExpression)
+	var err error = nil
+	var podSelectorCount, VMSelectorCount = 0, 0
```
dantingl, line 232: sorry, but I think we should use `vmSelectorCount`

pkg/nsx/services/firewall.go
timdengyun: After thought, I think we had better to do pre-check for operatorIn case and put nsx limit validation in the last step after expression build completed. Since operatorIn is a special case which is tightly connected to logic of the next step matchexpression build.  After separation , it can decouple logic for expression build and expression validation.

pkg/nsx/services/firewall.go
timdengyun: Thanks for advice. I'm changing all error handling process as we discussed

pkg/nsx/services/firewall.go
timdengyun: This limit design for In operator is following NCP's design. but Since nsx limit, Operator In could get more complex. 
There is a lot of case to handle since the code need to make a cross product, and then make a OR operator with each member of product.  so for the example above, it would be (k1 == v1 AND k2 == v2) OR  {....}.
it easier to lead to create a big nested expressions. Ideally, we can use cross product number to check, if this, we can support more cases for operator IN, but not fully

pkg/nsx/services/firewall.go
timdengyun: Thanks for suggestion.

pkg/nsx/services/firewall.go
timdengyun: Thanks

pkg/nsx/services/firewall.go
timdengyun: Thanks for advice.

pkg/nsx/services/firewall.go
timdengyun: Sure.

pkg/nsx/services/firewall.go
timdengyun: Copy that.

pkg/nsx/services/firewall.go
timdengyun: sure, will update it.

pkg/nsx/services/firewall.go
```diff
@@ -281,18 +363,209 @@ func (service *SecurityPolicyService) addOperatorIfNeeded(expressions *data.List
...
+	return &mergedMatchExpressions
+}
+
+// Todo: check limit for mixed namespaceSelector, podSelector, VMSelector
+// Given NSX currently doesn't support 'In' LabelSelector, to keep design simple,
+// only allow just one 'In' LabelSelector in matchExpressions with at most of five values in it.
+func (service *SecurityPolicyService) validateSelectorExpressions(matchLabels map[string]string,
+	matchExpressions []metav1.LabelSelectorRequirement, member_type string) (int, error) {
+	var exprInCount = len(matchLabels)
+	var mexprInOpcount = 0
```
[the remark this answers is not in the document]
timdengyun: Thanks for correction.

pkg/nsx/services/firewall.go
```diff
@@ -281,18 +363,209 @@ func (service *SecurityPolicyService) addOperatorIfNeeded(expressions *data.List
...
 		)
 		expressions.Add(expression)
 	}
 }
 
+// NSX understand the multiple values w.r.t a key in a joined string manner
+// this function iterates over input matchExpressions LabelSelectorRequirement
+// with same operator and Key, and merges them into one and values to a joined string
+// e.g. {'NotIn': [{'k1': ['a1', 'a2', 'a3']}, {'k1': ['a2', 'a3', 'a4']}]}
+//   => {'NotIn': [{'k1': 'a1, a2, a3, a4'}]
```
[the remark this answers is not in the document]
timdengyun: Yes,  it is exactly what will be build at last.

pkg/util/utils.go
```diff
@@ -87,3 +87,21 @@ func Sha1(data string) string {
...
+			allKeys[item] = true
+			list = append(list, item)
+		}
+	}
+	return list
+}
+
+func MergeStr(strSlice1 []string, strSlice2 []string) []string {
+	list := append(strSlice1, strSlice2...)
+	meregedList := RemoveDuplicateStr(list)
```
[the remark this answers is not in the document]
timdengyun: That's a great idea, and set data structure will automatically bring removing duplication members in the set.

pkg/nsx/services/firewall.go
```diff
@@ -281,18 +363,209 @@ func (service *SecurityPolicyService) addOperatorIfNeeded(expressions *data.List
...
+	isMerged := make([]bool, len(matchExpressions))
+	var meregedSelector metav1.LabelSelectorRequirement
+	var i, j = 0, 0
+	var isFound = false
+
+	for i = 0; i < len(matchExpressions); i++ {
+		if isMerged[i] {
+			continue
+		}
+		for j = i + 1; j < len(matchExpressions); j++ {
```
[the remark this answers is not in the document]
timdengyun: Thanks for advice, I tried to the above example and maybe it's not exact we what we want. since we want to build union  Values set for the same operator and key value. the Data structure will have to  iterate each Values list of match expression list. but we can put Values in a set when  same operator and key value, using set union to create a merged set to remove duplicated  value from union set. it's will be easier and time complexity also be reduced since there is no nested-loop any more.

pkg/nsx/services/firewall.go
```diff
@@ -281,18 +381,242 @@ func (service *SecurityPolicyService) addOperatorIfNeeded(expressions *data.List
...
+	var mexprInValueCount = 0
+	var err error = nil
+	var errorMsg string = ""
+
+	for _, expr := range matchExpressions {
+		if expr.Operator == metav1.LabelSelectorOpIn {
+			mexprInOpCount++
+			mexprInValueCount += len(expr.Values)
+			if mexprInOpCount > MaxMatchExpressionInOp {
+				errorMsg = fmt.Sprintf("Count of Opeator IN expressions: %d exceed limit of %d",
```
heypnus, line 448: typo: `Opeator` -> `operator`
And I think here we need to use the lower case. Currently, the output looks like:
`Reason:   Error occurred while processing the Security Policy CR. Please check the config and try again. Error: Count of Operator IN expressions: 2 exceed limit of 1`

pkg/nsx/services/firewall.go
```diff
@@ -281,18 +381,242 @@ func (service *SecurityPolicyService) addOperatorIfNeeded(expressions *data.List
...
+					mexprInOpCount, MaxMatchExpressionIn)
+				break
+
+			} else if mexprInValueCount > MaxMatchExpressionInValues {
+				errorMsg = fmt.Sprintf("Values of Opeator IN expressions: %d exceed limit of %d",
```
heypnus, line 453: ditto

pkg/nsx/services/firewall.go
```diff
@@ -281,18 +381,242 @@ func (service *SecurityPolicyService) addOperatorIfNeeded(expressions *data.List
...
+			service.addOperatorIfNeeded(expressions, "AND")
+			expression := service.buildExpression(
+				"Condition", memberType,
+				fmt.Sprintf("%s|", expr.Key),
+				"Tag", "", "NOTEQUALS",
+			)
+			expressions.Add(expression)
+
+		default:
+			errorMsg = fmt.Sprintf("Invalid Opeator: %s IN matchExpressions", expr.Operator)
```
heypnus, line 610: ditto

pkg/nsx/services/firewall.go
```diff
@@ -125,8 +142,24 @@ func (service *SecurityPolicyService) buildPolicyGroup(obj *v1alpha1.SecurityPol
...
 	for i, target := range appliedTo {
-		service.updateTargetExpressions(obj, &target, &policyGroup, i)
+		count, err = service.updateTargetExpressions(obj, &target, &policyGroup, i)
+		if err == nil {
+			targetGroupCount += count
+		} else {
+			return nil, "", err
+		}
+	}
+	log.V(1).Info("Build policy target group criteria count:", "targetGroupCount", targetGroupCount)
```
heypnus, line 156: build -> build
And don't append `:` in the structure log

pkg/nsx/services/firewall.go
```diff
@@ -82,7 +94,12 @@ func (service *SecurityPolicyService) buildSecurityPolicy(obj *v1alpha1.Security
 	policyPriority := int64(obj.Spec.Priority)
 	nsxSecurityPolicy.SequenceNumber = &policyPriority
 
-	policyGroup, policyGroupPath, _ := service.buildPolicyGroup(obj)
+	policyGroup, policyGroupPath, err := service.buildPolicyGroup(obj)
+	if err != nil {
+		log.Error(err, "failed to build policy groups")
```
heypnus, line 99: groups -> group
Because we only create ONE policy-level group for one security policy CR.

pkg/nsx/services/firewall.go
```diff
@@ -260,17 +327,50 @@ func (service *SecurityPolicyService) buildConjOperator(op string) *data.StructV
...
+			"",
+			map[string]data.DataValue{
+				"expressions":   expressions,
+				"resource_type": data.NewStringValue("NestedExpression"),
+			},
+		)
+		*policyExpression = append(*policyExpression, expressionFrame)
+		return expressions
+	} else {
+		return nil
```
heypnus, line 343: This cannot return nil(although I'm not sure if the case of returning nil will happen) because the returned var will invoke `Add` method. At least it should be an empty list?

pkg/nsx/services/firewall.go
```diff
@@ -281,18 +381,242 @@ func (service *SecurityPolicyService) addOperatorIfNeeded(expressions *data.List
...
+		_, exists := labelSelectorMap[d.Operator]
+		if !exists {
+			labelSelectorMap[d.Operator] = map[string][]string{}
+		}
+		_, exists = labelSelectorMap[d.Operator][d.Key]
+
+		if !exists {
+			labelSelectorMap[d.Operator][d.Key] = append(labelSelectorMap[d.Operator][d.Key], d.Values...)
+		} else {
+			labelSelectorMap[d.Operator][d.Key] = append(labelSelectorMap[d.Operator][d.Key], d.Values...)
```
heypnus, line 418: line 418 and 416 are duplicated.

pkg/nsx/services/firewall.go
heypnus: And you interrupt the check loop... Then in the negative case, it will never report "**2** exceed 1", won't be "**3/4/5/6**... exceed 1", then printing `mexprInOpCount` won't be useful, right?
My suggestion is still that you check the total number in the end, i.e. check the number after the for loop(line 458).

pkg/nsx/services/firewall.go
timdengyun: Thanks for correction.

pkg/nsx/services/firewall.go
timdengyun: Thanks

pkg/nsx/services/firewall.go
```diff
@@ -125,9 +134,17 @@ func (service *SecurityPolicyService) buildPolicyGroup(obj *v1alpha1.SecurityPol
 		return nil, "ANY", nil
 	}
 
+	targetGroupCount := 0
+	var err error = nil
```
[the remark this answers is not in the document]
timdengyun: corrected.

pkg/nsx/services/firewall.go
timdengyun: remove all append : in the structure log.

pkg/nsx/services/firewall.go
timdengyun: Change to lower case words in the log.

pkg/nsx/services/firewall.go
timdengyun: Agreed, Thanks, corrected.

pkg/nsx/services/firewall.go
timdengyun: Thanks for suggestion. I've updated code as suggested.

pkg/nsx/services/firewall.go
```diff
@@ -457,56 +930,144 @@ func (service *SecurityPolicyService) updatePeerExpressions(obj *v1alpha1.Securi
...
+			nsMatchExpressions := &peer.NamespaceSelector.MatchExpressions
+
+			// Validate expressions for POD/VM Selectors
+			mergedMatchExpressions = service.mergeSelectorMatchExpression(*matchExpressions)
+			opInValueCount, err = service.validateSelectorOpIn(*mergedMatchExpressions)
+
+			nsMergedMatchExpressions := service.mergeSelectorMatchExpression(*nsMatchExpressions)
+			nsOpInValCount, opErr := service.validateSelectorOpIn(*nsMergedMatchExpressions)
+
+			if err == nil && opErr == nil {
```
heypnus, line 1011: Please change line 1011 to `if err == nil || opErr == nil` to reduce the indent for line 1012~1033 for the readibility.

pkg/nsx/services/firewall.go
```diff
@@ -457,56 +930,144 @@ func (service *SecurityPolicyService) updatePeerExpressions(obj *v1alpha1.Securi
...
-		)
-		expressions.Add(expression)
+	if peer.PodSelector != nil || peer.VMSelector != nil || peer.NamespaceSelector != nil {
+		if !mixedCriteria {
+			service.updateExpressionsMatchLables(matchLabels, memberType, expressions)
+
+			if matchExpressions != nil {
+				mergedMatchExpressions = service.mergeSelectorMatchExpression(*matchExpressions)
+				opInValueCount, err = service.validateSelectorOpIn(*mergedMatchExpressions)
+				if err == nil {
```
heypnus, line 1049: ditto

pkg/nsx/services/firewall.go
```diff
@@ -581,7 +1159,7 @@ func (service *SecurityPolicyService) CreateOrUpdateSecurityPolicy(obj *v1alpha1
 
 	if len(nsxSecurityPolicy.Scope) == 0 {
 		// warning
-		log.Info("SecurityPolicy %s has empty policy-level appliedTo")
```
heypnus, line 584: Thanks for the fix. I found it but forgot to fix...

pkg/nsx/services/firewall.go
```diff
@@ -599,7 +1177,7 @@ func (service *SecurityPolicyService) CreateOrUpdateSecurityPolicy(obj *v1alpha1
 	} else {
 		err = service.createOrUpdateGroups(groupsClient, nsxGroups)
 		if err != nil {
-			log.Error(err, "failed to create or update groups", "nsxSecurityPolicy.Id", nsxSecurityPolicy.Id)
+			log.Error(err, "failed to create or update group", "nsxSecurityPolicy.Id", nsxSecurityPolicy.Id)
```
heypnus, line 1180: I think it should be `groups`, not `group`. I think this case is NOT similar to my previous comment https://github.com/vmware-tanzu/nsx-operator/pull/38#discussion_r793307770.

pkg/nsx/services/firewall.go
```diff
@@ -201,45 +234,93 @@ func (service *SecurityPolicyService) buildBasicTags(obj *v1alpha1.SecurityPolic
...
+				err = service.updateExpressionsMatchExpression(*mergedMatchExpressions, matchLabels,
+					&group.Expression, clusterExpression, tagValueExpression, memberType, expressions)
+				if err != nil {
+					log.Error(err, "build label selector matchExpressions failed")
+					return 0, err
+				}
+			} else {
+				log.Error(err, "validate operator 'IN' in label selector matchExpressions failed")
+				return 0, err
+			}
```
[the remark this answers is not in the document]
timdengyun: That's really good suggestion. simplified.

pkg/nsx/services/firewall.go
```diff
@@ -201,45 +234,93 @@ func (service *SecurityPolicyService) buildBasicTags(obj *v1alpha1.SecurityPolic
...
+			"Tag", "EQUALS", "EQUALS",
 		)
 		expressions.Add(nsExpression)
-		service.updatePortExpressions(target.VMSelector.MatchLabels, expressions)
+
+		tagValueExpression = nsExpression
+		memberType = "SegmentPort"
+		matchLabels = target.VMSelector.MatchLabels
+		matchExpressions = &target.VMSelector.MatchExpressions
+		matchLabelsCount = len(matchLabels)
```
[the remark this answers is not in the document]
timdengyun: Thanks, remove dup for matchLabelsCount and memberType as well

pkg/nsx/services/firewall.go
```diff
@@ -201,45 +234,93 @@ func (service *SecurityPolicyService) buildBasicTags(obj *v1alpha1.SecurityPolic
 	return tags
 }
 
-func (service *SecurityPolicyService) updateTargetExpressions(obj *v1alpha1.SecurityPolicy, target *v1alpha1.SecurityPolicyTarget, group *model.Group, idx int) {
+func (service *SecurityPolicyService) updateTargetExpressions(obj *v1alpha1.SecurityPolicy, target *v1alpha1.SecurityPolicyTarget, group *model.Group, idx int) (int, error) {
+	var err error = nil
+	var errorMsg string = ""
+	var tagValueExpression *data.StructValue = nil
+	var memberType string
```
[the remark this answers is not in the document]
timdengyun: Thanks, remove dup of memberType and code is  simplified

pkg/nsx/services/firewall.go
```diff
@@ -281,18 +391,320 @@ func (service *SecurityPolicyService) addOperatorIfNeeded(expressions *data.List
...
+
+	if len(errorMsg) != 0 {
+		err = errors.New(errorMsg)
+	}
+	return mexprInValueCount, err
+}
+
+func (service *SecurityPolicyService) validateSelectorExpressions(matchLabelsCount int,
+	matchExpressionsCount int, opInValueCount int, memberType string) (int, error) {
+	var exprCount = matchLabelsCount
```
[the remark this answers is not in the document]
timdengyun: well, I think we just use input arg:matchLabelsCount instead of add an extra exprCount, will be better.

pkg/nsx/services/firewall.go
timdengyun: Definitely right direction to go!! thanks

pkg/nsx/services/firewall.go
timdengyun: updated as well.

pkg/nsx/services/firewall.go
timdengyun: you are welcome:)

pkg/nsx/services/firewall.go
timdengyun: Thanks for explain.

heypnus (approved):

PR #42, opened 2022-01-20
Add thumbprint verify

TaoZou1:
Add thumbprint validation, thumbprint number should be 1 or equal to endpoint number if want to verify it.
When tls dial, it will calculate the thumbprint and compare it with local thumbprint.

### DIFF
--- a/pkg/nsx/cluster.go
+++ b/pkg/nsx/cluster.go
@@ -4,9 +4,13 @@
 package nsx
 
 import (
+	"crypto/sha256"
 	"crypto/tls"
+	"errors"
 	"fmt"
+	"net"
 	"net/http"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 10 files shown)


pkg/nsx/cluster.go
```diff
@@ -89,14 +93,65 @@ func (cluster *Cluster) NewRestConnector() (*policyclient.RestConnector, *Header
...
+		if tpCount > 1 {
+			for index, ep := range cluster.endpoints {
+				if ep.Host() == host {
+					thumbprint = cluster.config.Thumbprint[index]
+					break
+				}
+			}
+		}
+		config := &tls.Config{
+			InsecureSkipVerify: true,
```
dantingl, line 112: Do we need to set `InsecureSkipVerify` as false when thumbprint is there

pkg/nsx/cluster.go
```diff
@@ -89,14 +93,65 @@ func (cluster *Cluster) NewRestConnector() (*policyclient.RestConnector, *Header
...
+		host := strings.Split(addr, ":")[0]
+		var thumbprint string
+		tpCount := len(cluster.config.Thumbprint)
+		if tpCount == 1 {
+			thumbprint = cluster.config.Thumbprint[0]
+		}
+		if tpCount > 1 {
+			for index, ep := range cluster.endpoints {
+				if ep.Host() == host {
+					thumbprint = cluster.config.Thumbprint[index]
```
dantingl, line 106: thumbprint and endpoints are matched in order, right?

TaoZou1: yes

pkg/nsx/cluster.go
TaoZou1: no, even we verify the thumbprint, the InsecureSkipVerify is true, that verify here is provided by http lib, not include function we added.

dantingl (approved):

PR #43, opened 2022-01-21
Add UT for garbage collector

TaoZou1:
Use gomonkey to mock struct method
Use gomock to mock interface method
Update gc function so it could be tested


pkg/mock/nsxop.ini
```diff
@@ -0,0 +1,45 @@
+[DEFAULT]
+
+debug = True
```
dantingl, line 3: Please remove options which are not used in current operator code

TaoZou1: done

dantingl (approved):

PR #45, opened 2022-01-24
Add priority range in SecurityPolicy definition

dantingl:


heypnus (commented):
I think you should add
```
// +kubebuilder:validation:Minimum=0
// +kubebuilder:validation:Maximum=1000
```
into `securitypolicy_types.go` instead of directly modifying the yaml file because this file is generated by `make manifests`. refer to https://book.kubebuilder.io/reference/markers/crd-validation.html

PR #46, opened 2022-01-25
Fix nsx requests not using token while token exists

TaoZou1:


pkg/config/config.go
```diff
@@ -150,10 +151,13 @@ func (operatorConfig *NSXOperatorConfig) validate() error {
 	return operatorConfig.validateVersion()
 }
 
+func (operatorConfig *NSXOperatorConfig) GetTokenProvider() auth.TokenProvider {
+	return tokenProvider
+}
```
heypnus, line 156: This function looks a little strange because it returned a variable that is not initialized/changed by itself.
And initializing the provider (I mean `tokenProvider, _ = jwt.NewTokenProvider(...)`) in a validation function also looks unordinary...
Do you have an assumption that `validateVersion` must be executed before `GetTokenProvider` is invoked?
Could you please re-think the overall logic?

TaoZou1: the first step is to validate the configuration, the validateVersion  is the first time to connect to nsx-t. if version is greater than 3.2.0, then creating the cluster. 
the validate() of NSXOperatorConfig is invoked by NewNSXOperatorConfigFromFile().
once the NewNSXOperatorConfigFromFile() finished, the tokenProvider should be ready.

pkg/nsx/auth/jwt/tesclient.go
```diff
@@ -75,6 +75,6 @@ func (client *TESClient) ExchangeJWT(samlToken string, useOldAudience bool) (str
 		log.Error(tesErr, "failed to exchange JWT")
 		return "", errors.New(msg)
 	}
-
+	log.V(1).Info("retrive JWT")
```
dantingl, line 78: retrieve？

TaoZou1: exchanged

heypnus (approved):

PR #50, opened 2022-02-08
Fix SecurityPolicyPort issues

heypnus:
This patch will fix 2 issues:

1. start of portRange was mistakenly set into the sourcePorts property
   in NSX-T.
2. json.Marshal(rule) cannot detect the change of SecurityPolicyPort.

### DIFF
--- a/pkg/nsx/services/firewall.go
+++ b/pkg/nsx/services/firewall.go
@@ -13,6 +13,7 @@ import (
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx"
 	"github.com/vmware-tanzu/nsx-operator/pkg/util"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/data/serializers/cleanjson"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra/domains"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


pkg/nsx/services/firewall.go
```diff
@@ -535,22 +536,28 @@ func (service *SecurityPolicyService) buildRuleDstGroup(obj *v1alpha1.SecurityPo
...
 		serviceEntry := data.NewStructValue(
 			"",
 			map[string]data.DataValue{
 				"source_ports":      sourcePorts,
 				"destination_ports": destinationPorts,
 				"l4_protocol":       data.NewStringValue(string(port.Protocol)),
 				"resource_type":     data.NewStringValue("L4PortSetServiceEntry"),
+				// adding the following default values to make it easy when compare the existing object from store and the new built object
+				"marked_for_delete": data.NewBooleanValue(false),
+				"overridden":        data.NewBooleanValue(false),
```
dantingl, line 560: Could we add a simplify struct for service entry like other resources?

heypnus: Yes, theoretically I think we can. However, if we simplify this property, we need to dive into `rule.ServiceEntries[i].source_ports/destination_ports/l4_protocol/resource_type`, i.e. about 3 layers of indexes. I thought doing it won't make other things can reuse it so I didn't do it. As I commented, this is a tricky method to "make it easy", do you think it's acceptable?

dantingl: ok. Later if we have more similar cases, maybe we find a better and consistent way.

dantingl (approved):

pkg/nsx/services/firewall.go
heypnus: Thanks for understanding. If we want to solve it in a common way in the future, maybe need to consider make the change in `store.go`, i.e. reject some unused options when saving the object into the store.

PR #51, opened 2022-02-09
Add MatchExpression UnitTests

timdengyun:
This patch is to add MatchExpression UnitTests, fix some typo
and a minor code refactor in updateMixedExpressionsMatchExpression

### DIFF
--- a/pkg/nsx/services/firewall_test.go
+++ b/pkg/nsx/services/firewall_test.go
@@ -47,7 +47,7 @@ var (
 	tagValueNS                   = "ns1"
 	tagValuePolicyCRName         = "spA"
 	tagValuePolicyCRUID          = "uidA"
-	tagValuePodSelectorHash      = "6fb11ac7a06285f0ed64a85310de99cf5bf423d0"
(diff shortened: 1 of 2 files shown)


pkg/nsx/services/firewall_test.go
```diff
@@ -63,7 +63,35 @@ var (
...
+					PodSelector: &metav1.LabelSelector{
+						MatchLabels: map[string]string{"pod_selector_1": "pod_value_1"},
+						MatchExpressions: []metav1.LabelSelectorRequirement{
+							{
+								Key:      "k1",
+								Operator: metav1.LabelSelectorOpIn,
+								Values: []string{
+									"a1",
+									"a2",
+								},
```
heypnus, line 75: The indentation looks a little weird. It looks that you tried using multiple spaces to align the values, but not all of them actually align. And there are several such situations below.

timdengyun: Thanks for bringing this attention. I've re-formatted the code to align each value filed.

heypnus (approved):

PR #54, opened 2022-02-09
Auto decrement page size when 60576 errcode appears

zhengxiexie:
When it reports 60576 error code, meaning the content of one page is too large to transport, of queryGroup, querySecurityPolicy and queryGroup, auto decrement by 100 records and retry, and it goes on.

### DIFF
--- a/pkg/nsx/services/store.go
+++ b/pkg/nsx/services/store.go
@@ -6,8 +6,11 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/vmware/vsphere-automation-sdk-go/lib/vapi/std/errors"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 
+	util2 "github.com/vmware-tanzu/nsx-operator/pkg/nsx/util"
 	"github.com/vmware-tanzu/nsx-operator/pkg/util"
 )
 
@@ -19,7 +22,7 @@ const (
 )
 
 var (
-	pageSize int64 = 10 // TODO consider a appropriate page size
+	PageSize int64 = 1000
 )
 
 func securityPolicyCRUIDScopeIndexFunc(obj interface{}) ([]string, error) {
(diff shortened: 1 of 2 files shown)


pkg/nsx/services/store.go
```diff
@@ -78,8 +81,15 @@ func queryGroup(service *SecurityPolicyService, wg *sync.WaitGroup, fatalErrors
 	defer wg.Done()
 	queryParam := fmt.Sprintf("%s:%s", resourceType, resourceTypeGroup) + " AND " + queryTagCondition(service)
 	var cursor *string = nil
+	pageSize := PageSize
 	for {
-		response, err := service.NSXClient.QueryClient.List(queryParam, cursor, nil, nil, nil, nil)
+		response, err := service.NSXClient.QueryClient.List(queryParam, cursor, nil, &pageSize, nil, nil)
+		err = filterError(err)
```
TaoZou1, line 87: I'm not sure if all the query has the same page size limit. if yes, when you find the right page size, then each time you could use it instead of try again.
so PageSize should be updated to the right one.
it seems that you could find the right page size at the first time, then other query no need to try again.

pkg/nsx/services/store.go
```diff
@@ -166,3 +190,33 @@ func queryRule(service *SecurityPolicyService, wg *sync.WaitGroup, fatalErrors c
...
+
+func decrementPageSize(pageSize *int64) {
+	*pageSize -= 100
+	if int(*pageSize) <= 0 {
+		*pageSize = 1
+	}
+}
+
+func filterError(err error) error {
+	var typeConverter = bindings.NewTypeConverter()
```
TaoZou1, line 202: this function looks like to transfer err to other format error, not filter

pkg/nsx/util/errors.go
```diff
@@ -550,3 +550,11 @@ func CreateConnectionError(host string) *ConnectionError {
 	nsxErr.msg = m
 	return nsxErr
 }
+
+type PageMaxError struct {
```
TaoZou1, line 554: all the errors here before are only used in transport.go for retry. 
you may add comments to mark it 's for client usage.

pkg/nsx/services/store.go
```diff
@@ -166,3 +190,33 @@ func queryRule(service *SecurityPolicyService, wg *sync.WaitGroup, fatalErrors c
 		}
 	}
 }
+
+func decrementPageSize(pageSize *int64) {
+	*pageSize -= 100
+	if int(*pageSize) <= 0 {
+		*pageSize = 1
```
TaoZou1, line 197: it seems 1 is too small for the page size.

pkg/nsx/services/store.go
zhengxiexie: @TaoZou1 The max default size is 1000, 60576 error code may appear dynamically, so only have to try again, it is like python code.

pkg/nsx/util/errors.go
zhengxiexie: ok

pkg/nsx/services/store.go
zhengxiexie: I think it seldom hit here, unless error appears when pagesize=100, it can't happen.

TaoZou1 (approved):

PR #55, opened 2022-02-10
Ensure all criterion mixed, part 1

heypnus:


pkg/nsx/services/firewall.go
```diff
@@ -263,7 +264,7 @@ func (service *SecurityPolicyService) updateTargetExpressions(obj *v1alpha1.Secu
 	expressions := service.buildGroupExpression(&group.Expression)
 
 	clusterExpression := service.buildExpression(
-		"Condition", memberType,
```
dantingl, line 266: Could you add comments to say why we use segment here, but use segment port for NamespaceSelector

heypnus: updated

PR #59, opened 2022-02-18
Support all namespace selector in rule group

timdengyun:
During building peer expressions for rule destination or source group,
we need to support one special case:
- namespaceSelector: {}
it means all namespace in a cluster are selected.
This patch is to support this special case.

### DIFF
--- a/pkg/nsx/services/firewall.go
+++ b/pkg/nsx/services/firewall.go
@@ -976,6 +976,7 @@ func (service *SecurityPolicyService) updatePeerExpressions(obj *v1alpha1.Securi
 	if peer.PodSelector == nil && peer.VMSelector == nil && peer.NamespaceSelector != nil && peer.NamespaceSelector.Size() > 0 {
 		clusterMemberType = "SegmentPort"
 	}
+
 	clusterExpression := service.buildExpression(
 		"Condition", clusterMemberType,
 		fmt.Sprintf("%s|%s", util.TagScopeNCPCluster, service.getCluster()),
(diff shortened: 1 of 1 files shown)


pkg/nsx/services/firewall.go
```diff
@@ -1033,13 +1034,31 @@ func (service *SecurityPolicyService) updatePeerExpressions(obj *v1alpha1.Securi
...
 		if !mixedNsSelector {
+			if peer.NamespaceSelector.Size() == 0 {
+				// Since expressions list in model.Group must follow criteria from NSXT:
+				// 1. A non-empty expression list, must be of odd size
+				// 2. An expression list size is equal to or greater than 3
+				// 3. In a list, with indices starting from 0, all non-conjunction expressions must be at even indices
+				// Hence, add one more SegmentPort member condition to meet the criteria aforementioned
+				service.addOperatorIfNeeded(expressions, "AND")
+				clusterSegPortExpression := service.buildExpression(
+					"Condition", "SegmentPort",
```
timdengyun, line 1045: I've tested if putting a duplicated Segment Cluster tag condition here, NSX will  report a "Duplicate expressions specified" error. So, SegmentPort member type is right way.

Also, I've tried to remove NestedExpression from  expressions of group, but it has to be an expression list to meet ceratain criteria, so this workaround fix is only way if we continue to user this group structure.

pkg/nsx/services/firewall.go
```diff
@@ -1033,13 +1034,31 @@ func (service *SecurityPolicyService) updatePeerExpressions(obj *v1alpha1.Securi
 	}
 	if peer.NamespaceSelector != nil {
 		if !mixedNsSelector {
+			if peer.NamespaceSelector.Size() == 0 {
```
dantingl, line 1037: Do we need to check whether vmSelector and podSelector are nil here?

timdengyun: It's not necessary to check here, since here logic is a branch for a non mixed criteria, which means both vmSelector and podSelector are nil. If either one is non nil, it will be a mixed criteria and go to code line 1062.

pkg/nsx/services/firewall.go
```diff
@@ -1033,13 +1034,31 @@ func (service *SecurityPolicyService) updatePeerExpressions(obj *v1alpha1.Securi
...
+
 			tagValueExpression = nil
 			memberType = "Segment"
 			matchLabels = peer.NamespaceSelector.MatchLabels
 			matchExpressions = &peer.NamespaceSelector.MatchExpressions
 			matchLabelsCount = len(matchLabels)
 			// NamespaceSelector has one more built-in labels
 			matchLabelsCount += ClusterTagCount
+			if peer.NamespaceSelector.Size() == 0 {
+				matchLabelsCount += 1
```
dantingl, line 1060: Do you think we should put `matchLabelsCount += 1` in above block(before line 1049)

timdengyun: since code line 1056 will reassign matchLabelsCount value, So simple move matchLabelsCount += 1 in before line 1049 will not work, but we can create else branch to handle difference conditions here.

heypnus (dismissed):

pkg/nsx/services/firewall.go
```diff
@@ -1033,13 +1034,30 @@ func (service *SecurityPolicyService) updatePeerExpressions(obj *v1alpha1.Securi
...
+				matchLabelsCount = ClusterTagCount + 1
+				matchExpressionsCount = 0
+			} else {
+				tagValueExpression = nil
+				memberType = "Segment"
+				matchLabels = peer.NamespaceSelector.MatchLabels
+				matchExpressions = &peer.NamespaceSelector.MatchExpressions
+				matchLabelsCount = len(matchLabels)
+				// NamespaceSelector has one more built-in labels
+				matchLabelsCount += ClusterTagCount
```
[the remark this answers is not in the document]
timdengyun: Thanks a lot for advice.

heypnus (approved):

PR #60, opened 2022-02-18
Update error package

TaoZou1:
Add error ut
Add more check for get version

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -236,7 +236,7 @@ func (nsxVersion *NsxVersion) validate() error {
 		return err
 	}
 	if !nsxVersion.featureSupported() {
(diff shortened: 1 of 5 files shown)


pkg/config/config.go
```diff
@@ -305,6 +305,11 @@ func (nsxVersion *NsxVersion) getVersion(host string, userName string, password
 		log.Error(err, "failed to get nsx version")
 		return err
 	}
+	if resp.StatusCode > 300 {
```
dantingl, line 308: Do you think we should check whether the status code is exact 200 or 201, to detect if it's a failure?

TaoZou1: ok

dantingl (approved):

PR #63, opened 2022-02-24
Refactor getVersion, move from config to cluster

TaoZou1:
Refactor updateAuthInfo
Refactor handling http response
Update some UT

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -64,6 +64,10 @@ func main() {
 		Scheme: mgr.GetScheme(),
 	}
 	nsxClient := nsx.GetClient(cf)
+	if nsxClient == nil {
+		setupLog.Error(err, "unable to get nsx client")
+		os.Exit(1)
+	}
 	if service, err := services.InitializeSecurityPolicy(nsxClient, cf); err != nil {
 		setupLog.Error(err, "unable to init securitypolicy service", "controller", "SecurityPolicy")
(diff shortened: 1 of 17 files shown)


pkg/config/config.go
```diff
@@ -148,47 +138,40 @@ func (operatorConfig *NSXOperatorConfig) validate() error {
 		return err
 	}
 	// TODO, verify if user&pwd, cert, jwt has any of them provided
-
-	return operatorConfig.validateVersion()
+	return nil
 }
 
+// it's not thread safe
 func (operatorConfig *NSXOperatorConfig) GetTokenProvider() auth.TokenProvider {
```
dantingl, line 145: Why do we still need to create/get token provider in config.go?

TaoZou1: it's used to NewConfig. since config only cares about TokenProvider, so create token provider in OperatorConfig.

cmd/main.go
```diff
@@ -64,6 +64,9 @@ func main() {
 		Scheme: mgr.GetScheme(),
 	}
 	nsxClient := nsx.GetClient(cf)
+	if nsxClient == nil {
+		os.Exit(1)
```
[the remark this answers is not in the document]
TaoZou1: done

dantingl (approved):

PR #67, opened 2022-03-01
Add duplicate expression validation for Op In

timdengyun:
Since operator 'In' not support in NSX-T, so it will be translated into
    an equivalent 'EQUALS' expression like:
      "member_type" : "Segment",
      "key" : "Tag",
      "operator" : "EQUALS",
      "scope_operator" : "EQUALS",
      "value" : "app|qe",
    This 'EQUALS' expression is the same format with matchLabels' one.
    So, if matchLabels have defined the same key:value with matchExpressions
    operator 'In', there will be duplicate expressions in the generated group
    criteria. This error only happens for single POD, VM or Namespace Selector
    not for mixed selector, since mixed selector have different member_type
    in the expressions.

This patch is to add duplicate expression validation of operator 'In'
for matchLabels and matchExpression. Once duplicate expression checked,
 it updates Security Policy CR status with error message.

### DIFF
--- a/pkg/nsx/services/firewall.go
+++ b/pkg/nsx/services/firewall.go
@@ -308,10 +308,10 @@ func (service *SecurityPolicyService) updateTargetExpressions(obj *v1alpha1.Secu
 		if matchExpressions != nil {
 			mergedMatchExpressions = service.mergeSelectorMatchExpression(*matchExpressions)
 			matchExpressionsCount = len(*mergedMatchExpressions)
-			opInValueCount, err = service.validateSelectorOpIn(*mergedMatchExpressions)
+			opInValueCount, err = service.validateSelectorOpIn(*mergedMatchExpressions, matchLabels)
 
 			if err != nil {
-				log.Error(err, "validate operator 'IN' in label selector matchExpressions failed")
+				log.Error(err, "validate operator 'In' in label selector matchExpressions failed")
 				return 0, 0, err
 			}
 			err = service.updateExpressionsMatchExpression(*mergedMatchExpressions, matchLabels,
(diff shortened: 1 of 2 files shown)


pkg/nsx/services/firewall.go
```diff
@@ -452,24 +452,41 @@ func (service *SecurityPolicyService) mergeSelectorMatchExpression(matchExpressi
...
-		errorMsg = fmt.Sprintf("count of operator 'IN' expressions %d exceed limit of %d",
+		errorMsg = fmt.Sprintf("count of operator 'In' expressions %d exceed limit of %d",
 			mexprInOpCount, MaxMatchExpressionIn)
 	} else if mexprInValueCount > MaxMatchExpressionInValues {
-		errorMsg = fmt.Sprintf("count of values list for operator 'IN' expressions %d exceed limit of %d",
+		errorMsg = fmt.Sprintf("count of values list for operator 'In' expressions %d exceed limit of %d",
 			mexprInValueCount, MaxMatchExpressionInValues)
+	} else if exists {
+		// matchLabels can only be duplicated with matchExpressions Operator 'In' expression
+		// Since only Operator 'In is equivalent to key-value condition
```
dantingl, line 482: `Operator` -> `operator`
`'In` -> `'In'`

pkg/nsx/services/firewall.go
```diff
@@ -452,24 +452,41 @@ func (service *SecurityPolicyService) mergeSelectorMatchExpression(matchExpressi
...
+		for _, value := range matchExpressions[opInIndex].Values {
+			if matchLabels[matchExpressions[opInIndex].Key] == value {
+				errorMsg = fmt.Sprintf("duplicate expression - %s:%s specified in both matchLabels and matchExpressions Operator 'In'",
```
dantingl, line 485: `Operator` -> `operator`

pkg/nsx/services/firewall.go
```diff
@@ -1064,20 +1081,20 @@ func (service *SecurityPolicyService) updatePeerExpressions(obj *v1alpha1.Securi
...
-			opInValueCount, err = service.validateSelectorOpIn(*mergedMatchExpressions)
+			opInValueCount, err = service.validateSelectorOpIn(*mergedMatchExpressions, matchLabels)
 
 			nsMergedMatchExpressions := service.mergeSelectorMatchExpression(*nsMatchExpressions)
-			nsOpInValCount, opErr := service.validateSelectorOpIn(*nsMergedMatchExpressions)
+			nsOpInValCount, opErr := service.validateSelectorOpIn(*nsMergedMatchExpressions, nsMatchLabels)
 
 			if err != nil || opErr != nil {
-				log.Error(err, "validate Operator 'IN' in label selector matchExpressions failed")
+				log.Error(err, "validate Operator 'In' in label selector matchExpressions failed")
```
dantingl, line 1090: Operator -> operator

timdengyun: Thanks, Fixed

pkg/nsx/services/firewall.go
timdengyun: Thanks, corrected.

dantingl (approved):

PR #70, opened 2022-03-16
Pass insecure parameter to NewTokenProvider

TaoZou1:


pkg/config/config.go
```diff
@@ -16,6 +17,7 @@ import (
 //TODO replace to yaml
 const (
 	nsxOperatorDefaultConf = "/etc/nsx-operator/nsxop.ini"
+	vcHostCACertPath       = "/etc/vmware/wcp/tls/vmca.pem"
```
dantingl, line 20: I guess this is the default path, is there a config option for the path?

TaoZou1: no, ncp uses it directly.

pkg/config/config.go
```diff
@@ -61,6 +63,7 @@ type VCConfig struct {
 	VCEndPoint string `ini:"vc_endpoint"`
 	SsoDomain  string `ini:"sso_domain"`
 	HttpsPort  int    `ini:"https_port"`
+	VCInsecure bool   `ini:"insecure"`
```
dantingl, line 66: There is no VC `insecure` in configmap

pkg/config/config.go
TaoZou1: no

pkg/config/config.go
TaoZou1: use insecure in NsxConfig

pkg/config/config.go
```diff
@@ -150,8 +152,16 @@ func (operatorConfig *NSXOperatorConfig) GetTokenProvider() auth.TokenProvider {
 }
 
 func (operatorConfig *NSXOperatorConfig) createTokenProvider() error {
+	log.Info("try to load VC host CA")
+	vcCaCert, err := ioutil.ReadFile(vcHostCACertPath)
+	// If operatorConfig.VCInsecure is false, tls will the CA to verify the server
+	// certificate. If loading CA failed, tls will use the CA on local host
+	if err != nil {
+		log.Info("fail to load CA cert from file", "error", err)
+	}
```
dantingl, line 161: Do we load VC cert only when insecure = false?

pkg/config/config.go
```diff
@@ -150,8 +152,16 @@ func (operatorConfig *NSXOperatorConfig) GetTokenProvider() auth.TokenProvider {
 }
 
 func (operatorConfig *NSXOperatorConfig) createTokenProvider() error {
+	log.Info("try to load VC host CA")
```
dantingl, line 155: debug log level should be enough

TaoZou1: done

pkg/config/config.go
TaoZou1: yes. 
updated.

dantingl (approved):

PR #73, opened 2022-03-29
Set log level according to config

zhengxiexie:
If debug=True in the config, then the log level is DEBUG, else INFO.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -8,10 +8,11 @@ import (
 	"flag"
 	"io/ioutil"
 
-	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/auth"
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -83,37 +89,34 @@ func NewNSXOperatorConfigFromFile() (*NSXOperatorConfig, error) {
 	cfg := ini.Empty()
 	err := ini.ReflectFrom(cfg, nsxOperatorConfig)
 	if err != nil {
-		log.Error(err, "failed to load default NSX Operator configuration")
```
dantingl, line 86: Is it because error can be printed in upper layer, so you remove them all?

zhengxiexie: @dantingl  Exists mutual dependency between config and log, so let's firstly get log level from config and then use log afterwords.

dantingl (approved):

PR #74, opened 2022-04-18
add health metrics

lxiaopei:


cmd/main.go
```diff
@@ -103,3 +112,43 @@ func main() {
...
 	}
 }
+
+// Function for fetching nsx health status and feeding it to the prometheus metric.
+func getHealthStatus(lastSeenStatus *string) error {
+    status := "healthy"
+	if err := nsxClient.NSXChecker.CheckNSXHealth(); err != nil {
+		status := "unhealthy"
+	}
+	// Return without updating the version if it stayed the same since last time.
```
dantingl, line 122: Why not return the same status again?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,50 @@
...
+	"net/http"
+	"sync"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var (
+	log = logf.Log.WithName("metrics").WithName("metrics")
```
dantingl, line 13: One `WithName("metrics")` is enough?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,53 @@
...
+// objects with Prometheus and installs the Prometheus http handler at the
+// default context.
+func RegisterMetricsAndInstallHandler(m *http.ServeMux, metricsAddr string) {
+	registry := prometheus.NewRegistry()
+	register(registry)
+	m.Handle("/metrics", promhttp.HandlerFor(registry, promhttp.HandlerOpts{ErrorHandling: promhttp.ContinueOnError}))
+	address := net.JoinHostPort("", metricsAddr)
+	http.ListenAndServe(address, m)
+	log.Info("Registered /metrics with prometheus")
+}
```
heypnus, line 53: kubebuilder has exposed the metrics port 8080 by default so I think we can reuse it. You can refer to https://book.kubebuilder.io/reference/metrics.html#publishing-additional-metrics

lxiaopei: thanks, Qian. So we just use the metrics server started by controller.

pkg/metrics/metrics.go
lxiaopei: yes, removed the last one

cmd/main.go
lxiaopei: if err happend, set the status to "unhealthy"

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,41 @@
...
+			Namespace: MetricNamespace,
+			Subsystem: MetricSubsystem,
+			Name:      HealthKey,
+			Help:      "Last health status for NSX-Operator. 1 for 'status' label with current status.",
+		},
+		[]string{"status"},
+	)
+)
+
+var registerMetrics sync.Once
```
heypnus, line 34: For `sync.Once`, I'm afraid we cannot invoke Register multiple times for different metrics. e.g.
```
metrics.Register(metrics.NSXOperatorHealthStats)
metrics.Register(metrics.NSXOperatorControllerSyncTotal)
metrics.Register(metrics.NSXOperatorControllerUpdateTotal)
...
```
will fail, right?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"sync"
+
+	"github.com/prometheus/client_golang/prometheus"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
+	log = logf.Log.WithName("metrics")
```
heypnus, line 12: It seems that `log` is unused in `metrics.go`, can we remove it?

lxiaopei: removed. thanks.

cmd/main.go
```diff
@@ -33,8 +39,9 @@ func init() {
 }
 
 func main() {
-	var probeAddr string
+	var probeAddr, metricsAddr string
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8384", "The address the probe endpoint binds to.")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8089", "The address the metrics endpoint binds to.")
```
heypnus, line 44: Please change the port to 8093 as we discussed.

lxiaopei: changed.

pkg/metrics/metrics.go
lxiaopei: What error do you encounter with the above? I did not test it with multiple metrics though.
I checked the doc and think we can set it 
func InitializePrometheusMetrics() {
	log.Info("Initializing prometheus metrics")
	Register(NSXOperatorHealthStats, NSXOperatorControllerSyncTotal, NSXOperatorControllerUpdateTotal)
}

heypnus: got it, thanks.

heypnus (approved):

PR #83, opened 2022-07-07
Support named port feature

zhengxiexie:
Support named port feature.
    Consider the below sceneries:
     * When a new added pod whose port name exists in security policy.
     * When a deleted pod whose port name exists in security policy.
     * When a pod's label is changed.
     * When a namespace's label is changed.


docs/security-policy.md
```diff
@@ -13,17 +13,16 @@ and Pods are attached to NSX-T segments, each workload has segment port on NSX-T
...
-configure security of workloads which are attached to NSX-T networking.
-Compared with standard K8s NetworkPolicies, SecurityPolicy CRD adds `VMSelector` to
-select VMs, it also introduces some rule syntaxes including Drop and Reject actions,
-priority, rule level selectors(`appliedTo`). 
+SecurityPolicy is Namespaced scope CRD, admins could apply CR within a namespace to configure security of workloads
+which are attached to NSX-T networking. Compared with standard K8s NetworkPolicies, SecurityPolicy CRD adds `vmSelector`
+to select VMs, it also introduces some rule syntax including Drop and Reject actions, priority, rule level
+selectors(`appliedTo`).
 
 An example of SecurityPolicy CR:
```
timdengyun, line 27: It's better to add a named port in this example since we want to cover usage of SecurityPolicy as much as possible in this one, including matchLables, matchExpressions, namespaceSelector, vmSelector,podSelector and named port etc. It will be a good reference for users.

cmd/main.go
```diff
@@ -8,61 +8,52 @@ import (
...
+	log                    logr.Logger
+	cf                     *config.NSXOperatorConfig
 )
 
 func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 	utilruntime.Must(v1alpha1.AddToScheme(scheme))
-}
-
-func main() {
```
timdengyun, line 38: In Golang, the init() function is reserved and is used for specific reasons. This function is meant to run before any other piece of code – even before the main() function, which supposedly is the first function from where execution of our program begins. So, I suggest we follow Golan convention to keep both init and main function. Although, semantics of current is the same as before.

docs/security-policy.md
```diff
@@ -95,7 +94,7 @@ to access through TCP with port 8000. The second rule allows the selected VMs to
 access Pods with label "role: dns" through UDP with port 53. The third and forth
 rules are to drop any other ingress and egress traffic to/from the selected VMs.
 
-Below are explainations for the fields:
```
timdengyun, line 98: Thank you, xie, for  correction of my type.

PR #85, opened 2022-07-26
Support named port feature

zhengxiexie:
Consider the below scenarios:
 * When a new added pod whose port name exists in security policy
 * When a deleted pod whose port name exists in security policy
 * When a pod's label is changed
 * When a namespace's label is changed
Ensure named port and endPort not coexist, endPort can only be defined if port is also defined, and both ports must be numeric.
Fix compare rule and compare group logic. When pod or namespace label change, the IPSet group may change, however,
the groups compare logic still returns equal, the group misses Expression field and should also use DataValueToJsonEncoder to compare groups.
Fix the bug no pod to be selected if only namespaceSelectors or podSelector.
Fix the bug no right pod to be selected when there is *and* relationship in one
target and *or* relationship in another target.


build/yaml/samples/named-port-sample/rule-out-rule-applied-to-with-dst-with-dup-port-multi.yml
```diff
@@ -0,0 +1,108 @@
...
+    - direction: out
+      action: allow
+      appliedTo:
+        - podSelector:
+            matchLabels:
+              role: web
+      destinations:
+        - podSelector:
+            matchLabels:
+              role: db
```
heypnus, line 102: I guess we can remove L100-L102 because its meaning is "select the pods with label role=db in the same namespace", which can not select the pods in namespace db or db2. There's similar issue in rule-out-rule-applied-to-with-dst-with-dup-port.yml

build/yaml/samples/named-port-sample/rule-out-rule-applied-to-with-dst-with-dup-port-multi.yml
```diff
@@ -0,0 +1,108 @@
+# DB
+# first run
+#  kubectl label ns db2 role=db
+#  test: kubectl exec -ti web-ip -n web -- nc db-ip 3306 -v
+#  kubectl exec -ti web-ip -n web -- nc db-ip 1234 -v
+#  nc db-ip 1234 -v
```
heypnus, line 6: Please polish the description, remove meaningless words and unify the formats.
The same issues in the other yamls.

build/yaml/samples/named-port-sample/rule-out-rule-applied-to-with-dst-with-dup-port-multi.yml
```diff
@@ -0,0 +1,108 @@
...
+  namespace: web
+spec:
+  priority: 10
+  appliedTo:
+    - podSelector:
+        matchLabels:
+          role: web
+  rules:
+    - direction: out
+      action: allow
```
heypnus, line 94: I remember the egress traffic is allowed by default, so if you want to verify it's this security policy to work, you need to add the "drop" rule below, like rule-out-rule-applied-to-with-dst-with-dup-port.yml

build/yaml/samples/named-port-sample/rule-in-policy-applied-to.yml
```diff
@@ -0,0 +1,61 @@
+#  test: kubectl exec -ti client -n client -- curl web ip
```
heypnus, line 1: typo in commit title and PR title! And in the commit message, I think the word should be "scenarios", not "sceneries".

heypnus (changes requested):

pkg/controllers/securitypolicy_associated_controller.go
```diff
@@ -0,0 +1,241 @@
...
+		if p, ok := e.ObjectOld.(*v1.Pod); ok {
+			if checkPodHasNamedPort(*p, "update") {
+				return true
+			}
+		}
+		if p, ok := e.ObjectNew.(*v1.Pod); ok {
+			if checkPodHasNamedPort(*p, "update") {
+				return true
+			}
+		}
```
heypnus, line 190: What's the meanings of ObjectOld and ObjectNew? Could you please add a comment to explain why we need to do the check L180~L191? For example, in which case the return value will be false?

pkg/controllers/securitypolicy_associated_controller.go
```diff
@@ -0,0 +1,241 @@
...
+
+func (e *EnqueueRequestForPod) Delete(deleteEvent event.DeleteEvent, l workqueue.RateLimitingInterface) {
+	e.Raw(deleteEvent, l)
+}
+
+func (e *EnqueueRequestForPod) Generic(genericEvent event.GenericEvent, l workqueue.RateLimitingInterface) {
+	e.Raw(genericEvent, l)
+}
+
+func (e *EnqueueRequestForPod) Raw(evt interface{}, q workqueue.RateLimitingInterface) {
```
heypnus, line 98: Why in some places you named the instance of `workqueue.RateLimitingInterface` to `l` (e.g. L94) and in the other places it's `q` (e.g. L98)? Could you please unify the var name for the same struct?

pkg/controllers/securitypolicy_associated_controller.go
```diff
@@ -0,0 +1,241 @@
...
+				if port.Port.Type == intstr.String {
+					if podPortNames.Has(port.Port.StrVal) {
+						shouldReconcile = true
+					}
+				}
+			}
+		}
+		if shouldReconcile {
+			log.Info("reconcile it", "namespace", securityPolicy.Namespace, "name", securityPolicy.Name)
+			q.Add(reconcile.Request{
```
heypnus, line 162: So this `q` is the work queue for security policy CR, right? I have this question because the enqueued info only includes the CR's name and namespace, I'm not sure how the reconciler knows the resource type of the equeued object.

pkg/controllers/securitypolicy_associated_controller_test.go
```diff
@@ -0,0 +1,570 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			e := &EnqueueRequestForNamespace{
+				Client: tt.fields.Client,
+			}
+			e.Create(tt.args.createEvent, tt.args.l)
+		})
+	}
+}
+
+func TestEnqueueRequestForPod_Raw1(t *testing.T) {
```
heypnus, line 145: `Raw1` -> `Raw`

pkg/controllers/securitypolicy_associated_controller_test.go
```diff
@@ -0,0 +1,570 @@
...
+	) error {
+		return nil
+	})
+	defer patches.Reset()
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			e := &EnqueueRequestForPod{}
+			e.Raw(tt.args.evt, tt.args.q)
+		})
+	}
```
heypnus, line 178: It seems that you only mock the `reconcileSecurityPolicy` and then invoke `Raw`, but didn't verify anything.

pkg/controllers/securitypolicy_associated_controller_test.go
```diff
@@ -0,0 +1,570 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "test-ns-new",
+			},
+		},
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		args    args
+		wantErr assert.ErrorAssertionFunc
```
heypnus, line 558: Where the `wantErr` used?

cmd/main.go
```diff
@@ -78,6 +78,7 @@ func main() {
 		log.Error(err, "unable to initialize securitypolicy service", "controller", "SecurityPolicy")
 		os.Exit(1)
 	} else {
+		service.Client = mgr.GetClient()
```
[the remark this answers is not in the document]
zhengxiexie: In expand.go, to List pod or ns.

pkg/controllers/securitypolicy_associated_controller.go
```diff
@@ -0,0 +1,241 @@
...
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
+)
+
+var ignoreNs = []string{"kube-", "vmware-"}
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/controllers/securitypolicy_associated_controller.go
zhengxiexie: ObjectOld and ObjectNew are defined by controller-runtime. Others I will comment on.

pkg/controllers/securitypolicy_associated_controller_test.go
```diff
@@ -0,0 +1,570 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			e := &EnqueueRequestForNamespace{
+				Client: tt.fields.Client,
+			}
+			e.Create(tt.args.createEvent, tt.args.l)
+		})
+	}
+}
```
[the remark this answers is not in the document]
zhengxiexie: ok, Thanks~~~

pkg/controllers/securitypolicy_associated_controller.go
```diff
@@ -0,0 +1,241 @@
...
+	case event.GenericEvent:
+		obj = et.Object.(*v1.Pod)
+	default:
+		log.Error(nil, "Unknown event type", "event", evt)
+	}
+	pod := obj.(*v1.Pod)
+	pods = append(pods, *pod)
+	err := reconcileSecurityPolicy(e.Client, pods, q)
+	if err != nil {
+		log.Error(err, "Failed to reconcile security policy")
```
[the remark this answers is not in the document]
zhengxiexie: The error occurs when listing security policy, I think there is no need to retry. Put the request to queue and the controller will be responsible to retry.

pkg/controllers/securitypolicy_associated_controller.go
```diff
@@ -0,0 +1,241 @@
...
+			for _, port := range rule.Ports {
+				if port.Port.Type == intstr.String {
+					if podPortNames.Has(port.Port.StrVal) {
+						shouldReconcile = true
+					}
+				}
+			}
+		}
+		if shouldReconcile {
+			log.Info("reconcile it", "namespace", securityPolicy.Namespace, "name", securityPolicy.Name)
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/controllers/securitypolicy_associated_controller.go
zhengxiexie: ok

pkg/controllers/securitypolicy_associated_controller.go
zhengxiexie: This is a good question, I think it is like our main logic `Reconcile` in security_controller.go, one queue maps to a specific CR, `obj := &v1alpha1.SecurityPolicy{}` here stands for security policy.

pkg/controllers/securitypolicy_associated_controller_test.go
zhengxiexie: ok

pkg/controllers/securitypolicy_associated_controller_test.go
zhengxiexie: Yes, because the Raw function returns no error, this test mainly ensures that it supports different events.

pkg/controllers/securitypolicy_associated_controller_test.go
zhengxiexie: Init by framework, used by `tt.wantErr`.

pkg/controllers/securitypolicy_associated_controller.go
```diff
@@ -0,0 +1,246 @@
+/* Copyright © 2022 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: Apache-2.0 */
```
dantingl, line 2: The file name looks weird to me, I did not find a controller defined here

pkg/controllers/securitypolicy_associated_controller.go
```diff
@@ -0,0 +1,246 @@
...
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
+)
+
+// Take these namespaces started with "kube-" or "vmware-" as system namespace.
```
dantingl, line 24: This assumption is wrong, please check `isCRRequestedInSystemNamespace`

pkg/controllers/securitypolicy_associated_controller.go
```diff
@@ -0,0 +1,246 @@
...
+		}
+	}
+	if !shouldReconcile {
+		log.Info("no pod in namespace is relevant", "namespace", obj.Namespace, "name", obj.Name)
+		return
+	}
+
+	err = reconcileSecurityPolicy(e.Client, podList.Items, l)
+	if err != nil {
+		log.Error(err, "Failed to reconcile security policy")
```
dantingl, line 77: Do we need retry for this failure?

pkg/controllers/securitypolicy_controller.go
```diff
@@ -217,6 +231,16 @@ func (r *SecurityPolicyReconciler) setupWithManager(mgr ctrl.Manager) error {
 			controller.Options{
 				MaxConcurrentReconciles: runtime.NumCPU(),
 			}).
+		Watches(
+			&source.Kind{Type: &v1.Namespace{}},
```
dantingl, line 235: May I know why we need to watch namespace event for named port?

pkg/controllers/securitypolicy_associated_controller.go
zhengxiexie: I think it can also be called a controller in terms of business functions, because the main function of this file is to retrigger the logic of reconciliation of security policy as soon as it is found that there is a change in the resources associated with the security policy.

pkg/controllers/securitypolicy_associated_controller.go
zhengxiexie: The error occurs only when listing security policy which rarely happens, I think there is no need to retry. Put the request to queue and the controller will be responsible to retry.

pkg/controllers/securitypolicy_controller.go
zhengxiexie: We have namespaceSelector
`   destinations:
        - podSelector:
            matchLabels:
              role: db
            matchExpressions:
              - {key: k1, operator: In, values: [a1,a2]}
              - {key: k2, operator: NotIn, values: [a3,a4]}
        - namespaceSelector:
            matchLabels:
              role: db`
, if one namespace's label satisfied the criteria for the selector, we should reconcile.

pkg/controllers/securitypolicy_associated_controller.go
zhengxiexie: ok, Thanks~~~

pkg/nsx/services/builder.go
```diff
@@ -639,7 +722,42 @@ func (service *SecurityPolicyService) buildRuleDstGroup(obj *v1alpha1.SecurityPo
 	return &ruleDstGroup, ruleDstGroupPath, err
 }
 
-func (service *SecurityPolicyService) buildPeerTags(obj *v1alpha1.SecurityPolicy, peers *[]v1alpha1.SecurityPolicyPeer, idx int) []model.Tag {
+func (service *SecurityPolicyService) buildRuleBasicInfo(obj *v1alpha1.SecurityPolicy,
+	rule *v1alpha1.SecurityPolicyRule, ruleIdx int, portIdx int, dupPortIdx int,
```
heypnus, line 726: Add comments to explain the function, especially the meaning of `portIdx` and `dupPortIdx`.

pkg/nsx/services/expand.go
```diff
@@ -0,0 +1,308 @@
...
+
+	if len(rule.Ports) == 0 {
+		rs, err := service.buildRuleBasicInfo(obj, rule, ruleIdx, 0, 0)
+		if err != nil {
+			return nil, nil, err
+		}
+		nsxRules = append(nsxRules, rs)
+	}
+	for portIdx, port := range rule.Ports {
+		gs, rs, err := service.expandRuleByPort(obj, rule, ruleIdx, port, portIdx)
```
heypnus, line 39: Please don't use ns/gs/rs/... as the variable names, it's too hard to understand the logic!
And the naming styles, `nsxRules` and `rs` are not consistent.

pkg/nsx/services/builder_test.go
```diff
@@ -406,7 +406,7 @@ func TestValidateSelectorExpressions(t *testing.T) {
 	assert.Equal(t, 2, totalCriteriaCount)
 	assert.Equal(t, 30, totalExprCount)
 
-	// Case: total count of expressions exceed NSX limit '15' in one criterion mixed criteria
+	// Case: total count of expressions exceed NSX limit '15' in one criteria mixed criteria
```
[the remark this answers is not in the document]
timdengyun: Maybe "in one criterion with mixed member type" is better

pkg/controllers/securitypolicy_controller.go
```diff
@@ -183,11 +196,12 @@ func (r *SecurityPolicyReconciler) updateSecurityPolicyStatusConditions(ctx *con
...
 }
 
-func (r *SecurityPolicyReconciler) mergeSecurityPolicyStatusCondition(ctx *context.Context, sec_policy *v1alpha1.SecurityPolicy, newCondition *v1alpha1.SecurityPolicyCondition) bool {
+func (r *SecurityPolicyReconciler) mergeSecurityPolicyStatusCondition(_ *context.Context, sec_policy *v1alpha1.SecurityPolicy, newCondition *v1alpha1.SecurityPolicyCondition) bool {
 	matchedCondition := getExistingConditionOfType(newCondition.Type, sec_policy.Status.Conditions)
 
 	if reflect.DeepEqual(matchedCondition, newCondition) {
-		log.V(2).Info("Conditions already match", "New Condition", newCondition, "Existing Condition", matchedCondition)
+		log.V(2).
+			Info("conditions already match", "New Condition", newCondition, "Existing Condition", matchedCondition)
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/services/builder_test.go
```diff
@@ -394,7 +394,7 @@ func TestValidateSelectorExpressions(t *testing.T) {
 	assert.Equal(t, 2, totalCriteriaCount)
 	assert.Equal(t, 10, totalExprCount)
 
-	// Case: total count of expressions exceed NSX limit '5' in one criterion based on same member type
+	// Case: total count of expressions exceed NSX limit '5' in one criteria based on same member type
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/services/builder_test.go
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/services/compare.go
```diff
@@ -49,22 +49,57 @@ func RulesEqual(existingRules []model.Rule, rules []model.Rule) (bool, []model.R
...
+		se1 := existingRules[i].ServiceEntries
+		se2 := rules[i].ServiceEntries
+		if len(se1) != len(se2) {
+			log.Info(
+				"service entry len diff",
+				"nsx service entry len",
+				len(se1),
+				"k8s service entry len",
+				len(se2),
+			)
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/services/firewall_test.go
```diff
@@ -36,14 +36,17 @@ var (
...
 	seq2                         = int64(2)
-	ruleNameWithPodSelector      = "rule-with-pod-selector"
-	ruleNameWithVMSelector       = "rule-with-VM-selector"
-	ruleNameWithNsSelector       = "rule-with-ns-selector"
-	ruleNameWithIpBlock          = "rule-with-ip-block"
+	ruleNameWithPodSelector      = "rule-with-pod-selector-0-0"
+	ruleNameWithVMSelector       = "rule-with-VM-selector-0-0"
+	ruleNameWithNsSelector       = "rule-with-ns-selector-0-0"
+	ruleNameWithIpBlock          = "rule-with-ip-block-0-0"
 	cidr                         = "192.168.1.1/24"
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/services/firewall_test.go
```diff
@@ -36,14 +36,17 @@ var (
...
 	ruleID0                      = "sp_uidA_0"
 	ruleID1                      = "sp_uidA_1"
 	ruleID2                      = "sp_uidA_2"
+	ruleIDPort0                  = "sp_uidA_0_0_0"
+	ruleIDPort1                  = "sp_uidA_1_0_0"
+	ruleIDPort2                  = "sp_uidA_2_0_0"
 	nsxDirectionIn               = "IN"
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/services/builder.go
zhengxiexie: ok

pkg/nsx/services/expand.go
zhengxiexie: ok

pkg/nsx/services/expand.go
```diff
@@ -0,0 +1,308 @@
...
+		map[string]data.DataValue{
+			"resource_type": data.NewStringValue("IPAddressExpression"),
+			"ip_addresses":  addresses,
+		},
+	)
+	ipSetGroup.Expression = append(ipSetGroup.Expression, blockExpression)
+	return &ipSetGroup
+}
+
+// Different direction rule decides different target of the traffic, we should carefully get
```
TaoZou1, line 225: maybe emphasize that named port only care destination selector. that means IN, rule/policy appliedTo, OUT rule's destination.

pkg/nsx/services/expand.go
```diff
@@ -0,0 +1,308 @@
...
+// the destination pod selector and namespaces.
+func (service *SecurityPolicyService) getPodSelector(obj *v1alpha1.SecurityPolicy,
+	rule *v1alpha1.SecurityPolicyRule,
+) (*client.ListOptions, []string, error) {
+	// Port means the target of traffic, so we should select the pod by rule direction,
+	// as for IN direction, we judge by rule's or policy's AppliedTo,
```
TaoZou1, line 231: appliedTo

pkg/controllers/securitypolicy_associated_controller.go
```diff
@@ -0,0 +1,250 @@
...
+		log.Error(nil, "unknown event type", "event", evt)
+	}
+	pod := obj.(*v1.Pod)
+	if isInSysNs, err := isNsInSystemNamespace(e.Client, pod.Namespace); err != nil {
+		return
+	} else if isInSysNs {
+		log.Info("pod is in system namespace, do nothing")
+		return
+	}
+	err := reconcileSecurityPolicy(e.Client, pods, q)
```
dantingl, line 120: put pod in pods?

pkg/controllers/securitypolicy_associated_controller.go
```diff
@@ -0,0 +1,250 @@
...
+		log.Error(err, "failed to list security policy")
+		return err
+	}
+
+	for _, securityPolicy := range spList.Items {
+		shouldReconcile := false
+		for _, rule := range securityPolicy.Spec.Rules {
+			for _, port := range rule.Ports {
+				if port.Port.Type == intstr.String {
+					if podPortNames.Has(port.Port.StrVal) {
```
dantingl, line 155: Do we need to check whether the securitypolicy would be applied to this pod?

pkg/controllers/securitypolicy_associated_controller.go
```diff
@@ -0,0 +1,250 @@
...
+						shouldReconcile = true
+					}
+				}
+			}
+		}
+		if shouldReconcile {
+			log.Info("reconcile security policy", "namespace", securityPolicy.Namespace, "name", securityPolicy.Name)
```
dantingl, line 162: reconcile security policy because of named port change

pkg/controllers/securitypolicy_associated_controller.go
```diff
@@ -0,0 +1,250 @@
...
+			})
+		}
+	}
+	return nil
+}
+
+var PredicateFuncsPod = predicate.Funcs{
+	CreateFunc: func(e event.CreateEvent) bool {
+		if p, ok := e.Object.(*v1.Pod); ok {
+			return checkPodHasNamedPort(*p, "create")
```
dantingl, line 177: Keep line 177, 183, 188, 196 the same

pkg/controllers/securitypolicy_associated_controller.go
```diff
@@ -0,0 +1,250 @@
...
+	},
+	UpdateFunc: func(e event.UpdateEvent) bool {
+		oldObj := e.ObjectOld.(*v1.Namespace)
+		newObj := e.ObjectNew.(*v1.Namespace)
+		log.Info("receive namespace event", "name", oldObj.Name)
+		if reflect.DeepEqual(oldObj.ObjectMeta.Labels, newObj.ObjectMeta.Labels) {
+			log.Info("label of namespace is not changed, ignore it", "name", oldObj.Name)
+			return false
+		}
+		if isSysNs, ok := oldObj.Annotations[WCP_SYSTEM_RESOURCE]; ok && strings.ToLower(isSysNs) == "true" {
```
dantingl, line 229: Same here, use a single function to check if it's system NS

docs/security-policy.md
```diff
@@ -139,4 +139,4 @@ Limitations of SecurityPolicy CR:
    given NSX-T does not support 'In' in NSGroup condition, so we use a workaround to support 'In' with limited counts.
 7. Max IP elements in one security policy: 4000
 8. Priority range of SecurityPolicy CR is [0, 1000].
-9. Do not support named port.
\ No newline at end of file
+9. Support named port.
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/controllers/securitypolicy_associated_controller.go
```diff
@@ -0,0 +1,250 @@
...
+// When a namespace's label is changed.
+// In summary, we could roughly think if the port name of security policy exists in the
+// new pod or old pod, we should reconcile the security policy.
+
+type EnqueueRequestForNamespace struct {
+	Client client.Client
+}
+
+func (e *EnqueueRequestForNamespace) Create(_ event.CreateEvent, _ workqueue.RateLimitingInterface) {
+	log.V(1).Info("namespace create event, do nothing")
```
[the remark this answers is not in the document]
zhengxiexie: If namespace create event occurs but there is no pod in it, we can ignore it. If there is pod in it, then pod create event
would trigger the logic.

pkg/controllers/securitypolicy_associated_controller.go
```diff
@@ -0,0 +1,250 @@
...
+			return false
+		}
+		return true
+	},
+	DeleteFunc: func(e event.DeleteEvent) bool {
+		return false
+	},
+}
+
+func isNsInSystemNamespace(c client.Client, ns string) (bool, error) {
```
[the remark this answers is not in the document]
zhengxiexie: The arg is different, it's a bit difficult to put it in one place.

pkg/controllers/securitypolicy_associated_controller.go
```diff
@@ -0,0 +1,250 @@
...
+	case event.DeleteEvent:
+		obj = et.Object.(*v1.Pod)
+	case event.GenericEvent:
+		obj = et.Object.(*v1.Pod)
+	default:
+		log.Error(nil, "unknown event type", "event", evt)
+	}
+	pod := obj.(*v1.Pod)
+	if isInSysNs, err := isNsInSystemNamespace(e.Client, pod.Namespace); err != nil {
+		return
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/controllers/securitypolicy_associated_controller.go
zhengxiexie: Oh, Thanks~~~

pkg/controllers/securitypolicy_associated_controller.go
zhengxiexie: I think putting the request into the queue, and letting the reconcile logic check is a good way.

pkg/controllers/securitypolicy_associated_controller.go
zhengxiexie: ok

pkg/controllers/securitypolicy_associated_controller.go
zhengxiexie: UpdateFunc could be deleted.

pkg/controllers/securitypolicy_associated_controller.go
zhengxiexie: ok

pkg/controllers/securitypolicy_associated_controller.go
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/controllers/securitypolicy_associated_controller.go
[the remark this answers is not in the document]
heypnus: How does the controller know there's an error when listing security policy and then retry? I saw the error was not in the returned value of the func `Raw`. And in func `reconcileSecurityPolicy`, when the client fails to list security policy, it returns before operating on workqueue.

pkg/nsx/services/expand.go
```diff
@@ -0,0 +1,310 @@
...
+		if len(obj.Spec.AppliedTo) > 0 {
+			for _, target := range obj.Spec.AppliedTo {
+				if target.PodSelector != nil {
+					labelSelector, err = meta1.LabelSelectorAsSelector(target.PodSelector)
+					if err != nil {
+						return nil, nil, err
+					}
+				}
+			}
+		} else if len(rule.AppliedTo) > 0 {
```
dantingl, line 254: Is it possible `rule.AppliedTo` and `obj.Spec.AppliedTo` both have items?

pkg/nsx/services/compare.go
```diff
@@ -49,22 +49,57 @@ func RulesEqual(existingRules []model.Rule, rules []model.Rule) (bool, []model.R
...
+				simplifyRule(&rules[i]),
+			)
 			isEqual = false
 			break
 		}
+
+		se1 := existingRules[i].ServiceEntries
+		se2 := rules[i].ServiceEntries
+		if len(se1) != len(se2) {
+			log.Info(
```
dantingl, line 76: I don't think the way to print log in so many lines, is it golint style?

pkg/nsx/services/builder_test.go
```diff
@@ -466,7 +466,7 @@ func TestValidateSelectorOpIn(t *testing.T) {
 	assert.NotEqual(t, nil, err)
 	assert.Equal(t, 6, opInValueCount)
 
-	// Case: matchLabels has duplication expression with match-expression operator 'In'
+	// Case: matchLabels has duplication expression with matchexpression operator 'In'
```
dantingl, line 469: matchExpressions

pkg/controllers/securitypolicy_associated_controller.go
heypnus: I think the file structure style with `xxx_associated_controller.go` is not good for our future expansion, because if we think that security policy is the core, and pod and namespace are satellites, then multiple cores may have the same satellites, which will result in the intersection of resource types. E.g. In the future, if we also implement the K8s network policy in NSX operator, we also need to "retrigger the logic of reconciliation of network policy as soon as it is found that there is a change in the resources associated with the network policy". If we also added the `networkpolicy_associated_controller.go`, we should also add the duplicate codes for the duplicate resource types pod and namespace in the new associated controller? So I personally prefer to put `EnqueueRequestForNamespace` in namespace controller, put `EnqueueRequestForPod` in pod controller and the other logic about workflow to securitypolicy_controller.go.

pkg/util/utils.go
```diff
@@ -59,6 +60,25 @@ func RemoveDuplicateStr(strSlice []string) []string {
 	return resultStr
 }
 
+func MergeAddressByPort(portAddressOriginal []PortAddress) []PortAddress {
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/util/utils.go
```diff
@@ -232,8 +244,21 @@ func HandleHTTPResponse(response *http.Response, result interface{}, debug bool)
...
 	}
 	if err := json.Unmarshal(body, result); err != nil {
-		log.Error(err, "Error converting HTTP response to result", "result type", result)
+		log.Error(err, "error converting HTTP response to result", "result type", result)
 		return err, body
 	}
 	return nil, body
 }
+
+func IsSystemNamespace(c client.Client, ns string, obj *v1.Namespace) (bool, error) {
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/services/expand.go
```diff
@@ -0,0 +1,310 @@
...
+	ruleDirection, err := getRuleDirection(rule)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if ruleDirection == "IN" {
+		if len(obj.Spec.AppliedTo) > 0 {
+			for _, target := range obj.Spec.AppliedTo {
+				if target.PodSelector != nil {
+					labelSelector, err = meta1.LabelSelectorAsSelector(target.PodSelector)
```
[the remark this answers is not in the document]
zhengxiexie: type SecurityPolicyTarget struct {
	// VMSelector uses label selector to select VMs.
	VMSelector *metav1.LabelSelector `json:"vmSelector,omitempty"`
	// PodSelector uses label selector to select Pods.
	PodSelector *metav1.LabelSelector `json:"podSelector,omitempty"`
}

The definition is vmSelector or podSelector, do we permit multiple podSelector?

pkg/nsx/services/expand.go
zhengxiexie: obj level is preferred to rule level

pkg/nsx/services/compare.go
zhengxiexie: golines style.

pkg/nsx/services/builder_test.go
zhengxiexie: ok

pkg/nsx/services/expand.go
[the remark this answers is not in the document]
zhengxiexie: I think it's user's responsibility to ensure only one podSelector in target.

pkg/controllers/securitypolicy_associated_controller.go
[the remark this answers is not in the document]
zhengxiexie: Let me try to put the pod in the queue again, perhaps it can solve it.

pkg/controllers/securitypolicy_associated_controller.go
zhengxiexie: ok, it makes sense, let me think about how to rename it.

zhengxiexie: Done, pls re-check

pkg/controllers/securitypolicy_associated_controller.go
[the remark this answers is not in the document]
zhengxiexie: As discussed, the controller-runtime seems not to support retry associated events, so let's just record the error log (which rarely happens) for now.

pkg/controllers/securitypolicy_associated_controller.go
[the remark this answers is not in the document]
dantingl: Then why do not ignore namespace update event?

pkg/nsx/services/expand.go
dantingl: Then you need to check rule.AppliedTo first?

zhengxiexie: I mean, obj level is first, so the order is right.

zhengxiexie: Doc says.

**appliedTo**: is a list of policy targets to apply rules. As the CRD is namespaced
scope, `vmSelector` or `podSelector` will be selected from the Namespace where the
CR is created. `vmSelector` and `podSelector` cannot be in one entry as it would
not select any workload. We can also have `appliedTo` in each rule entry, but if
there is policy level `appliedTo`, it will take precedence over rule level.

pkg/nsx/util/utils.go
```diff
@@ -29,6 +28,14 @@ type ErrorDetail struct {
 	Details            string
 }
 
+// PortAddress is used when named port is specified.
+type PortAddress struct {
+	// Port is the port number.
+	Port int `json:"port"`
+	// IPs is a list of IPs.
```
timdengyun, line 35: IPs is a list of IPs associated to port number.

pkg/controllers/securitypolicy/namespace_controller.go
```diff
@@ -0,0 +1,90 @@
...
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	util2 "github.com/vmware-tanzu/nsx-operator/pkg/util"
+)
+
+// We should consider the below sceneries:
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/controllers/securitypolicy/namespace_controller.go
```diff
@@ -0,0 +1,90 @@
...
+// When a namespace's label is changed, perhaps as to the pods in it, we should reconcile the corresponding security policy.
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/mock/controller-runtime/client/mock_workqueue.go
```diff
@@ -0,0 +1,125 @@
+package mock_client
+
+import (
+	reflect "reflect"
+	"time"
+
+	gomock "github.com/golang/mock/gomock"
+)
+
+// MockInterface is a mock of Interface interface.
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/util/utils.go
```diff
@@ -63,3 +72,31 @@ func ToUpper(obj interface{}) string {
...
+	} else if err := c.Get(context.Background(), types.NamespacedName{Namespace: ns, Name: ns}, nsObj); err != nil {
+		return false, client.IgnoreNotFound(err)
+	}
+	if isSysNs, ok := nsObj.Annotations[wcpSystemResource]; ok && strings.ToLower(isSysNs) == "true" {
+		return true, nil
+	}
+	return false, nil
+}
+
+// CheckPodHasNamedPort checks if the pod has a named port, it filters the pod events
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/util/utils.go
```diff
@@ -63,3 +72,31 @@ func ToUpper(obj interface{}) string {
...
+func CheckPodHasNamedPort(pod v1.Pod, reason string) bool {
+	for _, container := range pod.Spec.Containers {
+		for _, port := range container.Ports {
+			if port.Name != "" {
+				log.V(1).Info(fmt.Sprintf("%s pod %s has a named port %s", reason, pod.Name, port.Name))
+				return true
+			}
+		}
+	}
+	log.V(1).Info(fmt.Sprintf("%s pod %s has no named", reason, pod.Name))
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/controllers/securitypolicy/pod_controller.go
```diff
@@ -0,0 +1,123 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	util2 "github.com/vmware-tanzu/nsx-operator/pkg/util"
+)
+
+// We should consider the below sceneries:
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/controllers/securitypolicy/pod_controller.go
```diff
@@ -0,0 +1,123 @@
...
+	default:
+		log.Error(nil, "unknown event type", "event", evt)
+	}
+
+	pod := obj.(*v1.Pod)
+	if isInSysNs, err := util2.IsSystemNamespace(e.Client, pod.Namespace, nil); err != nil {
+		log.Error(err, "failed to fetch namespace", "namespace", pod.Namespace)
+		return
+	} else if isInSysNs {
+		log.V(2).Info("pod is in system namespace, do nothing")
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/util/utils.go
zhengxiexie: ok

pkg/controllers/securitypolicy/pod_controller.go
```diff
@@ -0,0 +1,123 @@
...
+	},
+	UpdateFunc: func(e event.UpdateEvent) bool {
+		oldObj := e.ObjectOld.(*v1.Pod)
+		newObj := e.ObjectNew.(*v1.Pod)
+		log.V(1).Info("receive pod update event", "namespace", oldObj.Namespace, "name", oldObj.Name)
+		if reflect.DeepEqual(oldObj.ObjectMeta.Labels, newObj.ObjectMeta.Labels) {
+			log.V(1).Info("label of pod is not changed, ignore it", "name", oldObj.Name)
+			return false
+		}
+		if util2.CheckPodHasNamedPort(*oldObj, "update") {
```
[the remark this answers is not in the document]
zhengxiexie: We can remove checking the old, since the old and new both have the same port.

pkg/nsx/services/firewall.go
```diff
@@ -125,10 +134,12 @@ func (service *SecurityPolicyService) createOrUpdateGroups(nsxGroups []model.Gro
 			return err
 		}
 		err = service.GroupStore.Add(group)
+		log.V(2).Info("Add group to store", "group", group.Id)
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/services/compare.go
```diff
@@ -14,14 +14,15 @@ func SecurityPolicyEqual(existingSecurityPolicy *model.SecurityPolicy, securityP
 	if string(s1) == string(s2) {
 		return true
 	}
-	log.Info("security policy diff", "nsx sp", s1, "k8s sp", s2)
+	log.Info("compare security policy", "nsx security policy", simplifySecurityPolicy(existingSecurityPolicy),
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/services/builder.go
```diff
@@ -44,14 +54,23 @@ func (service *SecurityPolicyService) buildSecurityPolicy(obj *v1alpha1.Security
...
 		if err != nil {
 			log.Error(err, "failed to build rule and groups", "rule", rule, "ruleIndex", ruleIdx)
 			return nil, nil, err
 		}
-		nsxRules = append(nsxRules, *nsxRule)
-		for _, ruleGroup := range *ruleGroups {
-			nsxGroups = append(nsxGroups, ruleGroup)
+		for _, nsxRule := range expandRules {
+			if nsxRule != nil {
+				nsxRules = append(nsxRules, *nsxRule)
```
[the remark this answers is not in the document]
zhengxiexie: The inner logic doesn't produce the duplicated rules, however, the groups do.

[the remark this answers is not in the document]
timdengyun: ok. I see. Thanks

pkg/nsx/services/compare.go
```diff
@@ -2,10 +2,11 @@ package services
```
dantingl, line 2: Put builder.go and compare.go into a securitypolicy subdir?

pkg/nsx/services/expand.go
```diff
@@ -0,0 +1,303 @@
...
+			}
+		}
+	} else if ruleDirection == "OUT" {
+		if len(rule.Destinations) > 0 {
+			var labelSelectors []client.ListOptions
+			var namespaceSelectors []client.ListOptions
+			for _, target := range rule.Destinations {
+				labelSelector := client.ListOptions{}
+				namespaceSelector := client.ListOptions{}
+				if target.PodSelector != nil {
```
timdengyun, line 249: Do we need to handle target.PodSelector == nil?

docs/security-policy.md
```diff
@@ -138,5 +138,4 @@ Limitations of SecurityPolicy CR:
 6. In one NSGroup group, supports only one 'In' with at most of five values in MatchExpressions,
    given NSX-T does not support 'In' in NSGroup condition, so we use a workaround to support 'In' with limited counts.
 7. Max IP elements in one security policy: 4000
-8. Priority range of SecurityPolicy CR is [0, 1000].
-9. Do not support named port.
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/services/compare.go
```diff
@@ -14,14 +14,15 @@ func SecurityPolicyEqual(existingSecurityPolicy *model.SecurityPolicy, securityP
 	if string(s1) == string(s2) {
 		return true
 	}
-	log.Info("security policy diff", "nsx sp", s1, "k8s sp", s2)
+	log.Info("compare security policy", "nsx security policy",
```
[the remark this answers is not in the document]
zhengxiexie: OK, -> log.V(1)

pkg/nsx/services/compare.go
```diff
@@ -49,40 +50,85 @@ func RulesEqual(existingRules []model.Rule, rules []model.Rule) (bool, []model.R
...
+		// now we fix this problem by comparing them separately.
+		r1, _ := json.Marshal(simplifyRule(&existingRules[i]))
+		r2, _ := json.Marshal(simplifyRule(&rules[i]))
+		if string(r1) != string(r2) {
+			log.Info(
+				"compare rule",
+				"nsx rule",
+				simplifyRule(&existingRules[i]),
+				"k8s rule",
+				simplifyRule(&rules[i]),
```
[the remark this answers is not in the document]
zhengxiexie: log.V(1).Info(
				"rules differ",
				"current NSX rule", simplifyRule(&existingRules[i]),
				"desired NSX rule", simplifyRule(&rules[i]))

pkg/nsx/services/compare.go
zhengxiexie: If so,  I think we should move all the files in services to securitypolicy subdir ?

pkg/nsx/services/builder.go
```diff
@@ -291,201 +317,183 @@ func (service *SecurityPolicyService) buildPolicyGroupPath(obj *v1alpha1.Securit
...
-				log.Error(err, "failed to build rule source groups")
+	for _, nsxRule := range nsxRules {
+		if ruleDirection == "IN" {
+			nsxRuleSrcGroup, nsxRuleSrcGroupPath, nsxRuleDstGroupPath, err = service.buildRuleInGroup(
+				obj,
+				rule,
+				nsxRule,
+				ruleIdx,
+			)
+			if err != nil {
```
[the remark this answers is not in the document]
zhengxiexie: I think we'd better print the error log in the outermost layer only once, otherwise, there is the same error info everywhere in the log.

pkg/nsx/services/builder.go
```diff
@@ -291,201 +317,183 @@ func (service *SecurityPolicyService) buildPolicyGroupPath(obj *v1alpha1.Securit
...
-		if len(rule.Destinations) > 0 {
-			nsxRuleDstGroup, nsxRuleDstGroupPath, err = service.buildRuleDstGroup(obj, rule, ruleIdx)
-			if err == nil {
-				ruleGroups = append(ruleGroups, *nsxRuleDstGroup)
-			} else {
-				log.Error(err, "failed to build rule destination groups")
+			ruleGroups = append(ruleGroups, nsxRuleSrcGroup)
+		} else if ruleDirection == "OUT" {
+			nsxRuleDstGroup, nsxRuleSrcGroupPath, nsxRuleDstGroupPath, err = service.buildRuleOutGroup(obj, rule, nsxRule, ruleIdx)
+			if err != nil {
```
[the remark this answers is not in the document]
zhengxiexie: ditto

pkg/nsx/services/builder.go
```diff
@@ -639,7 +689,38 @@ func (service *SecurityPolicyService) buildRuleDstGroup(obj *v1alpha1.SecurityPo
...
+		SequenceNumber: &sequence,
+		Action:         &ruleAction,
+		Services:       []string{"ANY"},
+		Tags:           service.buildBasicTags(obj),
+	}
+	log.V(1).Info("built rule basic info", "nsxRule", nsxRule)
+	return &nsxRule, nil
+}
+
+func (service *SecurityPolicyService) BuildPeerTags(obj *v1alpha1.SecurityPolicy, peers *[]v1alpha1.SecurityPolicyPeer, idx int) []model.Tag {
```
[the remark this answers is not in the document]
zhengxiexie: Because here,
				patches := gomonkey.ApplyMethod(reflect.TypeOf(s), "BuildPeerTags",
		func(s *SecurityPolicyService, v *v1alpha1.SecurityPolicy, p *[]v1alpha1.SecurityPolicyPeer, i int) []model.Tag {
			peerTags := []model.Tag{
				{Scope: nil, Tag: nil},
			}
			return peerTags
		})
I have to change it to upper case, otherwise, it can't gomonkey it.

pkg/nsx/services/compare.go
```diff
@@ -14,14 +14,15 @@ func SecurityPolicyEqual(existingSecurityPolicy *model.SecurityPolicy, securityP
...
 	}
-	log.Info("security policy diff", "nsx sp", s1, "k8s sp", s2)
+	log.Info("security policies differ", "nsx security policy", simplifySecurityPolicy(existingSecurityPolicy),
+		"k8s security policy", simplifySecurityPolicy(securityPolicy))
 	return false
 }
 
 func RulesEqual(existingRules []model.Rule, rules []model.Rule) (bool, []model.Rule) {
 	// sort the rules by id, otherwise expandRule may return different results, only the sequence of the
 	// rule is different, so sort by port number, and it avoids the needless updates.
```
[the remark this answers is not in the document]
zhengxiexie: Outdated.
Now remove the sort logic, refactored, no need anymore.

pkg/nsx/services/expand.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+	log.V(2).Info("built rule by service entry", "rule", nsxRule)
+	return nsxGroups, nsxRule, nil
+}
+
+// Resolve a named port to port number by rule and policy selector.
+// e.g. "http" -> [{"80":['1.1.1.1', '2.2.2.2']}, {"443":['3.3.3.3']}]
+func (service *SecurityPolicyService) resolveNamedPort(obj *v1alpha1.SecurityPolicy, rule *v1alpha1.SecurityPolicyRule,
+	spPort v1alpha1.SecurityPolicyPort) ([]util2.PortAddress, error) {
+	var address []util2.PortAddress
```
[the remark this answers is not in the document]
zhengxiexie: You mean the method name ?

pkg/nsx/services/builder.go
```diff
@@ -521,20 +533,30 @@ func (service *SecurityPolicyService) buildRuleAppliedGroupByRule(obj *v1alpha1.
...
+		)
 		if err == nil {
 			ruleGroupCount += criteriaCount
 			ruleGroupTotalExprCount += totalExprCount
 		} else {
 			return nil, "", err
 		}
 	}
-	log.V(2).Info("build rule applied group criteria", "total criteria", ruleGroupCount, "total expressions of criteria", ruleGroupTotalExprCount)
+	log.V(2).
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/services/builder.go
```diff
@@ -567,20 +593,30 @@ func (service *SecurityPolicyService) buildRuleSrcGroup(obj *v1alpha1.SecurityPo
...
+		)
 		if err == nil {
 			ruleSrcGroupCount += criteriaCount
 			ruleSrcGroupTotalExprCount += totalExprCount
 		} else {
 			return nil, "", err
 		}
 	}
-	log.V(1).Info("build rule source group criteria", "total criteria", ruleSrcGroupCount, "total expressions of criteria", ruleSrcGroupTotalExprCount)
+	log.V(1).
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/services/builder.go
```diff
@@ -614,20 +654,30 @@ func (service *SecurityPolicyService) buildRuleDstGroup(obj *v1alpha1.SecurityPo
...
+		)
 		if err == nil {
 			ruleDstGroupCount += criteriaCount
 			ruleDstGroupTotalExprCount += totalExprCount
 		} else {
 			return nil, "", err
 		}
 	}
-	log.V(1).Info("build rule destination group criteria", "total criteria", ruleDstGroupCount, "total expressions of criteria", ruleDstGroupTotalExprCount)
+	log.V(1).
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/services/expand.go
```diff
@@ -0,0 +1,299 @@
...
+	ipSetGroup.Expression = append(ipSetGroup.Expression, blockExpression)
+	return &ipSetGroup
+}
+
+// Different direction rule decides different target of the traffic, we should carefully get
+// the destination pod selector and namespaces. Named port only cares about the destination
+// pod selector, so we use policy's AppliedTo or rule's AppliedTo when "IN" direction and
+// rule's DestinationSelector when "OUT" direction.
+func (service *SecurityPolicyService) getPodSelectors(obj *v1alpha1.SecurityPolicy, rule *v1alpha1.SecurityPolicyRule) ([]client.ListOptions, error) {
+	// Port means the target of traffic, so we should select the pod by rule direction,
```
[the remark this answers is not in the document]
zhengxiexie: I think we only care about the target port here.
We say target port = the destination of traffic by convention.

pkg/nsx/services/expand.go
```diff
@@ -0,0 +1,303 @@
...
+						namespaceSelector.Namespace = nsItem.Name
+						namespaceSelectors = append(namespaceSelectors, namespaceSelector)
+					}
+				} else {
+					namespaceSelector.Namespace = obj.Namespace
+					namespaceSelectors = append(namespaceSelectors, namespaceSelector)
+				}
+			}
+			// calculate the union of labelSelectors and namespaceSelectors
+			for _, nsSelector := range namespaceSelectors {
```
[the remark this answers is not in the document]
zhengxiexie: fixed.

pkg/nsx/services/expand.go
zhengxiexie: The logic considers this.

pkg/nsx/services/expand.go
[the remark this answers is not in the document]
timdengyun: No, I mean  "var address []util2.PortAddress"  could be "var portAddress []util2.PortAddress"

pkg/nsx/services/expand.go
```diff
@@ -0,0 +1,317 @@
...
+					for _, nsItem := range ns.Items {
+						namespaceSelector.Namespace = nsItem.Name
+						namespaceSelectors = append(namespaceSelectors, namespaceSelector)
+					}
+				} else {
+					namespaceSelector.Namespace = obj.Namespace
+					namespaceSelectors = append(namespaceSelectors, namespaceSelector)
+				}
+			}
+			if len(labelSelectors) > 0 && len(namespaceSelectors) > 0 {
```
timdengyun, line 271: Considering this case:

apiVersion: apps/v1
kind: Deployment
metadata:
  name: mysql
  namespace: db2
  labels:
    role: db
spec:
  selector:
    matchLabels:
      role: db
  template:
    metadata:
      labels:
        role: db
    spec:
      containers:
        - image: "harbor-repo.vmware.com/dockerhub-proxy-cache/gophernet/netcat:latest"
          command: ["nc", "-l", "-p", "3306"]
          name: mysql2
          ports:
            - containerPort: 3306
              name: mysql-port

apiVersion: apps/v1
kind: Deployment
metadata:
  name: mysql
  namespace: db3
  labels:
    role: db3
spec:
  selector:
    matchLabels:
      role: db3
  template:
    metadata:
      labels:
        role: db3
    spec:
      containers:
        - image: "harbor-repo.vmware.com/dockerhub-proxy-cache/gophernet/netcat:latest"
          command: ["nc", "-l", "-p", "3306"]
          name: mysql3
          ports:
            - containerPort: 3306
              name: mysql-port

kubectl label ns db2 role=db
kubectl label ns db3 role=db3

we have rule like:
  rules:
    - direction: out
      action: allow
      appliedTo:
        - podSelector:
            matchLabels:
              role: web
      destinations:
        - podSelector:
            matchLabels:
              role: db
          namespaceSelector:
            matchLabels:
              role: db
        - namespaceSelector:
            matchLabels:
              role: db3
      ports:
        - protocol: TCP
          port: mysql-port

If we print at code line #L274
log.Info("getPodSelectors", "nsSelector", nsSelector.Namespace)
log.Info("getPodSelectors", "labelSelector", labelSelector.LabelSelector)
output:
2022-09-07 15:19:00.349	INFO	services/expand.go:280	getPodSelectors	{"nsSelector": "db2"}
2022-09-07 15:19:00.349	INFO	services/expand.go:281	getPodSelectors	{"labelSelector": "role=db"}
2022-09-07 15:19:00.349	INFO	services/expand.go:280	getPodSelectors	{"nsSelector": "db3"}
2022-09-07 15:19:00.349	INFO	services/expand.go:281	getPodSelectors	{"labelSelector": "role=db"}

You can see podSelector for "role=db" also applied to "nsSelector": "db3". I think this is not we want. We should select all pods that match namespaceSelector "nsSelector": "db3", that means all pods in namespace db3.  
As a result, we are not able to select pods from db3 , the result is showing only pod from db2: 172.20.0.74  is selected. 

db2                   mysql-c744b6ff8-9mwdl                   1/1     Running   0                2m41s   172.20.0.74   10.173.83.168   <none>           <none>
db3                   mysql-66c999f85d-ks4d7                  1/1     Running   0                2m41s   172.20.0.90   10.173.83.168   <none>           <none>

Therefore, I suggest we might need to consider how to fix it .

pkg/nsx/services/expand.go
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/services/builder.go
```diff
@@ -12,13 +12,23 @@ import (
...
+
+	contains := func(groups []model.Group, group model.Group) bool {
+		for _, a := range groups {
+			if *a.Id == *group.Id {
+				return true
+			}
+		}
+		return false
+	}
+
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/services/expand.go
zhengxiexie: fixed.

timdengyun:
[Fix the bug no pod in namespace which has label role:db3 to be select](https://github.com/vmware-tanzu/nsx-operator/pull/85/commits/fa81ff03596b2079143b10e2d8ebb42b89ded831)
for this commit message: shall we rephrase this description,  the issue should be a case that multiple selector in one destination target.

pkg/nsx/services/expand.go
```diff
@@ -0,0 +1,305 @@
...
+					selector.LabelSelector = label
+					selector.Namespace = obj.Namespace
+					finalSelectors = append(finalSelectors, selector)
+				}
+			}
+		}
+	} else if ruleDirection == "OUT" {
+		if len(rule.Destinations) > 0 {
+			for _, target := range rule.Destinations {
+				var namespaceSelectors []client.ListOptions // ResolveNamespace may return multiple namespaces
```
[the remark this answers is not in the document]
timdengyun: I see. when matchlabel matches multiple namespaces, the  ResolveNamespace could return multiple namespaces

timdengyun (approved):

timdengyun (approved):

PR #87, opened 2022-08-16
Add VPCNetworkConfiguration CRD

lxiaopei:


jianjuns (commented):
Should not we define Golang types first? If we use kubebuilder(?),  YAMLs can be auto-generated?

lxiaopei:
yes Jianjun, Thanks. We use kubebuilder, I submitted the changes for Golang types for the CRD, then generate the manifests using controller-gen in this PR https://github.com/vmware-tanzu/nsx-operator/pull/91

PR #88, opened 2022-08-18
Support dynamic NSX version check

timdengyun:
SecurityPolicy service can only be activated from NSX 3.2.0 onwards.
This patch is to add nsx version check and activate SecurityPolicy service dynamically.
As a result, this dynamic NSX version check can support that NSX is upgraded
from 3.1.x to 3.2.x when nsx operator is running.

This patch also reads nsx version only from API request not from cached version,
it's to avoid cached version not updated or mismatched with nsx manager.

### DIFF
--- a/pkg/controllers/securitypolicy_controller.go
+++ b/pkg/controllers/securitypolicy_controller.go
@@ -37,9 +37,10 @@ const (
 )
 
 var (
-	log           = logf.Log.WithName("controller").WithName("securitypolicy")
-	resultNormal  = ctrl.Result{}
-	resultRequeue = ctrl.Result{Requeue: true}
+	log                     = logf.Log.WithName("controller").WithName("securitypolicy")
+	resultNormal            = ctrl.Result{}
+	resultRequeue           = ctrl.Result{Requeue: true}
+	resultRequeueAfter5mins = ctrl.Result{Requeue: true, RequeueAfter: 5 * time.Minute}
 )
 
 // SecurityPolicyReconciler SecurityPolicyReconcile reconciles a SecurityPolicy object
@@ -78,6 +79,15 @@ func (r *SecurityPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return resultNormal, client.IgnoreNotFound(err)
 	}
 
+	// Since SecurityPolicy service can only be activated from NSX 3.2.0 onwards,
+	// So need to check NSX version before starting SecurityPolicy reconcile
+	if !r.Service.NSXClient.NSXCheckVersionForSecurityPolicy() {
+		err := errors.New("NSX version check failed, SecurityPolicy feature is not supported")
+		updateFail(r, &ctx, obj, &err)
+		// if NSX version check fails, it will be put back to reconcile queue and be reconciled after 5 minutes
+		return resultRequeueAfter5mins, nil
+	}
+
 	if obj.ObjectMeta.DeletionTimestamp.IsZero() {
 		metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerUpdateTotal, METRIC_RES_TYPE)
 		if !controllerutil.ContainsFinalizer(obj, util.FinalizerName) {
@@ -99,7 +109,7 @@ func (r *SecurityPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 			err = errors.New("security Policy CR cannot be created in System Namespace")
 			log.Error(err, "", "securitypolicy", req.NamespacedName)
 			updateFail(r, &ctx, obj, &err)
-			return resultNormal, err
+			return resultNormal, nil
 		}
 
 		if err := r.Service.OperateSecurityPolicy(obj); err != nil {
--- a/pkg/nsx/client.go
+++ b/pkg/nsx/client.go
@@ -19,6 +19,10 @@ import (
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/ratelimiter"
 )
 
+const (
+	FeatureSecurityPolicy string = "SECURITY_POLICY"
+)
+
 type Client struct {
 	NsxConfig      *config.NSXOperatorConfig
 	RestConnector  *client.RestConnector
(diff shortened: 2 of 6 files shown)


pkg/controllers/securitypolicy_controller.go
```diff
@@ -78,6 +79,15 @@ func (r *SecurityPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return resultNormal, client.IgnoreNotFound(err)
 	}
 
+	// Since securitypolicy service can only be activated from nsx 3.2.0 onwards,
+	// So need to check nsx version before starting securitypolicy Reconcile
+	if !r.Service.NSXClient.CheckNsxVersion() {
+		err := errors.New("check Nsx Version failed, would retry after 5 minutes")
+		log.Error(err, "check Nsx Version failed, would retry after 5 minutes")
```
dantingl, line 86: It will print error message twice

pkg/nsx/client.go
```diff
@@ -63,18 +67,30 @@ func GetClient(cf *config.NSXOperatorConfig) *Client {
...
 	}
+	nsxVersionChecker := &NSXVersionChecker{
+		cluster:          cluster,
+		versionSupported: false,
+	}
+
+	if nsxVersion.featureSupported(minVersion) {
+		nsxVersionChecker.versionSupported = true
+	} else {
+		log.Info("warning: check nsx version failed", "nsx version", nsxVersion.NodeVersion)
```
dantingl, line 91: `log.Info("warning:` -> log.Error ?

dantingl, line 91: `check nsx version failed`: could we print current NSX version and the required version

pkg/controllers/securitypolicy_controller.go
```diff
@@ -78,6 +79,15 @@ func (r *SecurityPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return resultNormal, client.IgnoreNotFound(err)
 	}
 
+	// Since securitypolicy service can only be activated from nsx 3.2.0 onwards,
+	// So need to check nsx version before starting securitypolicy Reconcile
+	if !r.Service.NSXClient.CheckNsxVersion() {
+		err := errors.New("check Nsx Version failed, would retry after 5 minutes")
```
dantingl, line 85: `would retry after 5 minutes`: I don't think we need to say retry 5 minutes

zhengxiexie: how about turning it into a function to support various interval time?

pkg/nsx/client.go
```diff
@@ -63,18 +67,30 @@ func GetClient(cf *config.NSXOperatorConfig) *Client {
...
 	}
+
 	queryClient := search.NewQueryClient(restConnector(cluster))
 	groupClient := domains.NewGroupsClient(restConnector(cluster))
 	securityClient := domains.NewSecurityPoliciesClient(restConnector(cluster))
 	ruleClient := security_policies.NewRulesClient(restConnector(cluster))
 	nsxChecker := &NSXHealthChecker{
 		cluster: cluster,
 	}
+	nsxVersionChecker := &NSXVersionChecker{
```
TaoZou1, line 83: Line 83-93, Could you just call client.CheckNsxVersion() after Client created?

pkg/nsx/client.go
```diff
@@ -83,5 +99,29 @@ func GetClient(cf *config.NSXOperatorConfig) *Client {
...
+	}
+
+	nsxVersion, err := client.NSXVerChecker.cluster.GetVersion()
+	if err != nil {
+		log.Error(err, "get version error")
+		return false
+	}
+	err = nsxVersion.Validate()
+	if err != nil {
+		log.Error(err, "validate version error")
```
TaoZou1, line 118: print nsxVersion and minVersion

pkg/nsx/client.go
```diff
@@ -83,5 +99,29 @@ func GetClient(cf *config.NSXOperatorConfig) *Client {
 		SecurityClient: securityClient,
 		RuleClient:     ruleClient,
 		NSXChecker:     *nsxChecker,
+		NSXVerChecker:  *nsxVersionChecker,
+	}
+}
+
+func (client *Client) CheckNsxVersion() bool {
```
TaoZou1, line 106: how about add a parameter minVersion for CheckNsxVersion()?  using a global variable by nsxVersion.featureSupported maybe not easy to change later if each controller has its own minVersion

pkg/nsx/client.go
```diff
@@ -58,30 +66,55 @@ func GetClient(cf *config.NSXOperatorConfig) *Client {
...
+	}
+	err = nsxVersion.Validate()
+	if err != nil {
+		log.Error(err, "validate version error")
+		return false
+	}
+
+	if !nsxVersion.featureSupported(FeatureSecurityPolicy) {
+		err = errors.New("NSX version check failed")
+		log.Error(err, "Securitypolicy feature is not supported", "current version", nsxVersion.NodeVersion, "required version", nsx320Version)
```
dantingl, line 115: `Securitypolicy` -> SecurityPolicy?

pkg/nsx/client.go
```diff
@@ -58,30 +66,55 @@ func GetClient(cf *config.NSXOperatorConfig) *Client {
...
 		NsxConfig:      cf,
 		RestConnector:  restConnector(cluster),
 		QueryClient:    queryClient,
 		GroupClient:    groupClient,
 		SecurityClient: securityClient,
 		RuleClient:     ruleClient,
 		NSXChecker:     *nsxChecker,
+		NSXVerChecker:  *nsxVersionChecker,
+	}
+	nsxClient.NSXCheckVersionForSecurityPolicy()
```
dantingl, line 92: Add comments why not exiting here, or print a warning log?

timdengyun: Done

pkg/nsx/client.go
timdengyun: Done

dantingl (approved):

dantingl (approved):

PR #89, opened 2022-08-18
Support dynamic NSX version check

timdengyun:
SecurityPolicy service can only be activated from NSX 3.2.0 onwards.
This patch is to add nsx version check and activate SecurityPolicy service dynamically.
As a result, this dynamic NSX version check can support that NSX is upgraded
from 3.1.x to 3.2.x when nsx operator is running.

### DIFF
--- a/pkg/controllers/securitypolicy_controller.go
+++ b/pkg/controllers/securitypolicy_controller.go
@@ -28,7 +28,8 @@ import (
 )
 
 var (
-	log = logf.Log.WithName("controller").WithName("securitypolicy")
+	log                     = logf.Log.WithName("controller").WithName("securitypolicy")
+	resultRequeueAfter5mins = ctrl.Result{Requeue: true, RequeueAfter: 5 * time.Minute}
 )
 
 const (
@@ -43,6 +44,11 @@ type SecurityPolicyReconciler struct {
 	Service *services.SecurityPolicyService
 }
 
+func updateFail(r *SecurityPolicyReconciler, c *context.Context, o *v1alpha1.SecurityPolicy, e *error) {
+	r.setSecurityPolicyReadyStatusFalse(c, o, e)
+	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerUpdateFailTotal, METRIC_RES_TYPE)
+}
+
 func (r *SecurityPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	obj := &v1alpha1.SecurityPolicy{}
 	log.Info("reconciling securitypolicy CR", "securitypolicy", req.NamespacedName)
@@ -51,6 +57,15 @@ func (r *SecurityPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	if err := r.Client.Get(ctx, req.NamespacedName, obj); err != nil {
 		log.Error(err, "unable to fetch security policy CR", "req", req.NamespacedName)
 		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// Since securitypolicy service can only be activated from nsx 3.2.0 onwards,
+	// So need to check nsx version before starting securitypolicy Reconcile
+	if !r.Service.NSXClient.NSXCheckVersion() {
+		err := errors.New("NSX version check failed")
+		log.Error(err, "SecurityPolicy feature is not supported")
+		updateFail(r, &ctx, obj, &err)
+		return resultRequeueAfter5mins, nil
 	}
 
 	if obj.ObjectMeta.DeletionTimestamp.IsZero() {
--- a/pkg/nsx/client.go
+++ b/pkg/nsx/client.go
@@ -27,14 +27,18 @@ type Client struct {
 	SecurityClient domains.SecurityPoliciesClient
 	RuleClient     security_policies.RulesClient
 	NSXChecker     NSXHealthChecker
+	NSXVerChecker  NSXVersionChecker
 }
 
-var (
-	minVersion = [3]int64{3, 2, 0}
-)
+var minVersion = [3]int64{3, 2, 0}
 
 type NSXHealthChecker struct {
 	cluster *Cluster
+}
+
+type NSXVersionChecker struct {
+	cluster          *Cluster
+	versionSupported bool
 }
 
 func (ck *NSXHealthChecker) CheckNSXHealth(req *http.Request) error {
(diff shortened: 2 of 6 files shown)


pkg/controllers/securitypolicy_controller.go
```diff
@@ -53,6 +59,15 @@ func (r *SecurityPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	// Since securitypolicy service can only be activated from nsx 3.2.0 onwards,
+	// So need to check nsx version before starting securitypolicy Reconcile
+	if !r.Service.NSXClient.CheckNsxVersion() {
+		err := errors.New("check Nsx Version failed, would retry later")
+		log.Error(err, "check Nsx Version failed, would retry later")
```
timdengyun, line 66: @dantingl  It will print error message twice
for this question, I feel it won't print twice, it just print for each CR once

dantingl: I mean it will print "check Nsx Version failed, would retry later" twice.

pkg/controllers/securitypolicy_controller.go
```diff
@@ -71,7 +86,7 @@ func (r *SecurityPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 			err = errors.New("Security Policy CR cannot be created in System Namespace")
 			log.Error(err, "failed to create security policy CR", "securitypolicy", req.NamespacedName)
 			metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerUpdateFailTotal, METRIC_RES_TYPE)
-			return ctrl.Result{}, err
+			return ctrl.Result{}, nil
```
dantingl, line 89: Why update code here?

pkg/nsx/client.go
```diff
@@ -27,16 +27,20 @@ type Client struct {
...
-)
+var minVersion = [3]int64{3, 2, 0}
 
 type NSXHealthChecker struct {
 	cluster *Cluster
 }
 
+type NSXVersionChecker struct {
+	cluster          *Cluster
+	versionSupported bool
```
dantingl, line 41: Later we should change `versionSupported` and `minVersion` per features

pkg/controllers/securitypolicy_controller.go
dantingl, line 66: `check Nsx Version failed, would retry later` -> NSX version check failed, SecurityPolicy feature is not supported

pkg/controllers/securitypolicy_controller.go
timdengyun: I only revert it back, it's changed based on main branch's change, since it's a normal case, so shouldn't put nil return. otherwise, it will be re queued again.

pkg/controllers/securitypolicy_controller.go
```diff
@@ -53,6 +59,15 @@ func (r *SecurityPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	// Since securitypolicy service can only be activated from nsx 3.2.0 onwards,
+	// So need to check nsx version before starting securitypolicy Reconcile
+	if !r.Service.NSXClient.CheckNsxVersion() {
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/client.go
```diff
@@ -83,5 +100,29 @@ func GetClient(cf *config.NSXOperatorConfig) *Client {
...
+		log.Error(err, "get version error")
+		return false
+	}
+	err = nsxVersion.Validate()
+	if err != nil {
+		log.Error(err, "validate version error")
+		return false
+	}
+
+	if !nsxVersion.featureSupported(minVersion) {
```
heypnus, line 123: Not related to your patch but I doubt that if the func `featureSupported` can work as expected... What cases did you test? Did you verify it with real NSX or mocked NSX?

heypnus (changes requested):

pkg/controllers/securitypolicy_controller.go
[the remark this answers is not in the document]
zhengxiexie: @timdengyun  By the way, I think there is no need to check in every reconcile, how about cache a global variable, if it returns true and skip it in following reconciles.

heypnus (commented):
The logic LGTM.

pkg/controllers/securitypolicy_controller.go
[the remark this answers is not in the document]
timdengyun: Thanks xie. In fact, we have a flag var versionSupported stored inside if client
So, when checkNsxVersion function is called, it will check this flag firstly, it will work as a global var check.

pkg/nsx/client.go
```diff
@@ -58,30 +62,64 @@ func GetClient(cf *config.NSXOperatorConfig) *Client {
...
+		cluster:          cluster,
+		versionSupported: false,
+	}
+
+	if nsxVersion.featureSupported(minVersion) {
+		nsxVersionChecker.versionSupported = true
+	} else {
+		err := errors.New("NSX version check failed")
+		log.Error(err, "SecurityPolicy feature is not supported", "current version", nsxVersion.NodeVersion, "required version", minVersion)
+	}
```
dantingl, line 83: Why it print the similar log with line 96-99?

pkg/nsx/client.go
```diff
@@ -58,30 +62,64 @@ func GetClient(cf *config.NSXOperatorConfig) *Client {
...
 	nsxChecker := &NSXHealthChecker{
 		cluster: cluster,
 	}
-	return &Client{
+	nsxVersionChecker := &NSXVersionChecker{
+		cluster:          cluster,
+		versionSupported: false,
+	}
+
+	if nsxVersion.featureSupported(minVersion) {
```
dantingl, line 78: Where did you define `nsxVersion`?

pkg/nsx/client.go
timdengyun: They are redundant code and are removed. Thanks for reminder.

pkg/controllers/securitypolicy_controller.go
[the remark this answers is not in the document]
zhengxiexie: ok

dantingl (approved):

TaoZou1 (approved):

PR #96, opened 2022-08-24
Enhance nsx version check per feature

timdengyun:
This patch is to:
1. enhance nsx version check per feature.
2. read nsx version only from API request not from cached version,
   this is to avoid cached version not updated or mismatched with nsx manager.

### DIFF
--- a/pkg/nsx/cluster.go
+++ b/pkg/nsx/cluster.go
@@ -229,9 +229,6 @@ func (cluster *Cluster) Health() ClusterHealth {
 }
 
 func (cluster *Cluster) GetVersion() (*NsxVersion, error) {
-	if len(nsxVersion.NodeVersion) > 0 {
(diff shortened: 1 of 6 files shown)


dantingl (approved):

pkg/nsx/cluster.go
```diff
@@ -229,9 +229,6 @@ func (cluster *Cluster) Health() ClusterHealth {
 }
 
 func (cluster *Cluster) GetVersion() (*NsxVersion, error) {
-	if len(nsxVersion.NodeVersion) > 0 {
```
timdengyun, line 232: Remove local cached NodVersion  and read value by API request each time. This is to avoid cached value not changed even nsx manger is upgraded to new version.

heypnus (approved):

TaoZou1 (approved):

PR #97, opened 2022-08-24
add subnet/subnetset CRD

jwsui:
Add subnet and subnetset api, and generate manifests with kubebuilder.


pkg/apis/v1alpha1/subnet_types.go
```diff
@@ -0,0 +1,75 @@
...
+// SubnetSpec defines the desired state of Subnet
+type SubnetSpec struct {
+	// Size of subnet based upon estimated workload count.
+	// Defaults to 64.
+	// +kubebuilder:default:=64
+	IPV4SubnetSize string `json:"ipv4SubnetSize"`
+	// Access mode of subnet, accessible only from within VPC or from outside VPC.
+	// Defaults to private.
+	// +kubebuilder:default:=private
+	AccessMode string `json:"accessMode"`
```
lxiaopei, line 27: Could we also add "+kubebuilder:validation:Enum" to limit it can only be public or private?

pkg/apis/v1alpha1/subnet_types.go
```diff
@@ -0,0 +1,75 @@
...
+	SubnetReady    SubnetStatusCondition = "Ready"
+	SubnetNotReady SubnetStatusCondition = "NotReady"
+)
+
+// SubnetSpec defines the desired state of Subnet
+type SubnetSpec struct {
+	// Size of subnet based upon estimated workload count.
+	// Defaults to 64.
+	// +kubebuilder:default:=64
+	IPV4SubnetSize string `json:"ipv4SubnetSize"`
```
lxiaopei, line 23: maybe use "int" for the size?

pkg/apis/v1alpha1/subnet_types.go
```diff
@@ -0,0 +1,75 @@
...
+	// Access mode of subnet, accessible only from within VPC or from outside VPC.
+	// Defaults to private.
+	// +kubebuilder:default:=private
+	AccessMode string `json:"accessMode"`
+	// Subnet CIDR.
+	IPAddresses string `json:"ipAddresses,omitempty"`
```
lxiaopei, line 29: From NSX-T VPC subnet yaml, this field is an array with maxItems:2. Could you try the nsx subnet API that if we can specify 2 cidrs? If yes, we can add the below validations.
        // +kubebuilder:validation:MinItems=0
	// +kubebuilder:validation:MaxItems=2
ip_addresses:
      type: array
      maxItems: 2
      items:
        type: string
        format: ip-cidr-block
      title: CIDR

pkg/apis/v1alpha1/subnet_types.go
```diff
@@ -0,0 +1,75 @@
...
+	Type SubnetStatusCondition `json:"type"`
+	// Status defines status of condition type, True or False.
+	Status corev1.ConditionStatus `json:"status"`
+	// Reason shows a brief reason of condition.
+	Reason string `json:"reason,omitempty"`
+	// Message shows a human readable message about condition.
+	Message string `json:"message,omitempty"`
+}
+
+// SubnetStatus defines the observed state of Subnet
```
lxiaopei, line 44: Add "." at the end of the sentence.

pkg/apis/v1alpha1/subnetset_types.go
```diff
@@ -0,0 +1,77 @@
...
+	SubnetSetReady    SubnetSetStatusCondition = "Ready"
+	SubnetSetNotReady SubnetSetStatusCondition = "NotReady"
+)
+
+// SubnetSetSpec defines the desired state of SubnetSet
+type SubnetSetSpec struct {
+	// Size of subnet based upon estimated workload count.
+	// Defaults to 64.
+	// +kubebuilder:default:=64
+	IPV4SubnetSize string `json:"ipv4SubnetSize"`
```
lxiaopei, line 23: use int and add omitEmpty?

pkg/apis/v1alpha1/subnetset_types.go
```diff
@@ -0,0 +1,77 @@
...
+	// Access mode of subnet, accessible only from within VPC or from outside VPC.
+	// Defaults to private.
+	// +kubebuilder:default:=private
+	AccessMode string `json:"accessMode"`
```
lxiaopei, line 27: add omitEmpty?

pkg/apis/v1alpha1/subnet_types.go
lxiaopei: this is an optional field.

pkg/apis/v1alpha1/subnet_types.go
jwsui: Thanks, also add `isolated` in enum, not sure whether `isolate` is needed.

pkg/apis/v1alpha1/subnet_types.go
jwsui: Thanks, done.

pkg/apis/v1alpha1/subnet_types.go
jwsui: Thanks and fixed.

pkg/apis/v1alpha1/subnetset_types.go
jwsui: Thanks and fixed.

build/yaml/crd/nsx.vmware.com_subnets.yaml
```diff
@@ -0,0 +1,106 @@
...
+            description: SubnetSpec defines the desired state of Subnet
+            properties:
+              accessMode:
+                default: private
+                description: Access mode of subnet, accessible only from within VPC
+                  or from outside VPC. Defaults to private.
+                enum:
+                - private
+                - public
+                - isolated
```
dantingl, line 46: Do we have isolated type in the first version?

build/yaml/crd/nsx.vmware.com_subnets.yaml
```diff
@@ -0,0 +1,106 @@
...
+                type: string
+              ipAddresses:
+                description: Subnet CIDR.
+                items:
+                  type: string
+                maxItems: 2
```
dantingl, line 52: Why maxItems is 2, for v4 and v6?

build/yaml/crd/nsx.vmware.com_subnets.yaml
```diff
@@ -0,0 +1,106 @@
...
+                minItems: 0
+                type: array
+              ipv4SubnetSize:
+                default: 64
```
dantingl, line 56: The default subnet prefix is /28, which is different with size 64 here @lxiaopei

build/yaml/crd/nsx.vmware.com_subnets.yaml
jwsui: > Do we have isolated type in the first version?

Discussed with @lxiaopei , we are not sure in which scenario we'll use `isolated`. And we just list `isolated` here and call for comments about this field.

build/yaml/crd/nsx.vmware.com_subnets.yaml
jwsui: > The default subnet prefix is /28, which is different with size 64 here @lxiaopei

If the default prefix is /28, then default `ipv4SubnetSize` should be 16? @lxiaopei , could you help confirm?

jwsui: It seems that the default value is still 64 on api doc of nsx-t 4.1.0, not sure if I'm using the right version.

build/yaml/crd/nsx.vmware.com_subnets.yaml
jwsui: Not sure about this, @lxiaopei do you know about it?

build/yaml/crd/nsx.vmware.com_subnets.yaml
lxiaopei: For VPC Subnet API, the default subnet size is 64.  
I set the default subnet prefix in VPC NetworkConfiguration CR as 28 according to current T1 NSXNetworkConfig CR.
Do we need to keep them the same value?

build/yaml/crd/nsx.vmware.com_subnets.yaml
lxiaopei: yes, for ipv4 and ipv6.

build/yaml/crd/nsx.vmware.com_subnets.yaml
lxiaopei: Also changed default subnet prefix in VPC NetworkConfiguration CR  to 26 to be consistent with the default subnet size 64 here.

pkg/apis/v1alpha1/subnet_types.go
```diff
@@ -0,0 +1,78 @@
...
+	SubnetReady    SubnetStatusCondition = "Ready"
+	SubnetNotReady SubnetStatusCondition = "NotReady"
+)
+
+// SubnetSpec defines the desired state of Subnet
+type SubnetSpec struct {
+	// Size of subnet based upon estimated workload count.
+	// Defaults to 64.
+	// +kubebuilder:default:=64
+	IPV4SubnetSize int `json:"ipv4SubnetSize,omitempty"`
```
jianjuns, line 23: IPv4?

pkg/apis/v1alpha1/subnet_types.go
```diff
@@ -0,0 +1,78 @@
...
+	// Access mode of subnet, accessible only from within VPC or from outside VPC.
```
jianjuns, line 24: subnet -> Subnet (assuming we use upper camel case for CRD kinds).

pkg/apis/v1alpha1/subnet_types.go
```diff
@@ -0,0 +1,78 @@
...
+	// Defaults to private.
+	// +kubebuilder:default:=private
+	// +kubebuilder:validation:Enum=private;public;isolated
+	AccessMode string `json:"accessMode,omitempty"`
```
jianjuns, line 28: Should we define a type for AccessMode?

pkg/apis/v1alpha1/subnet_types.go
```diff
@@ -0,0 +1,78 @@
...
+	// Subnet CIDR.
+	// +kubebuilder:validation:MinItems=0
+	// +kubebuilder:validation:MaxItems=2
+	IPAddresses []string `json:"ipAddresses,omitempty"`
```
jianjuns, line 32: Why CIDR is a slice? Should not it be a single string (assuming we use two fields for v4 and v6 respectively)? What is the definition in NSX-T API?

pkg/apis/v1alpha1/subnet_types.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type SubnetStatusCondition string
+
+const (
+	SubnetReady    SubnetStatusCondition = "Ready"
+	SubnetNotReady SubnetStatusCondition = "NotReady"
```
jianjuns, line 15: Do we need a condition to express (permanent) failure?

pkg/apis/v1alpha1/subnet_types.go
```diff
@@ -0,0 +1,78 @@
...
+	// Reason shows a brief reason of condition.
+	Reason string `json:"reason,omitempty"`
+	// Message shows a human readable message about condition.
+	Message string `json:"message,omitempty"`
+}
+
+// SubnetStatus defines the observed state of Subnet.
+type SubnetStatus struct {
+	// Logical switch ID.
+	LsID       string            `json:"lsID"`
```
jianjuns, line 50: This should be NSX-T Subnet path? If so, name it NSXTPath?

pkg/apis/v1alpha1/subnetset_types.go
```diff
@@ -0,0 +1,78 @@
...
+	// Status defines status of condition type, True or False.
+	Status corev1.ConditionStatus `json:"status"`
+	// Reason shows a brief reason of condition.
+	Reason string `json:"reason,omitempty"`
+	// Message shows a human readable message about condition.
+	Message string `json:"message,omitempty"`
+}
+
+// SubnetItem defines subnet items of SubnetSet.
+type SubnetItem struct {
```
jianjuns, line 44: How about "SubnetInfo" or "SubnetStatusItem"?

pkg/apis/v1alpha1/subnetset_types.go
```diff
@@ -0,0 +1,78 @@
...
+	Type SubnetSetStatusCondition `json:"type"`
+	// Status defines status of condition type, True or False.
+	Status corev1.ConditionStatus `json:"status"`
+	// Reason shows a brief reason of condition.
+	Reason string `json:"reason,omitempty"`
+	// Message shows a human readable message about condition.
+	Message string `json:"message,omitempty"`
+}
+
+// SubnetItem defines subnet items of SubnetSet.
```
jianjuns, line 43: subnet -> Subnet

jianjuns: defines the observed state of a single Subnet of a SubnetSet

pkg/apis/v1alpha1/subnet_types.go
jwsui: Thanks for reminder, fixed.

pkg/apis/v1alpha1/subnetset_types.go
jwsui: Thanks and fixed.

pkg/apis/v1alpha1/subnet_types.go
lxiaopei: We use the LS instead of Subnet due to the restore requirements for PodVM on WCP.
But we can change to use path instead of ID.

pkg/apis/v1alpha1/subnet_types.go
lxiaopei: NSX Subnet API define ip_addresses as array for v4 and v6 as below
   ip_addresses:
      type: array
      maxItems: 2
      items:
        type: string
        format: ip-cidr-block
      title: CIDR

jianjuns: Ok. Then let us follow NSX-T API. BTW, what is the use of "title" in the NSX-T API spec?

pkg/apis/v1alpha1/subnet_types.go
```diff
@@ -0,0 +1,78 @@
...
+	// Size of Subnet based upon estimated workload count.
+	// Defaults to 64.
+	// +kubebuilder:default:=64
+	IPv4SubnetSize int `json:"ipv4SubnetSize,omitempty"`
+	// Access mode of Subnet, accessible only from within VPC or from outside VPC.
+	// Defaults to private.
+	// +kubebuilder:default:=private
+	// +kubebuilder:validation:Enum=private;public;isolated
+	AccessMode string `json:"accessMode,omitempty"`
+	// Subnet CIDR.
```
jianjuns, line 29: CIDR -> CIDRS, as it is a slice.

pkg/apis/v1alpha1/subnet_types.go
lxiaopei: I think it is a summary of the field. We can see in nsx policy api yaml files , e.g.
http://build-squid.eng.vmware.com/build/mts/release/bora-20375668/publish/linux-centos64-64/openapi/nsx_policy_api.yaml

pkg/apis/v1alpha1/subnet_types.go
```diff
@@ -0,0 +1,78 @@
+/* Copyright © 2021 VMware, Inc. All Rights Reserved.
```
lxiaopei, line 1: rebase the latest code to update the license.

pkg/apis/v1alpha1/subnet_types.go
```diff
@@ -0,0 +1,78 @@
...
+	Reason string `json:"reason,omitempty"`
+	// Message shows a human readable message about condition.
+	Message string `json:"message,omitempty"`
+}
+
+// SubnetStatus defines the observed state of Subnet.
+type SubnetStatus struct {
+	// Logical switch ID.
+	LsID       string            `json:"lsID"`
+	SubnetCIDR string            `json:"subnetCIDR"`
```
lxiaopei, line 51: based on above , shall we also change to ipAddresses with slice type?

pkg/apis/v1alpha1/subnet_types.go
lxiaopei: I got the point. The LS ID is needed in the NSX Subnet properties. In the CRD, we can change it to Subnet path. Could we change it to nsxSubnetPath? in VPC status, we have nsxVPCPath.

pkg/apis/v1alpha1/subnet_types.go
lxiaopei: I see for SecurityPolicy CR, it only has "Ready" condition with values: true/false/unknown.  Could we also use the same way for all other CRs?

pkg/apis/v1alpha1/subnet_types.go
jwsui: Thanks and defined AccessMode type.

pkg/apis/v1alpha1/subnet_types.go
jwsui: Thanks Xiaopei for the explanation, and change it to nsxtSubnetPath.

pkg/apis/v1alpha1/subnet_types.go
jwsui: Thanks and fixed.

pkg/apis/v1alpha1/subnet_types.go
jwsui: Thanks and done.

pkg/apis/v1alpha1/subnet_types.go
jwsui: Thanks and done.

pkg/apis/v1alpha1/subnet_types.go
```diff
@@ -0,0 +1,78 @@
...
+	Status corev1.ConditionStatus `json:"status"`
+	// Reason shows a brief reason of condition.
+	Reason string `json:"reason,omitempty"`
+	// Message shows a human readable message about condition.
+	Message string `json:"message,omitempty"`
+}
+
+// SubnetStatus defines the observed state of Subnet.
+type SubnetStatus struct {
+	NSXTSubnetPath string            `json:"nsxtSubnetPath"`
```
heypnus, line 50: I discussed this field with Jianjun and Xiaopei, we prefer to change it to `NSXPath`, then all CRDs can use the same field name, which looks more common. What do you think?

jwsui: Sure, I'm ok with this as there is no other nsx resource path in this crd. Change it to `NSXPath` in `SubnetStatus` and `SubnetsetStatus`. Thanks.

jianjuns: I think NSXAPIPath or NSXPolicyAPIPath is clearer. But let us choose one and use that for all CRDs.

jwsui: Ok, I'll align it with the name in struct VPCStatus(guess that it'd be `NSXPolicyPath`).

https://github.com/vmware-tanzu/nsx-operator/pull/106/files#r971438788

pkg/apis/v1alpha1/subnet_types.go
```diff
@@ -0,0 +1,78 @@
...
+	// Status defines status of condition type, True or False.
+	Status corev1.ConditionStatus `json:"status"`
+	// Reason shows a brief reason of condition.
+	Reason string `json:"reason,omitempty"`
+	// Message shows a human readable message about condition.
+	Message string `json:"message,omitempty"`
+}
+
+// SubnetStatus defines the observed state of Subnet.
+type SubnetStatus struct {
```
jianjuns, line 49: @lxiaopei @heypnus : could you make sure Status structs use a common name (I now prefer "NSXResourcePath") for the resource path in NSX-T API.

And could you define a common Condition struct to be shared by all CRDs? I saw all CRD Condition definitions are the same. Let me know if I am wrong, and you believe we may have different condition definitions for different CRDs.

pkg/apis/v1alpha1/subnet_types.go
jianjuns: In our implementation, do we have there different cases: 1) CRD validation or NSX-T API validation failed (and we will not retry for these cases); 2) NSX-T realization failed, but we are retrying; 3) NSX-T realization failed and we stop re-trying?

And "Ready" and "NotReady" are just opposites. Why we need both? Maybe "NotReady" actually means "Failure" (and we no longer re-try)?

pkg/apis/v1alpha1/subnet_types.go
lxiaopei: Yes, We define a common Condition struct in this patch https://github.com/vmware-tanzu/nsx-operator/pull/108, after this patch be merged, all CRDs can use the same Condition. @jianjuns   Thanks for the reminder.

For the path under Status, all the CRDs will use NSXResourcePath.

pkg/apis/v1alpha1/subnet_types.go
lxiaopei: I think the 3 cases above are all there.
But I think we can only have "Ready" condition type in common file https://github.com/vmware-tanzu/nsx-operator/pull/108/, the value can be True/False/Unknown.  Do you think it is OK? @jianjuns 
For specific CRDs, they can define their own ConditionType if needed in their _types.go.

jianjuns: Ok let us discuss there.

pkg/apis/v1alpha1/subnet_types.go
jwsui: Thanks for clarifying, updated to use `NSXResourcePath`.

jwsui:
```
In our implementation, do we have there different cases: 1) CRD validation or NSX-T API validation failed (and we will not retry for these cases); 2) NSX-T realization failed, but we are retrying; 3) NSX-T realization failed and we stop re-trying?

And "Ready" and "NotReady" are just opposites. Why we need both? Maybe "NotReady" actually means "Failure" (and we no longer re-try)?
```

Indeed, we don't need "Ready" and "NotReady" both. As pr [#108](https://github.com/vmware-tanzu/nsx-operator/pull/108) is merged, now I rebase the commit and use the common Condition struct instead in which we only define "Ready". Thanks Jianjun.

pkg/apis/v1alpha1/subnet_types.go
```diff
@@ -0,0 +1,59 @@
...
+// SubnetSpec defines the desired state of Subnet.
+type SubnetSpec struct {
+	// Size of Subnet based upon estimated workload count.
+	// Defaults to 64.
+	// +kubebuilder:default:=64
+	IPv4SubnetSize int `json:"ipv4SubnetSize,omitempty"`
+	// Access mode of Subnet, accessible only from within VPC or from outside VPC.
+	// Defaults to private.
+	// +kubebuilder:default:=private
+	// +kubebuilder:validation:Enum=private;public;isolated
```
lxiaopei, line 21: I do not think any use case for isolated subnet for the first release. @jianjuns  could we just use private/public subnet type? although NSX Subnet API supports private/public/isolated.

lxiaopei (commented):
only 1 minor comments, others LGTM

pkg/apis/v1alpha1/subnet_types.go
jianjuns: Yes, for manual defined CRD we can remove "isolated" for now.

jwsui: OK, removed `isolated`, thanks.

jianjuns (commented):
LGTM

lxiaopei (approved):
LGTM

PR #104, opened 2022-09-07
enhance log message for nsx version check

timdengyun:


pkg/nsx/client.go
```diff
@@ -89,10 +89,11 @@ func GetClient(cf *config.NSXOperatorConfig) *Client {
 		NSXChecker:     *nsxChecker,
 		NSXVerChecker:  *nsxVersionChecker,
 	}
-
+	// NSX version check will be restarted during SecurityPolicy reconcile
+	// So, it's unnecessary to exit even if failed in the first time
 	if !nsxClient.NSXCheckVersionForSecurityPolicy() {
```
TaoZou1, line 94: since it doesn't return nil here, do we still need to check the return value of nsxClient.NSXCheckVersionForSecurityPolicy()? the log inside nsxClient.NSXCheckVersionForSecurityPolicy() is enough. The log here seems redundant

timdengyun: Agreed. The checking return value is a little bit redundant, since log messages already are printed inside 
NSXCheckVersionForSecurityPolicy().

The solution for this redundancy is that we remove log message here or use the different log message. My choice is the latter one.  That is why I print a different log message. In fact, this approach is the same with one in patch for 4.0 and main:
https://github.com/vmware-tanzu/nsx-operator/pull/99/files#diff-294ccb5c11de7239479925afdfe26cda5336127b3e2badf2f076e4965bbb8894

TaoZou1 (approved):

PR #108, opened 2022-09-16
extract condition struct

jwsui:
Extract CRD condition struct so that all CRDs could share the same condition struct. Update CRD SecurityPolicy to use the common condition struct instead of defining by itselt.


pkg/apis/v1alpha1/condition.go
```diff
@@ -0,0 +1,22 @@
...
+import corev1 "k8s.io/api/core/v1"
+
+type ConditionType string
+
+const (
+	Ready    ConditionType = "Ready"
+	NotReady ConditionType = "NotReady"
+)
+
+// Condition defines condition of CRD.
```
heypnus, line 12: How about `Condition defines the condition of the custom resource.`? Because I think it's a condition of CR, not CRD.

jwsui: Indeed, fixed.

pkg/apis/v1alpha1/condition.go
```diff
@@ -0,0 +1,22 @@
+package v1alpha1
+
+import corev1 "k8s.io/api/core/v1"
+
+type ConditionType string
+
+const (
+	Ready    ConditionType = "Ready"
+	NotReady ConditionType = "NotReady"
```
lxiaopei, line 9: Ready status has value "True or False", do we still need "NotReady" Status?

pkg/apis/v1alpha1/condition.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+// Condition defines condition of custom resource.
+type Condition struct {
+	// Type defines condition type.
+	Type ConditionType `json:"type"`
+	// Status defines status of condition type, True or False.
+	Status corev1.ConditionStatus `json:"status"`
+	// Reason shows a brief reason of condition.
```
lxiaopei, line 18: Could we also add this field?

// Last time the condition transitioned from one status to another.
	// This should be when the underlying condition changed. If that is not known, then using the time when
	// the API field changed is acceptable.
	// +required
	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty"`

pkg/apis/v1alpha1/condition.go
```diff
@@ -0,0 +1,22 @@
...
+const (
+	Ready    ConditionType = "Ready"
+	NotReady ConditionType = "NotReady"
+)
+
+// Condition defines condition of custom resource.
+type Condition struct {
+	// Type defines condition type.
+	Type ConditionType `json:"type"`
+	// Status defines status of condition type, True or False.
```
lxiaopei, line 16: Status of the condition, one of True, False, Unknown.

pkg/apis/v1alpha1/securitypolicy_types.go
```diff
@@ -110,31 +110,9 @@ type SecurityPolicyPort struct {
...
-}
-
-// SecurityPolicyStatusCondition is an aspect of operator state.
-type SecurityPolicyStatusCondition string
-
-const (
-	// Ready indicates whether NSX Operator was able to successfully make
-	// a PATCH call to create/update Security Policy on NSX associated
-	// with the user created Security Policy CR
-	SecurityPolicyReady SecurityPolicyStatusCondition = "Ready"
```
lxiaopei, line 135: SecurityPolicyReady  is used in securitypolicy_controller.go, if removed here, need to change that file.

pkg/apis/v1alpha1/condition.go
jwsui: Indeed, removed `NotReady` as we don't need it anymore, thanks.

pkg/apis/v1alpha1/condition.go
jwsui: Added, but a minor question, we have kubebuilder label `+required` which means it's a a required field while we also have `omitempty` in json tag which indicate that it's option. Do we have conflict here？

pkg/apis/v1alpha1/condition.go
jwsui: Thanks, fixed.

pkg/apis/v1alpha1/securitypolicy_types.go
jwsui: My bad, didn't notice that we already have controller code for securitypolicy CRD. Updated `securitypolicy_controller.go` and `securitypolicy_controller_test.go`, thanks for reminder.

pkg/apis/v1alpha1/condition.go
lxiaopei: sorry, I think we can set // +optional for it.

pkg/apis/v1alpha1/condition.go
```diff
@@ -0,0 +1,29 @@
+package v1alpha1
```
lxiaopei, line 1: could we rename the file as condition_types.go?

lxiaopei (commented):
Did you try to generate binary using the changed files?

pkg/apis/v1alpha1/condition.go
jwsui: sure, fixed.

jwsui:
> Did you try to generate binary using the changed files?

Yes, generate binary successfully by `make build`.

jianjuns:
Please make sure the commit message contains enough information for readers to understand the change. I changed the PR title to: "Extract common CRD Status Condition struct"

jwsui:
> Please make sure the commit message contains enough information for readers to understand the change. I changed the PR title to: "Extract common CRD Status Condition struct"

Thanks Jianjun, updated the commit message.

lxiaopei:
LGTM

lxiaopei (approved):
LGTM

jwsui:
PR merged, thanks for help review.

PR #109, opened 2022-09-16
Enhance log message format

timdengyun:
This patch is to enhance nsx operator logger function. The new logger system uses go-logr lib that consolidates zap logger with more user-friendly info output.
Also, thew new logger could convert epoch timestamp to human-readable timestamp in the log message.

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -1,0 +1,66 @@
+package logger
+
+import (
+	"flag"
+	"time"
+
+	"github.com/go-logr/logr"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	zapcr "sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/config"
+)
+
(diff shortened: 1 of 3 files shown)


pkg/logger/logger.go
```diff
@@ -0,0 +1,66 @@
...
+	customCallerEncoder = func(caller zapcore.EntryCaller, enc zapcore.PrimitiveArrayEncoder) {
+		enc.AppendString(caller.TrimmedPath())
+	}
+)
+
+func init() {
+	flag.IntVar(&logLevel, "log-level", 0, "Use zap-core log system.")
+}
+
+func ZapLogger(cf *config.NSXOperatorConfig) logr.Logger {
```
TaoZou1, line 31: How about change to InitLogger(), we may use other log framework

timdengyun: That's a good suggestion. Right now, this func is specific for ZapLogger.  these code are corss ported from main branch. So, I feel that we could keep this func name as it is at this moment until main branch is changed later. We can change in v3.2.1 branch accordingly.

heypnus (approved):

PR #110, opened 2022-09-19
Fix bug when there is no security policy filtered in garbage collector

zhengxiexie:
Process exits reporting panic: runtime error: index out of range [0] with length 0.

### DIFF
--- a/pkg/nsx/services/firewall.go
+++ b/pkg/nsx/services/firewall.go
@@ -153,16 +153,29 @@ func (service *SecurityPolicyService) DeleteSecurityPolicy(obj interface{}) erro
 			return err
 		}
 	case types.UID:
-		// We can delete the security policy directly by its ID,
-		// It's related resources will be deleted automatically,
-		// So don't worry we have no nsxGroups and nsxRules.
 		indexResults, err := service.SecurityPolicyStore.ByIndex(util.TagScopeSecurityPolicyCRUID, string(sp))
 		if err != nil {
 			log.Error(err, "failed to get security policy", "UID", string(sp))
 			return err
 		}
+		if len(indexResults) == 0 {
+			log.Info("did not get security policy with index", "UID", string(sp))
+			return nil
+		}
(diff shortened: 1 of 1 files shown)


pkg/nsx/services/firewall.go
```diff
@@ -161,6 +161,10 @@ func (service *SecurityPolicyService) DeleteSecurityPolicy(obj interface{}) erro
 			log.Error(err, "failed to get security policy", "UID", string(sp))
 			return err
 		}
+		if len(indexResults) == 0 {
+			log.Info("failed to get security policy with index", "UID", string(sp))
```
heypnus, line 165: How about `did not get security policy with index`?
And in which case `len(indexResults)` will be 0?

zhengxiexie: ok
It's hard to figure out why there are garbage groups below when running pipelines, but for defensive programming， it's better to add this line.

https://github.com/vmware-tanzu/nsx-operator/blob/1b9f49970085a0893b44c91de81074e8d8c32694/pkg/nsx/services/store.go#L263

image

zhengxiexie: ok
I found the bug, because the associated groups are not deleted when deleting the security policy.

pkg/nsx/services/firewall.go
```diff
@@ -153,16 +153,30 @@ func (service *SecurityPolicyService) DeleteSecurityPolicy(obj interface{}) erro
...
+			log.Error(err, "failed to get groups", "UID", string(sp))
+			return err
+		}
+		if len(indexResults) == 0 {
+			log.Info("did not get groups with index", "UID", string(sp))
+		}
+		for _, group := range indexResults {
+			*nsxGroups = append(*nsxGroups, group.(model.Group))
+		}
+		nsxSecurityPolicy = &t
```
heypnus, line 179: Why we should copy line 166?

zhengxiexie: removed.

PR #111, opened 2022-09-21
IPPool CRD init

zhengxiexie:
Consider the stale IPPool in NCP, use v1alpha2.


pkg/apis/v1alpha2/condition_types.go
```diff
@@ -0,0 +1,29 @@
+package v1alpha2
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ConditionType string
```
lxiaopei, line 8: I think you can use the condition_types.go under v1alpha1? Why we re-define again here?

zhengxiexie: OK

timdengyun:
Please submit this patch to  https://github.com/vmware-tanzu/nsx-operator/tree/vpc_dev not to main branch.

PR #120, opened 2022-09-30
Fix mixing staleGroups with staleRules when deleting legacy resources

zhengxiexie:
The process exits reporting panic: runtime error: index out of range [0] with length 0. It's an obvious error.
When there are stale rules and no stale groups, it would definitely hit this crashing down.

### DIFF
--- a/pkg/nsx/services/firewall.go
+++ b/pkg/nsx/services/firewall.go
@@ -98,7 +98,7 @@ func (service *SecurityPolicyService) CreateOrUpdateSecurityPolicy(obj *v1alpha1
 	finalSecurityPolicy.Rules = finalRules
 
 	finalGroups := make([]model.Group, 0)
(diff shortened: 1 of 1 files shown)


lxiaopei:
Could you change your commit msg to be more specific?

timdengyun:
It's better that we can describe which scenario hit this error rather than just say"
Process exits reporting panic: runtime error: index out of range [0] with length 0.
It's an obvious error."

For other users, we can better understand when this error happen although it's obvious.

zhengxiexie:
> It's better that we can describe which scenario hit this error rather than just say" Process exits reporting panic: runtime error: index out of range [0] with length 0. It's an obvious error."
> 
> For other users, we can better understand when this error happen although it's obvious.

When there are stale rules and no stale groups, it would definitely hit this crashing down.

timdengyun (approved):

PR #127, opened 2022-10-19
Support named port feature

zhengxiexie:
Consider the below scenarios:
 * When a new added pod whose port name exists in security policy
 * When a deleted pod whose port name exists in security policy
 * When a pod's label is changed
 * When a namespace's label is changed Ensure named port and endPort not coexist, endPort can only be defined if port is also defined, and both ports must be numeric. Fix compare rule and compare group logic. When pod or namespace label change, the IPSet group may change, however, the groups compare logic still returns equal, the group misses Expression field and should also use DataValueToJsonEncoder to compare groups. Fix the bug no pod to be selected if only namespaceSelectors or podSelector. Fix the bug no right pod to be selected when there is *and* relationship in one target and *or* relationship in another target.


timdengyun:
It looks ok for me of this cheery pick from https://github.com/vmware-tanzu/nsx-operator/pull/85

timdengyun (approved):

PR #134, opened 2022-10-26
Modify builder.go, use third-party pointy to reduce many temporary va…

zhengxiexie:
…riables

Use pointy to get address of variables, in order to reduce many redundant lines.


lxiaopei (approved):

timdengyun (approved):
It looks ok to me.

PR #138, opened 2022-10-27
Add NSXServiceAccount Controller

gran-vmv:
Reconciler:
1. If NSXServiceAccount CR DeletionTimestamp is zero and Status.Phase is not "realized"
  a. Add NSXServiceAccount finalizer if not exists
  b. Generate certificate
  c. Create PrincipalIdentity with certificate and add to store
  d. Create ClusterControlPlane and add to store
  e. Create K8s Secret
  f. Update CR status

2. If NSXServiceAccount CR DeletionTimestamp is not zero and has NSXServiceAccount finalizer
  a. Delete K8s Secret
  b. Delete ClusterControlPlane and remove from store
  c. Delete PrincipalIdentity and remove from store
  d. Remove NSXServiceAccount finalizer

GC:
1. Get all CR UIDs from Tags of PrincipalIdentity and ClusterControlPlane from store
2. Get all NSXServiceAccount CRs
3. Get the UIDs which exists in NSX resource Tags and doesn't exist in CRs
4. Convert UIDs to Namespacednames
5. Run the same delete process in section Reconciler.2

Signed-off-by: gran <gran@vmware.com>

### DIFF
--- a/pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
+++ b/pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
@@ -1,0 +1,236 @@
+/* Copyright © 2022 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: Apache-2.0 */
+
+package nsxserviceaccount
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"runtime"
+	"time"
+
+	apimachineryruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	nsxvmwarecomv1alpha1 "github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
+	"github.com/vmware-tanzu/nsx-operator/pkg/controllers/common"
+	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
+	"github.com/vmware-tanzu/nsx-operator/pkg/metrics"
+	servicecommon "github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/nsxserviceaccount"
+)
+
+var (
+	log                     = logger.Log
+	ResultNormal            = common.ResultNormal
+	ResultRequeue           = common.ResultRequeue
+	ResultRequeueAfter5mins = common.ResultRequeueAfter5mins
+	MetricResType           = common.MetricResTypeNSXServiceAccount
+)
+
+// NSXServiceAccountReconciler reconciles a NSXServiceAccount object
+type NSXServiceAccountReconciler struct {
+	client.Client
+	Scheme  *apimachineryruntime.Scheme
+	Service *nsxserviceaccount.NSXServiceAccountService
+}
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.0/pkg/reconcile
+func (r *NSXServiceAccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	obj := &nsxvmwarecomv1alpha1.NSXServiceAccount{}
+	log.Info("reconciling CR", "nsxserviceaccount", req.NamespacedName)
+	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerSyncTotal, MetricResType)
+
+	if err := r.Client.Get(ctx, req.NamespacedName, obj); err != nil {
+		log.Error(err, "unable to fetch NSXServiceAccount CR", "req", req.NamespacedName)
+		return ResultNormal, client.IgnoreNotFound(err)
+	}
+
+	// Since NSXServiceAccount service can only be activated from NSX 4.1.0 onwards,
+	// So need to check NSX version before starting NSXServiceAccount reconcile
+	if !r.Service.NSXClient.NSXCheckVersionForNSXServiceAccount() {
+		err := errors.New("NSX version check failed, NSXServiceAccount feature is not supported")
+		updateFail(r, &ctx, obj, &err)
+		// if NSX version check fails, it will be put back to reconcile queue and be reconciled after 5 minutes
+		return ResultRequeueAfter5mins, nil
+	}
+
+	if obj.ObjectMeta.DeletionTimestamp.IsZero() {
+		metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerUpdateTotal, MetricResType)
+		if !controllerutil.ContainsFinalizer(obj, servicecommon.NSXServiceAccountFinalizerName) {
+			controllerutil.AddFinalizer(obj, servicecommon.NSXServiceAccountFinalizerName)
+			if err := r.Client.Update(ctx, obj); err != nil {
+				log.Error(err, "add finalizer", "nsxserviceaccount", req.NamespacedName)
+				updateFail(r, &ctx, obj, &err)
+				return ResultRequeue, err
+			}
+			log.V(1).Info("added finalizer on CR", "nsxserviceaccount", req.NamespacedName)
+		}
+
+		if obj.Status.Phase == nsxvmwarecomv1alpha1.NSXServiceAccountPhaseRealized {
+			return ResultNormal, nil
+		}
+		if err := r.Service.CreateOrUpdateNSXServiceAccount(ctx, obj); err != nil {
+			log.Error(err, "operate failed, would retry exponentially", "nsxserviceaccount", req.NamespacedName)
+			updateFail(r, &ctx, obj, &err)
+			return ResultRequeue, err
+		}
+		updateSuccess(r, &ctx, obj)
+	} else {
+		if controllerutil.ContainsFinalizer(obj, servicecommon.NSXServiceAccountFinalizerName) {
+			metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteTotal, MetricResType)
+			if err := r.Service.DeleteNSXServiceAccount(ctx, types.NamespacedName{
+				Namespace: obj.Namespace,
+				Name:      obj.Name,
+			}); err != nil {
+				log.Error(err, "deleting failed, would retry exponentially", "nsxserviceaccount", req.NamespacedName)
+				deleteFail(r, &ctx, obj, &err)
+				return ResultRequeue, err
+			}
+			controllerutil.RemoveFinalizer(obj, servicecommon.NSXServiceAccountFinalizerName)
+			if err := r.Client.Update(ctx, obj); err != nil {
+				log.Error(err, "removing finalizer failed, would retry exponentially", "nsxserviceaccount", req.NamespacedName)
+				deleteFail(r, &ctx, obj, &err)
+				return ResultRequeue, err
+			}
+			log.V(1).Info("removed finalizer", "nsxserviceaccount", req.NamespacedName)
+			deleteSuccess(r, &ctx, obj)
+		} else {
+			// only print a message because it's not a normal case
+			log.Info("finalizers cannot be recognized", "nsxserviceaccount", req.NamespacedName)
+		}
+	}
+
+	return ResultNormal, nil
+}
+
+// setupWithManager sets up the controller with the Manager.
+func (r *NSXServiceAccountReconciler) setupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&nsxvmwarecomv1alpha1.NSXServiceAccount{}).
+		WithEventFilter(predicate.Funcs{
+			DeleteFunc: func(e event.DeleteEvent) bool {
+				// Suppress Delete events to avoid filtering them out in the Reconcile function
+				return false
+			},
+		}).
+		WithOptions(
+			controller.Options{
+				MaxConcurrentReconciles: runtime.NumCPU(),
+			}).
+		Complete(r)
+}
+
+// Start setup manager and launch GC
+func (r *NSXServiceAccountReconciler) Start(mgr ctrl.Manager) error {
+	err := r.setupWithManager(mgr)
+	if err != nil {
+		return err
+	}
+
+	go r.GarbageCollector(make(chan bool), servicecommon.GCInterval)
+	return nil
+}
+
+// GarbageCollector collect NSXServiceAccount which has been removed from crd.
+// cancel is used to break the loop during UT
+func (r *NSXServiceAccountReconciler) GarbageCollector(cancel chan bool, timeout time.Duration) {
+	ctx := context.Background()
+	log.Info("garbage collector started")
+	for {
+		select {
+		case <-cancel:
+			return
+		case <-time.After(timeout):
+		}
+		nsxServiceAccountUIDSet := r.Service.ListNSXServiceAccountRealization()
+		if len(nsxServiceAccountUIDSet) == 0 {
+			continue
+		}
+		nsxServiceAccountList := &nsxvmwarecomv1alpha1.NSXServiceAccountList{}
+		err := r.Client.List(ctx, nsxServiceAccountList)
+		if err != nil {
+			log.Error(err, "failed to list NSXServiceAccount CR")
+			continue
+		}
+		gcSuccessCount, gcErrorCount := r.garbageCollector(nsxServiceAccountUIDSet, nsxServiceAccountList)
+		log.V(1).Info("gc collects NSXServiceAccount CR", "success", gcSuccessCount, "error", gcErrorCount)
+	}
+}
+
+func (r *NSXServiceAccountReconciler) garbageCollector(nsxServiceAccountUIDSet sets.String, nsxServiceAccountList *nsxvmwarecomv1alpha1.NSXServiceAccountList) (gcSuccessCount, gcErrorCount uint32) {
+	nsxServiceAccountCRUIDMap := map[string]types.NamespacedName{}
+	for _, nsxServiceAccount := range nsxServiceAccountList.Items {
+		nsxServiceAccountCRUIDMap[string(nsxServiceAccount.UID)] = types.NamespacedName{
+			Namespace: nsxServiceAccount.Namespace,
+			Name:      nsxServiceAccount.Name,
+		}
+	}
+
+	for nsxServiceAccountUID := range nsxServiceAccountUIDSet {
+		if _, ok := nsxServiceAccountCRUIDMap[nsxServiceAccountUID]; ok {
+			continue
+		}
+		log.V(1).Info("gc collects NSXServiceAccount CR", "UID", nsxServiceAccountUID)
+		namespacedName := r.Service.GetNSXServiceAccountNameByUID(nsxServiceAccountUID)
+		if namespacedName.Namespace == "" || namespacedName.Name == "" {
+			continue
+		}
+		metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteTotal, MetricResType)
+		err := r.Service.DeleteNSXServiceAccount(context.TODO(), namespacedName)
+		if err != nil {
+			gcErrorCount++
+			metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteFailTotal, MetricResType)
+		} else {
+			gcSuccessCount++
+			metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteSuccessTotal, MetricResType)
+		}
+	}
+	return
+}
+
+func (r *NSXServiceAccountReconciler) updateNSXServiceAccountStatus(ctx *context.Context, o *nsxvmwarecomv1alpha1.NSXServiceAccount, e *error) {
+	obj := o
+	if e != nil && *e != nil {
+		obj = o.DeepCopy()
+		obj.Status.Phase = nsxvmwarecomv1alpha1.NSXServiceAccountPhaseFailed
+		obj.Status.Reason = fmt.Sprintf("Error: %v", *e)
+	}
+	err := r.Client.Status().Update(*ctx, obj)
+	if err != nil {
+		log.Error(err, "update NSXServiceAccount failed", "Namespace", obj.Namespace, "Name", obj.Name, "Status", obj.Status)
+	} else {
+		log.V(1).Info("updated NSXServiceAccount", "Namespace", obj.Namespace, "Name", obj.Name, "Status", obj.Status)
+	}
+}
+
+func updateFail(r *NSXServiceAccountReconciler, c *context.Context, o *nsxvmwarecomv1alpha1.NSXServiceAccount, e *error) {
+	r.updateNSXServiceAccountStatus(c, o, e)
+	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerUpdateFailTotal, MetricResType)
+}
+
+func deleteFail(r *NSXServiceAccountReconciler, c *context.Context, o *nsxvmwarecomv1alpha1.NSXServiceAccount, e *error) {
+	r.updateNSXServiceAccountStatus(c, o, e)
+	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteFailTotal, MetricResType)
+}
+
+func updateSuccess(r *NSXServiceAccountReconciler, c *context.Context, o *nsxvmwarecomv1alpha1.NSXServiceAccount) {
+	r.updateNSXServiceAccountStatus(c, o, nil)
+	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerUpdateSuccessTotal, MetricResType)
+}
+
+func deleteSuccess(r *NSXServiceAccountReconciler, _ *context.Context, _ *nsxvmwarecomv1alpha1.NSXServiceAccount) {
+	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteSuccessTotal, MetricResType)
+}
--- a/pkg/nsx/cluster.go
+++ b/pkg/nsx/cluster.go
@@ -265,8 +265,12 @@ func (nsxVersion *NsxVersion) Validate() error {
 func (nsxVersion *NsxVersion) featureSupported(feature string) bool {
 	var minVersion [3]int64
 	validFeature := false
-	if feature == FeatureSecurityPolicy {
+	switch feature {
+	case FeatureSecurityPolicy:
 		minVersion = nsx320Version
+		validFeature = true
+	case FeatureNSXServiceAccount:
+		minVersion = nsx401Version
 		validFeature = true
 	}
 	if validFeature {
--- a/pkg/nsx/services/common/builder.go
+++ b/pkg/nsx/services/common/builder.go
@@ -1,0 +1,37 @@
+/* Copyright © 2022 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: Apache-2.0 */
+
+package common
+
+import (
+	"fmt"
+	"strings"
+
+	mpmodel "github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/model"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+)
+
+func QueryTagCondition(resourceType, cluster string) string {
+	return fmt.Sprintf("%s:%s AND tags.scope:%s AND tags.tag:%s",
+		ResourceType, resourceType,
+		strings.Replace(TagScopeCluster, "/", "\\/", -1),
+		strings.Replace(cluster, ":", "\\:", -1))
+}
+
+func ConvertTagsToMPTags(tags []model.Tag) []mpmodel.Tag {
+	mpTags := make([]mpmodel.Tag, len(tags))
+	for i := 0; i < len(tags); i++ {
+		mpTags[i].Scope = tags[i].Scope
+		mpTags[i].Tag = tags[i].Tag
+	}
+	return mpTags
+}
+
+func ConvertMPTagsToTags(mpTags []mpmodel.Tag) []model.Tag {
+	tags := make([]model.Tag, len(mpTags))
+	for i := 0; i < len(mpTags); i++ {
+		tags[i].Scope = mpTags[i].Scope
+		tags[i].Tag = mpTags[i].Tag
+	}
+	return tags
+}
--- a/pkg/nsx/services/common/types.go
+++ b/pkg/nsx/services/common/types.go
@@ -15,25 +15,30 @@ import (
 )
 
 const (
-	HashLength                   int    = 8
-	MaxTagLength                 int    = 256
-	TagScopeCluster              string = "nsx-op/cluster"
-	TagScopeNamespace            string = "nsx-op/namespace"
-	TagScopeSecurityPolicyCRName string = "nsx-op/security_policy_cr_name"
-	TagScopeSecurityPolicyCRUID  string = "nsx-op/security_policy_cr_uid"
-	TagScopeRuleID               string = "nsx-op/rule_id"
-	TagScopeGroupType            string = "nsx-op/group_type"
-	TagScopeSelectorHash         string = "nsx-op/selector_hash"
-	TagScopeNCPCluster           string = "ncp/cluster"
-	TagScopeNCPProject           string = "ncp/project"
-	TagScopeNCPVIFProject        string = "ncp/vif_project"
-	TagScopeNCPPod               string = "ncp/pod"
-	TagScopeNCPVNETInterface     string = "ncp/vnet_interface"
-	TagScopeVPCCRName            string = "nsx-op/vpc_cr_name"
-	TagScopeVPCCRUID             string = "nsx-op/vpc_cr_uid"
+	HashLength                      int    = 8
+	MaxTagLength                    int    = 256
+	MaxIdLength                     int    = 255
+	TagScopeCluster                 string = "nsx-op/cluster"
+	TagScopeNamespace               string = "nsx-op/namespace"
+	TagScopeSecurityPolicyCRName    string = "nsx-op/security_policy_cr_name"
+	TagScopeSecurityPolicyCRUID     string = "nsx-op/security_policy_cr_uid"
+	TagScopeRuleID                  string = "nsx-op/rule_id"
+	TagScopeGroupType               string = "nsx-op/group_type"
+	TagScopeSelectorHash            string = "nsx-op/selector_hash"
+	TagScopeNSXServiceAccountCRName string = "nsx-op/nsx_service_account_name"
+	TagScopeNSXServiceAccountCRUID  string = "nsx-op/nsx_service_account_uid"
+	TagScopeNCPCluster              string = "ncp/cluster"
+	TagScopeNCPProject              string = "ncp/project"
+	TagScopeNCPVIFProject           string = "ncp/vif_project"
+	TagScopeNCPPod                  string = "ncp/pod"
+	TagScopeNCPVNETInterface        string = "ncp/vnet_interface"
+	TagScopeVPCCRName               string = "nsx-op/vpc_cr_name"
+	TagScopeVPCCRUID                string = "nsx-op/vpc_cr_uid"
 
 	GCInterval    = 60 * time.Second
 	FinalizerName = "securitypolicy.nsx.vmware.com/finalizer"
+
+	NSXServiceAccountFinalizerName = "nsxserviceaccount.nsx.vmware.com/finalizer"
 )
 
 var (
@@ -42,6 +47,10 @@ var (
 	ResourceTypeGroup          = "Group"
 	ResourceTypeRule           = "Rule"
 	ResourceTypeVPC            = "VPC"
+	// ResourceTypeClusterControlPlane is used by NSXServiceAccountController
+	ResourceTypeClusterControlPlane = "clustercontrolplane"
+	// ResourceTypePrincipalIdentity is used by NSXServiceAccountController, and it is MP resource type.
+	ResourceTypePrincipalIdentity = "principalidentity"
 )
 
 type Service struct {
--- a/pkg/nsx/services/nsxserviceaccount/cluster.go
+++ b/pkg/nsx/services/nsxserviceaccount/cluster.go
@@ -1,0 +1,289 @@
+/* Copyright © 2022 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: Apache-2.0 */
+
+package nsxserviceaccount
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	mpmodel "github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/model"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/tools/cache"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
+	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
+	"github.com/vmware-tanzu/nsx-operator/pkg/util"
+)
+
+const (
+	siteId             = "default"
+	enforcementpointId = "default"
+	PortRestAPI        = "rest-api"
+	PortNSXRPCFwdProxy = "nsx-rpc-fwd-proxy"
+	SecretSuffix       = "-nsx-cert"
+	SecretCAName       = "ca.crt"
+	SecretCertName     = "tls.crt"
+	SecretKeyName      = "tls.key"
+)
+
+var (
+	log = logger.Log
+
+	isProtectedTrue = true
+	vpcRole         = "ccp_internal_operator"
+	readerPath      = "/"
+	readerRole      = "cluster_info_reader"
+
+	antreaClusterResourceType = "AntreaClusterControlPlane"
+	revision1                 = int64(1)
+)
+
+type NSXServiceAccountService struct {
+	common.Service
+	PrincipalIdentityStore   *PrincipalIdentityStore
+	ClusterControlPlaneStore *ClusterControlPlaneStore
+}
+
+// InitializeNSXServiceAccount sync NSX resources
+func InitializeNSXServiceAccount(service common.Service) (*NSXServiceAccountService, error) {
+	wg := sync.WaitGroup{}
+	wgDone := make(chan bool)
+	fatalErrors := make(chan error)
+
+	wg.Add(2)
+	nsxServiceAccountService := &NSXServiceAccountService{Service: service}
+
+	nsxServiceAccountService.SetUpStore()
+	go nsxServiceAccountService.InitializeResourceStore(&wg, fatalErrors, common.ResourceTypePrincipalIdentity, nsxServiceAccountService.PrincipalIdentityStore)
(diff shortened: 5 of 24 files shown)


pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -0,0 +1,278 @@
...
+	antreaClusterResourceType = "AntreaClusterControlPlane"
+	revision1                 = int64(1)
+)
+
+func init() {
+	Converter = bindings.NewTypeConverter()
+	Converter.SetMode(bindings.REST)
+}
+
+type NSXServiceAccountService struct {
```
timdengyun, line 56: Recently, we have a refactor for store resource, we defined a common resource store. 
 // ResourceStore is the store for resource, embed it to subclass
type ResourceStore struct {
	cache.Indexer        // the ultimate place to store the resource
	bindings.BindingType // used by converter to convert the resource
}

and user can to implemented specific store in own service like
type SecurityPolicyService struct {
	common.Service
	common.Service
	securityPolicyStore *SecurityPolicyStore
	ruleStore           *RuleStore
	groupStore          *GroupStore
}
}

So, we don't call cache.Indexer directly as a store, instead, we can encapsulate calling of cache.Indexer into
store's GetByKey and GetByIndex. 
You can refer to this PR:
https://github.com/vmware-tanzu/nsx-operator/pull/148
do you think you need to do this refactor as well?

timdengyun: https://github.com/vmware-tanzu/nsx-operator/pull/152, I have rebase recent vpc changes to main branch. So, right now, main branch include all changes. Give timeline constraints, I will be fine to keep current implementation.

timdengyun (approved):

timdengyun (approved):

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -0,0 +1,241 @@
...
+		case <-cancel:
+			return
+		case <-time.After(timeout):
+		}
+		nsxServiceAccountUIDSet := r.Service.ListNSXServiceAccountRealization()
+		if len(nsxServiceAccountUIDSet) == 0 {
+			continue
+		}
+		nsxServiceAccountList := &nsxvmwarecomv1alpha1.NSXServiceAccountList{}
+		err := r.Client.List(ctx, nsxServiceAccountList)
```
edwardbadboy, line 169: Is it possible to use a cache for NSXServiceAccount, and update the cache in sync with CR Added/Updated/Deleted events? Although NSXServiceAccount object is small, I feel better to avoid a regular list call for all NSXServiceAccounts.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -0,0 +1,241 @@
...
+			Namespace: nsxServiceAccount.Namespace,
+			Name:      nsxServiceAccount.Name,
+		}
+	}
+
+	for nsxServiceAccountUID := range nsxServiceAccountUIDSet {
+		if _, ok := nsxServiceAccountCRUIDMap[nsxServiceAccountUID]; ok {
+			continue
+		}
+		log.V(1).Info("GC collected NSXServiceAccount CR", "UID", nsxServiceAccountUID)
```
edwardbadboy, line 191: Maybe "collect" is better than "collected", because at this time it hasn't really deletes the resources yet.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	return
+}
+
+func (r *NSXServiceAccountReconciler) updateNSXServiceAccountStatus(ctx *context.Context, o *nsxvmwarecomv1alpha1.NSXServiceAccount, e *error) {
+	obj := o
+	if e != nil && *e != nil {
+		obj = o.DeepCopy()
+		obj.Status.Phase = nsxvmwarecomv1alpha1.NSXServiceAccountPhaseFailed
+		obj.Status.Reason = obj.Status.Reason + fmt.Sprintf("Error: %v", *e)
```
edwardbadboy, line 214: Will this make Status.Reason longer and longer? Maybe we can keep just the latest reason.

pkg/apis/v1alpha1/nsxserviceaccount_types.go
```diff
@@ -0,0 +1,86 @@
...
+type NSXSecret struct {
+	Name      string `json:"name"`
+	Namespace string `json:"namespace"`
+}
+
+type NSXServiceAccountPhase string
+
+const (
+	NSXServiceAccountPhaseRealized   NSXServiceAccountPhase = "realized"
+	NSXServiceAccountPhaseInProgress NSXServiceAccountPhase = "inProgress"
```
edwardbadboy, line 47: I didn't find where NSXServiceAccountPhaseInProgress is set. Can we set it if deletionTimestamp=0 and phase!=NSXServiceAccountPhase? Or if the time for realizing NSXServiceAccount is really short, we can delete this NSXServiceAccountPhaseInProgress.

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -0,0 +1,278 @@
...
+	secretName := obj.Name + SecretSuffix
+	secretNamespace := obj.Namespace
+	if err := s.Client.Create(ctx, &v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secretName,
+			Namespace: secretNamespace,
+			// TODO: Add labels/annotations
+			Labels:          nil,
+			Annotations:     nil,
+			OwnerReferences: nil,
```
edwardbadboy, line 160: Should it set ownerReference to the corresponding NSXSA CR?

pkg/util/crypto.go
```diff
@@ -0,0 +1,76 @@
...
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"math/big"
+	"time"
+)
+
+const (
+	DefaultRSABits            = 2048
+	DefaultValidDays          = 3650
```
edwardbadboy, line 19: Maybe add a comment: for now the ClusterControlPlane API doesn't support rotating certificate. We set long valid time  to avoid certificate expiration.

gran-vmv: Added.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
gran-vmv: Sure. We only set Reason here thus keeping latest reason makes sense.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
gran-vmv: collected -> collects

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
gran-vmv: The cache is used in kubebuilder.
https://book.kubebuilder.io/architecture.html

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -0,0 +1,241 @@
...
+				Namespace: obj.Namespace,
+				Name:      obj.Name,
+			}); err != nil {
+				log.Error(err, "deletion failed, would retry exponentially", "nsxserviceaccount", req.NamespacedName)
+				deleteFail(r, &ctx, obj, &err)
+				return ResultRequeue, err
+			}
+			controllerutil.RemoveFinalizer(obj, servicecommon.NSXServiceAccountFinalizerName)
+			if err := r.Client.Update(ctx, obj); err != nil {
+				log.Error(err, "deletion failed, would retry exponentially", "nsxserviceaccount", req.NamespacedName)
```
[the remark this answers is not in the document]
gran-vmv: Fixed.

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -0,0 +1,278 @@
...
+		cascade := true
+		if err := s.NSXClient.ClusterControlPlanesClient.Delete(siteId, enforcementpointId, normalizedClusterName, &cascade); err != nil {
+			log.Error(err, "Failed to delete", "ClusterControlPlane", normalizedClusterName)
+			return err
+		}
+		s.ClusterControlPlaneStore.Delete(ccp)
+	}
+
+	// delete PI
+	if piobj, exist, err := s.PrincipalIdentityStore.GetByKey(normalizedClusterName); err != nil {
```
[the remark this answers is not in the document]
gran-vmv: This is a cache get, thus it is OK.
We need to get an obj to find the ID for a MP object, since we don't store this ID in upstream resources.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
edwardbadboy: Got it. Thanks. Found explanations in controller-runtime package doc.

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: Sure, please check if my implementation is expected.

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -0,0 +1,278 @@
...
+	// delete Secret
+	secretName := namespacedName.Name + SecretSuffix
+	secretNamespace := namespacedName.Namespace
+	if err := s.Client.Delete(ctx, &v1.Secret{ObjectMeta: metav1.ObjectMeta{Name: secretName, Namespace: secretNamespace}}); err != nil && !errors.IsNotFound(err) {
+		log.Error(err, "Failed to delete", "secret", secretName, "namespace", secretNamespace)
+		return err
+	}
+
+	// delete ClusterControlPlane
+	if ccp, err := s.NSXClient.ClusterControlPlanesClient.Get(siteId, enforcementpointId, normalizedClusterName); err == nil {
```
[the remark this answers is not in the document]
gran-vmv: This GET operation is not needed since ClusterControlPlanes deletion bug is fixed now. Removed this.

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+		s.ClusterControlPlaneStore.Delete(ccp)
+	}
+
+	// delete PI
+	if piobj, exist, err := s.PrincipalIdentityStore.GetByKey(normalizedClusterName); err != nil {
+		return err
+	} else if exist {
+		pi := piobj.(mpmodel.PrincipalIdentity)
+		if err := s.NSXClient.PrincipalIdentitiesClient.Delete(*pi.Id); err != nil {
```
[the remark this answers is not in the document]
gran-vmv: No, to delete PI won't delete certificate.
Added certificate deletion here.

pkg/apis/v1alpha1/nsxserviceaccount_types.go
```diff
@@ -0,0 +1,86 @@
...
+type NSXProxyEndpointAddress struct {
+	Hostname string `json:"hostname,omitempty"`
+	//+kubebuilder:validation:Format=ip
+	IP string `json:"ip,omitempty"`
+}
+
+type NSXProxyProtocol string
+
+const (
+	NSXProxyProtocolTCP NSXProxyProtocol = "TCP"
```
zhengxiexie, line 24: It shows Unused constant 'NSXProxyProtocolTCP' in my IDE.

gran-vmv: Yes. This is CRD definition and it is not used yet.
Note: CRD related PRs are already merged to main branch.

pkg/nsx/services/common/builder.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+func QueryTagCondition(resourceType, cluster string) string {
+	return fmt.Sprintf("%s:%s AND tags.scope:%s AND tags.tag:%s",
+		ResourceType, resourceType,
+		strings.Replace(TagScopeCluster, "/", "\\/", -1),
+		strings.Replace(cluster, ":", "\\:", -1))
+}
+
+func ConvertTagsToMPTags(tags []model.Tag) []mpmodel.Tag {
```
zhengxiexie, line 21: Since only serviceaccount use it, perhaps it's better to put them in nsxserviceaccount/builder.go ?

pkg/nsx/services/common/builder.go
```diff
@@ -0,0 +1,37 @@
...
+	mpTags := make([]mpmodel.Tag, len(tags))
+	for i := 0; i < len(tags); i++ {
+		mpTags[i].Scope = tags[i].Scope
+		mpTags[i].Tag = tags[i].Tag
+	}
+	return mpTags
+}
+
+func ConvertMPTagsToTags(mpTags []mpmodel.Tag) []model.Tag {
```
zhengxiexie, line 30: ditto

pkg/nsx/services/common/builder.go
gran-vmv: ConvertTagsToMPTags and ConvertMPTagsToTags is common functions, which might be used by all MP API calls in future. Thus I prefer to put them into common module.

pkg/nsx/services/nsxserviceaccount/store.go
```diff
@@ -0,0 +1,130 @@
...
+	"sync"
+
+	mpmodel "github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/model"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
+	nsxutil "github.com/vmware-tanzu/nsx-operator/pkg/nsx/util"
+)
+
+func queryPrincipalIdentity(service *NSXServiceAccountService, wg *sync.WaitGroup, fatalErrors chan error) {
```
zhengxiexie, line 18: You can rebase the latest vpc_dev branch, and reuse function InitializeResourceStore, refer to [https://github.com/vmware-tanzu/nsx-operator/blob/vpc_dev/pkg/nsx/services/securitypolicy/firewall.go#L56](url)

pkg/nsx/services/nsxserviceaccount/store.go
```diff
@@ -0,0 +1,130 @@
...
+			break
+		}
+		c, _ := strconv.Atoi(*cursor)
+		if int64(c) >= *response.ResultCount {
+			break
+		}
+	}
+}
+
+func queryClusterControlPlane(service *NSXServiceAccountService, wg *sync.WaitGroup, fatalErrors chan error) {
```
zhengxiexie, line 57: ditto

pkg/nsx/services/nsxserviceaccount/store.go
```diff
@@ -0,0 +1,130 @@
...
+	case mpmodel.PrincipalIdentity:
+		return *v.Name, nil
+	default:
+		return "", errors.New("keyFunc doesn't support unknown type")
+	}
+}
+
+// indexFunc is used to get index of a resource, usually, which is the UID of the CR controller reconciles,
+// index is used to filter out resources which are related to the CR
+func IndexFunc(obj interface{}) ([]string, error) {
```
zhengxiexie, line 110: Keep it lowercase since it is only used inside the package?

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -0,0 +1,289 @@
...
+	"github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
+	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
+	"github.com/vmware-tanzu/nsx-operator/pkg/util"
+)
+
+const (
+	siteId             = "default"
+	enforcementpointId = "default"
+	PortRestAPI        = "rest-api"
```
zhengxiexie, line 30: Unused constant 'PortRestAPI'

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -0,0 +1,289 @@
...
+	PortNSXRPCFwdProxy = "nsx-rpc-fwd-proxy"
```
zhengxiexie, line 31: ditto

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -0,0 +1,289 @@
...
+	SecretSuffix       = "-nsx-cert"
+	SecretCAName       = "ca.crt"
```
zhengxiexie, line 33: ditto

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -0,0 +1,289 @@
...
+	vpcRole         = "ccp_internal_operator"
+	readerPath      = "/"
+	readerRole      = "cluster_info_reader"
+
+	antreaClusterResourceType = "AntreaClusterControlPlane"
+	revision1                 = int64(1)
+)
+
+func init() {
+	Converter = bindings.NewTypeConverter()
```
zhengxiexie, line 52: You can use [https://github.com/vmware-tanzu/nsx-operator/blob/vpc_dev/pkg/nsx/services/common/types.go#L50](url)

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -0,0 +1,289 @@
...
+					Role: &vpcRole,
+				}},
+			}},
+			CertificatePem: &cert,
+			Tags:           common.ConvertTagsToMPTags(s.buildBasicTags(obj)),
+		})
+		if err != nil {
+			return err
+		}
+		s.PrincipalIdentityStore.Add(pi)
```
zhengxiexie, line 130: Unhandled error

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -0,0 +1,289 @@
...
+			Revision:     &revision1,
+			ResourceType: &antreaClusterResourceType,
+			Certificate:  &cert,
+			VhcPath:      &vpcPath,
+			Tags:         s.buildBasicTags(obj),
+		})
+		if err != nil {
+			return err
+		}
+		s.ClusterControlPlaneStore.Add(ccp)
```
zhengxiexie, line 146: ditto

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -0,0 +1,289 @@
...
+		return err
+	}
+
+	// delete ClusterControlPlane
+	cascade := true
+	if err := s.NSXClient.ClusterControlPlanesClient.Delete(siteId, enforcementpointId, normalizedClusterName, &cascade); err != nil {
+		log.Error(err, "Failed to delete", "ClusterControlPlane", normalizedClusterName)
+		return err
+	}
+	s.ClusterControlPlaneStore.Delete(model.ClusterControlPlane{Id: &normalizedClusterName})
```
zhengxiexie, line 209: ditto

pkg/nsx/services/nsxserviceaccount/store.go
gran-vmv: This is MP resource and has some gap to use common store. I'll refactor queryClusterControlPlane.

pkg/util/crypto.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	derBytes, err := x509.CreateCertificate(rand.Reader, &template, &template, &priv.PublicKey, priv)
+	if err != nil {
+		log.Error(err, "Failed to create certificate")
+		return "", "", err
+	}
+
+	certOut := &bytes.Buffer{}
+	pem.Encode(certOut, &pem.Block{Type: "CERTIFICATE", Bytes: derBytes})
```
zhengxiexie, line 72: Unhandled error

pkg/util/crypto.go
```diff
@@ -0,0 +1,77 @@
...
+	keyOut := &bytes.Buffer{}
+	privBytes := x509.MarshalPKCS1PrivateKey(priv)
+	pem.Encode(keyOut, &pem.Block{Type: "RSA PRIVATE KEY", Bytes: privBytes})
```
zhengxiexie, line 75: ditto

pkg/nsx/services/common/types.go
```diff
@@ -15,30 +15,39 @@ import (
...
+
+	NSXServiceAccountFinalizerName = "nsxserviceaccount.nsx.vmware.com/finalizer"
 )
 
 var (
 	ResourceType               = "resource_type"
 	ResourceTypeSecurityPolicy = "SecurityPolicy"
 	ResourceTypeGroup          = "Group"
 	ResourceTypeRule           = "Rule"
+	// Used by NSXServiceAccountController
```
zhengxiexie, line 47: This format is better
ResourceTypeClusterControlPlane = "clustercontrolplane" // Used by NSXServiceAccountController
Otherwise 
image

pkg/nsx/services/nsxserviceaccount/store.go
gran-vmv: Fixed.

pkg/nsx/services/nsxserviceaccount/store.go
gran-vmv: Refactored.

pkg/nsx/services/nsxserviceaccount/store.go
gran-vmv: Refactored.

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: This is for external use and will be used by operator in future.

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: This is for external use and will be used by operator in future.

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: This is for external use and will be used by operator in future.

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: Removed this unused convertor.

pkg/nsx/services/common/types.go
gran-vmv: Refactored here.

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: Do we need to handle error here? We believe that no error should occur for these calls.

pkg/nsx/cluster.go
```diff
@@ -265,9 +265,13 @@ func (nsxVersion *NsxVersion) Validate() error {
 func (nsxVersion *NsxVersion) featureSupported(feature string) bool {
 	var minVersion [3]int64
 	validFeature := false
-	if feature == FeatureSecurityPolicy {
+	switch feature {
+	case FeatureSecurityPolicy:
 		minVersion = nsx320Version
 		validFeature = true
+	case FeatureNSXServiceAccount:
+		minVersion = nsx410Version
```
edwardbadboy, line 273: minVersion should be changed to NSX 4.0.1. The VPC based multi-tenancy is available in NSX 4.0.1 already.

gran-vmv: Fixed.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+		updateSuccess(r, &ctx, obj)
+	} else {
+		if controllerutil.ContainsFinalizer(obj, servicecommon.NSXServiceAccountFinalizerName) {
+			metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteTotal, MetricResType)
+			if err := r.Service.DeleteNSXServiceAccount(ctx, types.NamespacedName{
+				Namespace: obj.Namespace,
+				Name:      obj.Name,
+			}); err != nil {
+				log.Error(err, "removing finalizer failed, would retry exponentially", "nsxserviceaccount", req.NamespacedName)
```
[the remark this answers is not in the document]
gran-vmv: "removing finalizer failed, would retry exponentially" -> "deleting failed, would retry exponentially" is OK, since nsxserviceaccount is mentioned after this string.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -0,0 +1,241 @@
...
+			Name:      nsxServiceAccount.Name,
+		}
+	}
+
+	for nsxServiceAccountUID := range nsxServiceAccountUIDSet {
+		if _, ok := nsxServiceAccountCRUIDMap[nsxServiceAccountUID]; ok {
+			continue
+		}
+		log.V(1).Info("GC collects NSXServiceAccount CR", "UID", nsxServiceAccountUID)
+		metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteTotal, MetricResType)
```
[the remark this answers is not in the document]
gran-vmv: Thanks. FIxed.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -0,0 +1,241 @@
...
+		namespacedName := r.Service.GetNSXServiceAccountNameByUID(nsxServiceAccountUID)
+		if namespacedName.Namespace == "" || namespacedName.Name == "" {
+			continue
+		}
+		err := r.Service.DeleteNSXServiceAccount(context.TODO(), nsxServiceAccountCRUIDMap[nsxServiceAccountUID])
+		if err != nil {
+			gcSuccessCount++
```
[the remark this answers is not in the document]
gran-vmv: Added V(1) log in the caller. This count is also needed by UT.
Swapped gcSuccessCount/gcErrorCount because current logic is wrong.

pkg/apis/v1alpha1/nsxserviceaccount_types.go
lxiaopei: do we still need NSXServiceAccountPhaseInProgress?

gran-vmv: Current not used, but this commit is merged to main branch. We can ignore CRD definition part in this PR.

gran-vmv: Cherry-picked CRD PRs from main to vpc_dev branch, this part is no longer in this PR.

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -0,0 +1,289 @@
...
+		BindingType: model.ClusterControlPlaneBindingType(),
+	}}
+}
+
+func (s *NSXServiceAccountService) CreateOrUpdateNSXServiceAccount(ctx context.Context, obj *v1alpha1.NSXServiceAccount) error {
+	clusterName := s.getClusterName(obj.Namespace, obj.Name)
+	normalizedClusterName := util.NormalizeId(clusterName)
+	// TODO: Support default project name in config as project
+	project := s.NSXConfig.CoeConfig.Cluster
+	vpcName := obj.Namespace
```
lxiaopei, line 98: need to check the default vpc name here.

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -0,0 +1,289 @@
...
+	vpcPath := fmt.Sprintf("/orgs/default/projects/%s/vpcs/%s", util.NormalizeId(project), vpcName)
```
lxiaopei, line 99: for vpcPath, do we get from VPC CR status or joint path here?

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -0,0 +1,289 @@
...
+	s.ClusterControlPlaneStore.Delete(model.ClusterControlPlane{Id: &normalizedClusterName})
+
+	// delete PI
+	if piobj := s.PrincipalIdentityStore.GetByKey(normalizedClusterName); piobj != nil {
+		pi := piobj.(mpmodel.PrincipalIdentity)
+		if err := s.NSXClient.PrincipalIdentitiesClient.Delete(*pi.Id); err != nil {
+			log.Error(err, "Failed to delete", "PrincipalIdentity", *pi.Name)
+			return err
+		}
+		if pi.CertificateId != nil && *pi.CertificateId != "" {
```
lxiaopei, line 222: if there is case: PI is deleted from NSX but not from store yet, but certificate is not deleted. 
will s.NSXClient.PrincipalIdentitiesClient.Delete raise error for a already deleted PI id?

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: @lxiaopei @edwardbadboy 
Do we need to change the source of `project`/`vpcName`/`vpcPath`? Current implementation hardcoded them above.

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: No, most DELETE NSXAPIs will return `200 OK` if the `Id` is not present.

Thus we delete the resource directly for CCP and PI.

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: `project` is NSX cluster name.
`vpcName` is NSXServiceAccount CR namespace

edwardbadboy: Hi @lxiaopei ,

Do you know how supervisor K8s admin specifies which VPC to use for workload cluster in an SV namespace?

If the workload cluster is using just the default VPC for the namespace, the Antrea-NSX integration should also use the default VPC in the namespace.

If supervisor cluster allows admin to specify different VPCs for different workload clusters in an SV namespace, Antrea-NSX integration should follow the same way.

In short, Antrea-NSX integration should always specify the same VPC as the workload cluster node VM is.

gran-vmv: Note:
We have `Spec.VPCName` in `NSXServiceAccount` CR

pkg/nsx/services/nsxserviceaccount/cluster.go
lxiaopei: thanks, then this LGTM.

pkg/nsx/services/nsxserviceaccount/cluster.go
lxiaopei: for the first version there will be only 1 default VPC created for each namespace. 
for the default VPC name, it may not be the same as Namespace, might be "<namespace>-default-vpc".
As long as it will keep consistence with VPC implementation, it would be OK.

gran-vmv: OK. We need to fix here after the default VPC name is determined. Is this correct?

@edwardbadboy Do we need to check `Spec.VPCName` in `NSXServiceAccount` CR in this release?

gran-vmv: Do we finished the design for default VPC name?

lxiaopei: you can use "<namespace>-default-vpc" for now.

gran-vmv: Corrected default VPC name and added the check for `Spec.VPCName`

edwardbadboy: Hi Ran,

If NCP and NSX-operator now support creating and reconciling new VPC CRs in a namespace, we can keep this Spec.VPCName. This VPCName should refer to the VPC CR name in the same namespace. If NCP and NSX-operator don't support custom VPCs (only support the namespace-default VPC), we can remove this VPCName field.

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: Fixed. Using `obj.Namespace + "-default-vpc"`

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: Got it. Removed Spec.VPCName verification. This version we only support hard-coded default VPC per NS.

gran-vmv:
@lxiaopei Can we prepare to merge this PR since no new comments for a long time?
Integration test is done by @liu4480 at 2 months ago.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -0,0 +1,242 @@
...
+		log.Error(err, "unable to fetch NSXServiceAccount CR", "req", req.NamespacedName)
+		return ResultNormal, client.IgnoreNotFound(err)
+	}
+
+	// Since NSXServiceAccount service can only be activated from NSX 4.1.0 onwards,
+	// So need to check NSX version before starting NSXServiceAccount reconcile
+	if !r.Service.NSXClient.NSXCheckVersionForNSXServiceAccount() {
+		err := errors.New("NSX version check failed, NSXServiceAccount feature is not supported")
+		updateFail(r, &ctx, obj, &err)
+		// if NSX version check fails, it will be put back to reconcile queue and be reconciled after 5 minutes
```
lxiaopei, line 67: I do not understand this, if NSX version check fails, why we re-queue  and try to reconcile again?

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -0,0 +1,242 @@
...
+				return ResultRequeue, err
+			}
+			log.V(1).Info("added finalizer on CR", "nsxserviceaccount", req.NamespacedName)
+		}
+
+		if obj.Status.Phase == nsxvmwarecomv1alpha1.NSXServiceAccountPhaseRealized {
+			return ResultNormal, nil
+		}
+		if err := r.Service.CreateOrUpdateNSXServiceAccount(ctx, obj); err != nil {
+			if errors.As(err, &util2.RestrictionError{}) {
```
lxiaopei, line 87: could we rename util2 -> nsxutil?

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -0,0 +1,289 @@
...
+		Indexer:     cache.NewIndexer(keyFunc, cache.Indexers{common.TagScopeNSXServiceAccountCRUID: indexFunc}),
+		BindingType: model.ClusterControlPlaneBindingType(),
+	}}
+}
+
+func (s *NSXServiceAccountService) CreateOrUpdateNSXServiceAccount(ctx context.Context, obj *v1alpha1.NSXServiceAccount) error {
+	clusterName := s.getClusterName(obj.Namespace, obj.Name)
+	normalizedClusterName := util.NormalizeId(clusterName)
+	// TODO: Support default project name in config as project
+	project := s.NSXConfig.CoeConfig.Cluster
```
lxiaopei, line 97: in this PR https://github.com/vmware-tanzu/nsx-operator/pull/183 will contain Project info.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -0,0 +1,242 @@
...
+		if _, ok := nsxServiceAccountCRUIDMap[nsxServiceAccountUID]; ok {
+			continue
+		}
+		log.V(1).Info("GC collects NSXServiceAccount CR", "UID", nsxServiceAccountUID)
+		namespacedName := r.Service.GetNSXServiceAccountNameByUID(nsxServiceAccountUID)
+		if namespacedName.Namespace == "" || namespacedName.Name == "" {
+			continue
+		}
+		metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteTotal, MetricResType)
+		err := r.Service.DeleteNSXServiceAccount(context.TODO(), nsxServiceAccountCRUIDMap[nsxServiceAccountUID])
```
lxiaopei, line 198: nsxServiceAccountCRUIDMap[nsxServiceAccountUID] should have no value, right?

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -0,0 +1,289 @@
...
+		pi := obj.(mpmodel.PrincipalIdentity)
+		for _, tag := range pi.Tags {
+			switch *tag.Scope {
+			case common.TagScopeNamespace:
+				namespacedName.Namespace = *tag.Tag
+			case common.TagScopeNSXServiceAccountCRName:
+				namespacedName.Name = *tag.Tag
+			}
+			if namespacedName.Name != "" && namespacedName.Namespace != "" {
+				return
```
lxiaopei, line 259: we do not need return namespaceName?

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
gran-vmv: Do we support NSX upgrade when nsx-operator is running? If yes, we should requeue.

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: In line 243, we described the return value name is `namespacedName`, thus we can return it with a simple `return` command without parameters.

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: Do we need `s.NSXConfig.CoeConfig.Cluster` as a fallback? Or can we guarantee `nsxt_project` will exist always?

lxiaopei: Currently for NSXServiceAccount, you do not need a real NSX VPC there, right? you just need a VPCPath right?
If yes, I think you can use cluster s Project for now. With VPC networking supported, projects need to be created by user and set as input when creating namespace.

pkg/nsx/services/nsxserviceaccount/cluster.go
lxiaopei: ok ,thanks.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
gran-vmv: Fixed.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
gran-vmv: Nice catch. Fixed this code issue and updated UT for this.

pkg/nsx/services/nsxserviceaccount/cluster.go
lxiaopei: OK. thanks.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
lxiaopei: OK, checked with QE that we can upgrade NSX during NCP Pod running.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -0,0 +1,242 @@
...
+				return ResultRequeue, err
+			}
+			log.V(1).Info("added finalizer on CR", "nsxserviceaccount", req.NamespacedName)
+		}
+
+		if obj.Status.Phase == nsxvmwarecomv1alpha1.NSXServiceAccountPhaseRealized {
+			return ResultNormal, nil
+		}
+		if err := r.Service.CreateOrUpdateNSXServiceAccount(ctx, obj); err != nil {
+			if errors.As(err, &nsxutil.RestrictionError{}) {
```
lxiaopei, line 87: I do not see RestrictionError raised from CreateOrUpdateNSXServiceAccount func.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -0,0 +1,242 @@
...
+			if err := r.Client.Update(ctx, obj); err != nil {
+				log.Error(err, "removing finalizer failed, would retry exponentially", "nsxserviceaccount", req.NamespacedName)
+				deleteFail(r, &ctx, obj, &err)
+				return ResultRequeue, err
+			}
+			log.V(1).Info("removed finalizer", "nsxserviceaccount", req.NamespacedName)
+			deleteSuccess(r, &ctx, obj)
+		} else {
+			// only print a message because it's not a normal case
+			log.Info("finalizers cannot be recognized", "nsxserviceaccount", req.NamespacedName)
```
lxiaopei, line 118: for abnormal case, shall we use error instead of info?

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
gran-vmv: Thanks. This check is not needed for nsxserviceaccount. Removed.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -0,0 +1,242 @@
...
+func (r *NSXServiceAccountReconciler) updateNSXServiceAccountStatus(ctx *context.Context, o *nsxvmwarecomv1alpha1.NSXServiceAccount, e *error) {
+	obj := o
+	if e != nil && *e != nil {
+		obj = o.DeepCopy()
+		obj.Status.Phase = nsxvmwarecomv1alpha1.NSXServiceAccountPhaseFailed
+		obj.Status.Reason = fmt.Sprintf("Error: %v", *e)
+	}
+	err := r.Client.Status().Update(*ctx, obj)
+	if err != nil {
+		log.Error(err, "updated NSXServiceAccount failed", "Namespace", obj.Namespace, "Name", obj.Name, "Status", obj.Status)
```
lxiaopei, line 219: updated -> update

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
gran-vmv: In securitypolicy controller, we already use log.info for missing finalizer case. I think we should keep the behavior same.
https://github.com/vmware-tanzu/nsx-operator/blob/32e9730b482c58d4706b7b7e6998d0ae0de6bd89/pkg/controllers/securitypolicy/securitypolicy_controller.go#L155

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+func (s *NSXServiceAccountService) DeleteNSXServiceAccount(ctx context.Context, namespacedName types.NamespacedName) error {
+	clusterName := s.getClusterName(namespacedName.Namespace, namespacedName.Name)
+	normalizedClusterName := util.NormalizeId(clusterName)
+	// delete Secret
+	secretName := namespacedName.Name + SecretSuffix
+	secretNamespace := namespacedName.Namespace
+	if err := s.Client.Delete(ctx, &v1.Secret{ObjectMeta: metav1.ObjectMeta{Name: secretName, Namespace: secretNamespace}}); err != nil && !errors.IsNotFound(err) {
+		log.Error(err, "Failed to delete", "secret", secretName, "namespace", secretNamespace)
```
lxiaopei, line 203: some log starts with lowercase, some with uppercase, could you ensure be consistent?

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
gran-vmv: Thanks. Fixed.

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: Fixed.
All logs start with lowercase except proper noun.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
lxiaopei: OK

lxiaopei (approved):
LGTM.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -0,0 +1,236 @@
...
+	Scheme  *apimachineryruntime.Scheme
+	Service *nsxserviceaccount.NSXServiceAccountService
+}
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.0/pkg/reconcile
+func (r *NSXServiceAccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
akutz, line 51: Minor nit, the function receivers do not need to be by address since the type's fields are all by address.

PR #140, opened 2022-10-28
Fix bug of logger, log-level doesn't take effect

zhengxiexie:
In the process of refactoring structure, we use a global logger, however, that PR leads to log-level not taking effect. The basic reason is there existing mutual dependency between main and logger.

### DIFF
--- a/pkg/controllers/securitypolicy/securitypolicy_controller.go
+++ b/pkg/controllers/securitypolicy/securitypolicy_controller.go
@@ -38,7 +38,7 @@ import (
 )
 
 var (
-	log                     = logger.Log
+	log                     = logger.Log()
 	ResultNormal            = common.ResultNormal
 	ResultRequeue           = common.ResultRequeue
 	ResultRequeueAfter5mins = common.ResultRequeueAfter5mins
--- a/pkg/nsx/services/securitypolicy/firewall.go
+++ b/pkg/nsx/services/securitypolicy/firewall.go
@@ -16,7 +16,7 @@ import (
 )
 
 var (
-	log                       = logger.Log
+	log                       = logger.Log()
 	MarkedForDelete           = true
 	EnforceRevisionCheckParam = false
 	Converter                 *bindings.TypeConverter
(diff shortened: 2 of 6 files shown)


timdengyun:
Did this PR https://github.com/vmware-tanzu/nsx-operator/pull/130  introduce this bug ?

zhengxiexie:
> Did this PR #130 introduce this bug ?

Yes.

zhengxiexie:
> > > Did this PR #130 introduce this bug ?
> > 
> > 
> > Yes.
> 
> That will be better since it won't break v4.0.0 branch.

It won't.

timdengyun:
> 

If this, we just roll back changes on logger to make firewall and securitypolicyController also use the separated logger as origin, it should be be ok, there is no need to bring more changes here.

timdengyun:
> > > > Did this PR #130 introduce this bug ?
> > > 
> > > 
> > > Yes.
> > 
> > 
> > That will be better since it won't break v4.0.0 branch.
> 
> It won't.


If this, we just roll back changes on logger to make firewall and securitypolicyController also use the separated logger as origin, it should be be ok, there is no need to bring more changes here.

pkg/nsx/services/securitypolicy/firewall.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 var (
-	log                       = logger.Log
+	log                       = logger.Log()
```
timdengyun, line 19: log                       = logf.Log.WithName("service").WithName("firewall")

pkg/controllers/securitypolicy/securitypolicy_controller.go
```diff
@@ -38,7 +38,7 @@ import (
 )
 
 var (
-	log                     = logger.Log
+	log                     = logger.Log()
```
timdengyun, line 41: log                     = logf.Log.WithName("controller").WithName("securitypolicy") as original

timdengyun:
> In the process of refactoring structure, we use a global logger, however, that PR leads to log-level not taking effect. The basic reason is there existing mutual dependency between main and logger.

-> there exists mutual dependency between main and logger.

PR #141, opened 2022-10-31
Fix bug of logger, log-level doesn't take effect

zhengxiexie:
This statement logf.SetLogger(logger.ZapLogger())`` must called after flag.Parse`,
otherwise, log-level option is always the default value, so the the log-level we pass from command line
doesn't take effect as expected.
Besides, before this PR, there are redundant options prefixed with -zap in
command line, we don't need those, actually.


timdengyun:
In the future, we had better copy commit message to problem description as well, which will help readers to understand this PR even without checking git log or commit message.

timdengyun (approved):

PR #153, opened 2022-11-22
Keep files formatted with go fmt tool

timdengyun:
This PR is to format files in order to keep file format consistency with the one generated by go fmt tool during make.


zhengxiexie (approved):
LGTM

PR #158, opened 2022-11-28
Implement vpcnetworkconfig crd

lxiaopei:


timdengyun:
please merge patch to vpc_dev instead of main branch

seanpang-vmware:
Should we have a VPCNetworkConfigStore to save all the configs for others to query?

zhengxiexie:
I see that there is cache.go in vpcnetworkconfig, why not use store.go like other CR?

zhengxiexie:
As discussed, we could deprecate this PR, and put all the logic in the VPC controller as the security policy does.[https://github.com/vmware-tanzu/nsx-operator/blob/main/pkg/controllers/securitypolicy/securitypolicy_controller.go#L243](url)

pkg/controllers/vpcnetworkconfig/vpcnetworkconfiguration_controller.go
```diff
@@ -0,0 +1,94 @@
...
+	var vpcNetworkConfig v1alpha1.VPCNetworkConfiguration
+	err := r.Client.Get(ctx, req.NamespacedName, &vpcNetworkConfig)
+	var isDeleted bool
+	if err != nil {
+		isDeleted = apierrors.IsNotFound(err)
+		if !isDeleted {
+			klog.InfoS("Unable to fetch VPCNetworkConfiguration", "vpcNetworkConfiguration", req.NamespacedName.String(), "err", err)
+			return ctrl.Result{}, err
+		} else {
+			vpcNetworkConfigurationObj, exist := r.installedVPCNetworkConfigs.GetByKey(req.NamespacedName.String())
```
TaoZou1, line 49: logic here is different with other controller. The other controller will return error if it is not IsNotFound(err). Any reason here?

pkg/controllers/vpcnetworkconfig/cache.go
```diff
@@ -0,0 +1,80 @@
+/* Copyright © 2022 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: Apache-2.0 */
+
+package vpcnetworkconfig
```
TaoZou1, line 4: There is no store.go for this controller, how about other controller needs to access the cache?

timdengyun: please hold off review,  it's possible to move this PR to vpc_dev branch merge.

lxiaopei:
After discussion with @seanpang-vmware  and @zhengxiexie , this PR is not needed anymore. We will watch VPCNetworkConfiguration CRD in VPC controller.

PR #113, opened 2022-09-27
Add NSXServiceAccount CRD

gran-vmv:
Signed-off-by: gran <gran@vmware.com>

PR #160, opened 2022-12-05
IPPool controller added

zhengxiexie:


pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,213 @@
...
+
+func (service *IPPoolService) Operate(nsxIPPool *model.IpAddressPool, nsxIPSubnets []*model.IpAddressPoolBlockSubnet, IPPoolUpdated bool, IPPoolSubnetsUpdated bool) error {
+	if !IPPoolUpdated && !IPPoolSubnetsUpdated {
+		return nil
+	}
+	infraIPPool, err := service.WrapHierarchyIPPool(nsxIPPool, nsxIPSubnets)
+	if err != nil {
+		return err
+	}
+	err = service.NSXClient.ProjectInfraClient.Patch("default", "project-1", *infraIPPool, &EnforceRevisionCheckParam)
```
lxiaopei, line 111: we need to depend on "type" in IPPool CRD to see where the NSX IP Pool be created, under global /infra for "public type" or  project /projects/project/id/infra for "private type".

zhengxiexie: OK, I'll check it.

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,183 @@
...
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	obj := &v1alpha2.IPPool{}
+	log.Info("reconciling ippool CR", "ippool", req.NamespacedName)
+	if err := r.Client.Get(ctx, req.NamespacedName, obj); err != nil {
+		log.Error(err, "unable to fetch ippool CR", "req", req.NamespacedName)
+		return resultNormal, client.IgnoreNotFound(err)
+	}
+	if obj.ObjectMeta.DeletionTimestamp.IsZero() {
+		if !controllerutil.ContainsFinalizer(obj, servicecommon.FinalizerName) {
+			controllerutil.AddFinalizer(obj, servicecommon.FinalizerName)
```
heypnus, line 61: The value of `FinalizerName` is `"securitypolicy.nsx.vmware.com/finalizer"`, which is specific for security policy, I think.

pkg/nsx/client.go
```diff
@@ -72,10 +82,15 @@ func GetClient(cf *config.NSXOperatorConfig) *Client {
 	cluster, _ := NewCluster(c)
 
 	queryClient := search.NewQueryClient(restConnector(cluster))
+	projectQueryClient := projectsearch.NewQueryClient(restConnector(cluster))
 	groupClient := domains.NewGroupsClient(restConnector(cluster))
 	securityClient := domains.NewSecurityPoliciesClient(restConnector(cluster))
 	ruleClient := security_policies.NewRulesClient(restConnector(cluster))
 	infraClient := nsx_policy.NewInfraClient(restConnector(cluster))
+	projectInfraClient := projects.NewInfraClient(restConnector(cluster))
+	ippoolClient := infra.NewIpPoolsClient(restConnector(cluster))
```
heypnus, line 91: `ippoolClient` -> `ipPoolClient`

pkg/nsx/services/common/store.go
```diff
@@ -152,3 +152,42 @@ func (service *Service) InitializeResourceStore(wg *sync.WaitGroup, fatalErrors
...
+
+	tagScopeClusterKey := strings.Replace(TagScopeCluster, "/", "\\/", -1)
+	tagScopeClusterValue := strings.Replace(service.NSXClient.NsxConfig.Cluster, ":", "\\:", -1)
+	tagParam := fmt.Sprintf("tags.scope:%s AND tags.tag:%s", tagScopeClusterKey, tagScopeClusterValue)
+	resourceParam := fmt.Sprintf("%s:%s", ResourceType, resourceTypeValue)
+	queryParam := resourceParam + " AND " + tagParam
+
+	var cursor *string = nil
+	for {
+		response, err := service.NSXClient.ProjectQueryClient.List("default", "project-1", queryParam, cursor, nil, Int64(PageSize), nil, nil)
```
heypnus, line 169: temporary parameters for debugging?

pkg/nsx/services/common/store.go
```diff
@@ -152,3 +152,42 @@ func (service *Service) InitializeResourceStore(wg *sync.WaitGroup, fatalErrors
 		}
 	}
 }
+
+// InitializeProjectResourceStore is the method to query all the various resources from nsx-t side and
+// save them to the store, we could use it to cache all the resources when process starts.
+func (service *Service) InitializeProjectResourceStore(wg *sync.WaitGroup, fatalErrors chan error, resourceTypeValue string, store Store) {
```
heypnus, line 158: It seems that the existing func `InitializeResourceStore` also does the same thing. What's the difference?

pkg/nsx/services/ippool/ippool.go
zhengxiexie: Done, leaved some TODOs.

cmd/main.go
```diff
@@ -8,6 +8,8 @@ import (
 	"os"
 	"time"
 
+	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
+
```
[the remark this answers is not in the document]
zhengxiexie: ok

build/yaml/crd/nsx.vmware.com_ippools.yaml
```diff
@@ -0,0 +1,160 @@
...
+spec:
+  group: nsx.vmware.com
+  names:
+    kind: IPPool
+    listKind: IPPoolList
+    plural: ippools
+    singular: ippool
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
```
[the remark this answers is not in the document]
zhengxiexie: Automatically generating is just an additional tool, we decide the definition, I think it's better to keep the evolution of the version for us to track the change though version 1 is not in nsx-operator, considering ncp is going to migrate to nsx-operator, we could keep all the redundant information.  @lxiaopei How do you think?

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,183 @@
+/* Copyright © 2021 VMware, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/vmware-tanzu/nsx-operator/pkg/metrics"
+	servicecommon "github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/ippool"
+)
+
+var (
+	log           = logger.Log
+	resultNormal  = common.ResultNormal
+	resultRequeue = common.ResultRequeue
+	MetricResType = common.MetricResTypeSecurityPolicy
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/controllers/ippool/ippool_controller.go
zhengxiexie: ok, Thanks~~

pkg/nsx/client.go
zhengxiexie: good

pkg/nsx/services/common/store.go
zhengxiexie: I updated it, and leaved TODOs.

pkg/nsx/services/common/store.go
zhengxiexie: I also refactored it at https://github.com/vmware-tanzu/nsx-operator/pull/172/files

pkg/nsx/services/ippool/builder.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
+	"github.com/vmware-tanzu/nsx-operator/pkg/util"
+)
+
+var (
+	Int64  = common.Int64
+	String = common.String
+)
+
+func (service *IPPoolService) BuildIPPool(IPPool *v1alpha2.IPPool) (*model.IpAddressPool, []*model.IpAddressPoolBlockSubnet) {
```
heypnus, line 18: Can we only return one variable ippool? The concern is that we'd better keep a unify common style. My thinking is that we'd better had a function template likes as follow:
```
func (service *<CRD name>Service) Build<CRD name>(<CRD name> *<api version>.<CRD name>) (*model.<NSX resource type>) {
    return &model.<NSX resource type>{
        // TODO: add properties here
    }
}
```
What's your opinion?

build/image/photon/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.17.7 as golang-build
+FROM golang:1.19.5 as golang-build
```
heypnus, line 1: Don't change it without solid confirm.

pkg/apis/v1alpha2/ippool_types.go
```diff
@@ -0,0 +1,79 @@
...
+	// Subnets defines subnets allocation result.
+	Subnets []SubnetResult `json:"subnets"`
+	// Conditions defines current state of the IPPool.
+	Conditions []v1alpha1.Condition `json:"conditions"`
+}
+
+// SubnetRequest defines the subnet allocation request.
+type SubnetRequest struct {
+	// PrefixLength defines prefix length for this subnet.
+	// +optional
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func containsString(source []string, target string) bool {
+	for _, item := range source {
+		if item == target {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,183 @@
...
+	if obj.ObjectMeta.DeletionTimestamp.IsZero() {
+		if !controllerutil.ContainsFinalizer(obj, servicecommon.IPPoolFinalizerName) {
+			controllerutil.AddFinalizer(obj, servicecommon.IPPoolFinalizerName)
+			if err := r.Client.Update(ctx, obj); err != nil {
+				return resultRequeue, err
+			}
+			log.V(1).Info("added finalizer on ippool CR", "ippool", req.NamespacedName)
+		}
+
+		subnetCidrUpdated, ipPoolSubnetsUpdated, err := r.Service.CreateOrUpdateIPPool(obj)
```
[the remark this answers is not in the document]
zhengxiexie: I have tried several ways to do this logic and found this one is the most concise, this is different from security policy.
subnetCidrUpdated and ipPoolSubnetsUpdated are very vital to the status updated and Service.FullyRealized(obj) is also vital to control if we would put back obj to queue since some subnets are not realized. In this way, we can return partial realized subnets to user in advance even though partial are not realized. We leverage resultRequeue to implement gradually realizing.

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,183 @@
...
+		if err != nil {
+			log.Error(err, "operate failed, would retry exponentially", "ippool", req.NamespacedName)
+			return resultRequeue, err
+		}
+		if !r.Service.FullyRealized(obj) {
+			if subnetCidrUpdated || ipPoolSubnetsUpdated {
+				err = r.Client.Status().Update(ctx, obj)
+				if err != nil {
```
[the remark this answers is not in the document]
zhengxiexie: Yes, I have not done the metric, I will supply later.

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,183 @@
...
+					return resultRequeue, err
+				}
+				log.Info("successfully reconcile ippool CR", "ippool", obj)
+			} else {
+				log.Info("full realized already, and resources are not changed, skip updating them", "obj", obj)
+			}
+		}
+	} else {
+		if containsString(obj.GetFinalizers(), servicecommon.IPPoolFinalizerName) {
```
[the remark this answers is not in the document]
zhengxiexie: good

pkg/nsx/services/ippool/builder.go
zhengxiexie: Then security policy also requires change, much effort needed, perhaps.

build/image/photon/Dockerfile
zhengxiexie: Oh, I don't know when make mess here.

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+
+		log.V(2).Info("ippool garbage collector", "nsxIPPoolSet", nsxIPPoolSet, "CRIPPoolSet", CRIPPoolSet)
+
+		for elem := range nsxIPPoolSet {
+			if CRIPPoolSet.Has(elem) {
+				continue
+			}
+			log.Info("GC collected ip pool CR", "UID", elem)
+			err = r.Service.DeleteIPPool(types.UID(elem))
```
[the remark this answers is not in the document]
zhengxiexie: OK, I will supply metric today.

pkg/apis/v1alpha2/ippool_types.go
```diff
@@ -0,0 +1,78 @@
...
+type IPPoolList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []IPPool `json:"items"`
+}
+
+// IPPoolSpec defines the desired state of IPPool.
+type IPPoolSpec struct {
+	// Type defines the type of this IPPool, public or private.
+	// +kubebuilder:validation:Enum=public;private
```
lxiaopei, line 36: pls change to external or private.

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,225 @@
...
+	if !IPPoolUpdated && !IPPoolSubnetsUpdated {
+		return nil
+	}
+	infraIPPool, err := service.WrapHierarchyIPPool(nsxIPPool, nsxIPSubnets)
+	if err != nil {
+		return err
+	}
+
+	// Get IPPool Type from nsxIPPool
+	IPPoolType := "public"
```
lxiaopei, line 109: use a constant instead of hard-coded string.

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	// TODO: decide org and project by ns, public or private to do
+	if IPPoolType == "private" {
+		// TODO: Get the org name and project name from IPPool's namespace
+		err = service.NSXClient.ProjectInfraClient.Patch("default", "project-1", *infraIPPool, &EnforceRevisionCheckParam)
+	} else if IPPoolType == "public" {
+		err = service.NSXClient.InfraClient.Patch(*infraIPPool, &EnforceRevisionCheckParam)
+	}
+	if err != nil {
```
lxiaopei, line 124: with the above API, I suppose only IPPool is created? which part of code is to create the block subnets for the IpPool?

pkg/apis/v1alpha2/ippool_types.go
zhengxiexie: ok

pkg/nsx/services/ippool/ippool.go
zhengxiexie: ok

pkg/nsx/services/ippool/ippool.go
zhengxiexie: In AcquireRealizedSubnetIP method.

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,249 @@
...
+		return resultNormal, client.IgnoreNotFound(err)
+	}
+
+	// TODO: Since IPPool v1alpha2 service can only be activated from NSX 4.1.0 onwards,
+	// So need to check NSX version before starting IPPool reconcile
+	//if !r.Service.NSXClient.NSXCheckVersionForStaticRoute() {
+	//	err := errors.New("NSX version check failed, IPPool v1alpha2 is not supported")
+	//	updateFail(r, &ctx, obj, &err)
+	//	if NSX version check fails, it will be put back to reconcile queue and be reconciled after 5 minutes
+	//return ResultRequeueAfter5mins, nil
```
lxiaopei, line 118: why we remove this?

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,252 @@
...
+	realizedSubnets, subnetCidrUpdated, e := service.AcquireRealizedSubnetIP(obj)
+	if e != nil {
+		return false, false, e
+	}
+	obj.Status.Subnets = realizedSubnets
+	return subnetCidrUpdated, ipPoolSubnetsUpdated, nil
+}
+
+func (service *IPPoolService) Operate(nsxIPPool *model.IpAddressPool, nsxIPSubnets []*model.IpAddressPoolBlockSubnet, IPPoolUpdated bool, IPPoolSubnetsUpdated bool) error {
+	if !IPPoolUpdated && !IPPoolSubnetsUpdated {
```
lxiaopei, line 105: can we change to if !(IPPoolUpdated  || IPPoolSubnetsUpdated )?

pkg/nsx/services/ippool/builder.go
```diff
@@ -0,0 +1,108 @@
...
+	return &model.IpAddressPool{
+		Id:          String(service.buildIPPoolID(IPPool)),
+		DisplayName: String(service.buildIPPoolName(IPPool)),
+		Tags:        service.buildIPPoolTags(IPPool),
+	}, service.buildIPSubnets(IPPool)
+}
+
+func (service *IPPoolService) buildIPPoolID(IPPool *v1alpha2.IPPool) string {
+	return strings.Join([]string{"ipc", string(IPPool.UID)}, "_")
```
lxiaopei, line 29: When create new IPPool CRD, I do not see UID filed in the CR.

pkg/nsx/services/ippool/builder.go
```diff
@@ -0,0 +1,108 @@
...
+		IPSubnet := service.buildIPSubnet(IPPool, subnetRequest)
+		if IPSubnet != nil {
+			IPSubnets = append(IPSubnets, IPSubnet)
+		}
+	}
+	return IPSubnets
+}
+
+func (service *IPPoolService) buildIPSubnetID(IPPool *v1alpha2.IPPool, subnetRequest *v1alpha2.SubnetRequest) string {
+	return strings.Join([]string{"ibs", string(IPPool.UID), subnetRequest.Name}, "_")
```
lxiaopei, line 58: ditto, no UID in IPPool CRD.

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,252 @@
...
+		realizedSubnet := v1alpha2.SubnetResult{Name: subnetRequest.Name}
+		for _, statusSubnet := range obj.Status.Subnets {
+			if statusSubnet.Name == subnetRequest.Name && statusSubnet.CIDR != "" {
+				realizedSubnet.CIDR = statusSubnet.CIDR
+				realized = true
+				break
+			}
+		}
+		if !realized {
+			cidr, err := service.acquireCidr(obj, &subnetRequest, 3)
```
lxiaopei, line 169: for the retry count, let's make a constant and be consistent for all resources to check the realization.

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+func (service *IPPoolService) AcquireRealizedSubnetIP(obj *v1alpha2.IPPool) ([]v1alpha2.SubnetResult, bool, error) {
+	var realizedSubnets []v1alpha2.SubnetResult
+	subnetCidrUpdated := false
+	for _, subnetRequest := range obj.Spec.Subnets {
+		// check if the subnet is already realized
+		realized := false
+		realizedSubnet := v1alpha2.SubnetResult{Name: subnetRequest.Name}
+		for _, statusSubnet := range obj.Status.Subnets {
```
lxiaopei, line 161: Shall we also consider the scenario that some subnets in status are not needed anymore?  meaning there are no such subnets in SubnetRequest.

pkg/controllers/ippool/ippool_controller.go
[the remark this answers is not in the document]
lxiaopei: I still don't get this. We only need to check if subnets are fully realized, if yes, update status to success, in other cases, requeue to reconcile again.  If ipPoolSubnetsUpdated true but subnetCidrUpdated false, I think we do not need to update IPPool CR again.

lxiaopei (commented):
Reminder to resolve the conflicts.

pkg/controllers/ippool/ippool_controller.go
zhengxiexie: It is TODO, we should reorganize NSXCheckVersion, otherwise, every CRD needs to create duplicated logic.

pkg/nsx/services/ippool/ippool.go
zhengxiexie: ok

pkg/nsx/services/ippool/builder.go
zhengxiexie: Isn't it in metav1.ObjectMeta like all others?

pkg/nsx/services/ippool/builder.go
zhengxiexie: ditto

pkg/nsx/services/ippool/ippool.go
zhengxiexie: ok

pkg/nsx/services/ippool/ippool.go
zhengxiexie: Actually considered. obj.Spec.Subnets are only the latest subnets required, the stale subnets are ignored before.

zhengxiexie:
> Reminder to resolve the conflicts.

ok

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,249 @@
...
+				log.Error(err, "add finalizer", "ippool", req.NamespacedName)
+				updateFail(r, &ctx, obj, &err)
+				return resultRequeue, err
+			}
+			log.V(1).Info("added finalizer on ippool CR", "ippool", req.NamespacedName)
+		}
+
+		subnetCidrUpdated, ipPoolSubnetsUpdated, err := r.Service.CreateOrUpdateIPPool(obj)
+		if err != nil {
+			log.Error(err, "operate failed, would retry exponentially", "ippool", req.NamespacedName)
```
lxiaopei, line 135: if this infinite retry? If in some case, create IP Pool failed due to nsx manager down or some other un-recoverable error, should we also retry infinitely?

pkg/nsx/services/ippool/ippool.go
lxiaopei: If there are some realized subnets in status, but  the subnet is deleted in obj.Spec.Subnets,  can the realized subnet in status be deleted from NSX?

zhengxiexie: Yes, it can.

pkg/controllers/ippool/ippool_controller.go
zhengxiexie: Yes, all the CRDs would retry infinitely since the event exists in the queue forever.

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,252 @@
...
+			if statusSubnet.Name == subnetRequest.Name && statusSubnet.CIDR != "" {
```
lxiaopei, line 162: if subnet in obj.Spec.Subnets does not change name, but change prefixLength, I suppose we also need to re-create the nsx subnet for it?

pkg/controllers/ippool/ippool_controller.go
lxiaopei: I mean for some cases, if retry won't work, then we do not need to requeue the IPPool CR.

pkg/nsx/services/ippool/ippool.go
zhengxiexie: Should we support this scenario? I think we only support adding or deleting the subnet and not the change of prefixLength.

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,249 @@
+/* Copyright © 2022 VMware, Inc. All Rights Reserved.
```
lxiaopei, line 1: 2022 -＞　2023

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+}
+
+func (r *Reconciler) setReadyStatusTrue(ctx *context.Context, ippool *v1alpha2.IPPool) {
+	conditions := []v1alpha1.Condition{
+		{
+			Type:    v1alpha1.Ready,
+			Status:  v1.ConditionTrue,
+			Message: "NSX IPPool has been successfully created/updated",
+			Reason:  "NSX API returned 200 response code for PATCH",
```
lxiaopei, line 93: do not need set Reason?

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,252 @@
...
+		Indexer:     cache.NewIndexer(keyFunc, cache.Indexers{common.TagScopeIPPoolCRUID: indexFunc}),
+		BindingType: model.IpAddressPoolBindingType(),
+	}}
+	ipPoolService.ipPoolBlockSubnetStore = &IPPoolBlockSubnetStore{ResourceStore: common.ResourceStore{
+		Indexer:     cache.NewIndexer(keyFunc, cache.Indexers{common.TagScopeIPPoolCRUID: indexFunc}),
+		BindingType: model.IpAddressPoolBlockSubnetBindingType(),
+	}}
+
+	go ipPoolService.InitializeResourceStore(&wg, fatalErrors, ResourceTypeIPPool, ipPoolService.ipPoolStore)
+	go ipPoolService.InitializeResourceStore(&wg, fatalErrors, ResourceTypeIPPoolBlockSubnet, ipPoolService.ipPoolBlockSubnetStore)
```
lxiaopei, line 52: For NSX VPC Subnet, there will also be IP Pool and Block Subnet created.  I suppose we do not want to search and store them in ipPoolService.ipPoolStore and ipPoolService.ipPoolBlockSubnetStore. 
So we need to search via tags?

pkg/nsx/services/ippool/ippool.go
zhengxiexie: As discussed, we don't support it.

pkg/controllers/ippool/ippool_controller.go
zhengxiexie: OK

pkg/controllers/ippool/ippool_controller.go
zhengxiexie: OK

pkg/nsx/services/ippool/ippool.go
zhengxiexie: Updated in https://github.com/vmware-tanzu/nsx-operator/pull/215

lxiaopei: I do not see ippoolService initialize resource store via tags in #215.

pkg/nsx/services/ippool/ippool.go
lxiaopei: OK, I suggest we add webhook to disallow user to update prefixLength field.

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,252 @@
...
+	case err := <-fatalErrors:
+		close(fatalErrors)
+		return ipPoolService, err
+	}
+	return ipPoolService, nil
+}
+
+func (service *IPPoolService) CreateOrUpdateIPPool(obj *v1alpha2.IPPool) (bool, bool, error) {
+	nsxIPPool, nsxIPSubnets := service.BuildIPPool(obj)
+	if nsxIPPool == nil || len(nsxIPSubnets) == 0 {
```
lxiaopei, line 70: is there possibility that nsxIPPool nil or nsxIPSubnets empty? BTW,  is it supported that https://github.com/vmware-tanzu/nsx-operator/blob/main/pkg/apis/v1alpha2/ippool_types.go#L40 , there is no SubnetRequest in IPPool CRD?

zhengxiexie: Thanks, there's no possibility that nsxIPPool can be nil, however, nsxIPSubnets can be empty since IPPool.Spec.Subnets are optional, I think it's better to remove this check.

zhengxiexie: Done, support when IPPool.Spec.Subnets is nil.

pkg/nsx/services/ippool/ippool.go
zhengxiexie: Done. Add
		{Scope: String(common.TagScopeIPSubnetOwner), Tag: String(ResourceTypeIPPool)},
You can do it in the same way.

pkg/nsx/services/ippool/ippool.go
zhengxiexie: OK

pkg/controllers/ippool/ippool_controller.go
zhengxiexie: I think it is a common feature for all CRDs, should we discuss it and classify the cases for all CRDs when we must requeue or not?

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,251 @@
...
+	if err != nil {
+		log.Error(err, "failed to get ip pool and ip pool subnets by UID", "UID", obj.UID)
+		return false, false, err
+	}
+	ipPoolSubnetsUpdated := false
+	ipPoolUpdated := common.CompareResource(IpAddressPoolToComparable(existingIPPool), IpAddressPoolToComparable(nsxIPPool))
+	changed, stale := common.CompareResources(IpAddressPoolBlockSubnetsToComparable(existingIPSubnets), IpAddressPoolBlockSubnetsToComparable(nsxIPSubnets))
+	changedIPSubnets, staleIPSubnets := ComparableToIpAddressPoolBlockSubnets(changed), ComparableToIpAddressPoolBlockSubnets(stale)
+	for i := len(staleIPSubnets) - 1; i >= 0; i-- {
+		staleIPSubnets[i].MarkedForDelete = &MarkedForDelete
```
dantingl, line 84: Here we cannot just set staleIPSubnets[i].MarkedForDelete = true?

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,251 @@
...
+	m, err := service.NSXClient.RealizedEntitiesClient.List(service.buildIPSubnetIntentPath(obj, subnetRequest), nil)
+	if err != nil {
+		return "", err
+	}
+	for _, realizedEntity := range m.Results {
+		if *realizedEntity.EntityType == "IpBlockSubnet" {
+			for _, attr := range realizedEntity.ExtendedAttributes {
+				if *attr.Key == "cidr" {
+					cidr := attr.Values[0]
+					log.V(1).Info("successfully realized ip subnet ip", "subnetRequest.Name", subnetRequest.Name, "cidr", cidr)
```
dantingl, line 221: successfully realized ip block subnet?

PROJECT
```diff
@@ -35,4 +35,18 @@ resources:
...
   path: github.com/vmware-tanzu/nsx-operator/pkg/api/v1alpha1
   version: v1alpha1
+- api:
+    crdVersion: v1
+    namespaced: true
+  domain: nsx.vmware.com
+  kind: IPPool
+  path: github.com/vmware-tanzu/nsx-operator/pkg/api/v1alpha2
+  version: v1alpha2
+- api:
```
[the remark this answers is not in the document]
zhengxiexie: Perhaps auto-generated, removed.

cmd/main.go
```diff
@@ -102,6 +107,31 @@ func StartNSXServiceAccountController(mgr ctrl.Manager, commonService common.Ser
...
+	}
+
+	// TODO: remove this after vpc is ready
+	if vpcService, err := vpc.InitializeVPC(commonService); err != nil {
+		log.Error(err, "failed to initialize vpc commonService", "controller", "vpc")
+		os.Exit(1)
+	} else {
+		commonctl.ServiceMediator.VPCService = vpcService
+	}
+	if err := ippoolReconcile.Start(mgr); err != nil {
```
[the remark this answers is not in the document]
zhengxiexie: OK, we can discuss it later.

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,249 @@
...
+	Scheme  *apimachineryruntime.Scheme
+	Service *ippool.IPPoolService
+}
+
+func deleteSuccess(r *Reconciler, _ *context.Context, _ *v1alpha2.IPPool) {
+	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteSuccessTotal, MetricResType)
+}
+
+func deleteFail(r *Reconciler, c *context.Context, o *v1alpha2.IPPool, e *error) {
+	r.setReadyStatusFalse(c, o, e)
```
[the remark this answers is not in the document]
zhengxiexie: > Do we need to set CR status for deletion, I guess CR will be deleted even nsx-operator fails to delete NSX resource.
No, when at this point,	
> controllerutil.RemoveFinalizer(obj, servicecommon.IPPoolFinalizerName)
controller-runtime would really remove it.
> And Message in setReadyStatusFalse is only about create/update.
Updated.

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,249 @@
...
+			Status:  v1.ConditionFalse,
+			Message: "NSX IPPool could not be created/updated",
+			Reason: fmt.Sprintf(
+				"error occurred while processing the IPPool CR. Error: %v",
+				*err,
+			),
+		},
+	}
+	ippool.Status.Conditions = conditions
+	if ippool.Status.Subnets == nil {
```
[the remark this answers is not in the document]
zhengxiexie: Yes, the ippool.Status.Subnets is the ultimate which contain the stale and updated subnets, so it override all.

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,249 @@
...
+	obj := &v1alpha2.IPPool{}
+	log.Info("reconciling ippool CR", "ippool", req.NamespacedName)
+	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerSyncTotal, MetricResType)
+	if err := r.Client.Get(ctx, req.NamespacedName, obj); err != nil {
+		log.Error(err, "unable to fetch ippool CR", "req", req.NamespacedName)
+		return resultNormal, client.IgnoreNotFound(err)
+	}
+
+	// TODO: Since IPPool v1alpha2 service can only be activated from NSX 4.1.0 onwards,
+	// So need to check NSX version before starting IPPool reconcile
```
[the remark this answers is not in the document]
zhengxiexie: `	//if !r.Service.NSXClient.NSXCheckVersionForStaticRoute() {
`
It is in another PR, I would uncomment it when the other PR checked in.

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,249 @@
...
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	obj := &v1alpha2.IPPool{}
+	log.Info("reconciling ippool CR", "ippool", req.NamespacedName)
+	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerSyncTotal, MetricResType)
+	if err := r.Client.Get(ctx, req.NamespacedName, obj); err != nil {
+		log.Error(err, "unable to fetch ippool CR", "req", req.NamespacedName)
+		return resultNormal, client.IgnoreNotFound(err)
+	}
+
+	// TODO: Since IPPool v1alpha2 service can only be activated from NSX 4.1.0 onwards,
```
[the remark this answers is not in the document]
zhengxiexie: Yun has done it.

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,249 @@
...
+	err := r.SetupWithManager(mgr)
+	if err != nil {
+		return err
+	}
+	go r.IPPoolGarbageCollector(make(chan bool), servicecommon.GCInterval)
+	return nil
+}
+
+// IPPoolGarbageCollector collect ippool which has been removed from crd.
+// cancel is used to break the loop during UT
```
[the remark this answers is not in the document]
zhengxiexie: All the CRDs copy it from securitypolicy which Tao implemented, @TaoZou1 do you have any comments?

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,251 @@
...
+		ns := service.GetIPPoolNamespace(nsxIPPool)
+		VPCInfo := commonctl.ServiceMediator.GetVPCInfo(ns)
+		if len(VPCInfo) == 0 {
+			err = util.NoEffectiveOption{Desc: "no effective org and project for ippool"}
+		} else {
+			err = service.NSXClient.ProjectInfraClient.Patch(VPCInfo[0].OrgID, VPCInfo[0].ProjectID, *infraIPPool,
+				&EnforceRevisionCheckParam)
+		}
+	} else if IPPoolType == common.IPPoolTypePublic {
+		err = service.NSXClient.InfraClient.Patch(*infraIPPool, &EnforceRevisionCheckParam)
```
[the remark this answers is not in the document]
zhengxiexie: Yes, refer buildIPSubnetIntentPath.

pkg/nsx/services/ippool/ippool.go
zhengxiexie: It must be a pointer.

pkg/nsx/services/ippool/ippool.go
zhengxiexie: OK

pkg/controllers/ippool/ippool_controller.go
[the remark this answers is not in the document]
dantingl: Then better remove TODO here

pkg/nsx/services/ippool/parse.go
```diff
@@ -0,0 +1,45 @@
...
+		if !realized {
+			unrealizedSubnets = append(unrealizedSubnets, subnetRequest.Name)
+		}
+	}
+	return unrealizedSubnets
+}
+
+func (service *IPPoolService) FullyRealized(obj *v1alpha2.IPPool) bool {
+	fullyRealized := true
+	for _, subnetRequest := range obj.Spec.Subnets {
```
TaoZou1, line 26: this function could reuse GetUnrealizedSubnetNames, if len(unrealizedSubnets) == 0, return true

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,251 @@
...
+	case *v1alpha2.IPPool:
+		nsxIPPool, nsxIPSubnets = service.BuildIPPool(o)
+		if err != nil {
+			log.Error(err, "failed to build ip pool", "IPPool", o)
+			return err
+		}
+	case types.UID:
+		nsxIPPool, nsxIPSubnets, err = service.indexedIPPoolAndIPPoolSubnets(o)
+		if err != nil {
+			log.Error(err, "failed to get ip pool and ip pool subnets by UID", "UID", o)
```
TaoZou1, line 196: if ip pool found, but ip pool subnet not found,  or just subnet found, no ip pool found, do we also need to delete resources?

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,249 @@
...
+				"error occurred while processing the IPPool CR. Error: %v",
+				*err,
+			),
+		},
+	}
+	ippool.Status.Conditions = conditions
+	if ippool.Status.Subnets == nil {
+		ippool.Status.Subnets = make([]v1alpha2.SubnetResult, 0)
+	}
+	e := r.Client.Status().Update(*ctx, ippool)
```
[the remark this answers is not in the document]
zhengxiexie: I think no need to check since there is little overhead to update the status.

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,249 @@
...
+	if e != nil {
+		log.Error(e, "unable to update IPPool status", "ippool", ippool)
+	}
```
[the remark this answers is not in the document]
zhengxiexie: it's better to log in Reconcile, since updateSuccess updateFail ... these functions are common, no need to log everywhere.

pkg/nsx/services/ippool/builder.go
```diff
@@ -0,0 +1,109 @@
...
+		{Scope: String(common.TagScopeCluster), Tag: String(getCluster(service))},
+		{Scope: String(common.TagScopeNamespace), Tag: String(IPPool.ObjectMeta.Namespace)},
+		{Scope: String(common.TagScopeIPPoolCRName), Tag: String(IPPool.ObjectMeta.Name)},
+		{Scope: String(common.TagScopeIPPoolCRUID), Tag: String(string(IPPool.UID))},
+		{Scope: String(common.TagScopeIPPoolCRType), Tag: String(IPPool.Spec.Type)},
+	}
+}
+
+func (service *IPPoolService) buildIPSubnets(IPPool *v1alpha2.IPPool) []*model.IpAddressPoolBlockSubnet {
+	var IPSubnets []*model.IpAddressPoolBlockSubnet
```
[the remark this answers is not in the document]
zhengxiexie: validate what?

pkg/nsx/services/ippool/parse.go
zhengxiexie: good idea.

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	if err := service.Operate(nsxIPPool, nsxIPSubnets, true, true); err != nil {
+		return err
+	}
+	log.V(1).Info("successfully deleted nsxIPPool", "nsxIPPool", nsxIPPool)
+	return nil
+}
+
+func (service *IPPoolService) acquireCidr(obj *v1alpha2.IPPool, subnetRequest *v1alpha2.SubnetRequest, retry int) (string, error) {
+	m, err := service.NSXClient.RealizedEntitiesClient.List(service.buildIPSubnetIntentPath(obj, subnetRequest), nil)
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,251 @@
...
+		Indexer:     cache.NewIndexer(keyFunc, cache.Indexers{common.TagScopeIPPoolCRUID: indexFunc}),
+		BindingType: model.IpAddressPoolBlockSubnetBindingType(),
+	}}
+
+	go ipPoolService.InitializeResourceStore(&wg, fatalErrors, ResourceTypeIPPool, nil, ipPoolService.ipPoolStore)
+	tags := []model.Tag{
+		{Scope: String(common.TagScopeIPSubnetOwner), Tag: String(ResourceTypeIPPool)},
+	}
+	go ipPoolService.InitializeResourceStore(&wg, fatalErrors, ResourceTypeIPPoolBlockSubnet, tags, ipPoolService.ipPoolBlockSubnetStore)
+
```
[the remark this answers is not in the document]
zhengxiexie: we can take it into consideration later.

pkg/nsx/services/ippool/ippool.go
zhengxiexie: we can take them as independent resources currently.

pkg/nsx/services/common/store.go
```diff
@@ -123,23 +123,30 @@ func TransError(err error) error {
...
 }
 
 // InitializeCommonStore is the common method used by InitializeResourceStore and InitializeVPCResourceStore
-func (service *Service) InitializeCommonStore(wg *sync.WaitGroup, fatalErrors chan error, org string, project string, resourceTypeValue string, store Store) {
+func (service *Service) InitializeCommonStore(wg *sync.WaitGroup, fatalErrors chan error, org string, project string, resourceTypeValue string, tags []model.Tag, store Store) {
 	defer wg.Done()
 
 	tagScopeClusterKey := strings.Replace(TagScopeCluster, "/", "\\/", -1)
 	tagScopeClusterValue := strings.Replace(service.NSXClient.NsxConfig.Cluster, ":", "\\:", -1)
 	tagParam := fmt.Sprintf("tags.scope:%s AND tags.tag:%s", tagScopeClusterKey, tagScopeClusterValue)
```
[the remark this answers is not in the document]
zhengxiexie: append to tagParam, include cluster already

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -0,0 +1,249 @@
...
+
+		subnetCidrUpdated, ipPoolSubnetsUpdated, err := r.Service.CreateOrUpdateIPPool(obj)
+		if err != nil {
+			log.Error(err, "operate failed, would retry exponentially", "ippool", req.NamespacedName)
+			updateFail(r, &ctx, obj, &err)
+			return resultRequeue, err
+		}
+		if !r.Service.FullyRealized(obj) {
+			if subnetCidrUpdated || ipPoolSubnetsUpdated || len(obj.Spec.Subnets) == 0 {
+				updateSuccess(r, &ctx, obj)
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,251 @@
...
+		close(fatalErrors)
+		return ipPoolService, err
+	}
+	return ipPoolService, nil
+}
+
+func (service *IPPoolService) CreateOrUpdateIPPool(obj *v1alpha2.IPPool) (bool, bool, error) {
+	nsxIPPool, nsxIPSubnets := service.BuildIPPool(obj)
+	existingIPPool, existingIPSubnets, err := service.indexedIPPoolAndIPPoolSubnets(obj.UID)
+	log.V(1).Info("existing ippool and ip subnets", "existingIPPool", existingIPPool, "existingIPSubnets", existingIPSubnets)
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/controllers/ippool/ippool_controller.go
zhengxiexie: TODO added

pkg/nsx/services/ippool/ippool.go
zhengxiexie: As disscussed, deferred.

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,256 @@
...
+	ipPoolService.ipPoolStore = &IPPoolStore{ResourceStore: common.ResourceStore{
+		Indexer:     cache.NewIndexer(keyFunc, cache.Indexers{common.TagScopeIPPoolCRUID: indexFunc}),
+		BindingType: model.IpAddressPoolBindingType(),
+	}}
+	ipPoolService.ipPoolBlockSubnetStore = &IPPoolBlockSubnetStore{ResourceStore: common.ResourceStore{
+		Indexer:     cache.NewIndexer(keyFunc, cache.Indexers{common.TagScopeIPPoolCRUID: indexFunc}),
+		BindingType: model.IpAddressPoolBlockSubnetBindingType(),
+	}}
+
+	go ipPoolService.InitializeResourceStore(&wg, fatalErrors, ResourceTypeIPPool, nil, ipPoolService.ipPoolStore)
```
lxiaopei, line 52: For VPCSubnet, it also has IP Pool, so this code will also initialize IPAddressPool for VPCSubnet in ippoolStore. Need to filter with tags.

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,256 @@
...
+		if len(VPCInfo) == 0 {
+			err = util.NoEffectiveOption{Desc: "no effective org and project for ippool"}
+		} else {
+			err = service.NSXClient.ProjectInfraClient.Patch(VPCInfo[0].OrgID, VPCInfo[0].ProjectID, *infraIPPool,
+				&EnforceRevisionCheckParam)
+		}
+	} else if IPPoolType == common.IPPoolTypePublic {
+		err = service.NSXClient.InfraClient.Patch(*infraIPPool, &EnforceRevisionCheckParam)
+	} else {
+		err = util.NoEffectiveOption{Desc: "no effective ippool type"}
```
lxiaopei, line 133: not valid IPpool type

pkg/nsx/services/ippool/builder.go
```diff
@@ -0,0 +1,109 @@
...
+func (service *IPPoolService) BuildIPPool(IPPool *v1alpha2.IPPool) (*model.IpAddressPool, []*model.IpAddressPoolBlockSubnet) {
+	return &model.IpAddressPool{
+		Id:          String(service.buildIPPoolID(IPPool)),
+		DisplayName: String(service.buildIPPoolName(IPPool)),
+		Tags:        service.buildIPPoolTags(IPPool),
+	}, service.buildIPSubnets(IPPool)
+}
+
+func (service *IPPoolService) buildIPPoolID(IPPool *v1alpha2.IPPool) string {
+	return strings.Join([]string{"ipc", string(IPPool.UID)}, "_")
```
lxiaopei, line 29: prefix "ipc" stands for?

pkg/nsx/services/ippool/builder.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (service *IPPoolService) buildIPSubnetTags(IPPool *v1alpha2.IPPool, subnetRequest *v1alpha2.SubnetRequest) []model.Tag {
+	return []model.Tag{
+		{Scope: String(common.TagScopeCluster), Tag: String(getCluster(service))},
+		{Scope: String(common.TagScopeNamespace), Tag: String(IPPool.ObjectMeta.Namespace)},
+		{Scope: String(common.TagScopeIPPoolCRName), Tag: String(IPPool.ObjectMeta.Name)},
+		{Scope: String(common.TagScopeIPPoolCRUID), Tag: String(string(IPPool.UID))},
+		{Scope: String(common.TagScopeIPSubnetName), Tag: String(subnetRequest.Name)},
+		{Scope: String(common.TagScopeIPSubnetOwner), Tag: String(ResourceTypeIPPool)},
```
lxiaopei, line 72: how about using tag "created_for": "ip_pool_crd" ?

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,262 @@
...
+func (service *IPPoolService) Operate(nsxIPPool *model.IpAddressPool, nsxIPSubnets []*model.IpAddressPoolBlockSubnet, IPPoolUpdated bool, IPPoolSubnetsUpdated bool) error {
+	if !(IPPoolUpdated || IPPoolSubnetsUpdated) {
+		return nil
+	}
+	infraIPPool, err := service.WrapHierarchyIPPool(nsxIPPool, nsxIPSubnets)
+	if err != nil {
+		return err
+	}
+	// Get IPPool Type from nsxIPPool
+	IPPoolType := common.IPPoolTypePublic
```
lxiaopei, line 113: in IPPool CR, the default value for type is Private, so change #113 to Private.

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,262 @@
...
+		}
+	}
+
+	if IPPoolType == common.IPPoolTypePrivate {
+		ns := service.GetIPPoolNamespace(nsxIPPool)
+		VPCInfo := commonctl.ServiceMediator.ListVPCInfo(ns)
+		if len(VPCInfo) == 0 {
+			err = util.NoEffectiveOption{Desc: "no effective org and project for ippool"}
+		} else {
+			err = service.NSXClient.ProjectInfraClient.Patch(VPCInfo[0].OrgID, VPCInfo[0].ProjectID, *infraIPPool,
```
lxiaopei, line 127: Can we use IpPoolsClient under project infra? I see other PRs use  Resource clients, for example StaticRoutesClient.

pkg/nsx/services/ippool/builder.go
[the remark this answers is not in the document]
lxiaopei: for example, prefix using IPv4 cannot be > 31.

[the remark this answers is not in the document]
zhengxiexie: Since only the cloud provider creates it, we can take all the validation logic into consideration later.

pkg/nsx/services/ippool/ippool.go
zhengxiexie: ok

pkg/nsx/services/ippool/builder.go
zhengxiexie: borrow it from ncp

pkg/nsx/services/ippool/ippool.go
zhengxiexie: ok

pkg/nsx/services/ippool/ippool.go
zhengxiexie: ProjectInfraClient use hierarchy mode

zhengxiexie:
Tao's suggestions:
Should we consider some Orphan subnets may exist？

pkg/nsx/services/ippool/ippool.go
zhengxiexie: fixed.

pkg/nsx/services/ippool/builder.go
zhengxiexie: fixed

pkg/nsx/services/ippool/builder.go
lxiaopei: ok, could we define a constant for the "ipc" prefix?

pkg/nsx/services/ippool/builder.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/vmware-tanzu/nsx-operator/pkg/util"
+)
+
+var (
+	Int64  = common.Int64
+	String = common.String
+)
+
+const (
+	GUESTCLUSTER = "guest_cluster"
```
lxiaopei, line 21: I do not have strong opinion for the value, need check with @dantingl

pkg/nsx/services/common/types.go
```diff
@@ -28,38 +28,53 @@ const (
...
 	TagScopeGroupType               string = "nsx-op/group_type"
 	TagScopeSelectorHash            string = "nsx-op/selector_hash"
-	TagScopeNSXServiceAccountCRName string = "nsx-op/nsx_service_account_name"
-	TagScopeNSXServiceAccountCRUID  string = "nsx-op/nsx_service_account_uid"
+	TagScopeIPPoolCRName            string = "nsx-op/ippool_cr_name"
+	TagScopeIPPoolCRUID             string = "nsx-op/ippool_cr_uid"
+	TagScopeIPPoolCRType            string = "nsx-op/ippool_cr_type"
+	TagScopeIPPoolCreatedFor        string = "nsx-op/ippool_created_for"
+	TagScopeIPSubnetName            string = "nsx-op/ipsubnet_cr_name"
+	TagScopeIPSubnetCreatedFor      string = "nsx-op/ipsubnet_created_for"
```
lxiaopei, line 36: can we use nsx-op/created_for as tag scope ?

pkg/nsx/services/ippool/builder.go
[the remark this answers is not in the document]
lxiaopei: could you summarize all the items which you plan to consider later in following patch so that we do not forget them?

pkg/nsx/services/ippool/ippool.go
```diff
@@ -0,0 +1,265 @@
...
+	nsxIPSubnets := make([]*model.IpAddressPoolBlockSubnet, 0)
+	switch o := obj.(type) {
+	case *v1alpha2.IPPool:
+		nsxIPPool, nsxIPSubnets = service.BuildIPPool(o)
+		if err != nil {
+			log.Error(err, "failed to build ip pool", "IPPool", o)
+			return err
+		}
+	case types.UID:
+		nsxIPPool, nsxIPSubnets, err = service.indexedIPPoolAndIPPoolSubnets(o)
```
lxiaopei, line 198: is there possibility that IPPool was deleted from nsx store but NSX block subnet was not deleted?

zhengxiexie: I think not, since mark deleted in ippool and IP block subnets and used in a singe hierarchy patch, the operation is atmic.

pkg/nsx/services/common/types.go
```diff
@@ -28,38 +28,53 @@ const (
...
 	TagScopeNCPProject              string = "ncp/project"
 	TagScopeNCPVIFProject           string = "ncp/vif_project"
 	TagScopeNCPPod                  string = "ncp/pod"
 	TagScopeNCPVNETInterface        string = "ncp/vnet_interface"
 	TagScopeVPCCRName               string = "nsx-op/vpc_cr_name"
 	TagScopeVPCCRUID                string = "nsx-op/vpc_cr_uid"
-	TagScopeSubnetPortCRName        string = "nsx-op/subnetport_cr_name"
-	TagScopeSubnetPortCRUID         string = "nsx-op/subnetport_cr_uid"
-	LabelDefaultSubnetSet           string = "nsxoperator.vmware.com/default-subnetset-for"
+	TagScopeNSXServiceAccountCRName string = "nsx-op/nsx_service_account_name"
```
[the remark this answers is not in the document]
zhengxiexie: Oh, my bad, fixed.

pkg/nsx/services/ippool/builder.go
zhengxiexie: OK

pkg/nsx/services/ippool/builder.go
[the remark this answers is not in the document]
zhengxiexie: OK, I have #TODO

pkg/nsx/services/common/types.go
zhengxiexie: ok

pkg/nsx/services/vpc/vpc.go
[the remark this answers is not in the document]
zhengxiexie: I have added	// TODO: remove this after vpc is ready

pkg/nsx/services/ippool/ippool.go
lxiaopei: Yea, the operation to mark for delete for IP Pool and IPPool subnet is in a single hierarchy patch, but the operation to operate ip pool store and ip pool subnet store is not.  
I think we need to add a TODO for this corner case and fix it in following patch.

zhengxiexie: ok

lxiaopei (commented):
Let's track TODOs in next patch. LGTM.

lxiaopei:
could you squash all your commits into one?

lxiaopei (approved):

PR #165, opened 2022-12-22
Implement subnet/subnetset CRD

jwsui:
Add subnet/subnetset controller to reconcile subnet/subnetset CRD. And add subnet service to perform NSX subnet CRUD operations.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -20,7 +20,9 @@ import (
 	nsxserviceaccountcontroller "github.com/vmware-tanzu/nsx-operator/pkg/controllers/nsxserviceaccount"
 	securitypolicycontroller "github.com/vmware-tanzu/nsx-operator/pkg/controllers/securitypolicy"
 	staticroutecontroller "github.com/vmware-tanzu/nsx-operator/pkg/controllers/staticroute"
+	"github.com/vmware-tanzu/nsx-operator/pkg/controllers/subnet"
 	"github.com/vmware-tanzu/nsx-operator/pkg/controllers/subnetport"
+	"github.com/vmware-tanzu/nsx-operator/pkg/controllers/subnetset"
 	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
 	"github.com/vmware-tanzu/nsx-operator/pkg/metrics"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx"
@@ -136,6 +138,13 @@ func main() {
 	// Start the NSXServiceAccount controller.
 	if cf.EnableAntreaNSXInterworking {
 		StartNSXServiceAccountController(mgr, commonService)
+	}
+	// Start subnet/subnetset controller.
+	if err := subnet.StartSubnetController(mgr, commonService); err != nil {
+		os.Exit(1)
+	}
+	if err := subnetset.StartSubnetSetController(mgr, commonService); err != nil {
+		os.Exit(1)
 	}
 
 	subnetport.StartSubnetPortController(mgr, commonService)
--- a/pkg/controllers/subnet/subnet_controller.go
+++ b/pkg/controllers/subnet/subnet_controller.go
@@ -1,0 +1,230 @@
+package subnet
+
+import (
+	"context"
+	"reflect"
+	"runtime"
+
+	v1 "k8s.io/api/core/v1"
+	apimachineryruntime "k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
+	"github.com/vmware-tanzu/nsx-operator/pkg/controllers/common"
+	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
+	"github.com/vmware-tanzu/nsx-operator/pkg/metrics"
+	servicecommon "github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/subnet"
+)
+
+var (
+	log                     = logger.Log
+	ResultNormal            = common.ResultNormal
+	ResultRequeue           = common.ResultRequeue
+	ResultRequeueAfter5mins = common.ResultRequeueAfter5mins
+	MetricResTypeSubnet     = common.MetricResTypeSubnet
+)
+
+// SubnetReconciler reconciles a SubnetSet object
+type SubnetReconciler struct {
+	Client  client.Client
+	Scheme  *apimachineryruntime.Scheme
+	Service *subnet.SubnetService
+}
+
+func (r *SubnetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	obj := &v1alpha1.Subnet{}
+	log.Info("reconciling subnet CR", "subnet", req.NamespacedName)
+	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerSyncTotal, MetricResTypeSubnet)
+
+	if err := r.Client.Get(ctx, req.NamespacedName, obj); err != nil {
+		log.Error(err, "unable to fetch Subnet CR", "req", req.NamespacedName)
+		return ResultNormal, client.IgnoreNotFound(err)
+	}
+
+	if obj.ObjectMeta.DeletionTimestamp.IsZero() {
+		metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerUpdateTotal, MetricResTypeSubnet)
+		if !controllerutil.ContainsFinalizer(obj, servicecommon.FinalizerName) {
+			controllerutil.AddFinalizer(obj, servicecommon.FinalizerName)
+			if err := r.Client.Update(ctx, obj); err != nil {
+				log.Error(err, "add finalizer", "subnet", req.NamespacedName)
+				updateFail(r, &ctx, obj)
+				return ResultRequeue, err
+			}
+			log.V(1).Info("added finalizer on subnet CR", "subnet", req.NamespacedName)
+		}
+		if _, err := r.Service.CreateOrUpdateSubnet(obj, nil); err != nil {
+			log.Error(err, "operate failed, would retry exponentially", "subnet", req.NamespacedName)
+			updateFail(r, &ctx, obj)
+			return ResultRequeue, err
+		}
+		if err := r.updateSubnetStatus(obj); err != nil {
+			log.Error(err, "update subnet status failed, would retry exponentially", "subnet", req.NamespacedName)
+			updateFail(r, &ctx, obj)
+			return ResultRequeue, err
+		}
+		updateSuccess(r, &ctx, obj)
+	} else {
+		if controllerutil.ContainsFinalizer(obj, servicecommon.FinalizerName) {
+			metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteTotal, MetricResTypeSubnet)
+			if err := r.Service.DeleteSubnet(*obj); err != nil {
+				log.Error(err, "deletion failed, would retry exponentially", "subnet", req.NamespacedName)
+				deleteFail(r, &ctx, obj)
+				return ResultRequeue, err
+			}
+			controllerutil.RemoveFinalizer(obj, servicecommon.FinalizerName)
+			if err := r.Client.Update(ctx, obj); err != nil {
+				log.Error(err, "deletion failed, would retry exponentially", "subnet", req.NamespacedName)
+				deleteFail(r, &ctx, obj)
+				return ResultRequeue, err
+			}
+			log.V(1).Info("removed finalizer", "subnet", req.NamespacedName)
+			deleteSuccess(r, &ctx, obj)
+		} else {
+			log.Info("finalizers cannot be recognized", "subnet", req.NamespacedName)
+		}
+	}
+	return ctrl.Result{}, nil
+}
+
+func (r *SubnetReconciler) updateSubnetStatus(obj *v1alpha1.Subnet) error {
+	nsxSubnet := r.Service.SubnetStore.GetByKey(string(obj.GetUID()))
+	obj.Status.IPAddresses = obj.Status.IPAddresses[:0]
+	statusList, err := r.Service.GetSubnetStatus(nsxSubnet)
+	if err != nil {
+		return err
+	}
+	for _, status := range statusList {
+		obj.Status.IPAddresses = append(obj.Status.IPAddresses, *status.NetworkAddress)
+	}
+	obj.Status.NSXResourcePath = *nsxSubnet.Path
+	return nil
+}
+
+func (r *SubnetReconciler) setSubnetReadyStatusTrue(ctx *context.Context, subnet *v1alpha1.Subnet) {
+	newConditions := []v1alpha1.Condition{
+		{
+			Type:    v1alpha1.Ready,
+			Status:  v1.ConditionTrue,
+			Message: "NSX Subnet has been successfully created/updated",
+			Reason:  "SubnetCreated",
+		},
+	}
+	r.updateSubnetStatusConditions(ctx, subnet, newConditions)
+}
+
+func (r *SubnetReconciler) setSubnetReadyStatusFalse(ctx *context.Context, subnet *v1alpha1.Subnet) {
+	newConditions := []v1alpha1.Condition{
+		{
+			Type:    v1alpha1.Ready,
+			Status:  v1.ConditionFalse,
+			Message: "NSX Subnet could not be created/updated",
+			Reason:  "SubnetNotReady",
+		},
+	}
+	r.updateSubnetStatusConditions(ctx, subnet, newConditions)
+}
+
+func (r *SubnetReconciler) updateSubnetStatusConditions(ctx *context.Context, subnet *v1alpha1.Subnet, newConditions []v1alpha1.Condition) {
+	conditionsUpdated := false
+	for i := range newConditions {
+		if r.mergeSubnetStatusCondition(ctx, subnet, &newConditions[i]) {
+			conditionsUpdated = true
+		}
+	}
+	if conditionsUpdated {
+		r.Client.Status().Update(*ctx, subnet)
+		log.V(1).Info("updated Subnet", "Name", subnet.Name, "Namespace", subnet.Namespace,
+			"New Conditions", newConditions)
+	}
+}
+
+func (r *SubnetReconciler) mergeSubnetStatusCondition(ctx *context.Context, subnet *v1alpha1.Subnet, newCondition *v1alpha1.Condition) bool {
+	matchedCondition := getExistingConditionOfType(newCondition.Type, subnet.Status.Conditions)
+
+	if reflect.DeepEqual(matchedCondition, newCondition) {
+		log.V(2).Info("conditions already match", "New Condition", newCondition, "Existing Condition", matchedCondition)
+		return false
+	}
+
+	if matchedCondition != nil {
+		matchedCondition.Reason = newCondition.Reason
+		matchedCondition.Message = newCondition.Message
+		matchedCondition.Status = newCondition.Status
+	} else {
+		subnet.Status.Conditions = append(subnet.Status.Conditions, *newCondition)
+	}
+	return true
+}
+
+func getExistingConditionOfType(conditionType v1alpha1.ConditionType, existingConditions []v1alpha1.Condition) *v1alpha1.Condition {
+	for i := range existingConditions {
+		if existingConditions[i].Type == conditionType {
+			return &existingConditions[i]
+		}
+	}
+	return nil
+}
+
+func updateFail(r *SubnetReconciler, c *context.Context, o *v1alpha1.Subnet) {
+	r.setSubnetReadyStatusFalse(c, o)
+	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerUpdateFailTotal, MetricResTypeSubnet)
+}
+
+func deleteFail(r *SubnetReconciler, c *context.Context, o *v1alpha1.Subnet) {
+	r.setSubnetReadyStatusFalse(c, o)
+	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteFailTotal, MetricResTypeSubnet)
+}
+
+func updateSuccess(r *SubnetReconciler, c *context.Context, o *v1alpha1.Subnet) {
+	r.setSubnetReadyStatusTrue(c, o)
+	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerUpdateSuccessTotal, MetricResTypeSubnet)
+}
+
+func deleteSuccess(r *SubnetReconciler, _ *context.Context, _ *v1alpha1.Subnet) {
+	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteSuccessTotal, MetricResTypeSubnet)
+}
+
+func (r *SubnetReconciler) setupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&v1alpha1.Subnet{}).
+		WithEventFilter(predicate.Funcs{
+			DeleteFunc: func(e event.DeleteEvent) bool {
+				// Suppress Delete events to avoid filtering them out in the Reconcile function
+				return false
+			},
+		}).
+		WithOptions(
+			controller.Options{
+				MaxConcurrentReconciles: runtime.NumCPU(),
+			}).
+		Complete(r)
+}
+
+func StartSubnetController(mgr ctrl.Manager, commonService servicecommon.Service) error {
+	subnetReconciler := &SubnetReconciler{
+		Client: mgr.GetClient(),
+		Scheme: mgr.GetScheme(),
+	}
+	subnetReconciler.Service = subnet.GetSubnetService(commonService)
+	common.ServiceMediator.SubnetService = subnetReconciler.Service
+	if err := subnetReconciler.Start(mgr); err != nil {
+		log.Error(err, "failed to create controller", "controller", "Subnet")
+		return err
+	}
+	return nil
+}
+
+// Start setup manager
+func (r *SubnetReconciler) Start(mgr ctrl.Manager) error {
+	err := r.setupWithManager(mgr)
+	if err != nil {
+		return err
+	}
+	return nil
+}
--- a/pkg/controllers/subnetset/subnetport_handler.go
+++ b/pkg/controllers/subnetset/subnetport_handler.go
@@ -1,0 +1,88 @@
+package subnetset
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+)
+
+// SubnetPortHandler supports lazy-creation of Subnet, the first Subnet won't
+// be created until there is a SubnetPort attached to it.
+// - SubnetPort creation: get available Subnet for the SubnetPort, create new
+//   Subnet if necessary.
+// - SubnetPort deletion: if recycling Subnet is required, delete Subnets without
+// SubnetPort attached to it.
+
+type SubnetPortHandler struct {
+	Reconciler *SubnetSetReconciler
+}
+
+//TODO Remove this handler when confirmed that SubnetPort could get allocated subnet via the interface
+// subnetservice.GetAvailableSubnet
+
+// Create allocates Subnet for SubnetPort from SubnetSet.
+func (h *SubnetPortHandler) Create(e event.CreateEvent, _ workqueue.RateLimitingInterface) {
+	log.V(4).Info("SubnetPort generic event, do nothing")
+	//subnetPort := e.Object.(*v1alpha1.SubnetPort)
+	//if subnetPort.Spec.Subnet != "" {
+	//	// Two possible scenarios:
+	//	// - 1. User uses `.Spec.Subnet` directly instead of `.Spec.SubnetSet`.
+	//	// - 2. Subnet has been allocated and `.Spec.Subnet` is rendered by SubnetPortHandler.
+	//	return
+	//}
+	//subnetSet := &v1alpha1.SubnetSet{}
+	//key := types.NamespacedName{
+	//	Namespace: subnetPort.GetNamespace(),
+	//	Name:      subnetPort.Spec.SubnetSet,
+	//}
+	//if err := h.Reconciler.Client.Get(context.Background(), key, subnetSet); err != nil {
+	//	log.Error(err, "failed to get SubnetSet", "ns", key.Namespace, "name", key.Name)
+	//	return
+	//}
+	//log.Info("allocating Subnet for SubnetPort")
+	//vpcList := &v1alpha1.VPCList{}
+	//if err := h.Reconciler.Client.List(context.Background(), vpcList, client.InNamespace(subnetPort.GetNamespace())); err != nil {
+	//	log.Error(err, fmt.Sprintf("failed to get VPC under namespace: %s.\n", subnetPort.GetNamespace()))
+	//	return
+	//}
+	//vpcInfo, err := servicecommon.ParseVPCResourcePath(vpcList.Items[0].Status.NSXResourcePath)
+	//if err != nil {
+	//	log.Error(err, "failed to resolve VPC info")
+	//	return
+	//}
+	//_, err = h.Reconciler.getAvailableSubnet(subnetSet, &vpcInfo)
+	//if err != nil {
+	//	log.Error(err, "failed to allocate Subnet")
+	//}
+	// TODO return subnetport id to caller.
+}
+
+// Delete TODO Implement this method if required to recycle Subnet without SubnetPort attached.
+func (h *SubnetPortHandler) Delete(e event.DeleteEvent, _ workqueue.RateLimitingInterface) {
+	log.V(4).Info("SubnetPort generic event, do nothing")
+}
+
+func (h *SubnetPortHandler) Generic(_ event.GenericEvent, _ workqueue.RateLimitingInterface) {
+	log.V(4).Info("SubnetPort generic event, do nothing")
+}
+
+func (h *SubnetPortHandler) Update(_ event.UpdateEvent, _ workqueue.RateLimitingInterface) {
+	log.V(4).Info("SubnetPort update event, do nothing")
+}
+
+var SubnetPortPredicate = predicate.Funcs{
+	CreateFunc: func(e event.CreateEvent) bool {
+		return true
+	},
+	UpdateFunc: func(e event.UpdateEvent) bool {
+		return false
+	},
+	DeleteFunc: func(e event.DeleteEvent) bool {
+		// TODO When recycling Subnet is required, return true.
+		return false
+	},
+	GenericFunc: func(genericEvent event.GenericEvent) bool {
+		return false
+	},
+}
--- a/pkg/controllers/subnetset/subnetset_controller.go
+++ b/pkg/controllers/subnetset/subnetset_controller.go
@@ -1,0 +1,236 @@
+package subnetset
+
+import (
+	"context"
+	"reflect"
+	"runtime"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	apimachineryruntime "k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
+	"github.com/vmware-tanzu/nsx-operator/pkg/controllers/common"
+	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
+	"github.com/vmware-tanzu/nsx-operator/pkg/metrics"
+	servicecommon "github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/subnet"
+)
+
+var (
+	log                     = logger.Log
+	ResultNormal            = common.ResultNormal
+	ResultRequeue           = common.ResultRequeue
+	ResultRequeueAfter5mins = common.ResultRequeueAfter5mins
+	MetricResTypeSubnetSet  = common.MetricResTypeSubnetSet
+	//TODO rename this
+	defaultSubnet = "default-subnet"
+)
+
+// SubnetSetReconciler reconciles a SubnetSet object
+type SubnetSetReconciler struct {
+	Client  client.Client
+	Scheme  *apimachineryruntime.Scheme
+	Service *subnet.SubnetService
+}
+
+func (r *SubnetSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	obj := &v1alpha1.SubnetSet{}
+	log.Info("reconciling subnetset CR", "subnetset", req.NamespacedName)
+	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerSyncTotal, MetricResTypeSubnetSet)
+
+	if err := r.Client.Get(ctx, req.NamespacedName, obj); err != nil {
+		log.Error(err, "unable to fetch subnetset CR", "req", req.NamespacedName)
+		return ResultNormal, client.IgnoreNotFound(err)
+	}
+
+	if obj.ObjectMeta.DeletionTimestamp.IsZero() {
+		metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerUpdateTotal, MetricResTypeSubnetSet)
+		if !controllerutil.ContainsFinalizer(obj, servicecommon.FinalizerName) {
+			controllerutil.AddFinalizer(obj, servicecommon.FinalizerName)
+			if err := r.Client.Update(ctx, obj); err != nil {
+				log.Error(err, "add finalizer", "subnetset", req.NamespacedName)
+				updateFail(r, &ctx, obj)
+				return ResultRequeue, err
+			}
+			log.V(1).Info("added finalizer on subnetset CR", "subnetset", req.NamespacedName)
+		}
+		updateSuccess(r, &ctx, obj)
+	} else {
+		if controllerutil.ContainsFinalizer(obj, servicecommon.FinalizerName) {
+			metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteTotal, MetricResTypeSubnetSet)
+			if err := r.Service.DeleteSubnetForSubnetSet(*obj, false); err != nil {
+				log.Error(err, "deletion failed, would retry exponentially", "subnetset", req.NamespacedName)
+				deleteFail(r, &ctx, obj)
+				return ResultRequeue, err
+			}
+			controllerutil.RemoveFinalizer(obj, servicecommon.FinalizerName)
+			if err := r.Client.Update(ctx, obj); err != nil {
+				log.Error(err, "deletion failed, would retry exponentially", "subnetset", req.NamespacedName)
+				deleteFail(r, &ctx, obj)
+				return ResultRequeue, err
+			}
+			log.V(1).Info("removed finalizer", "subnetset", req.NamespacedName)
+			deleteSuccess(r, &ctx, obj)
+		} else {
+			log.Info("finalizers cannot be recognized", "subnetset", req.NamespacedName)
+		}
+	}
+	return ctrl.Result{}, nil
+}
+
+func updateFail(r *SubnetSetReconciler, c *context.Context, o *v1alpha1.SubnetSet) {
+	r.setSubnetSetReadyStatusFalse(c, o)
+	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerUpdateFailTotal, MetricResTypeSubnetSet)
+}
+
+func deleteFail(r *SubnetSetReconciler, c *context.Context, o *v1alpha1.SubnetSet) {
+	r.setSubnetSetReadyStatusFalse(c, o)
+	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteFailTotal, MetricResTypeSubnetSet)
+}
+
+func updateSuccess(r *SubnetSetReconciler, c *context.Context, o *v1alpha1.SubnetSet) {
+	r.setSubnetSetReadyStatusTrue(c, o)
+	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerUpdateSuccessTotal, MetricResTypeSubnetSet)
+}
+
+func deleteSuccess(r *SubnetSetReconciler, _ *context.Context, _ *v1alpha1.SubnetSet) {
+	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteSuccessTotal, MetricResTypeSubnetSet)
+}
+
+func (r *SubnetSetReconciler) setSubnetSetReadyStatusTrue(ctx *context.Context, subnetset *v1alpha1.SubnetSet) {
+	newConditions := []v1alpha1.Condition{
+		{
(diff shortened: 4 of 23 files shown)


pkg/nsx/services/subnet/subnet.go
```diff
@@ -0,0 +1,146 @@
...
+		SubnetStore: &SubnetStore{
+			ResourceStore: common.ResourceStore{
+				Indexer:     cache.NewIndexer(keyFunc, cache.Indexers{common.TagScopeSubnetCRUID: indexFunc}),
+				BindingType: model.VpcSubnetBindingType(),
+			},
+		},
+	}
+
+	wg.Add(1)
+	go subnetService.InitializeResourceStore(&wg, fatalErrors, ResourceTypeSubnet, subnetService.SubnetStore)
```
zhengxiexie, line 60: Perhaps this would not properly function, refer [https://github.com/vmware-tanzu/nsx-operator/pull/160/files#diff-56a01862f16c1cc9248df82cc17bcd7adcb1fe24442da3fb078cfabee9b9c7e9R158](url)

pkg/nsx/services/subnet/subnet.go
```diff
@@ -0,0 +1,146 @@
...
+type SubnetService struct {
+	common.Service
+	SubnetStore *SubnetStore
+}
+
+var subnetService *SubnetService
+
+// GetSubnetService get singleton SubnetService instance, subnet/subnetset controller share the same instance.
+// GetSubnetService isn't thread-safe.
+func GetSubnetService(service common.Service) *SubnetService {
```
zhengxiexie, line 33: No lock?  refer [https://refactoring.guru/design-patterns/singleton/go/example](url)

jwsui: There is no lock as the function wasn't expected to be thread-safe. But it's better to add a lock for this, thanks.

pkg/nsx/services/subnet/compare.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
+)
+
+type (
+	Subnet     model.VpcSubnet
+	Comparable common.Comparable
```
zhengxiexie, line 12: It should be
type Comparable = common.Comparable
because we have no other function added to it.

pkg/controllers/subnetset/namespace_handler.go
```diff
@@ -0,0 +1,43 @@
...
+	"context"
+
+	"k8s.io/client-go/util/retry"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
+)
+
```
zhengxiexie, line 13: It's better to add some comments to explain why we should add extra handler, like [https://github.com/vmware-tanzu/nsx-operator/blob/main/pkg/controllers/securitypolicy/pod_controller.go#L19](url)

zhengxiexie: Besides, the name handler is suitable, but how about complying with the security policy and renaming it to controller?

pkg/controllers/subnetset/vpc_handler.go
```diff
@@ -0,0 +1,75 @@
...
+	"k8s.io/client-go/util/retry"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
+)
+
+var defaultSubnets = []string{"DefaultVMSubnetSet", "DefaultPodSubnetSet"}
+
```
zhengxiexie, line 16: ditto

pkg/controllers/subnetset/namespace_handler.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+}
+
+func (h *NamespaceHandler) Generic(_ event.GenericEvent, _ workqueue.RateLimitingInterface) {
+	log.V(4).Info("namespace generic event, do nothing")
+}
+
+func (h *NamespaceHandler) Update(_ event.UpdateEvent, _ workqueue.RateLimitingInterface) {
+	log.V(4).Info("namespace update event, do nothing")
+}
```
zhengxiexie, line 43: I wonder if you need any predicate function like [https://github.com/vmware-tanzu/nsx-operator/blob/main/pkg/controllers/securitypolicy/pod_controller.go#L92](url)

pkg/controllers/subnetset/vpc_handler.go
```diff
@@ -0,0 +1,75 @@
...
+			log.Error(err, "failed to create SubnetSet", "Namespace", ns, "Name", subnetSet)
+		}
+	}
+}
+
+func (h *VPCHandler) Generic(_ event.GenericEvent, _ workqueue.RateLimitingInterface) {
+	log.V(4).Info("VPC generic event, do nothing")
+}
+
+func (h *VPCHandler) Update(_ event.UpdateEvent, _ workqueue.RateLimitingInterface) {
```
zhengxiexie, line 73: ditto

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -0,0 +1,225 @@
...
+	return ctrl.Result{}, nil
+}
+
+func (r *SubnetReconciler) updateSubnetStatus(obj *v1alpha1.Subnet) error {
+	nsxSubnets := r.Service.SubnetStore.GetByIndex(servicecommon.TagScopeSubnetCRUID, string(obj.UID))
+	if len(nsxSubnets) == 0 {
+		return errors.New("failed to get subnet from store")
+	}
+	obj.Status.IPAddresses = make([]string, len(nsxSubnets[0].IpAddresses))
+	for i, ip := range nsxSubnets[0].IpAddresses {
```
lxiaopei, line 102: for subnet auto allocated from VPC IPBlock, the NSX Subnet does not include ip_addressed field,
how do we ensure there is value for nsxSubnet.IpAddresses?

pkg/nsx/services/subnet/subnet.go
jwsui: Thanks, predicates added.

pkg/nsx/services/subnet/subnet.go
jwsui: Lock added.

pkg/controllers/subnetset/namespace_handler.go
jwsui: NamespaceHandler is removed and comments are added to VPCHandler and SubnetPortHandler. As for the name, I prefer `handler` personally as `handler` is considered as a part of `controller/reconciler`.

pkg/controllers/subnetset/vpc_handler.go
jwsui: Comments added, thanks.

pkg/controllers/subnetset/namespace_handler.go
jwsui: Thanks for the suggestion, predicate can help.

pkg/controllers/subnetset/vpc_handler.go
jwsui: Predicates added.

pkg/nsx/services/subnet/wrap.go
```diff
@@ -0,0 +1,140 @@
...
+	configChildValue, errors := NewConverter().ConvertToVapi(configChild, model.ChildDhcpStaticBindingConfigBindingType())
+	if len(errors) > 0 {
+		return nil, errors[0]
+	}
+	return []*data.StructValue{configChildValue.(*data.StructValue)}, nil
+}
+
+// wrapSegmentPort wraps SegmentPort as children of VPCSubnet.
+// A reserved function for creating SegmentPorts, currently not used.
+func (service *SubnetService) wrapSegmentPort(port model.SegmentPort) ([]*data.StructValue, error) {
```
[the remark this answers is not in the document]
jwsui: Agree, will remove this when we have a final decision.

pkg/nsx/services/subnet/wrap.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	dataValue, errors := NewConverter().ConvertToVapi(childSubnet, model.ChildVpcSubnetBindingType())
+	if len(errors) > 0 {
+		return nil, errors[0]
+	}
+	return []*data.StructValue{dataValue.(*data.StructValue)}, nil
+}
+
+// wrapSegmentPort wraps SegmentPort as children of VPCSubnet.
+// A reserved function for creating SegmentPorts, currently not used.
```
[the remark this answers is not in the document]
jwsui: Thanks, done.

pkg/nsx/services/subnet/wrap.go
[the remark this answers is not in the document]
jwsui: But there is another benefit of hierarchy api that it can help handle dependency of api endpoint.

pkg/controllers/subnet/subnet_controller.go
jwsui: Indeed, there is no ip_addresses for subnet auto allocated from VPC IPBlock. For this case, how about retrieving subnet status by api `orgs/{{org_id}}/projects/{{pro_id}}/vpcs/{{vpc_id}}/subnets/{{subnet_id}}/status`? Using field `network_address` and a sample result will be like this.

```
        {
            "static_ip_pool_path": "xxx/xxx",
            "static_ip_pool_range": "10.0.0.75-10.0.0.78",
            "gateway_address": "10.0.0.65/28",
            "network_address": "10.0.0.64/28",
            "dhcp_server_address": "10.0.0.66/28",
            "dhcp_ranges": "10.0.0.67-10.0.0.74",
            "ip_address_type": "IPV4"
        }

```

lxiaopei: Yea, we need to retrieve subnet CIDR and update it under status.

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -0,0 +1,376 @@
...
+			controllerutil.AddFinalizer(obj, servicecommon.FinalizerName)
+			if err := r.Client.Update(ctx, obj); err != nil {
+				log.Error(err, "add finalizer", "subnetset", req.NamespacedName)
+				updateFail(r, &ctx, obj, &err)
+				return ResultRequeue, err
+			}
+			log.V(1).Info("added finalizer on subnetset CR", "subnetset", req.NamespacedName)
+		}
+
+		// TODO Only for local test, Create default Subnet for SubnetSet.
```
lxiaopei, line 71: As we discussed, there is a design change, we only create the first NSX subnet when first PodVM and VM created.

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -0,0 +1,376 @@
...
+}
+
+// listSubnets lists all subnets under subnetset
+func (r *SubnetSetReconciler) listSubnets(subnetset *v1alpha1.SubnetSet) (*v1alpha1.SubnetList, error) {
+	subnets := &v1alpha1.SubnetList{}
+	if err := retry.OnError(retry.DefaultRetry, func(err error) bool {
+		return err != nil
+	}, func() error {
+		if err := r.Client.List(context.Background(), subnets,
+			client.MatchingFields{"metadata.ownerReferences": string(subnetset.UID)}); err != nil {
```
lxiaopei, line 187: for SubnetSet CR, do we create Subnet CR? I thought we create NSX subnet directly.

pkg/nsx/services/subnet/subnet.go
```diff
@@ -0,0 +1,164 @@
...
+	if err = service.SubnetStore.Operate(&subnetCopy); err != nil {
+		return err
+	}
+	log.Info("successfully deleted  nsxSubnet", "nsxSubnet", nsxSubnet)
+	return nil
+}
+
+// GetAvailableIPNum TODO: need refactor
+func (service *SubnetService) GetAvailableIPNum(subnet *v1alpha1.Subnet) (int64, error) {
+	if subnet.Spec.DHCPConfig.EnableDHCP {
```
lxiaopei, line 150: there is an API GET https://&lt;policy-mgr&gt;/policy/api/v1/orgs/org-1/projects/project-1/vpcs/vpc-1/subnets/subnet-1/ip-pools/default-ipv4, which can return pool_usage in the response. Could you investigate that if we can use that API to check the subnet usage?

lxiaopei:
also remember to squash all commits into a single commit for this patch.

pkg/controllers/subnetset/subnetset_controller.go
jwsui: yes, in current implementation, subnetset controller creates a subnet CR that belongs to subnetset and delegate the creation of NSX subnet to subnet controller. subnetset controller doesn't create NSX subnet directly.

pkg/controllers/subnetset/subnetset_controller.go
jwsui: yes, this is only for local test and will be removed finally.

pkg/nsx/services/subnet/subnet.go
jwsui: ok, I think you mean `/parent_path/subnets/subnet-1/ip-pools/static-ipv4-default`. There is a question that should we enable DHCP for subnet? Subnet's ip-pool is separated into `static-ipv4-default` ip-pool and DHCP ip-pool if DHCP of subnet is enabled.

jwsui: Updated to use static ip-pool `static-ipv4-default` to calculate available IP number as DHCP is not enabled in Subnet.

pkg/controllers/subnet/subnet_controller.go
jwsui: Thanks, code has been updated to retrieve subnet CIDR from api `../subnet/status` for scenario that IP addresses are auto allocated from VPC IPBlock.

jwsui:
> also remember to squash all commits into a single commit for this patch.

Thanks for reminder, separate commits could help debug in development， and will squash all commits when it's ready to merge.

pkg/nsx/services/subnet/compare.go
jwsui: Thanks for reminder, done.

pkg/nsx/services/subnet/subnet.go
```diff
@@ -0,0 +1,198 @@
...
+	if err != nil {
+		log.Error(err, "WrapHierarchySubnet failed")
+		return err
+	}
+	if err = service.NSXClient.OrgRootClient.Patch(*orgRoot, &EnforceRevisionCheckParam); err != nil {
+		return err
+	}
+	// Get Subnet from NSX after patch operation as NSX renders several fields like `path`/`parent_path`.
+	if *nsxSubnet, err = service.NSXClient.SubnetsClient.Get(orgID, projectID, vpcID, *nsxSubnet.Id); err != nil {
+		return err
```
lxiaopei, line 117: For VPC Subnet, after create it via Policy API, do we need to check its realization state? Is there possibility that the Policy intent succeeds, but the Subnet is realized failed?

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -0,0 +1,271 @@
...
+	return ctrl.Result{}, nil
+}
+
+func (r *SubnetReconciler) updateSubnetStatus(obj *v1alpha1.Subnet) error {
+	nsxSubnets := r.Service.SubnetStore.GetByIndex(servicecommon.TagScopeSubnetCRUID, string(obj.UID))
+	if len(nsxSubnets) == 0 {
+		return errors.New("failed to get subnet from store")
+	}
+	obj.Status.IPAddresses = obj.Status.IPAddresses[:0]
+	if len(obj.Spec.IPAddresses) > 0 {
```
lxiaopei, line 111: do we know if the specified IP_address Subnet is realized successfully or not?

pkg/controllers/subnetset/subnetport_handler.go
```diff
@@ -0,0 +1,81 @@
...
+// SubnetPort attached to it.
+
+type SubnetPortHandler struct {
+	Reconciler *SubnetSetReconciler
+}
+
+// Create allocates Subnet for SubnetPort from SubnetSet.
+func (h *SubnetPortHandler) Create(e event.CreateEvent, _ workqueue.RateLimitingInterface) {
+	subnetPort := e.Object.(*v1alpha1.SubnetPort)
+	if subnetPort.Spec.Subnet != "" {
```
lxiaopei, line 28: if SubnetPort specified by Subnet, is this logic handled by Qian's SubnetPort controller?

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -0,0 +1,360 @@
...
+		return nil
+	}); err != nil {
+		log.Error(err, "failed to create subnet", "Namespace", subnetset.Namespace, "Name", name)
+		return nil, err
+	}
+	return subnet, nil
+}
+
+// getAvailableSubnets returns available Subnet under SubnetSet, and creates Subnet if necessary.
+func (r *SubnetSetReconciler) getAvailableSubnet(subnetSet *v1alpha1.SubnetSet) (*v1alpha1.Subnet, error) {
```
lxiaopei, line 139: For SubnetSet CR, besides create NSXSubnet, to release unused NSXSubnet, Could we also consider how to release unused NSXSubnet? for example, we can take a simple way, delete NSXSubnet if no Port on them with a conservative timeout.

pkg/nsx/services/subnet/subnet.go
jwsui: thanks, let me try this.

pkg/controllers/subnetset/subnetset_controller.go
jwsui: https://github.com/vmware-tanzu/nsx-operator/pull/165/files#diff-959aec2764d2baff7b43c068f97bee2f29c1c0c7e8bc8554569346cdf32499ccR54

how about recycling subnet when handing deletion of subnetport?

pkg/controllers/subnetset/subnetport_handler.go
jwsui: I'll confirm with Qian. But I think in this case, SubnetsetController shouldn't handle the subnet creation as the subnet already exists.

pkg/controllers/subnet/subnet_controller.go
jwsui: We don't check currently, I'll try check the api and try.

pkg/controllers/subnetset/subnetset_controller.go
lxiaopei: do you mean recycling the subnet when subnetport deleted from subnet and no ports on it immediately?  I think this might cause subnet deletion and re-creation.

pkg/controllers/subnetset/subnetport_handler.go
```diff
@@ -0,0 +1,81 @@
...
+		// Two possible scenarios:
+		// - 1. User uses `.Spec.Subnet` directly instead of `.Spec.SubnetSet`.
+		// - 2. Subnet has been allocated and `.Spec.Subnet` is rendered by SubnetPortHandler.
+		return
+	}
+	subnetSet := &v1alpha1.SubnetSet{}
+	key := types.NamespacedName{
+		Namespace: subnetPort.GetNamespace(),
```
wsquan171, line 36: For system VPC, the VPC CR will be created in one of system NS, but subnet port created in any of the system NS without a SubnetSet in spec should be matched to the same default subnet set created by vpc handler in the same NS as the system VPC CR.

While I don't think customers are allowed to create subnetset CR in system VPC, can we also confirm the behavior for non-default subnet set created in system NS? Should it be allowed for cross-NS sharing (suppose all of them are created in the same NS as system VPC). cc @lxiaopei

pkg/controllers/subnetset/subnetset_controller.go
lxiaopei: Yea, we can take this way.

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -0,0 +1,279 @@
...
+			deleteSuccess(r, &ctx, obj)
+		} else {
+			log.Info("finalizers cannot be recognized", "subnet", req.NamespacedName)
+		}
+	}
+	return ctrl.Result{}, nil
+}
+
+func (r *SubnetReconciler) updateSubnetStatus(obj *v1alpha1.Subnet) error {
+	nsxSubnets := r.Service.SubnetStore.GetByIndex(servicecommon.TagScopeSubnetCRUID, string(obj.UID))
```
lxiaopei, line 108: Can we use SubnetStore.GetByKey() to return the nsxSubnet using Id?

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -0,0 +1,279 @@
...
+	if err := r.Client.Get(ctx, req.NamespacedName, obj); err != nil {
+		log.Error(err, "unable to fetch Subnet CR", "req", req.NamespacedName)
+		return ResultNormal, client.IgnoreNotFound(err)
+	}
+
+	vpcList := &v1alpha1.VPCList{}
+	if err := r.Client.List(context.Background(), vpcList, client.InNamespace(obj.Namespace)); err != nil {
+		log.Error(err, fmt.Sprintf("failed to get VPC under namespace: %s.\n", obj.Namespace))
+		return ResultRequeue, err
+	}
```
lxiaopei, line 58: Better to check if len(vpcList)>0 for negative case.

pkg/controllers/subnet/subnet_controller.go
jwsui: I'm afraid we can't. In our previous discussion, we decided that subnetset controller creates nsx subnet directly instead of create subnet cr. As a result, the nsx subnet id is generated randomly, so in this case, we can't assume that we already know the nsx subnet id.

jwsui: https://github.com/vmware-tanzu/nsx-operator/pull/165/files#diff-e23a4dc07d0797a88d5a31f4250c8a7a2f249275eee2ee87507090be993d3f75R28

pkg/nsx/services/subnet/subnet.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+	log.Info("successfully updated nsxSubnet", "nsxSubnet", nsxSubnet)
+	return nil
+}
+
+func (service *SubnetService) DeleteSubnet(obj client.Object, orgID, projectID, vpcID string) error {
+	nsxSubnets := service.SubnetStore.GetByIndex(common.TagScopeSubnetCRUID, string(obj.GetUID()))
+	for _, nsxSubnet := range nsxSubnets {
+		if err := service.DeleteIPAllocation(orgID, projectID, vpcID, *nsxSubnet.Id); err != nil {
```
lxiaopei, line 142: I do not think we should forcely delete all allocations then deleting the Subnet. If we found it returns error " it has children or it is being referenced by other objects" when deleting the VPC Subnet,  it is expected.

pkg/controllers/subnetset/subnetport_handler.go
jwsui: Thanks for reminder, will confirm the bahaviour.

pkg/nsx/services/subnet/subnet.go
jwsui: Rebased and updated, thanks.

pkg/controllers/subnetset/subnetset_controller.go
jwsui: Updated to create subnet directly.

pkg/nsx/services/subnet/subnet.go
jwsui: Already added a service to check realization status.

pkg/controllers/subnet/subnet_controller.go
jwsui: Thanks, done.

pkg/nsx/services/subnet/subnet.go
jwsui: Done, thanks.

pkg/controllers/subnetset/subnetset_controller.go
jwsui: Marking as resolved as GC will be in another patch.

pkg/nsx/services/subnet/subnet.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+		if err = service.SubnetStore.Operate(&subnetCopy); err != nil {
+			return err
+		}
+		log.Info("successfully deleted nsxSubnet", "nsxSubnet", nsxSubnet)
+	}
+	return nil
+}
+
+func (service *SubnetService) DeleteIPAllocation(orgID, projectID, vpcID, subnetID string) error {
```
heypnus, line 169: In which case this function will be used?

pkg/controllers/subnet/subnet_controller.go
jwsui: Make sense, updated, thanks.

pkg/nsx/services/subnet/builder.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func (service *SubnetService) buildSubnet(obj client.Object, tags []model.Tag) (*model.VpcSubnet, error) {
+	tags = append(tags, service.buildBasicTags(obj)...)
+	var nsxSubnet *model.VpcSubnet
+	switch o := obj.(type) {
+	case *v1alpha1.Subnet:
+		nsxSubnet = &model.VpcSubnet{
+			//TODO How to compare subnets when using random uuid.
+			Id:             String(uuid.NewString()),
```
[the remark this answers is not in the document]
jwsui: Ok, done.

pkg/nsx/services/subnet/subnet.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	uid := string(obj.GetUID())
+	nsxSubnet, err := service.buildSubnet(obj, tags)
+	if err != nil {
+		log.Error(err, "failed to build Subnet")
+		return "", err
+	}
+	switch obj.(type) {
+	case *v1alpha1.Subnet:
+		nsxSubnets := service.SubnetStore.GetByIndex(common.TagScopeSubnetCRUID, uid)
```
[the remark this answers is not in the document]
jwsui: Added TODO as mark.

pkg/nsx/services/subnet/builder.go
```diff
@@ -0,0 +1,110 @@
...
+			AccessMode:     String(string(o.Spec.AccessMode)),
+			DhcpConfig:     service.buildDHCPConfig(&o.Spec.DHCPConfig),
+			Tags:           tags,
+			AdvancedConfig: service.buildAdvancedConfig(&o.Spec.AdvancedConfig),
+		}
+	case *v1alpha1.SubnetSet:
+		nsxSubnet = &model.VpcSubnet{
+			//TODO How to compare subnets when using random uuid.
+			Id:             String(uuid.NewString()),
```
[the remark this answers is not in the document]
jwsui: Ok, done.

pkg/nsx/services/subnet/subnet.go
jwsui: IP allocation will block deletion of subnet, so in previous implementation, nsx-operator delete ip allocation before deleing subnet. But @lxiaopei mentioned that it's ok when we hit deletion error when deleting subnet with allocation, so the function is not used for now.

But I think in GC, we may need it to ensure subnet can be deleted, so I keep the function here.

pkg/nsx/services/subnet/subnet.go
```diff
@@ -0,0 +1,242 @@
...
+		return "", err
+	}
+	log.Info("successfully updated nsxSubnet", "nsxSubnet", nsxSubnet)
+	return *nsxSubnet.Id, nil
+}
+
+func (service *SubnetService) DeleteSubnet(obj client.Object, vpcInfo *common.VPCResourceInfo) error {
+	nsxSubnets := service.SubnetStore.GetByIndex(common.TagScopeSubnetCRUID, string(obj.GetUID()))
+	for _, nsxSubnet := range nsxSubnets {
+		nsxSubnet.MarkedForDelete = &MarkedForDelete
```
lxiaopei, line 150: is it possible that we check if there is subnet ports on the subnet before invoking NSX API to delete the subnet?
If still subnet ports on it, re-enqueue the event.

pkg/nsx/services/subnet/wrap.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	dataValue, errors := NewConverter().ConvertToVapi(childSubnet, model.ChildVpcSubnetBindingType())
+	if len(errors) > 0 {
+		return nil, errors[0]
+	}
+	return []*data.StructValue{dataValue.(*data.StructValue)}, nil
+}
+
+// wrapDHCPStaticBindingConfig wraps DHCPStaticBindingConfig as children of VPCSubnet.
+func (service *SubnetService) wrapDHCPStaticBindingConfig(config model.DhcpStaticBindingConfig) ([]*data.StructValue, error) {
```
lxiaopei, line 113: where is this func used?

pkg/controllers/subnetset/subnetport_handler.go
```diff
@@ -0,0 +1,91 @@
...
+		Name:      subnetPort.Spec.SubnetSet,
+	}
+	if err := h.Reconciler.Client.Get(context.Background(), key, subnetSet); err != nil {
+		log.Error(err, "failed to get SubnetSet", "ns", key.Namespace, "name", key.Name)
+		return
```
lxiaopei, line 44: if SubnetSet not found, will the handler retry?

pkg/controllers/subnetset/subnetport_handler.go
```diff
@@ -0,0 +1,91 @@
...
+	if err := h.Reconciler.Client.List(context.Background(), vpcList, client.InNamespace(subnetPort.GetNamespace())); err != nil {
+		log.Error(err, fmt.Sprintf("failed to get VPC under namespace: %s.\n", subnetPort.GetNamespace()))
+		return
+	}
+	vpcInfo, err := servicecommon.ParseVPCResourcePath(vpcList.Items[0].Status.NSXResourcePath)
+	if err != nil {
+		log.Error(err, "failed to resolve VPC info")
+		return
+	}
+	_, err = h.Reconciler.getAvailableSubnet(subnetSet, &vpcInfo)
```
lxiaopei, line 57: if new nsx subnet created for the first pod attached to the SubnetSet, we also need to update the SubnetSet status with the NSX subnet info.

jwsui: According to today's discussion, we may not need subnetport handler any more.
But you are right that we need to find a way to update subnetset status after a new subnet is created, will consider how to implement this, thanks.

pkg/nsx/services/subnet/subnet.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+	return subnetService, nil
+}
+
+// TODO Test update of VpcSubnet(eg. update tags)
+func (service *SubnetService) CreateOrUpdateSubnet(obj client.Object, vpcInfo *common.VPCResourceInfo, tags []model.Tag) (string, error) {
+	value, ok := obj.GetAnnotations()[common.TagScopeSubnetType]
+	if ok {
+		tags = append(tags, model.Tag{
+			Scope: String(common.TagScopeSubnetType),
```
dantingl, line 100: Why we need add subnet type in tag?

cmd/main.go
```diff
@@ -142,6 +144,9 @@ func main() {
 	if cf.EnableAntreaNSXInterworking {
 		StartNSXServiceAccountController(mgr, commonService)
 	}
+	// Start subnet/subnetset controller.
+	subnet.StartSubnetController(mgr, commonService)
```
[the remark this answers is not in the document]
heypnus: OK, I'll use this way.

pkg/nsx/services/subnet/subnet.go
jwsui: Not needed any more, use label instead according to today's discusssion.

cmd/main.go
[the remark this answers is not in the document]
jwsui: Ok, done.

pkg/controllers/subnet/subnet_controller.go
jwsui: > do we know if the specified IP_address Subnet is realized successfully or not?

In latest change, we've added logic to check realize status of Subnet.

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -0,0 +1,272 @@
...
+	return nil
+}
+
+func (r *SubnetReconciler) setSubnetReadyStatusTrue(ctx *context.Context, subnet *v1alpha1.Subnet) {
+	newConditions := []v1alpha1.Condition{
+		{
+			Type:    v1alpha1.Ready,
+			Status:  v1.ConditionTrue,
+			Message: "NSX Subnet has been successfully created/updated",
+			Reason:  "NSX API returned 200 response code for PATCH",
```
lxiaopei, line 154: I checked K8s common reasons, they use brief phase  "Reason:  "InstanceDeletionInProgress",
		Message: "CustomResource deletion is in progress",", for example  https://github.com/kubernetes/kubernetes/blob/2c6c4566eff972d6c1320b5f8ad795f88c822d09/staging/src/k8s.io/apiextensions-apiserver/pkg/controller/finalizer/crd_finalizer.go#L129 and 
https://github.com/kubernetes/kubernetes/blob/2c6c4566eff972d6c1320b5f8ad795f88c822d09/pkg/controller/controller_utils.go#L414
Could we also follow the same convention?
Also I think we do not need to say "NSX API return 200" , just say "SubnetCreated"

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -0,0 +1,257 @@
...
+	return nil
+}
+
+func (r *SubnetSetReconciler) setSubnetSetReadyStatusTrue(ctx *context.Context, subnetset *v1alpha1.SubnetSet) {
+	newConditions := []v1alpha1.Condition{
+		{
+			Type:    v1alpha1.Ready,
+			Status:  v1.ConditionTrue,
+			Message: "NSX SubnetSet has been successfully created/updated",
+			Reason:  "All subnets are ready",
```
lxiaopei, line 143: ditto for reason and message.

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -0,0 +1,257 @@
...
+)
+
+// SubnetSetReconciler reconciles a SubnetSet object
+type SubnetSetReconciler struct {
+	Client  client.Client
+	Scheme  *apimachineryruntime.Scheme
+	Service *subnet.SubnetService
+}
+
+func (r *SubnetSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
lxiaopei, line 46: For SubnetSet, we also need to make sure when NS labels changed, update existing subnet tags with new NS labels. this is needed for NetworkPolicy and SecurityPolicy.

pkg/nsx/services/common/types.go
```diff
@@ -34,11 +34,18 @@ const (
 	TagScopeNCPVNETInterface        string = "ncp/vnet_interface"
 	TagScopeVPCCRName               string = "nsx-op/vpc_cr_name"
 	TagScopeVPCCRUID                string = "nsx-op/vpc_cr_uid"
+	// TagScopeSubnetCRType indicates that NSX Subnet is linked to Subnet CR or SubnetSet CR.
+	TagScopeSubnetCRType string = "nsx-op/subnet_cr_type"
+	TagScopeSubnetCRUID  string = "nsx-op/subnet_cr_uid"
 
-	GCInterval    = 60 * time.Second
-	FinalizerName = "securitypolicy.nsx.vmware.com/finalizer"
+	LabelDefaultSubnetSet    string = "nsxoperator.vmware.com/default-subnetset-for"
```
lxiaopei, line 41: how about DefaultSubnetSetLabelKey, DefaultPodSubnetSetLabelValue and DefaultVMSubnetSetLabelValue?

pkg/nsx/services/subnet/builder.go
```diff
@@ -0,0 +1,109 @@
...
+	Int64  = common.Int64
+	Bool   = common.Bool
+)
+
+func getCluster(service *SubnetService) string {
+	return service.NSXConfig.Cluster
+}
+
+func (service *SubnetService) buildSubnet(obj client.Object, tags []model.Tag) (*model.VpcSubnet, error) {
+	tags = append(tags, service.buildBasicTags(obj)...)
```
lxiaopei, line 23: we also need to add NS labels to NSX subnet tags, do you plan to support that in this patch or following patch?

pkg/nsx/services/subnet/builder.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (service *SubnetService) buildDHCPConfig(obj *v1alpha1.DHCPConfig) *model.DhcpConfig {
+	// Subnet DHCP is used by AVI, not needed for now. We need to explicitly mark enableDhcp = false,
+	// otherwise Subnet will use DhcpConfig inherited from VPC.
+	dhcpConfig := &model.DhcpConfig{
+		//DhcpRelayConfigPath: String(obj.DHCPRelayConfigPath),
+		//DhcpV4PoolSize:      Int64(int64(obj.DHCPV4PoolSize)),
+		//DhcpV6PoolSize:      Int64(int64(obj.DHCPV6PoolSize)),
+		//DnsClientConfig:     service.buildDNSClientConfig(&obj.DNSClientConfig),
+		EnableDhcp: Bool(false),
```
lxiaopei, line 57: when enableDHCP: false, will only static ip pool created for the subnet?

pkg/nsx/services/subnet/subnet.go
```diff
@@ -0,0 +1,236 @@
...
+	uid := string(obj.GetUID())
+	nsxSubnet, err := service.buildSubnet(obj, tags)
+	if err != nil {
+		log.Error(err, "failed to build Subnet")
+		return "", err
+	}
+	switch obj.(type) {
+	case *v1alpha1.Subnet:
+		//TODO Additional check for len(nsxSubnets)==1 for improvement.
+		nsxSubnets := service.SubnetStore.GetByIndex(common.TagScopeSubnetCRUID, uid)
```
lxiaopei, line 106: Seems GetByKey can only retrieve one object.

pkg/nsx/services/subnet/subnet.go
```diff
@@ -0,0 +1,236 @@
...
+		changed := false
+		if len(nsxSubnets) == 0 {
+			changed = true
+		} else {
+			changed = common.CompareResource(SubnetToComparable(&nsxSubnets[0]), SubnetToComparable(nsxSubnet))
+		}
+		if !changed {
+			log.Info("subnet not changed, skip updating", "subnet.Id", *nsxSubnets[0].Id)
+			return *nsxSubnets[0].Id, nil
+		}
```
lxiaopei, line 116: add default for switch obj.(type)?

pkg/nsx/services/subnet/subnet.go
jwsui: Thanks, done.

pkg/nsx/services/subnet/builder.go
jwsui: yes

pkg/nsx/services/subnet/builder.go
jwsui: I think the interface is ready, will do this in following patch.

pkg/nsx/services/common/types.go
jwsui: @heypnus 's patch had already added this label, and I have rebased her patch.

pkg/controllers/subnetset/subnetset_controller.go
jwsui: Yes, I remember that we will do this in following patch?

> For SubnetSet, we also need to make sure when NS labels changed, update existing subnet tags with new NS labels. this is needed for NetworkPolicy and SecurityPolicy.

pkg/controllers/subnetset/subnetport_handler.go
jwsui: This handler should be removed as this will be handled on subnetport side.

pkg/nsx/services/subnet/wrap.go
jwsui: not used currently, it can be removed.

pkg/nsx/services/subnet/subnet.go
jwsui: Done, using type assertion instead.

pkg/controllers/subnet/subnet_controller.go
jwsui: Thanks, done.

pkg/controllers/subnetset/subnetset_controller.go
jwsui: Thanks, done.

pkg/controllers/subnetset/subnetport_handler.go
jwsui: Done.

pkg/controllers/subnetset/subnetport_handler.go
jwsui: Marking as resolved as this will be handled in on subnetport side.

pkg/nsx/services/subnet/subnet.go
jwsui: It will require another api call for subnet ports in subnet service, so I would prefer to check if there are subnet ports left from subnet store as we can regard store as the source of truth. How do you think?

pkg/nsx/services/subnet/subnet.go
```diff
@@ -0,0 +1,266 @@
...
+		log.Error(err, "WrapHierarchySubnet failed")
+		return "", err
+	}
+	if err = service.NSXClient.OrgRootClient.Patch(*orgRoot, &EnforceRevisionCheckParam); err != nil {
+		return "", err
+	}
+	// Get Subnet from NSX after patch operation as NSX renders several fields like `path`/`parent_path`.
+	if *nsxSubnet, err = service.NSXClient.SubnetsClient.Get(vpcInfo.OrgID, vpcInfo.ProjectID, vpcInfo.VPCID, *nsxSubnet.Id); err != nil {
+		return "", err
+	}
```
heypnus, line 129: Please log the error before returning it. Because this function `CreateOrUpdateSubnet` may be indirectly invoked by `GetAvailableSubnet`. Imagine that when we got a standalone error `"com.vmware.vapi.std.errors.invalid_request"` from `GetAvailableSubnet`, locating the issue will be hard.

jwsui: Make sense, done.

pkg/nsx/services/subnet/builder.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (service *SubnetService) buildBasicTags(obj client.Object) []model.Tag {
+	tags := []model.Tag{
+		{
+			Scope: String(common.TagScopeCluster),
+			Tag:   String(getCluster(service)),
+		},
+		{
+			Scope: String(common.TagScopeSubnetCRUID),
+			Tag:   String(string(obj.GetUID())),
```
heypnus, line 76: Maybe we also need the subnet/subnetset CR name in the NSX tags?
```
# grep -E 'Tag.*Name.*nsx-op/.*_name' pkg/nsx/services/common/types.go
        TagScopeSecurityPolicyCRName    string = "nsx-op/security_policy_cr_name"
        TagScopeNSXServiceAccountCRName string = "nsx-op/nsx_service_account_name"
        TagScopeSubnetPortCRName        string = "nsx-op/subnetport_cr_name"
```

pkg/nsx/services/subnet/builder.go
```diff
@@ -0,0 +1,99 @@
...
+	return service.NSXConfig.Cluster
+}
+
+func (service *SubnetService) buildSubnet(obj client.Object, tags []model.Tag) (*model.VpcSubnet, error) {
+	tags = append(tags, service.buildBasicTags(obj)...)
+	var nsxSubnet *model.VpcSubnet
+	switch o := obj.(type) {
+	case *v1alpha1.Subnet:
+		nsxSubnet = &model.VpcSubnet{
+			Id:         String(string(o.GetUID())),
```
heypnus, line 29: For now, the NSX subnet's name is uid because we didn't explicitly set the display name. I think it'd better to constructed by name instead of uid.

pkg/nsx/services/subnet/store_test.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func Test_IndexFunc(t *testing.T) {
+	id, tag, scope := "test_id", "test_tag", "nsx-op/subnet_cr_uid"
+	subnet := model.VpcSubnet{
+		Id:   &id,
+		Tags: []model.Tag{{Tag: &tag, Scope: &scope}},
+	}
+	t.Run("1", func(t *testing.T) {
+		got, _ := indexFunc(subnet)
```
heypnus, line 42: Don't forget to fix the UT failure before merging the patch: `vet: pkg/nsx/services/subnet/store_test.go:42:13: undeclared name: indexFunc`

pkg/nsx/services/subnet/builder.go
jwsui: ok, done.

pkg/nsx/services/subnet/builder.go
jwsui: ok, done.

pkg/nsx/services/common/types.go
```diff
@@ -58,6 +66,8 @@ var (
 	ResourceTypeClusterControlPlane = "clustercontrolplane"
 	// ResourceTypePrincipalIdentity is used by NSXServiceAccountController, and it is MP resource type.
 	ResourceTypePrincipalIdentity = "principalidentity"
+	ResourceTypeVPC               = "VPC"
```
timdengyun, line 69: ResourceTypeVPC should be "Vpc" instead of "VPC"

pkg/nsx/services/subnet/store_test.go
jwsui: Thanks, done.

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -0,0 +1,245 @@
...
+	//cache := mgr.GetCache()
+	//indexFunc := func(obj client.Object) []string {
+	//	if len(obj.(*v1alpha1.Subnet).ObjectMeta.OwnerReferences) == 0 {
+	//		return []string{"no-owner"}
+	//	}
+	//	return []string{string(obj.(*v1alpha1.Subnet).ObjectMeta.OwnerReferences[0].UID)}
+	//}
+	//if err := cache.IndexField(context.Background(), &v1alpha1.Subnet{}, "metadata.ownerReferences", indexFunc); err != nil {
+	//	panic(err)
+	//}
```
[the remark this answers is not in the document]
jwsui: Thanks, done.

pkg/nsx/services/common/types.go
jwsui: Thanks, this should not be used, removed this as we already have ResourceTypeVpc

pkg/controllers/subnetset/subnetport_handler.go
jwsui: Marking as resolved as this part of logic will be handled in Pod/SubnetPort controller, the subnetport_handler will be removed.

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -0,0 +1,236 @@
...
+		Watches(
+			&source.Kind{Type: &v1alpha1.SubnetPort{}},
+			&SubnetPortHandler{Reconciler: r},
+			builder.WithPredicates(SubnetPortPredicate)).
+		Complete(r)
+}
+
+// GarbageCollector collect Subnet which there is no port attached on it.
+// cancel is used to break the loop during UT
+func (r *SubnetSetReconciler) GarbageCollector(cancel chan bool, timeout time.Duration) {
```
lxiaopei, line 190: seems GC can delete subnets which do not have ports on it, can it also handle there is no SubnetSet CR in K8s, but in nsx  subnet store, still has the NSX subnets for the SubnetSet CR?

lxiaopei (commented):
LGTM. Let's track the below TODO in next PR:
1. default subnet size and access mode need to get from vpc networkconfig CR
2. listen namespace and add label for subnet/subnetset
3. Subnet/SubnetSet GC

lxiaopei (dismissed):

pkg/controllers/subnetset/subnetset_controller.go
jwsui: > seems GC can delete subnets which do not have ports on it, can it also handle there is no SubnetSet CR in K8s, but in nsx subnet store, still has the NSX subnets for the SubnetSet CR?

No, it can't handle this case currently. We can also mark a TODO here.

jwsui:
> LGTM. Let's track the below TODO in next PR:
> 
> 1. default subnet size and access mode need to get from vpc networkconfig CR
> 2. listen namespace and add label for subnet/subnetset
> 3. Subnet/SubnetSet GC

Add GC for case that nsx subnet exist in store but no subnetset CR.

lxiaopei (approved):

PR #175, opened 2023-01-17
Update SubnetSet/Subnet field

jwsui:
Add `VPC` field in SubnetSet/Subnet CR to specify the VPC that subnet belongs to.


jwsui:
Currently subnet will be created under the default VPC of namespace, and doesn't need user to specify vpc in subnet cr. So closing this pr.

PR #178, opened 2023-02-13
remove applied to namespace

lxiaopei:
As discussed, we will remove appliedToNamespaces filed in
    VPCNetworkConfiguration CR, a label “nsx.vmware.com/vpc-network-configuration” will be added on Namespace
    to specify the VPCNetworkConfiguration it uses.


jianjuns (approved):
Could you define the NS label too?

lxiaopei:
> Could you define the NS label too?

How about "ncp/vpc_network_config_crd" for the label added on Namespace to specify the VPCNetworkConfiguration CR name?

jianjuns:
> > Could you define the NS label too?
> 
> How about "ncp/vpc_network_config_crd" for the label added on Namespace to specify the VPCNetworkConfiguration CR name?

Again, please never use "ncp". Something like "nsx.vmware.com/vpc-network-configuration"

lxiaopei:
OK, thanks Jianjun.

heypnus (approved):

PR #179, opened 2023-02-16
Expose getVPCsByNamespace to service level

seanpang-vmware:
The service level communication should not directly operate its store member, expose this method to VPC service.

### DIFF
--- a/pkg/nsx/services/vpc/vpc.go
+++ b/pkg/nsx/services/vpc/vpc.go
@@ -55,3 +55,7 @@ func InitializeVPC(service common.Service) (*VPCService, error) {
 
 	return VPCService, nil
 }
+
+func (s *VPCService) GetVPCsByNamespace(namespace string) []model.Vpc {
+	return s.vpcStore.GetVPCsByNamespace(namespace)
+}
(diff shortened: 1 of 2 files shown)


pkg/nsx/services/vpc/vpc.go
```diff
@@ -55,3 +55,7 @@ func InitializeVPC(service common.Service) (*VPCService, error) {
 
 	return VPCService, nil
 }
+
+func (s *VPCService) GetVPCsByNamespace(namespace string) []model.Vpc {
```
timdengyun, line 59: This function is a method of VPCService, but from caller perspective, we have to hold a 
vpc service firstly and call this function. It's good that we have this function inside VPCService. we also need to consider how to expose vpcservice to users.

timdengyun (approved):

pkg/nsx/services/vpc/vpc.go
seanpang-vmware: you are correct, i have raised the question in our slack channel to discuss how would we handle inter-service communication.

zhengxiexie:
LGTM， besides, I wanna ask how to get org and project by namespace.  @seanpang-vmware @lxiaopei

zhengxiexie (approved):

PR #181, opened 2023-02-20
Implement service mediator

zhengxiexie:


seanpang-vmware (approved):
LGTM

cmd/main.go
```diff
@@ -63,6 +64,7 @@ func StartSecurityPolicyController(mgr ctrl.Manager, commonService common.Servic
 		os.Exit(1)
 	} else {
 		securityReconcile.Service = securityService
+		commonctl.ServiceMediator.SecurityPolicyService = securityService
```
heypnus, line 67: So the `ServiceMediator.SecurityPolicyService` is initialized after `InitializeSecurityPolicy`. If `ServiceMediator.SecurityPolicyService` is referred before it is initialized, what will happen? In the other word, how we ensure that the `ServiceMediator.SecurityPolicyService` is only used after it is initialized?

zhengxiexie: Got your point. Since we start all CR controllers in sequence or in parallel,  in each controller, we initialize all services which request all the belonging resources from the nsx-t side, we can conclude that **at least at the initialization period**, there is no inter reference(like mutual interdependence).
Anyway, this is a good consideration about the hidden corner, and though we can't think of any scenario, we could discuss it again in case we meet it in the following integrating process of all CRDs.

PR #184, opened 2023-02-27
Upgrade some packages due to security issues

TaoZou1:
1. Since some packages require go 1.18 above so upgrade golang to 1.19.
2. Golang 1.19 doesn't use 'go get', update Makefile to use 'go install'
3. Update mock since interface changed
4. Update some UT case


dantingl (approved):

jwsui (approved):
LGTM

PR #191, opened 2023-03-21
Add GetVPCInfo method to ServiceMediator

zhengxiexie:
Create the common method used by other modules. When using gomonkey to test the method, encounter the problem that the mocked method doesn't take effect, fix it by adding -gcflags=all=-l option to prohibit inline optimization.


pkg/nsx/services/common/types.go
```diff
@@ -69,3 +69,8 @@ var (
 	String = pointy.String // address of string
 	Int64  = pointy.Int64  // address of int64
 )
+
+type OrgProject struct {
+	Org     string
```
TaoZou1, line 74: OrgId， ProjectId？

zhengxiexie: OK

pkg/nsx/services/mediator/mediator.go
```diff
@@ -14,3 +23,25 @@ type ServiceMediator struct {
...
+// so we return a slice of OrgProject.
+func (serviceMediator *ServiceMediator) GetOrgProjectVPC(ns string) []common.OrgProjectVPC {
+	var orgProjectVPC []common.OrgProjectVPC
+	vpcs := serviceMediator.GetVPCsByNamespace(ns) // Transparently call the VPCService.GetVPCsByNamespace method
+	for _, v := range vpcs {
+		matches := reExp.FindStringSubmatch(*v.Path)
+		if len(matches) == 4 {
+			org := matches[1]
+			project := matches[2]
+			vpc_ := matches[3]
```
lxiaopei, line 40: why add "_" suffix for "vpc_" variable?

pkg/nsx/services/mediator/mediator.go
```diff
@@ -14,3 +23,25 @@ type ServiceMediator struct {
 	*securitypolicy.SecurityPolicyService
 	*vpc.VPCService
 }
+
+// GetOrgProjectVPC is a common method, extracting the org and project string from vpc path of the VPC model.
+// VPC path looks like "/orgs/default/projects/project-1/vpcs/vpc-1",
+// Since other modules only know namespace, this is the only entry point to get org and project.
+// Currently, we only support one vpc per namespace, but we may support multiple vpcs per namespace in the future,
+// so we return a slice of OrgProject.
```
lxiaopei, line 31: OrgProject -> OrgProjectVPC?

zhengxiexie: ok

pkg/nsx/services/mediator/mediator.go
zhengxiexie: IDE reminds 
image
so add this trick.

pkg/nsx/services/common/types.go
```diff
@@ -69,3 +69,9 @@ var (
 	String = pointy.String // address of string
 	Int64  = pointy.Int64  // address of int64
 )
+
+type OrgProjectVPC struct {
```
dantingl, line 73: The struct name is a little weird to me..

zhengxiexie: It consists of org, project, and vpc, which is a common struct used often.

pkg/nsx/services/mediator/mediator.go
```diff
@@ -14,3 +23,25 @@ type ServiceMediator struct {
 	*securitypolicy.SecurityPolicyService
 	*vpc.VPCService
 }
+
+// GetOrgProjectVPC is a common method, extracting the org and project string from vpc path of the VPC model.
+// VPC path looks like "/orgs/default/projects/project-1/vpcs/vpc-1",
```
heypnus, line 28: From the comment, the input should be a string. I didn't get why you use the ns as the input, I think it's not common enough.

zhengxiexie: The comment explains how to extract, and the input is indeed a namespace.

dantingl (approved):

dantingl (approved):

PR #192, opened 2023-03-22
Fix bug which can't delete legacy rules and groups

zhengxiexie:
Delete the legacy rule and group at last when updating security policy. Split large file.
Based on v4.0.0 branch.


pkg/controllers/securitypolicy/securitypolicy_controller.go
```diff
@@ -124,12 +123,7 @@ func (r *SecurityPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 			return resultNormal, nil
 		}
 
-		if err := r.Service.OperateSecurityPolicy(obj); err != nil {
-			if errors.As(err, &util2.RestrictionError{}) {
-				log.Error(err, err.Error(), "securitypolicy", req.NamespacedName)
-				updateFail(r, &ctx, obj, &err)
-				return resultNormal, nil
-			}
```
heypnus, line 132: Why reverted this change?

pkg/nsx/services/builder.go
```diff
@@ -1033,7 +1033,8 @@ func (service *SecurityPolicyService) matchExpressionOpInExist(
 // Currently NSX only supports 'EQUALS' not 'In'. So, we have to make each value to be AND with other expressions
 // and finally produce a union set to translate from K8s 'In' to NSX EQUALS'.
 // e.g. - {key: k1, operator: NotIn, values: [a1,a2]}
-//      - {key: k2, operator: In, values: [a3,a4]}
+//   - {key: k2, operator: In, values: [a3,a4]}
+//
```
heypnus, line 1037: The reason of introducing this change?

zhengxiexie: image
I see it is from 2022.6.10, @timdengyun , could you confirm it? Refactored a long time ago.

pkg/controllers/securitypolicy/securitypolicy_controller.go
zhengxiexie: Refactored a long time ago, on vpc_dev is CreateOrUpdateSecurityPolicy,

coverage.cov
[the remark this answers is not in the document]
zhengxiexie: Removed.

pkg/apis/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
```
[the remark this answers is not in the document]
zhengxiexie: Removed.

pkg/nsx/services/builder.go
timdengyun: what changes need to be confirmed? you mean @zhengxiexie @heypnus  mergeSelectorMachExpresssion func?

zhengxiexie: //   - {key: k2, operator: In, values: [a3,a4]}
@timdengyun should this line leaved or removed?

timdengyun: We noted in main branch: https://github.com/vmware-tanzu/nsx-operator/blob/eb5ba400e574c2e3553df1e4310758f1ee185a83/pkg/nsx/services/securitypolicy/builder.go#L1012
we still have - {key: k2, operator: In, values: [a3,a4]}, not sure why it was removed in the builder.go.

We need to keep this line in the comments section. Since the example here is to explain how two expressions
on  operator NotIn and operator IN  are needed to be translated into  two equivalent expressions:
{k1 NotIn [a1,a2]} AND {k2 EQUALS a3} OR {k1 NotIn [a1,a2]} AND {k2 EQUALS a4}

// e.g. - {key: k1, operator: NotIn, values: [a1,a2]}
//   - {key: k2, operator: In, values: [a3,a4]}
// The above two expressions will be translated to:
// => {k1 NotIn [a1,a2]} AND {k2 EQUALS a3} OR {k1 NotIn [a1,a2]} AND {k2 EQUALS a4}

PR #200, opened 2023-03-30
Install go tool by using go install instead of go get after golang 1.18

zhengxiexie:
After golang 1.18, go get does not install binary by default, so controller-gen is not installed. Here should use go install.


jwsui (approved):
LGTM

PR #207, opened 2023-04-27
Implement cleaning up all nsx resources

zhengxiexie:
Just run `./bin/clean -cluster=''  -thumbprint="" -log-level=0 -vc-user="" -vc-passwd="" -vc-endpoint="" -vc-sso-domain="" -vc-https-port=443  -mgr-ip=""
` to clean up all. Leverage the existing module to realize the logic more conveniently.

### DIFF
--- a/cmd_clean/main.go
+++ b/cmd_clean/main.go
@@ -1,0 +1,65 @@
+/* Copyright © 2023 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: Apache-2.0 */
+
+package main
+
+import (
+	"flag"
+	"os"
+
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/clean"
+	"github.com/vmware-tanzu/nsx-operator/pkg/config"
+	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
+)
+
+// usage:
+// ./bin/clean -cluster=''  -thumbprint="" -log-level=0 -vc-user="" -vc-passwd="" -vc-endpoint="" -vc-sso-domain="" -vc-https-port=443  -mgr-ip=""
+
+var (
+	log         = logger.Log
+	cf          *config.NSXOperatorConfig
+	mgrIp       string
+	vcEndpoint  string
+	vcUser      string
+	vcPasswd    string
+	vcSsoDomain string
+	vcHttpsPort int
+	thumbprint  string
+	caFile      string
+	cluster     string
+)
+
+func main() {
+	flag.StringVar(&vcEndpoint, "vc-endpoint", "", "nsx manager ip")
+	flag.StringVar(&vcSsoDomain, "vc-sso-domain", "", "nsx manager ip")
+	flag.StringVar(&mgrIp, "mgr-ip", "", "nsx manager ip")
+	flag.StringVar(&vcUser, "vc-user", "", "nsx username")
+	flag.StringVar(&vcPasswd, "vc-passwd", "", "nsx password")
+	flag.IntVar(&vcHttpsPort, "vc-https-port", 443, "nsx password")
+	flag.StringVar(&thumbprint, "thumbprint", "", "nsx thumbprint")
+	flag.StringVar(&caFile, "ca-file", "", "ca file")
+	flag.StringVar(&cluster, "cluster", "", "cluster name")
+	flag.IntVar(&config.LogLevel, "log-level", 0, "Use zap-core log system.")
+	flag.Parse()
+
+	logf.SetLogger(logger.ZapLogger())
+	cf = config.NewNSXOpertorConfig()
+	cf.NsxApiManagers = []string{mgrIp}
+	cf.VCUser = vcUser
+	cf.VCPassword = vcPasswd
+	cf.VCEndPoint = vcEndpoint
+	cf.SsoDomain = vcSsoDomain
+	cf.HttpsPort = vcHttpsPort
+	cf.Thumbprint = []string{thumbprint}
+	cf.CaFile = []string{caFile}
+	cf.Cluster = cluster
+
+	err := clean.Clean(cf)
+	if err != nil {
+		log.Error(err, "failed to clean nsx resources")
+		os.Exit(1)
+	}
+	os.Exit(0)
+}
--- a/pkg/clean/clean.go
+++ b/pkg/clean/clean.go
@@ -1,0 +1,61 @@
+/* Copyright © 2023 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: Apache-2.0 */
+
+package clean
+
+import (
+	"fmt"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/config"
+	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx"
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/securitypolicy"
+)
+
+var log = logger.Log
+
+// Clean cleans up NSX resources,
+// including security policy, static route, subnet, subnet port, subnet set, vpc, ip pool, nsx service account
+// it is usually used when nsx-operator is uninstalled and remove all the resources created by nsx-operator
+// return error if any, return nil if no error
+func Clean(cf *config.NSXOperatorConfig) error {
+	log.Info("starting NSX cleanup")
+	if cleanupServices, err := InitializeCleanupServices(cf); err != nil {
+		return fmt.Errorf("failed to initialize cleanup service: %w", err)
+	} else {
+		for _, cleanupService := range cleanupServices {
+			if err := cleanupService.Cleanup(); err != nil {
+				return fmt.Errorf("failed to clean up: %w", err)
+			}
+		}
+	}
+	log.Info("cleanup NSX resources successfully")
+	return nil
+}
+
+// InitializeCleanupServices initializes all the CR services
+func InitializeCleanupServices(cf *config.NSXOperatorConfig) ([]cleanup, error) {
+	var cleanupServices []cleanup
+
+	nsxClient := nsx.GetClient(cf)
+	if nsxClient == nil {
+		return nil, fmt.Errorf("failed to get nsx client")
+	}
+
+	var commonService = common.Service{
+		NSXClient: nsxClient,
+		NSXConfig: cf,
+	}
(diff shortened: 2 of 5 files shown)


cmd_clean/main.go
```diff
@@ -0,0 +1,10 @@
+/* Copyright © 2022 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: Apache-2.0 */
```
dantingl, line 2: how about put cleanup_main.go into cmd directory

zhengxiexie: how about this structure, since only one main.go in one directory.
repo
  +----cmd
  |     +----- cmd1
  |     |        +------ main.go
  |     |
  |     +----- cmd2
  |              +------ main.go
  |
  +----others.go
  +----libs.go
  +---- ...

dantingl: Yes, this looks good to me, similar with what Antrea does https://github.com/antrea-io/antrea/tree/main/cmd

pkg/clean/clean.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/vmware-tanzu/nsx-operator/pkg/config"
+	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx"
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/securitypolicy"
+)
+
+var (
+	log = logger.Log
+	cf  *config.NSXOperatorConfig
```
dantingl, line 21: If we provide an interface to cleanup resources, we need parameters instead of ConfigMap as input

pkg/clean/clean.go
```diff
@@ -0,0 +1,78 @@
...
+		os.Exit(1)
+	}
+	err = securityPolicyService.Cleanup()
+
+	if err != nil {
+		log.Error(err, "failed to clean up")
+		os.Exit(1)
+	}
+	log.Info("successfully NSX cleanup")
+	os.Exit(0)
```
ihgann, line 77: For a CLI, this works great, but the `os.Exit` calls make this not possible to consume as a library. Could this be structured differently to be able to call into a `Clean()` function, or something similar, that simply returns `error` responses instead of calling into `os.Exit`?

zhengxiexie: Sure, we can return the err.

pkg/clean/clean.go
```diff
@@ -0,0 +1,84 @@
...
+
+	if securityPolicyService, err = securitypolicy.InitializeSecurityPolicy(commonService); err != nil {
+		return fmt.Errorf("failed to initialize securitypolicy service: %v", err)
+	}
+	err = securityPolicyService.Cleanup()
+	if err != nil {
+		return fmt.Errorf("failed to clean up: %v", err)
+	}
+
+	log.Info("successfully to cleanup NSX resources")
```
seanpang-vmware, line 82: cleanup NSX resources successfully

pkg/clean/clean.go
```diff
@@ -0,0 +1,84 @@
...
+	cf.VCPassword = vcPasswd
+	cf.VCEndPoint = vcEndpoint
+	cf.SsoDomain = vcSsoDomain
+	cf.HttpsPort = vcHttpsPort
+	cf.Thumbprint = []string{thumbprint}
+	cf.NsxApiCertFile = vmcaCertFile
+	cf.Cluster = cluster
+}
+
+func Clean() error {
```
ihgann, line 59: Can you include a Godoc that expresses:

1. What this function does
2. What success state should look like
3. When an error might return, and how those errors should be handled? Why error `types` can be returned here?

pkg/clean/clean.go
```diff
@@ -0,0 +1,84 @@
...
+		NSXClient: nsxClient,
+		NSXConfig: cf,
+	}
+
+	if securityPolicyService, err = securitypolicy.InitializeSecurityPolicy(commonService); err != nil {
+		return fmt.Errorf("failed to initialize securitypolicy service: %v", err)
+	}
+	err = securityPolicyService.Cleanup()
+	if err != nil {
+		return fmt.Errorf("failed to clean up: %v", err)
```
ihgann, line 79: nit: Use the `%w` directive instead of `%v`. This will wrap the error instead, meaning we retain the details of the original error.

pkg/clean/clean.go
```diff
@@ -0,0 +1,84 @@
...
+	cf.HttpsPort = vcHttpsPort
+	cf.Thumbprint = []string{thumbprint}
+	cf.NsxApiCertFile = vmcaCertFile
+	cf.Cluster = cluster
+}
+
+func Clean() error {
+	log.Info("starting NSX cleanup")
+	var err error
+	var securityPolicyService *securitypolicy.SecurityPolicyService
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/clean/clean.go
zhengxiexie: ok

pkg/clean/clean.go
zhengxiexie: ok

pkg/clean/clean.go
zhengxiexie: ok

seanpang-vmware:
could we use an service interface which contains a Cleanup method and let each service implement this interface, and in main method, we only operate the interface instead of service instance?

zhengxiexie:
> could we use an service interface which contains a Cleanup method and let each service implement this interface, and in main method, we only operate the interface instead of service instance?

OK

cmd_clean/main.go
```diff
@@ -0,0 +1,65 @@
+/* Copyright © 2022 VMware, Inc. All Rights Reserved.
```
ihgann, line 1: nit: 2023?

cmd_clean/main.go
```diff
@@ -0,0 +1,65 @@
...
+	cf = config.NewNSXOpertorConfig()
+	cf.NsxApiManagers = []string{mgrIp}
+	cf.VCUser = vcUser
+	cf.VCPassword = vcPasswd
+	cf.VCEndPoint = vcEndpoint
+	cf.SsoDomain = vcSsoDomain
+	cf.HttpsPort = vcHttpsPort
+	cf.Thumbprint = []string{thumbprint}
+	cf.NsxApiCertFile = vmcaCertFile
+	cf.Cluster = cluster
```
ihgann, line 57: Just a Q: Is this the recommended way to build this type of config?

cmd_clean/main.go
zhengxiexie: Thanks~~

cmd_clean/main.go
zhengxiexie: Hey, do you have any other way? I'm glad to know.

cmd_clean/main.go
```diff
@@ -0,0 +1,65 @@
...
+var (
+	log          = logger.Log
+	cf           *config.NSXOperatorConfig
+	mgrIp        string
+	vcEndpoint   string
+	vcUser       string
+	vcPasswd     string
+	vcSsoDomain  string
+	vcHttpsPort  int
+	thumbprint   string
```
dantingl, line 29: Could we also support nsx ca cert as input, we will remove SHA1 thumbprint support soon

zhengxiexie: fixed.

cmd_clean/main.go
```diff
@@ -0,0 +1,65 @@
...
+
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/clean"
+	"github.com/vmware-tanzu/nsx-operator/pkg/config"
+	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
+)
+
+// usage:
+// ./bin/clean -cluster='domain-c9:47b14f07-33b4-478b-8025-a6ae16e81be7'  -thumbprint="0E:CA:B5:98:D4:46:8C:19:4A:54:4A:67:7C:7B:9D:22:0F:78:76:5E" -log-level=0 -vc-user="administrator@vsphere.local" -vc-passwd="RAZKssvew+FyR7+7" -vc-endpoint="sc2-10-187-150-214.nimbus.eng.vmware.com" -vc-sso-domain="vsphere.local" -vc-https-port=443  -mgr-ip="10.187.157.154"
```
dantingl, line 18: Please use fake values or remove this example as it has testbed info

cmd_clean/main.go
```diff
@@ -0,0 +1,65 @@
...
+
+func main() {
+	flag.StringVar(&vcEndpoint, "vc-endpoint", "", "nsx manager ip")
+	flag.StringVar(&vcSsoDomain, "vc-sso-domain", "", "nsx manager ip")
+	flag.StringVar(&mgrIp, "mgr-ip", "", "nsx manager ip")
+	flag.StringVar(&vcUser, "vc-user", "", "nsx username")
+	flag.StringVar(&vcPasswd, "vc-passwd", "", "nsx password")
+	flag.IntVar(&vcHttpsPort, "vc-https-port", 443, "nsx password")
+	flag.StringVar(&thumbprint, "thumbprint", "", "nsx thumbprint")
+	flag.StringVar(&nsxCertFile, "nsx-cert-file", "", "vmca cert file")
```
dantingl, line 42: It's not vmvc cert, right?

zhengxiexie: Yes, wrong word.

cmd_clean/main.go
zhengxiexie: ok

seanpang-vmware (dismissed):

cmd_clean/main.go
```diff
@@ -0,0 +1,65 @@
...
+	logf.SetLogger(logger.ZapLogger())
+	cf = config.NewNSXOpertorConfig()
+	cf.NsxApiManagers = []string{mgrIp}
+	cf.VCUser = vcUser
+	cf.VCPassword = vcPasswd
+	cf.VCEndPoint = vcEndpoint
+	cf.SsoDomain = vcSsoDomain
+	cf.HttpsPort = vcHttpsPort
+	cf.Thumbprint = []string{thumbprint}
+	cf.NsxApiCertFile = nsxCertFile
```
dantingl, line 56: should be ca_file

dantingl: @TaoZou1 correct me if I'm wrong

zhengxiexie: fixed.

dantingl (approved):

cmd_clean/main.go
TaoZou1: yes

PR #212, opened 2023-05-10
Support HA mode

zhengxiexie:
Leverage controller-runtime builtin leader-election to support HA.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -33,6 +33,8 @@ var (
 	probeAddr, metricsAddr string
 	log                    = logger.Log
 	cf                     *config.NSXOperatorConfig
+	nsxOperatorNamespace   = "default"
+	enableHA               = true
 )
 
 func init() {
@@ -49,6 +51,14 @@ func init() {
 	if err != nil {
 		log.Error(err, "load config file error")
 		os.Exit(1)
+	}
+
+	if os.Getenv("NSX_OPERATOR_NAMESPACE") != "" {
+		nsxOperatorNamespace = os.Getenv("NSX_OPERATOR_NAMESPACE")
+	}
+
+	if cf.EnableHA == false {
+		enableHA = false
 	}
 
 	if metrics.AreMetricsExposed(cf) {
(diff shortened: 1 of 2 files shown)


cmd/main.go
```diff
@@ -96,10 +96,12 @@ func main() {
...
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
-		Scheme:                 scheme,
-		HealthProbeBindAddress: probeAddr,
-		MetricsBindAddress:     metricsAddr,
-		LeaderElectionID:       "nsx-operator",
+		Scheme:                  scheme,
+		HealthProbeBindAddress:  probeAddr,
+		MetricsBindAddress:      metricsAddr,
+		LeaderElection:          true,
+		LeaderElectionNamespace: "default",
```
dantingl, line 103: What is the election namespace used for? Will it create any resource in the namespace

zhengxiexie: Yes, it would create `lease` lock in the namespace, currently put it in default namespace.

cmd/main.go
```diff
@@ -96,10 +96,12 @@ func main() {
...
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
-		Scheme:                 scheme,
-		HealthProbeBindAddress: probeAddr,
-		MetricsBindAddress:     metricsAddr,
-		LeaderElectionID:       "nsx-operator",
+		Scheme:                  scheme,
+		HealthProbeBindAddress:  probeAddr,
+		MetricsBindAddress:      metricsAddr,
+		LeaderElection:          true,
```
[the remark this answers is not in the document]
zhengxiexie: Does it exist in NCP configmap now? I didn't see this option.

cmd/main.go
dantingl: Could we put it in nsx-operator namespace?

zhengxiexie: The namespace of WCP and vanilla k8s is different, so I put in the default.

timdengyun: yes, in WCP, the ncp is in vmware-system-nsx and in k8s, its in nsx-system
but, we can also change k8s env setup and put ncp and nsx-oprator in vmware-system-nsx as well. 
In such case, we just use ncp from photon image.

cmd/main.go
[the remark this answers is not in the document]
zhengxiexie: OK, I'll add it.

cmd/main.go
zhengxiexie: OK, read it from env

cmd/main.go
```diff
@@ -33,6 +33,7 @@ var (
 	probeAddr, metricsAddr string
 	log                    = logger.Log
 	cf                     *config.NSXOperatorConfig
+	ncpNamespace           = "default"
```
dantingl, line 36: call it namespace or nsxOperatorNamespace

zhengxiexie: ok

dantingl (approved):

PR #213, opened 2023-05-12
Fix makefile error after introducing client code produced by code-generator

zhengxiexie:
Add ./pkg/apis/nsx.vmware.com/v1alpha1
Fix copying files differences on linux and mac.


Makefile
```diff
@@ -40,7 +40,8 @@ help: ## Display this help.
 
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=build/yaml/crd/
+	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./pkg/apis/v1alpha1/" output:crd:artifacts:config=build/yaml/crd/
+	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./pkg/apis/v1alpha2/" output:crd:artifacts:config=build/yaml/crd/
```
timdengyun, line 44: why we have tow different version: apis/v1alpha1 and apis/v1alpha2?
Also, I noted there is one more new folder pkg/apis/nsx.vmware.com? what is purpose of this folder?

timdengyun:
> Make generated first, then make build, otherwise it reports error when clone new repo.

please add comments on why change make manifest.

Makefile
zhengxiexie: Talk as discussed. IPPool has 2 version.

hack/update-codegen.sh
```diff
@@ -18,7 +18,7 @@ CODEGEN_PKG=$(go env GOMODCACHE)/k8s.io/code-generator@v0.27.1
 rm -fr "${APIS:?}/${GROUP:?}"
 rm -fr ./pkg/client
 mkdir -p "${APIS}/${GROUP}/${VERSION}"
-cp -r "${APIS}/${VERSION}/" "${APIS}/${GROUP}/${VERSION}"
+cp -r "${APIS}/${VERSION}"/* "${APIS}/${GROUP}/${VERSION}/"
```
timdengyun, line 21: please help explain why we made such change here.

timdengyun: please change PR title: Let make build rely on make generated to reflect what your PR  does

zhengxiexie: `cp -r "${APIS}/${VERSION}/" "${APIS}/${GROUP}/${VERSION}"`
on linux, it copy the directory to dst regardless if the dst is the same as src, which is false,
on mac,  it copy the files in the directory to dst, which is true.
`cp -r "${APIS}/${VERSION}"/* "${APIS}/${GROUP}/${VERSION}/"`
escape the differences.

timdengyun: so, as we talked, 
which is false, you means this operation is incorrect, not we expected, is it？ 

The new change here:
cp -r "${APIS}/${VERSION}"/* "${APIS}/${GROUP}/${VERSION}/"
this will always copy files in the directory to dst and skip difference.

Makefile
timdengyun: do we need to update yaml as well?

timdengyun (commented):
Please see new comments https://github.com/vmware-tanzu/nsx-operator/pull/213#discussion_r1197639133

Makefile
zhengxiexie: Should we put the CRD of NCP defined earlier in pkg/apis/v1alpha1/? Otherwise, when `make manifests`, there is only v1alpha2 ippool in build/yaml/crd/nsx.vmware.com_ippools.yaml.
I manually added the v1alpha1 in the yaml, so it would override it, and only v1alpha2 was left behind. @timdengyun @heypnus @dantingl

hack/update-codegen.sh
zhengxiexie: Yes.

Makefile
timdengyun: I think if we need two versions of  ippool, we might need to put it in both.  if that, make manifests also need to put two versions' yaml in the different in build/yaml/crd, otherwise, the nsx.vmware.com_ippools.yaml will be override by the v1alpha2.

zhengxiexie: As discussed, we can only keep v1alpha2 for temp.

timdengyun (approved):

PR #215, opened 2023-05-24
Support query nsx-t resources by tags

zhengxiexie:
For example, SubnetCR and IPPoolCR both need query IpAddressPoolBlockSubnet we should filter them by tags.


pkg/nsx/services/common/store_test.go
```diff
@@ -177,5 +177,5 @@ func Test_InitializeResourceStore(t *testing.T) {
 		})
 	defer patches2.Reset()
 
-	service.InitializeResourceStore(&wg, fatalErrors, ResourceTypeRule, ruleStore)
+	service.InitializeResourceStore(&wg, fatalErrors, ResourceTypeRule, nil, ruleStore)
```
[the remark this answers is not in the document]
zhengxiexie: OK

pkg/nsx/services/common/store_test.go
```diff
@@ -177,5 +177,7 @@ func Test_InitializeResourceStore(t *testing.T) {
...
+	mTag, mScope := TagScopeNamespace, "11111"
+	service.InitializeResourceStore(&wg, fatalErrors, ResourceTypeRule, []model.Tag{{Tag: &mTag, Scope: &mScope}}, ruleStore)
```
dantingl, line 182: Please add verification logic in UT, not only here but any other places as well

zhengxiexie: OK

dantingl (approved):

PR #218, opened 2023-06-08
Update NSXServiceAccount CRD with conditions

gran-vmv:
Use Conditions instead of Phase/Reason for NSXServiceAccount.Status


pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -287,3 +288,35 @@ func (s *NSXServiceAccountService) GetNSXServiceAccountNameByUID(uid string) (na
...
+			conditions = append(conditions, *existingConditions[i].DeepCopy())
+		}
+	}
+	conditions = append(conditions, metav1.Condition{
+		Type:               v1alpha1.ConditionTypeRealized,
+		Status:             realizedStatus,
+		ObservedGeneration: generation,
+		LastTransitionTime: metav1.Now(),
+		Reason:             reason,
+		Message:            "",
```
liu4480, line 307: Message is not used here, Message is a human readable message indicating details about the transition. This may be an empty string. What if condition is false, will Message be empty?

gran-vmv: Yes, unless we want to design a human reading message for this case.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller_test.go
```diff
@@ -606,8 +654,15 @@ func TestNSXServiceAccountReconciler_updateNSXServiceAccountStatus(t *testing.T)
...
 					Status: nsxvmwarecomv1alpha1.NSXServiceAccountStatus{
-						Phase:          nsxvmwarecomv1alpha1.NSXServiceAccountPhaseFailed,
-						Reason:         "Error: test error",
+						Phase:  nsxvmwarecomv1alpha1.NSXServiceAccountPhaseFailed,
+						Reason: "Error: test error",
+						Conditions: []metav1.Condition{
+							{
+								Type:   nsxvmwarecomv1alpha1.ConditionTypeRealized,
+								Status: metav1.ConditionFalse,
+								Reason: "Error: test error",
```
liu4480, line 663: I think reason should contain a programmatic identifier indicating the reason for the condition's last transition, it should like some CONST, other than a sentense, and this should fill in message.

gran-vmv: Not get your point. This is a unit test for `updateNSXServiceAccountStatus` func. The input error is `"test error"` and the output reason is `"Error: test error"` and the func works as expected.
Currently we don't have anything to fill `Message` section thus we keep a similar implementation as `Phase`/`Reason`, unless we want to use `Message` section.

pkg/apis/v1alpha1/nsxserviceaccount_types.go
```diff
@@ -46,18 +46,22 @@ const (
...
-	Reason         string                 `json:"reason,omitempty"`
-	VPCPath        string                 `json:"vpcPath,omitempty"`
-	NSXManagers    []string               `json:"nsxManagers,omitempty"`
-	ProxyEndpoints NSXProxyEndpoint       `json:"proxyEndpoints,omitempty"`
-	ClusterID      string                 `json:"clusterID,omitempty"`
-	ClusterName    string                 `json:"clusterName,omitempty"`
-	Secrets        []NSXSecret            `json:"secrets,omitempty"`
+	// Deprecated: Use Condition instead.
+	Phase NSXServiceAccountPhase `json:"phase,omitempty"`
+	// Deprecated: Use Condition instead.
```
edwardbadboy, line 56: Is there a way to mark Phase and Reason deprecated in CRD, too?

After some googling, I found the following example.

```
// +kubebuilder:deprecatedversion:warning="example.com/v1alpha1 CronTab is deprecated" 
YourField FieldType `json:"cronTab,omitempty"`
```

build/yaml/crd/nsx.vmware.com_nsxserviceaccounts.yaml
```diff
@@ -45,11 +45,80 @@ spec:
                 type: string
               clusterName:
                 type: string
+              conditions:
+                items:
+                  description: "Condition contains details for one aspect of the current
```
edwardbadboy, line 50: This description seems too general and it's for developer. Is it a way to change it?

pkg/nsx/services/nsxserviceaccount/cluster.go
edwardbadboy: Human readable message will be helpful in debugging realization failure. If it's possible, we can put the direct error message which causes the failure here.

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -181,6 +181,7 @@ func (s *NSXServiceAccountService) CreateOrUpdateNSXServiceAccount(ctx context.C
 	// update NSXServiceAccountStatus
 	obj.Status.Phase = v1alpha1.NSXServiceAccountPhaseRealized
 	obj.Status.Reason = "Success."
```
edwardbadboy, line 183: I think Reason and Condition.Reason should be single camel case word without "." period.

If the Antrea-NSX supervisor registration code doesn't rely on this period, let's remove the period.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller_test.go
edwardbadboy: I think Reason should be more like enum. Such as "RealizationSuccess", "CannotReachNSX", "UnexpectedNSXErrorResponse", ... . "Error: test error" is more like a string for Message.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -207,6 +209,7 @@ func (r *NSXServiceAccountReconciler) updateNSXServiceAccountStatus(ctx *context
 		obj = o.DeepCopy()
 		obj.Status.Phase = nsxvmwarecomv1alpha1.NSXServiceAccountPhaseFailed
 		obj.Status.Reason = fmt.Sprintf("Error: %v", *e)
+		obj.Status.Conditions = nsxserviceaccount.GenerateNSXServiceAccountConditions(obj.Status.Conditions, obj.Generation, metav1.ConditionFalse, fmt.Sprintf("Error: %v", *e))
```
edwardbadboy, line 212: Better to put this error detail in message, and use some high-level causes in Reason . Reason should be enum. Reason is like the error type, message is like the error message. If it's hard to identify error type, let's use RealizationError as the only Reason enum.

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -287,3 +288,35 @@ func (s *NSXServiceAccountService) GetNSXServiceAccountNameByUID(uid string) (na
...
+		case v1alpha1.ConditionTypeRealized:
+		default:
+			conditions = append(conditions, *existingConditions[i].DeepCopy())
+		}
+	}
+	conditions = append(conditions, metav1.Condition{
+		Type:               v1alpha1.ConditionTypeRealized,
+		Status:             realizedStatus,
+		ObservedGeneration: generation,
+		LastTransitionTime: metav1.Now(),
```
edwardbadboy, line 305: If the condition status doesn't change, should it copy the LastTransitionTime from the existing ConditionTypeRealized condition?

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: Currently we put the error to reason, thus it is duplicated if we put it to message.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller_test.go
gran-vmv: OK. I think we can only put the module/step which get the error, but cannot report NSX error type since it is useless to collect the full list of NSX error.

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: OK, let's use Message instead of Reason.

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: No, this time means the timestamp NSXServiceAccount controller process this NSXServiceAccount, and this func will be called only when controller finished the process.

pkg/apis/v1alpha1/nsxserviceaccount_types.go
gran-vmv: I added this comment but found it make no effect on the generated code/yaml and the IDE highlights.

build/yaml/crd/nsx.vmware.com_nsxserviceaccounts.yaml
gran-vmv: It's generated from the Condition type, and I can remove it manually. Not sure how to make kubebuilder ignore the comments of Condition when generating yaml.

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -287,3 +288,34 @@ func (s *NSXServiceAccountService) GetNSXServiceAccountNameByUID(uid string) (na
...
+	return conditions
+}
+
+func IsNSXServiceAccountRealized(status v1alpha1.NSXServiceAccountStatus) bool {
+	for _, condition := range status.Conditions {
+		if condition.Type == v1alpha1.ConditionTypeRealized && condition.Status == metav1.ConditionTrue {
+			return true
+		}
+	}
+	if status.Phase == v1alpha1.NSXServiceAccountPhaseRealized {
```
liu4480, line 317: maybe use this line instead
```
return status.Phase == v1alpha1.NSXServiceAccountPhaseRealized
```

liu4480:
@edwardbadboy once this PR is merged, we need to update references in addons controllers

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
gran-vmv: Use Message instead of Reason. Reason is not used not since we only have 2 status currently and it can be covered by Phase.

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -287,3 +288,34 @@ func (s *NSXServiceAccountService) GetNSXServiceAccountNameByUID(uid string) (na
...
+		ObservedGeneration: generation,
+		LastTransitionTime: metav1.Now(),
+		Message:            message,
+	})
+	return conditions
+}
+
+func IsNSXServiceAccountRealized(status v1alpha1.NSXServiceAccountStatus) bool {
+	for _, condition := range status.Conditions {
+		if condition.Type == v1alpha1.ConditionTypeRealized && condition.Status == metav1.ConditionTrue {
```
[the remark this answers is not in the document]
gran-vmv: Not get your point
`condition.Type == v1alpha1.ConditionTypeRealized && condition.Status == metav1.ConditionTrue`
means we have a condition type is realized and the status is true, which means this nsxsa is realized.

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: Fixed.

pkg/nsx/services/nsxserviceaccount/cluster.go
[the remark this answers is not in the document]
liu4480: I mean Conditions is a slice, if there are more than one element in this slice, it should not return upon only checking the first condition.

pkg/nsx/services/nsxserviceaccount/cluster.go
liu4480: it looks like you did not change it yet

gran-vmv: This part is already merged. Do we need to change it since it is deprecated now.

pkg/nsx/services/nsxserviceaccount/cluster.go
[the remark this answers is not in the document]
gran-vmv: This code will check all conditions until a condition matches type and status, thus it works.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
edwardbadboy: According to https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition , "reason" is required. Since we only have two states, maybe use string enum such as "RealizationSuccess" "RealizationError".

pkg/nsx/services/nsxserviceaccount/cluster.go
edwardbadboy: Yes. Let's remove the ending period from "Success." if Antrea-NSX registration code doesn't rely on this period.

pkg/nsx/services/nsxserviceaccount/cluster.go
edwardbadboy: According to my understanding of the doc comment in https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition ,  LastTransitionTime should not be updated if the status doesn't change.

> // lastTransitionTime is the last time the condition transitioned from one status to another.
> // This should be when the underlying condition changed.  If that is not known, then using the time when the API field changed is acceptable.

edwardbadboy:
@liu4480 AntreaConfig controller doesn't inspect the NSXSA.status, adding a field in NSXSA will not have impact on AntreaConfig controller. We should update the reference but it's not urgent.

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: Removed the period.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
gran-vmv: Fixed.

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: Fixed. Will use existing LastTransitionTime if status is not changed.

liu4480 (approved):

edwardbadboy (approved):

PR #219, opened 2023-06-19
Change Vpc type define

seanpang-vmware:
Change VPC type define from VPC to Vpc according to nsx resource definition.

### DIFF
--- a/pkg/nsx/services/common/types.go
+++ b/pkg/nsx/services/common/types.go
@@ -46,7 +46,7 @@ var (
 	ResourceTypeSecurityPolicy = "SecurityPolicy"
 	ResourceTypeGroup          = "Group"
 	ResourceTypeRule           = "Rule"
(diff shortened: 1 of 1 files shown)


pkg/nsx/services/common/types.go
```diff
@@ -46,7 +46,7 @@ var (
 	ResourceTypeSecurityPolicy = "SecurityPolicy"
 	ResourceTypeGroup          = "Group"
 	ResourceTypeRule           = "Rule"
-	ResourceTypeVPC            = "VPC"
+	ResourceTypeVPC            = "Vpc"
```
timdengyun, line 49: It's good that we use ResourceTypeVpc as well to follow our naming convention.

seanpang-vmware: done

timdengyun (approved):

PR #220, opened 2023-06-19
Change Resource Type according to common definition

seanpang-vmware:


pkg/nsx/services/vpc/vpc.go
```diff
@@ -12,7 +12,7 @@ import (
 
 var (
 	log             = logger.Log
-	ResourceTypeVPC = common.ResourceTypeVPC
+	ResourceTypeVPC = common.ResourceTypeVpc
```
timdengyun, line 15: do we need a new var, or why not just directly use common.ResourceTypeVpc as resource type where necessary?

timdengyun (approved):

PR #221, opened 2023-06-19
Upgrade golang to latest 1.19.10

zhengxiexie:
Fix the CVEs prior to 1.19.10.


timdengyun:
please merge this PR to v4.1.0 instead of v4.1.1, also when upgrading golang version, you need also change go.mod and go.sum as in:
https://github.com/vmware-tanzu/nsx-operator/commit/0c2a7a60377fbf16842568c6a1cbdd8985950023

PR #223, opened 2023-06-20
Upgrade golang to latest 1.19.10

zhengxiexie:
Upgrade golang to v1.19.10 to escape the CVEs prior to 1.19.10.


timdengyun (commented):
Why this patch doesn't change controller-runtime version?
Do we know what controller-runtime version we are using?

zhengxiexie:
> Why this patch doesn't change controller-runtime version? Do we know what controller-runtime version we are using?

https://github.com/vmware-tanzu/nsx-operator/blob/4e2cd1d11000899527d11628fef92b8720097af7/go.mod#L38
Was v0.14.0 already.

timdengyun:
pleas also check do we need to change pkg/mock/controller-runtime/client/mocks.go? like we did in https://github.com/vmware-tanzu/nsx-operator/pull/222?

timdengyun (approved):

PR #229, opened 2023-07-05
Set EnableHA default to true

zhengxiexie:
If user forgets to set ha.enable but set replica 2, it would be confusing,
so set EnableHA default true to avoid this undefined behaviour.
Besides, NCP ha.enable default value is true, keep it aligned.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -4,7 +4,6 @@
 package main
 
 import (
-	"flag"
 	"os"
 	"time"
 
@@ -29,39 +28,32 @@ import (
 )
 
 var (
-	scheme                 = runtime.NewScheme()
-	probeAddr, metricsAddr string
-	log                    = logger.Log
-	cf                     *config.NSXOperatorConfig
-	nsxOperatorNamespace   = "default"
-	enableHA               = true
+	scheme               = runtime.NewScheme()
+	log                  = logger.Log
+	cf                   *config.NSXOperatorConfig
+	nsxOperatorNamespace = "default"
 )
 
 func init() {
+	var err error
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 	utilruntime.Must(v1alpha1.AddToScheme(scheme))
-	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8384", "The address the probe endpoint binds to.")
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8093", "The address the metrics endpoint binds to.")
 	config.AddFlags()
-	flag.Parse()
-	var err error
-
-	logf.SetLogger(logger.ZapLogger())
 	cf, err = config.NewNSXOperatorConfigFromFile()
 	if err != nil {
 		log.Error(err, "load config file error")
 		os.Exit(1)
 	}
 
+	logf.SetLogger(logger.ZapLogger())
+
 	if os.Getenv("NSX_OPERATOR_NAMESPACE") != "" {
 		nsxOperatorNamespace = os.Getenv("NSX_OPERATOR_NAMESPACE")
 	}
 
-	if cf.EnableHA == nil || *cf.EnableHA == true {
-		enableHA = true
+	if cf.HAEnabled() {
 		log.Info("HA mode enabled")
 	} else {
-		enableHA = false
 		log.Info("HA mode disabled")
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/config/config.go
```diff
@@ -132,6 +132,9 @@ func NewNSXOperatorConfigFromFile() (*NSXOperatorConfig, error) {
 	if err != nil {
 		return nil, err
 	}
+	if !cfg.Section("ha").HasKey("enable") {
+		nsxOperatorConfig.HAConfig.EnableHA = true
+	}
```
gran-vmv, line 137: Do you want to use pointer for this?
```
type HAConfig struct {
	EnableHA *bool `ini:"enable"`
}
```

timdengyun: WCP forgets to set ha.enable but set replica 2, it would be confusing, so set EnableHA default true to avoid this undefined behaviour.

Please remove WCP in comments, it's internal words not for public.

zhengxiexie: Now is

zhengxiexie: > WCP forgets to set ha.enable but set replica 2, it would be confusing, so set EnableHA default true to avoid this undefined behaviour.
> 
> Please remove WCP in comments, it's internal words not for public.

OK

timdengyun: > If user forgets to set ha.enable but set replica 2, it would be confusing, so set EnableHA default true to avoid this undefined behavior.

We can say  "NCP ha.enable default value is true, so setting EnableHA default true as well to align with NCP default setting.

dantingl:
Could you also remove the default value in main.go?

pkg/config/config.go
gran-vmv: > Now is
> 
> type HAConfig struct { EnableHA bool `ini:"enable"` }
> 
> Like others in this file.

In golang, if you want to check if an option is unset, we should use pointer in the target struct, then this parameter will be `nil` when it is not set or invalid.

zhengxiexie:
> Could you also remove the default value in main.go?

ok

pkg/config/config.go
zhengxiexie: Thanks, it is more Golang taste in this way, updated.

zhengxiexie: > > If user forgets to set ha.enable but set replica 2, it would be confusing, so set EnableHA default true to avoid this undefined behavior.
> 
> We can say "NCP ha.enable default value is true, so setting EnableHA default true as well to align with NCP default setting.

Appended.

pkg/config/config.go
```diff
@@ -132,6 +132,9 @@ func NewNSXOperatorConfigFromFile() (*NSXOperatorConfig, error) {
 	if err != nil {
 		return nil, err
 	}
+	if !cfg.Section("ha").HasKey("enable") {
```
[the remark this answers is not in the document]
liu4480: I mean if replicas is 1, does HA enable takes effect?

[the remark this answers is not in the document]
zhengxiexie: Yes, and the process gets the lock and run, if another process joins, it would wait.

cmd/main.go
```diff
@@ -34,7 +34,7 @@ var (
 	log                    = logger.Log
 	cf                     *config.NSXOperatorConfig
 	nsxOperatorNamespace   = "default"
-	enableHA               = true
+	enableHA               = false
```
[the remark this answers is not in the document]
zhengxiexie: ok

pkg/mock/nsxop.ini
```diff
@@ -6,3 +6,6 @@ nsx_api_managers = 127.0.0.1
 nsx_api_password = admin
```
timdengyun, line 6: DEFAULT]
debug = True
[nsx_v3]
nsx_api_user = admin
nsx_api_password = admin
[ha]
#enable=true
[k8s]
[coe]
cluster = k8scl-one

It's good that we change nsxop.ini file as the formal ini format, so that others can also follow this way to add more UTS if necessary.

cmd/main.go
```diff
@@ -57,8 +57,12 @@ func init() {
 		nsxOperatorNamespace = os.Getenv("NSX_OPERATOR_NAMESPACE")
 	}
 
-	if cf.EnableHA == false {
+	if cf.EnableHA == nil || *cf.EnableHA == true {
```
timdengyun, line 60: Maybe we can warp this  EnableHA initialization into a function and put in config.go, and then we can use UT to test this function.

timdengyun: Moreover, in the future, I'm thinking we also need follow this way to initialize the config options in a a central place like in config.go instead of in main

pkg/mock/nsxop.ini
zhengxiexie: OK

dantingl (approved):

gran-vmv (approved):
LGTM. Could you also add other changes to title/description? It seems we added some refactor in this PR.

pkg/config/config.go
```diff
@@ -38,6 +40,13 @@ type NSXOperatorConfig struct {
 	*HAConfig
 }
 
+func (operatorConfig *NSXOperatorConfig) HAEnabled() bool {
+	if operatorConfig.EnableHA == nil || *operatorConfig.EnableHA == true {
```
TaoZou1, line 44: How about set default value in NewNSXOpertorConfig?
It semes not good idea after NewNSXOpertorConfig, it still have some filed not initialized.
So you could also using EnableHA directly instead of HAEnabled().
and EnableHA type could also be bool.

PR #236, opened 2023-07-21
Add RealizeState service

jwsui:
Add service to implement checking realize status of entity with retries.

### DIFF
--- a/pkg/nsx/client.go
+++ b/pkg/nsx/client.go
@@ -18,6 +18,7 @@ import (
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra/domains"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra/domains/security_policies"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra/sites/enforcement_points"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/orgs/projects/infra/realized_state"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/search"
 
 	"github.com/vmware-tanzu/nsx-operator/pkg/config"
@@ -39,6 +40,7 @@ type Client struct {
 	RuleClient                 security_policies.RulesClient
 	InfraClient                nsx_policy.InfraClient
 	ClusterControlPlanesClient enforcement_points.ClusterControlPlanesClient
+	RealizedEntitiesClient     realized_state.RealizedEntitiesClient
 
 	MPQueryClient             mpsearch.QueryClient
 	CertificatesClient        trust_management.CertificatesClient
@@ -90,6 +92,7 @@ func GetClient(cf *config.NSXOperatorConfig) *Client {
 	ruleClient := security_policies.NewRulesClient(restConnector(cluster))
 	infraClient := nsx_policy.NewInfraClient(restConnector(cluster))
 	clusterControlPlanesClient := enforcement_points.NewClusterControlPlanesClient(restConnector(cluster))
+	realizedEntitiesClient := realized_state.NewRealizedEntitiesClient(restConnector(cluster))
 
 	mpQueryClient := mpsearch.NewQueryClient(restConnector(cluster))
 	certificatesClient := trust_management.NewCertificatesClient(restConnector(cluster))
@@ -114,6 +117,7 @@ func GetClient(cf *config.NSXOperatorConfig) *Client {
 		InfraClient:    infraClient,
 
 		ClusterControlPlanesClient: clusterControlPlanesClient,
+		RealizedEntitiesClient:     realizedEntitiesClient,
 
 		MPQueryClient:             mpQueryClient,
 		CertificatesClient:        certificatesClient,
--- a/pkg/nsx/services/realizestate/realize_state.go
+++ b/pkg/nsx/services/realizestate/realize_state.go
@@ -1,0 +1,60 @@
+/* Copyright © 2023 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: Apache-2.0 */
+
+package realizestate
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
+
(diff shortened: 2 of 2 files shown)


pkg/nsx/client.go
```diff
@@ -90,6 +92,7 @@ func GetClient(cf *config.NSXOperatorConfig) *Client {
 	ruleClient := security_policies.NewRulesClient(restConnector(cluster))
 	infraClient := nsx_policy.NewInfraClient(restConnector(cluster))
 	clusterControlPlanesClient := enforcement_points.NewClusterControlPlanesClient(restConnector(cluster))
+	realizedStateClient := realized_state.NewRealizedEntitiesClient(restConnector(cluster))
```
heypnus, line 95: I prefer to use `realizedEntitiesClient` as the var name because it will be easier to match the struct in vsphere-automation-sdk-go (there're many lookalike structs, e.g. `NewRealizedEntitiesClient` and `NewRealizedEntitiyClient`.. So we'd better to use the full name to distinguish them more carefully)

pkg/nsx/services/realizestate/realize_state.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+}
+
+// CheckRealizeState allows the caller to check realize status of entityType with retries.
+// backoff defines the maximum retries and the wait interval between two retries.
+func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, orgID, projectID, intentPath, entityType string) error {
+	return retry.OnError(backoff, func(err error) bool {
+		return err != nil
+	}, func() error {
+		results, err := service.NSXClient.RealizedStateClient.List(orgID, projectID, intentPath, nil)
```
heypnus, line 32: In my code, I imported `"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra/realized_state"`, so the `List` function only needs the path, e.g.
`realizedPort, err := service.NSXClient.RealizedEntitiesClient.List(*nsxSubnetPort.Path, nil)`
Which one do you prefer to?

pkg/nsx/services/realizestate/realize_state.go
```diff
@@ -0,0 +1,47 @@
...
+		for _, result := range results.Results {
+			if *result.EntityType != entityType {
+				continue
+			}
+			if *result.State == "REALIZED" {
+				return nil
+			}
+			return errors.New(fmt.Sprintf("%s not realized", entityType))
+		}
+		return errors.New(fmt.Sprintf("%s not realized", entityType))
```
heypnus, line 45: Can we return the original error retrieved at L32? E.g. we need to delete the subnet port in GenericPolicyRealizedResource_STATE_ERROR but don't need to delete it in the other unrealized state.

pkg/nsx/client.go
jwsui: Make sense, will update the name, thanks.

pkg/nsx/services/realizestate/realize_state.go
jwsui: I didn't try `infra/realize_state`, but I think for VPC, we should always use the client under `orgs/projects/infra`?

pkg/nsx/services/realizestate/realize_state.go
jwsui: The original error is returned in L34?

jwsui: > Can we return the original error retrieved at L32? E.g. we need to delete the subnet port in GenericPolicyRealizedResource_STATE_ERROR but don't need to delete it in the other unrealized state.

Got your point, but I found that `GenericPolicyRealizedResource_STATE_ERROR` is string type(not error type), so it can't be used directly. So how about returning errors.New(*result.state) and using `error.Error() == GenericPolicyRealizedResource_STATE_ERROR` to make the judgement?

jwsui: > Can we return the original error retrieved at L32? E.g. we need to delete the subnet port in GenericPolicyRealizedResource_STATE_ERROR but don't need to delete it in the other unrealized state.

Got your point, but I found that `GenericPolicyRealizedResource_STATE_ERROR` is string type(not error type), so it can't be used directly. So how about returning errors.New(*result.state) and using `error.Error() == GenericPolicyRealizedResource_STATE_ERROR` to make the judgement?

pkg/nsx/services/realizestate/realize_state.go
heypnus: It's ok for me.

pkg/nsx/services/realizestate/realize_state.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+}
+
+func IsRealizeStateError(err error) bool {
+	return err.Error() == model.GenericPolicyRealizedResource_STATE_ERROR
+}
+
+// CheckRealizeState allows the caller to check realize status of entityType with retries.
+// backoff defines the maximum retries and the wait interval between two retries.
+func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, orgID, projectID, intentPath, entityType string) error {
```
heypnus, line 36: Is it possible to simplify the function from `func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, orgID, projectID, intentPath, entityType string) error {...}` to `func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, intentPath string) error {...}`? i.e.
1. get the org/project id from intentPath instead of parameters
2. entity type is not required, right?

jwsui: > Is it possible to simplify the function from `func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, orgID, projectID, intentPath, entityType string) error {...}` to `func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, intentPath string) error {...}`? i.e.
> 
> 1. get the org/project id from intentPath instead of parameters
> 2. entity type is not required, right?

For 1, I will rebase and update patch after we merged the patch of parsing parameters from path.

For 2, entity type is required in Line 46, return value of `RealizedEntitiesClient.List` is a list, containing realized status of all related intent types. So `entity type` is used to check the status of entity which we really care.

jwsui: Done.

lxiaopei (commented):
LGTM.

heypnus (approved):
LGTM

PR #238, opened 2023-07-26
Remove empty item from thumbprint/nsxapimanagers, in case user input more comma than expected

TaoZou1:


pkg/config/config.go
```diff
@@ -215,7 +215,20 @@ func (vcConfig *VCConfig) validate() error {
...
 
+func removeEmptyItem(source []string) []string {
+	target := make([]string, 0)
+	for _, value := range source {
+		if len(value) == 0 {
+			continue
+		}
+		target = append(target, value)
+	}
+	return target
```
lxiaopei, line 226: From what I learned from WCPSVC team these days, they prefer the below way, they said the performance is slightly faster.
target := make([]string, 0, len(source))
for i, value := range source {
    if len(value) !=0 {
    target[i] = value
}
}

TaoZou1: but we don't know the length of target, e.g source length is 4, but 3 of them are empty item. then target only has 1 item.

timdengyun:
Please close this PR and submit PR to main branch. From now on, we only keep code related to vpc feature in vpc branch. So, bug fix for previous release should be merged to main branch. 

Also, in main branch, we have config.test to add config UT test. so it's good that we add a UT for this new function as well to test extra comma case in ini file.

Finally, please keep PR name a little bit shorter and put more description in commit message.

PR #239, opened 2023-07-27
Remove empty items from thumbprint/nsxapimanagers

TaoZou1:
If user input more comma than expected, some of items value may be empty. Remove those empty items

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -15,7 +15,7 @@ import (
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/auth/jwt"
 )
 
-//TODO replace to yaml
+// TODO replace to yaml
 const (
 	nsxOperatorDefaultConf = "/etc/nsx-operator/nsxop.ini"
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -197,7 +197,20 @@ func (vcConfig *VCConfig) validate() error {
...
+	target := make([]string, 0)
+	for _, value := range source {
+		if len(value) == 0 {
+			continue
+		}
+		target = append(target, value)
+	}
+	return target
+}
+
```
lxiaopei, line 210: target := make([]string, 0, len(source))
	for i, value := range source {
		if len(value) != 0 {
			target[i] = value
		}
	}
	return target

timdengyun:
Do we need to add one UT to test this case?

pkg/config/config.go
TaoZou1: It doesn't work. target[i]=value will crash. unless
target := make([]string, len(source))

timdengyun (approved):

PR #240, opened 2023-07-27
Init logger before load operator config file

timdengyun:
Initializing logger before loading operator config file, otherwise, the log info will not be logged

Also, change log info level in validating config and go-fmt files

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -39,13 +39,13 @@ func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 	utilruntime.Must(v1alpha1.AddToScheme(scheme))
 	config.AddFlags()
+
+	logf.SetLogger(logger.ZapLogger())
 	cf, err = config.NewNSXOperatorConfigFromFile()
 	if err != nil {
 		log.Error(err, "load config file error")
 		os.Exit(1)
 	}
-
-	logf.SetLogger(logger.ZapLogger())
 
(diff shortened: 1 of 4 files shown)


TaoZou1 (dismissed):

cmd/main.go
```diff
@@ -60,6 +60,7 @@ func init() {
 	if metrics.AreMetricsExposed(cf) {
 		metrics.InitializePrometheusMetrics()
 	}
+	log.Info("completed NSX Operator initialization")
```
dantingl, line 63: I do not think this log is needed

pkg/config/config.go
```diff
@@ -235,11 +234,11 @@ func (nsxConfig *NsxConfig) validate() error {
 	}
 	tpCount := len(nsxConfig.Thumbprint)
 	if tpCount == 0 {
-		log.V(1).Info("no thumbprint provided")
+		log.Info("no thumbprint provided")
 		return nil
 	}
 	if tpCount == 1 {
-		log.V(1).Info("all endpoints share one thumbprint")
```
dantingl, line 242: I still think these two are V1 level..

cmd/main.go
timdengyun: Agreed, removed.

pkg/config/config.go
timdengyun: Revert to V(1)

TaoZou1 (approved):

PR #242, opened 2023-07-28
Init logger before load operator config file

timdengyun:
Initializing logger before loading operator config file, otherwise, the log info will not be logged.

Referring to https://github.com/vmware-tanzu/nsx-operator/pull/141 to fix log-level setting bug.

Also, change log info level in validating config and go-fmt files.


pkg/logger/logger.go
```diff
@@ -52,15 +52,13 @@ func ZapLogger(cf *config.NSXOperatorConfig) logr.Logger {
 	}
 	opts.BindFlags(flag.CommandLine)
 
-	if cf.Debug == true {
```
dantingl, line 55: do we still use Debug to set log level?

timdengyun: Since in main.go#L44, we have to create logger before config initialization, so have to   remove parameter cf *config.NSXOperatorConfig from function ZapLogger here.
main.go#L44
>logf.SetLogger(logger.ZapLogger())
>cf, err = config.NewNSXOperatorConfigFromFile()

Moreover, this is to align the changes with main branch.

With regarding to Debug flag setting from config file, we can search a way that can support dynamically change log level by setting logger options be reading config debug flag. This will be addressed in https://github.com/vmware-tanzu/nsx-operator/issues/245
So for this moment,this patch only resolve logger initialization problem before creating config

TaoZou1 (approved):

PR #260, opened 2023-08-07
Check EnableVPCNetwork and nsx version

zhengxiexie:
Start associated controllers only when EnableVPCNetwork is true and nsx version support it.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -132,22 +132,26 @@ func main() {
 		NSXConfig: cf,
 	}
 
-	// Start the security policy controller.
+	if cf.CoeConfig.EnableVPCNetwork && commonService.NSXClient.NSXCheckVersion(nsx.VPC) {
+		log.V(1).Info("VPC mode enabled")
+		// Start controllers which only supports VPC
+		// Start subnet/subnetset controller.
+		if err := subnet.StartSubnetController(mgr, commonService); err != nil {
+			os.Exit(1)
+		}
+		if err := subnetset.StartSubnetSetController(mgr, commonService); err != nil {
+			os.Exit(1)
+		}
+
+		staticroutecontroller.StartStaticRouteController(mgr, commonService)
+		subnetport.StartSubnetPortController(mgr, commonService)
+	}
+	// Start the security policy controller, it supports VPC and non VPC mode
 	StartSecurityPolicyController(mgr, commonService)
-	staticroutecontroller.StartStaticRouteController(mgr, commonService)
 	// Start the NSXServiceAccount controller.
 	if cf.EnableAntreaNSXInterworking {
 		StartNSXServiceAccountController(mgr, commonService)
 	}
-	// Start subnet/subnetset controller.
-	if err := subnet.StartSubnetController(mgr, commonService); err != nil {
-		os.Exit(1)
-	}
-	if err := subnetset.StartSubnetSetController(mgr, commonService); err != nil {
-		os.Exit(1)
-	}
-
-	subnetport.StartSubnetPortController(mgr, commonService)
 
 	if metrics.AreMetricsExposed(cf) {
 		go updateHealthMetricsPeriodically(nsxClient)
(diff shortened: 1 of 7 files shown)


cmd/main.go
```diff
@@ -130,7 +130,9 @@ func main() {
 	}
 
 	// Start the security policy controller.
-	StartSecurityPolicyController(mgr, commonService)
+	if cf.CoeConfig.EnableVPCNetwork && commonService.NSXClient.NSXCheckVersionForSecurityPolicy() {
```
dantingl, line 133: If EnableVPCNetwork is true, you need to check NSX version(>= 4.1.1). If NSX version is lower than 4.1.1, we cannot start VPC related controllers

zhengxiexie: Yes, we should complete the featureSupported func later, and apply this pattern to start every controller.
This is just a pattern here.

cmd/main.go
```diff
@@ -129,8 +129,11 @@ func main() {
 		NSXConfig: cf,
 	}
 
-	// Start the security policy controller.
+	// Start the security policy controller, it supports VPC and non VPC mode
 	StartSecurityPolicyController(mgr, commonService)
+	if cf.CoeConfig.EnableVPCNetwork && commonService.NSXClient.NSXCheckVersionFor(nsx.FeatureVPC) {
+		// Start controllers which only supports VPC
```
dantingl, line 135: Add a debug log about starting nsx-operator on VPC mode

cmd/main.go
```diff
@@ -129,8 +129,11 @@ func main() {
...
+	}
```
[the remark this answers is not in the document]
zhengxiexie: ok

cmd/main.go
zhengxiexie: ok

cmd/main.go
```diff
@@ -129,7 +129,11 @@ func main() {
...
 
-	// Start the security policy controller.
+	if cf.CoeConfig.EnableVPCNetwork && commonService.NSXClient.NSXCheckVersionFor(nsx.FeatureVPC) {
+		log.V(1).Info("VPC mode enabled")
+		// Start controllers which only supports VPC
+	}
+	// Start the security policy controller, it supports VPC and non VPC mode
 	StartSecurityPolicyController(mgr, commonService)
 	// Start the NSXServiceAccount controller.
 	if cf.EnableAntreaNSXInterworking {
```
timdengyun, line 139: StartNSXServiceAccountController might be used in non-vpc mode as well, need to confirm with GuRan.

cmd/main.go
```diff
@@ -129,7 +129,11 @@ func main() {
 		NSXConfig: cf,
 	}
 
-	// Start the security policy controller.
+	if cf.CoeConfig.EnableVPCNetwork && commonService.NSXClient.NSXCheckVersionFor(nsx.FeatureVPC) {
+		log.V(1).Info("VPC mode enabled")
+		// Start controllers which only supports VPC
+	}
+	// Start the security policy controller, it supports VPC and non VPC mode
```
timdengyun, line 136: if cf.CoeConfig.EnableVPCNetwork is false, we also need to start security policy controller because securitypolicy controller can also work in non vpc mode.

zhengxiexie: Yes, it is as expected now.

cmd/main.go
zhengxiexie: ditto.

dantingl (approved):

cmd/main.go
timdengyun: I see. I miss code line #L138.

cmd/main.go
```diff
@@ -132,7 +132,11 @@ func main() {
...
 	}
 
-	// Start the security policy controller.
+	if cf.CoeConfig.EnableVPCNetwork && commonService.NSXClient.NSXCheckVersion(nsx.VPC) {
+		log.V(1).Info("VPC mode enabled")
+		// Start controllers which only supports VPC
+	}
+	// Start the security policy controller, it supports VPC and non VPC mode
 	StartSecurityPolicyController(mgr, commonService)
 	staticroutecontroller.StartStaticRouteController(mgr, commonService)
```
timdengyun, line 141: For other vpc related controllers: StartStaticRouteController, StartSubnetController,StartSubnetSetController, .StartSubnetPortController need to put in the condition: if cf.CoeConfig.EnableVPCNetwork

dantingl (approved):

cmd/main.go
zhengxiexie: ok

PR #261, opened 2023-08-07
Refactor loglevel based on debug in configmap

zhengxiexie:
We hold these truths to be self-evident:
log.info -> info, log.v(1).info -> debug, log.error-> error, log.v(2).info-> develop level log

If config.Debug=true, then loglevel=1, if command is passed in loglevel and greater than debug loglevel, then the loglevel is overridden.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -46,7 +46,7 @@ func init() {
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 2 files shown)


timdengyun:
please close this one and fix the issue in the main branch firstly.

dantingl (approved):

Issue #245, opened 2023-07-28
Operator Config [DEFAULT] debug = True is not working as expecting DEBUG log message output


timdengyun:
When user set config option debug = True is not working as expected. The logger output is the same with default level:0

In current logger, user can input argument  -log-level int to change logger level, 
The default -log-level  is 0, which means info and Error log will be logged.

The -log-level is set as 1, which means info, Error and Debug log will be logged. 
For example: when -log-level is 1, we can see DEBUG message like:
2023-07-28 05:13:25.303	DEBUG	cmd/main.go:59	starting NSX Operator

The -log-level is set as 2, which means info, Error, Debug log and more less detailed message will be logged. 
Like output below:
2023-07-28 05:13:25.303	DEBUG	cmd/main.go:59	starting NSX Operator
2023-07-28 05:13:25.303	LEVEL(-2)	cmd/main.go:60	starting NSX Operator

So,  if setting debug = True in config file, which means  log level  should be 1. So,  we need to change log level is equal to or more than 1, depending -log-level set by user,  it should take bigger value of two inputs.

timdengyun:
we discussed, we only adopt 0-1 two level in our implementation, so we only yse  log.info, log.V(1).info and log.Error.
If user set  debug = True , we force -log-level is 1 no matter user input from command line for this one.

timdengyun:
@zhengxiexie  please help take a look at this issue.

zhengxiexie:
https://github.com/vmware-tanzu/nsx-operator/pull/261

PR #265, opened 2023-08-09
Refactor loglevel based on debug in configmap

zhengxiexie:
This PR is to fix Issue: https://github.com/vmware-tanzu/nsx-operator/issues/245

We hold these truths to be self-evident:
log.info -> info, log.v(1).info -> debug, log.error-> error, log.v(2).info-> develop level log

If config.Debug=true, then loglevel=1, if command is passed in loglevel and greater than debug loglevel, then the loglevel is overridden.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -40,12 +40,12 @@ func init() {
 	utilruntime.Must(v1alpha1.AddToScheme(scheme))
 	config.AddFlags()
 
-	logf.SetLogger(logger.ZapLogger())
 	cf, err = config.NewNSXOperatorConfigFromFile()
 	if err != nil {
-		log.Error(err, "load config file error")
 		os.Exit(1)
 	}
+
+	logf.SetLogger(logger.ZapLogger(cf))
 
 	if os.Getenv("NSX_OPERATOR_NAMESPACE") != "" {
 		nsxOperatorNamespace = os.Getenv("NSX_OPERATOR_NAMESPACE")
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -9,8 +9,8 @@ import (
 	"fmt"
 	"os"
 
+	"go.uber.org/zap"
 	ini "gopkg.in/ini.v1"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/auth"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/auth/jwt"
@@ -26,7 +26,7 @@ var (
 	LogLevel               int
 	ProbeAddr, MetricsAddr string
 	configFilePath         = ""
-	log                    = logf.Log.WithName("config")
+	configLog              *zap.SugaredLogger
 	tokenProvider          auth.TokenProvider
 )
 
@@ -38,6 +38,11 @@ type NSXOperatorConfig struct {
 	*K8sConfig
 	*VCConfig
 	*HAConfig
+}
+
+func init() {
+	zapLogger, _ := zap.NewProduction()
+	configLog = zapLogger.Sugar()
 }
 
 func (operatorConfig *NSXOperatorConfig) HAEnabled() bool {
@@ -112,7 +117,8 @@ func UpdateConfigFilePath(configFile string) {
 	configFilePath = configFile
 }
 
-func NewNSXOperatorConfigFromFile() (*NSXOperatorConfig, error) {
+func LoadConfigFromFile() (*NSXOperatorConfig, error) {
+	configLog.Infof("loading NSX Operator configuration file: %s", configFilePath)
 	nsxOperatorConfig := NewNSXOpertorConfig()
 
 	cfg := ini.Empty()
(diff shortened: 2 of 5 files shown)


pkg/config/config.go
```diff
@@ -109,6 +113,7 @@ func UpdateConfigFilePath(configFile string) {
 }
 
 func NewNSXOperatorConfigFromFile() (*NSXOperatorConfig, error) {
+	log.Info("Loading NSX Operator configuration file", "path", configFilePath)
```
dantingl, line 116: use lower case for log message? Loading -> loading

pkg/config/config.go
```diff
@@ -186,7 +191,7 @@ func (operatorConfig *NSXOperatorConfig) GetTokenProvider() auth.TokenProvider {
 }
 
 func (operatorConfig *NSXOperatorConfig) createTokenProvider() error {
-	log.V(1).Info("try to load VC host CA")
+	log.Info("try to load VC host CA")
```
dantingl, line 194: I guess they are debug log @TaoZou1

pkg/config/config.go
zhengxiexie: ok

pkg/config/config.go
zhengxiexie: fixed

pkg/config/config.go
```diff
@@ -186,7 +191,7 @@ func (operatorConfig *NSXOperatorConfig) GetTokenProvider() auth.TokenProvider {
 }
 
 func (operatorConfig *NSXOperatorConfig) createTokenProvider() error {
-	log.V(1).Info("try to load VC host CA")
+	log.Debug("try to load VC host CA")
```
timdengyun, line 194: please still keep log.V(1) or log.Info, we don't need two kinds logger format in our code.

pkg/config/config.go
```diff
@@ -8,8 +8,8 @@ import (
 	"flag"
 	"io/ioutil"
 
+	"go.uber.org/zap"
 	ini "gopkg.in/ini.v1"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
```
timdengyun, line 12: I understand circle dependency issue you worry about, we don't need to import "github.com/vmware-tanzu/nsx-operator/pkg/logger", I mean do we still use "sigs.k8s.io/controller-runtime/pkg/log"lib to create another logger, instead of using a zap logger, is it possible?

cmd/main.go
```diff
@@ -40,13 +40,13 @@ func init() {
 	utilruntime.Must(v1alpha1.AddToScheme(scheme))
 	config.AddFlags()
 
-	logf.SetLogger(logger.ZapLogger())
 	cf, err = config.NewNSXOperatorConfigFromFile()
 	if err != nil {
-		log.Error(err, "load config file error")
```
timdengyun, line 46: we still need to keep this error log.

timdengyun: For example, the error returned from in func NewNSXOperatorConfigFromFile
func NewNSXOperatorConfigFromFile() (*NSXOperatorConfig, error) {
	nsxOperatorConfig := NewNSXOpertorConfig()

	cfg := ini.Empty()
	err := ini.ReflectFrom(cfg, nsxOperatorConfig)
	if err != nil {
		return nil, err
	}
	cfg, err = ini.Load(configFilePath)
	if err != nil {
		return nil, err
	}
...
 these error only logged here.  So, we can consider use another logger to log config file info and errors. but this logger need be created before  NewNSXOperatorConfigFromFile and pass it to config file.

zhengxiexie: As discussed, wrap a function and log.error in it.

pkg/config/config.go
```diff
@@ -108,7 +112,8 @@ func UpdateConfigFilePath(configFile string) {
 	configFilePath = configFile
 }
 
-func NewNSXOperatorConfigFromFile() (*NSXOperatorConfig, error) {
+func LoadConfigFromFile() (*NSXOperatorConfig, error) {
+	log.Infof("loading NSX Operator configuration file: %s", configFilePath)
```
timdengyun, line 116: why log.Infof? not log.Info?

pkg/config/config.go
```diff
@@ -152,6 +157,15 @@ func NewNSXOperatorConfigFromFile() (*NSXOperatorConfig, error) {
 	return nsxOperatorConfig, nil
 }
 
+func NewNSXOperatorConfigFromFile() (*NSXOperatorConfig, error) {
```
timdengyun, line 160: It's good to add a UT case to test NewNSXOperatorConfigFromFile returning an error

pkg/config/config.go
zhengxiexie: zap.SugaredLogger , the usage is different.

pkg/config/config.go
zhengxiexie: done

pkg/config/config.go
timdengyun: It's still confused why log.Info is used uere
also in func:
func (vcConfig *VCConfig) validate() error {
etc.

we suppose you use a different logger in this config.go, if that is case, why some functions are using log.Info, however, some are using log.Infof?

log.Info is supposed to be used in other modules, log.Infof is using sugar logger which is specific used in config.go. If that is the case, we had better use a newname for logger in config.go,
like, configLog. It will be easier to distinguish loggers used in different modules.

zhengxiexie: OK, renamed.

timdengyun (approved):

PR #269, opened 2023-08-09
Support to create subnetport in subnetset

heypnus:
1. Adding index by subnet CR uid in subnetport store.
2. Adding function GetPortsOfSubnet in service mediator -- moved several functions from
   service to controller to support it.
3. Support to create subnetport in subnetset belongs to shared VPC or current namespace.
4. Updating vsphere-automation-sdk-go/services/nsxt to 0.9.8 to fix the relization      
   error for subnetport.
5. Updating the real value to subnetport.Status.LogicalSwitchID.

### DIFF
--- a/pkg/controllers/subnetport/subnetport_controller.go
+++ b/pkg/controllers/subnetport/subnetport_controller.go
@@ -10,11 +10,13 @@ import (
 	"os"
 	"reflect"
 	"runtime"
+	"strings"
 	"time"
 
 	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
 	"github.com/vmware-tanzu/nsx-operator/pkg/metrics"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apimachineryruntime "k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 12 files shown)


dantingl (approved):

pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -295,3 +303,126 @@ func updateSuccess(r *SubnetPortReconciler, c *context.Context, o *v1alpha1.Subn
 func deleteSuccess(r *SubnetPortReconciler, _ *context.Context, _ *v1alpha1.SubnetPort) {
 	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteSuccessTotal, MetricResTypeSubnetPort)
 }
+
+func (r *SubnetPortReconciler) GetSubnetPathForSubnetPort(obj *v1alpha1.SubnetPort) (string, error) {
+	subnetPath := ""
+	if len(obj.Spec.Subnet) > 0 {
+		subnet := &v1alpha1.Subnet{}
+		namespacedName := types.NamespacedName{
```
lxiaopei, line 311: This reminds me, when creating subnet/Subnetset in NS, we also need to find the VPC if VPC is shared across NSes.
I think we need a TODO.

dantingl (approved):

PR #273, opened 2023-08-10
Add ca_file support for tls

TaoZou1:

### DIFF
--- a/pkg/nsx/cluster.go
+++ b/pkg/nsx/cluster.go
@@ -6,23 +6,24 @@ package nsx
 import (
 	"crypto/sha1"
 	"crypto/tls"
+	"crypto/x509"
 	"errors"
 	"fmt"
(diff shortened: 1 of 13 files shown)


pkg/nsx/cluster.go
```diff
@@ -120,26 +121,67 @@ func (cluster *Cluster) getThumbprint(addr string) string {
 	return thumbprint
 }
 
+func (cluster *Cluster) getCaFile(addr string) string {
+	host := addr[:strings.Index(addr, ":")]
+	var cafile string
+	tpCount := len(cluster.config.CAFile)
```
dantingl, line 127: tp is short for thumbprint?

TaoZou1: yes

dantingl (approved):

PR #277, opened 2023-08-14
Fix errors when deleting ip block from ipblock store

seanpang-vmware:


pkg/nsx/services/vpc/vpc.go
```diff
@@ -188,14 +188,13 @@ func (service *VPCService) DeleteIPBlock(vpc model.Vpc) error {
 				vpcCRUid = *tag.Tag
 			}
 		}
-		log.Info("search ip block from store using index", "index", common.TagScopeVPCCRUID, "value", vpcCRUid)
+		log.Info("search ip block from store using index and path", "index", common.TagScopeVPCCRUID, "value", vpcCRUid, "path", block)
```
dantingl, line 191: v2 log level

pkg/nsx/services/vpc/vpc.go
```diff
@@ -188,14 +188,13 @@ func (service *VPCService) DeleteIPBlock(vpc model.Vpc) error {
...
-		// TODO: bugfix, seems using vpc cr uid can not get the ip blocks from cache, need further checking
-		ipblocks := service.IpblockStore.GetByIndex(common.TagScopeVPCCRUID, vpcCRUid)
-		if ipblocks != nil && len(ipblocks) != 0 {
-			log.Info("deleting ip blocks", "IPBlock", ipblocks[0])
-			b := ipblocks[0].(model.IpAddressBlock)
-			b.MarkedForDelete = &MarkedForDelete
-			service.IpblockStore.Operate(&ipblocks[0])
+		log.Info("search ip block from store using index and path", "index", common.TagScopeVPCCRUID, "value", vpcCRUid, "path", block)
+		// using index vpc cr id may get multiple ipblocks, add path to filter the correct one
+		ipblock := service.IpblockStore.GetByIndexAndPath(common.TagScopeVPCCRUID, vpcCRUid, block)
```
dantingl, line 193: I think you can add path as index, then you can just delete it from store directly

lxiaopei (approved):
LGTM.

PR #280, opened 2023-08-15
Move startup into controllers

dantingl:


cmd/main.go
```diff
@@ -195,31 +98,54 @@ func main() {
...
+			os.Exit(1)
+		}
+		if err := subnetport.StartSubnetPortController(mgr, commonService); err != nil {
+			log.Error(err, "failed to start SubnetPort controller")
+			os.Exit(1)
+		}
+		if err := ippool.StartIPPoolController(mgr, commonService); err != nil {
+			log.Error(err, "failed to start IPPool controller")
 			os.Exit(1)
 		}
```
seanpang-vmware, line 136: is it possible that we use a interface Controller, and contains a unified Start method, each controller implement the start method. then we can maintain a List[Controller] , use a for loop to startup all the Controllers, this code seems quite duplicate, and the parameters are almost the same.

zhengxiexie: yes, we could do it like Cleanup.

zhengxiexie: refer https://github.com/vmware-tanzu/nsx-operator/pull/330
could close this pr.

PR #281, opened 2023-08-15
Get default values of accessMode and ipv4SubnetSize from VPCNetworkCo…

zhengxiexie:
…nfiguration

Current fields `IPv4SubnetSize` and `AccessMode` are marked as default by kube-builder labels(// +kubebuilder:default:), and their default values are rendered by k8s api-server according to CRD definition. We need to change this behavior, and get default values of these two fields from VPCNetworkConfiguration CR. To implement this, we follow these steps:
- Remove kube-builder default labels of SubnetSet CRD definition.
- When `IPv4SubnetSize` and `AccessMode` are not set in spec of SubnetSet CRD, SubnetSet controller should: - Get current namespace, and get the VPCNetworkConfiguration that the namespace is using. - Get `IPv4SubnetSize` and `AccessMode` from VPCNetworkConfiguration and set them to SubnetSet CRD.

### DIFF
--- a/pkg/controllers/subnet/subnet_controller.go
+++ b/pkg/controllers/subnet/subnet_controller.go
@@ -3,6 +3,7 @@ package subnet
 import (
 	"context"
 	"errors"
+	"fmt"
 	"reflect"
 	"runtime"
 
@@ -17,6 +18,7 @@ import (
 
 	"github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
 	"github.com/vmware-tanzu/nsx-operator/pkg/controllers/common"
+	commonctl "github.com/vmware-tanzu/nsx-operator/pkg/controllers/common"
 	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
 	"github.com/vmware-tanzu/nsx-operator/pkg/metrics"
 	servicecommon "github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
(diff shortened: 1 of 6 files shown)


pkg/controllers/subnet/subnet_controller.go
```diff
@@ -59,6 +61,19 @@ func (r *SubnetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 			}
 			log.V(1).Info("added finalizer on subnet CR", "subnet", req.NamespacedName)
 		}
+		vpcNetworkConfig := commonctl.ServiceMediator.GetVPCNetworkConfigByNamespace(obj.Namespace)
+		if vpcNetworkConfig == nil {
+			err := fmt.Errorf("failed to find VPCNetworkConfig for namespace %s", obj.Namespace)
+			log.Error(err, "operate failed, would retry exponentially", "subnet", req.NamespacedName)
```
dantingl, line 67: `operate failed`: cannot get configuration for Subnet CR

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -59,6 +61,19 @@ func (r *SubnetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+			updateFail(r, &ctx, obj)
+			return ResultRequeue, err
+		}
+		if obj.Spec.AccessMode == "" {
+			obj.Spec.AccessMode = v1alpha1.AccessMode(vpcNetworkConfig.DefaultSubnetAccessMode)
+		}
+		if obj.Spec.IPv4SubnetSize == 0 {
+			obj.Spec.IPv4SubnetSize = vpcNetworkConfig.DefaultIPv4SubnetSize
+		}
```
dantingl, line 76: if obj.Spec.AccessMode == "" or obj.Spec.IPv4SubnetSize == 0 {
    vpcNetworkConfig := commonctl.ServiceMediator.GetVPCNetworkConfigByNamespace(obj.Namespace)
}

dantingl (commented):
Add UT?

pkg/controllers/subnet/subnet_controller.go
dantingl: We can also do this
log.Error(nil, "failed to find VPCNetworkConfig for Subnet CR", "subnet", req.NamespacedName, "namespace", obj.Namespace)

dantingl: Just saw it would return err in the end, so do we still need to log error here?

zhengxiexie: OK

pkg/controllers/subnet/subnet_controller.go
zhengxiexie: ok

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -55,6 +57,19 @@ func (r *SubnetSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerUpdateTotal, MetricResTypeSubnetSet)
 		if !controllerutil.ContainsFinalizer(obj, servicecommon.SubnetSetFinalizerName) {
 			controllerutil.AddFinalizer(obj, servicecommon.SubnetSetFinalizerName)
+			vpcNetworkConfig := commonctl.ServiceMediator.GetVPCNetworkConfigByNamespace(obj.Namespace)
```
dantingl, line 60: Same here, check obj.Spec.AccessMode and obj.Spec.IPv4SubnetSize first

zhengxiexie: ok

dantingl (approved):

PR #288, opened 2023-08-17
Support codegen for v1alpha2

zhengxiexie:
The repo misses the v1alpha2 for codegen.
IPPool v1alpha1 was implemented in NCP, nsx-operator is responsible for IPPool v1alpha2, the reason to keep the two versions in the definition of CRD is to keep the iteration history in the repository and comply with the update process k8s recommends.


pkg/apis/nsx.vmware.com/v1alpha1/vpc_types.go
```diff
@@ -39,23 +39,12 @@ type VPCStatus struct {
...
-}
-
-type CIDRsUsageInfo struct {
-	// Allocated IPs in the IP Block (including IPs used by other VPCs).
-	Allocated int `json:"allocated"`
-	// Used IP for the VPC in the IP Block (IPs used only in current VPC).
-	Used int `json:"used"`
-	// Total IP number in the IP Block.
-	Total int `json:"total"`
+	// Default SNAT IP for private Subnets.
```
dantingl, line 42: I think Xiaopei already made this change, do you need to rebase the code?

pkg/apis/nsx.vmware.com/v1alpha1/staticroute_types.go
```diff
@@ -31,7 +31,8 @@ type NextHop struct {
 
 // StaticRouteStatus defines the observed state of StaticRoute.
 type StaticRouteStatus struct {
-	Conditions []StaticRouteCondition `json:"conditions"`
+	Conditions      []StaticRouteCondition `json:"conditions"`
+	NSXResourcePath string                 `json:"nsxResourcePath"`
```
dantingl, line 35: I remember we removed nsxResourcePath

zhengxiexie: ok

zhengxiexie: ditto

pkg/apis/nsx.vmware.com/v1alpha2/ippool_types.go
```diff
@@ -0,0 +1,81 @@
...
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []IPPool `json:"items"`
+}
+
+// IPPoolSpec defines the desired state of IPPool.
+type IPPoolSpec struct {
+	// Type defines the type of this IPPool, public or private.
+	// +kubebuilder:validation:Enum=public;private
+	// +kubebuilder:default=private
+	Type string `json:"type"`
```
wyike, line 41: should we add json:"string,omitempty"` ?

zhengxiexie: There is no need, if no Type is defined, then use the default value and ippool_controller would use the default value to check whether we should get the type from vpc network config.

wyike: I assume you are saying there is no impact for the ippool controller logic in nsx operator, right.

But for other client that wants to create an ippool CR, If no omitempty supported on Type, the client always has to provide Type field with a valid value (Public or Private) when building ippool spec that will be sent to apiserver later.

cc @dantingl  for comments.

build/yaml/crd/nsx.vmware.com_subnetports.yaml
```diff
@@ -86,6 +85,7 @@ spec:
                     description: 'UID of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids'
                     type: string
                 type: object
+                x-kubernetes-map-type: atomic
```
timdengyun, line 88: why adding this field?

pkg/apis/nsx.vmware.com/v1alpha1/staticroute_types.go
timdengyun: why we still have nsxResourcePath here?

timdengyun:
> The repo misses the v1alpha2 for codegen.

Please add more description on this PR to help reader understand why we add two versions of APIS? what's the difference?

build/yaml/crd/nsx.vmware.com_subnetports.yaml
zhengxiexie: It's generated automatically.

zhengxiexie:
> > The repo misses the v1alpha2 for codegen.
> 
> Please add more description on this PR to help reader understand why we add two versions of APIS? what's the difference?

ok

pkg/apis/nsx.vmware.com/v1alpha2/ippool_types.go
```diff
@@ -0,0 +1,81 @@
...
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []IPPool `json:"items"`
+}
+
+// IPPoolSpec defines the desired state of IPPool.
+type IPPoolSpec struct {
+	// Type defines the type of this IPPool, Public or Private.
+	// +optional
+	Type string `json:"type"`
```
wyike, line 41: Hi Xie, would you also add omitempty here in nsx.vmware.com/v1alpha2? I notice you add to the other v1alpha2. Thanks!

zhengxiexie: OK

pkg/apis/nsx.vmware.com/v1alpha1/staticroute_types.go
zhengxiexie: waiting for another PR, could ignore this.

timdengyun:
It looks good to me

wyike (approved):

pkg/apis/nsx.vmware.com/v1alpha1/nsxserviceaccount_types.go
```diff
@@ -46,23 +46,35 @@ const (
 	NSXServiceAccountPhaseRealized   NSXServiceAccountPhase = "realized"
 	NSXServiceAccountPhaseInProgress NSXServiceAccountPhase = "inProgress"
 	NSXServiceAccountPhaseFailed     NSXServiceAccountPhase = "failed"
+
```
dantingl, line 49: Why we also update service account CRD?

pkg/apis/nsx.vmware.com/v1alpha1/subnet_types.go
```diff
@@ -12,15 +12,11 @@ type AccessMode string
 // SubnetSpec defines the desired state of Subnet.
 type SubnetSpec struct {
 	// Size of Subnet based upon estimated workload count.
-	// Defaults to 64.
```
dantingl, line 15: Why these are removed?

dantingl (approved):

PR #292, opened 2023-08-18
ADD GC for subnet

TaoZou1:
ADD GC for subnet

### DIFF
--- a/pkg/controllers/subnet/subnet_controller.go
+++ b/pkg/controllers/subnet/subnet_controller.go
@@ -6,11 +6,13 @@ import (
 	"fmt"
 	"reflect"
 	"runtime"
+	"time"
 
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 	v1 "k8s.io/api/core/v1"
 	apimachineryruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -278,5 +280,49 @@ func (r *SubnetReconciler) Start(mgr ctrl.Manager) error {
 	if err != nil {
 		return err
 	}
+	go r.GarbageCollector(make(chan bool), servicecommon.GCInterval)
 	return nil
+}
+
+func (r *SubnetReconciler) GarbageCollector(cancel chan bool, timeout time.Duration) {
+	ctx := context.Background()
+	log.Info("subnet garbage collector started")
+	for {
+		select {
+		case <-cancel:
+			return
+		case <-time.After(timeout):
+		}
+		nsxSubnetList := r.Service.ListSubnetCreatedByCR()
+		if len(nsxSubnetList) == 0 {
+			continue
+		}
+
+		crdSubnetList := &v1alpha1.SubnetList{}
+		err := r.Client.List(ctx, crdSubnetList)
+		if err != nil {
+			log.Error(err, "failed to list subnet CR")
+			continue
+		}
+
+		crdSubnetIDs := sets.NewString()
+		for _, sr := range crdSubnetList.Items {
+			crdSubnetIDs.Insert(string(sr.UID))
+		}
+
+		for _, elem := range nsxSubnetList {
+			if crdSubnetIDs.Has(*elem.Id) {
+				continue
+			}
+
+			log.Info("GC collected Subnet CR", "UID", elem)
+			metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteTotal, common.MetricResTypeSubnet)
+			err = r.Service.DeleteSubnet(elem)
+			if err != nil {
+				metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteFailTotal, common.MetricResTypeSubnet)
+			} else {
+				metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteSuccessTotal, common.MetricResTypeSubnet)
+			}
+		}
+	}
 }
--- a/pkg/nsx/services/subnet/store.go
+++ b/pkg/nsx/services/subnet/store.go
@@ -3,8 +3,9 @@ package subnet
 import (
 	"errors"
 
-	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
 )
 
 // keyFunc is used to get the key of a resource, usually, which is the ID of the resource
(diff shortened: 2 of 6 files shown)


dantingl:
We also need to take care of SubnetSet GC

pkg/nsx/services/subnet/subnet.go
```diff
@@ -180,6 +180,30 @@ func (service *SubnetService) DeleteSubnet(nsxSubnet model.VpcSubnet) error {
 	return nil
 }
 
+func (service *SubnetService) ListSubnet() []model.VpcSubnet {
+	subnets := service.SubnetStore.List()
+	subnetSet := []model.VpcSubnet{}
```
lxiaopei, line 185: can we use a different variable name instead of "subnetSet"? it is confusing.

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -242,5 +244,58 @@ func (r *SubnetReconciler) Start(mgr ctrl.Manager) error {
...
+func (r *SubnetReconciler) GarbageCollector(cancel chan bool, timeout time.Duration) {
+	ctx := context.Background()
+	log.Info("subnet garbage collector started")
+	for {
+		select {
+		case <-cancel:
+			return
+		case <-time.After(timeout):
+		}
+		nsxSubnetList := r.Service.ListSubnet()
```
lxiaopei, line 260: Do we need to list all nsx subnets or just their IDs?

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -242,5 +244,58 @@ func (r *SubnetReconciler) Start(mgr ctrl.Manager) error {
...
+		}
+
+		crdSubnetList := &v1alpha1.SubnetList{}
+		err := r.Client.List(ctx, crdSubnetList)
+		if err != nil {
+			log.Error(err, "failed to list subnet CR")
+			continue
+		}
+
+		crdSubnet := sets.NewString()
```
lxiaopei, line 272: can we change to crdSubnetIDs?

pkg/nsx/services/subnet/subnet.go
```diff
@@ -180,6 +180,30 @@ func (service *SubnetService) DeleteSubnet(nsxSubnet model.VpcSubnet) error {
...
+	}
+	return subnetSet
+}
+
+// check if subnet belongs to a subnetset, if yes, check if that subnetset still exists
+func (service *SubnetService) IsOrphanSubnet(subnet model.VpcSubnet, subnetsets *v1alpha1.SubnetSetList) bool {
+	for _, tag := range subnet.Tags {
+		if *tag.Scope == common.TagScopeSubnetSetCRId {
+			subnetset_name := *tag.Tag
+			for _, subnetset := range subnetsets.Items {
```
lxiaopei, line 197: I am worried about the performance for the code.
Could we find another effecient way to check if the nsx subnet's reference SubnetSet CR exists or not?

pkg/nsx/services/subnet/subnet.go
```diff
@@ -180,6 +180,30 @@ func (service *SubnetService) DeleteSubnet(nsxSubnet model.VpcSubnet) error {
...
+		subnetSet = append(subnetSet, subnet.(model.VpcSubnet))
+	}
+	return subnetSet
+}
+
+// check if subnet belongs to a subnetset, if yes, check if that subnetset still exists
+func (service *SubnetService) IsOrphanSubnet(subnet model.VpcSubnet, subnetsets *v1alpha1.SubnetSetList) bool {
+	for _, tag := range subnet.Tags {
+		if *tag.Scope == common.TagScopeSubnetSetCRId {
+			subnetset_name := *tag.Tag
```
lxiaopei, line 196: subnetset_id

pkg/controllers/subnet/subnet_controller.go
TaoZou1: done

pkg/nsx/services/subnet/subnet.go
TaoZou1: if we had a subnetset Store, the performance will be better.

pkg/nsx/services/subnet/subnet.go
TaoZou1: done

pkg/nsx/services/subnet/subnet.go
TaoZou1: done

pkg/controllers/subnet/subnet_controller.go
TaoZou1: subnet, GC will check its tag later

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -242,5 +244,58 @@ func (r *SubnetReconciler) Start(mgr ctrl.Manager) error {
...
+			log.Error(err, "failed to list subnet CR")
+			continue
+		}
+
+		crdSubnetIDs := sets.NewString()
+		for _, sr := range crdSubnetList.Items {
+			crdSubnetIDs.Insert(string(sr.UID))
+		}
+
+		crdSubnetSetIDs := r.Service.ListSubnetSetID(ctx)
```
dantingl, line 277: This should be put in subnetset controller?

TaoZou1: both subnet/subnetset controller have the logic to GC the subnet.
They will have lots of duplicate code.
subnet controller also needs to check if subnet belongs to subnetset.
subnetset controller needs to list all subnet.
Only one place to handle subnet GC will make it simple.

pkg/nsx/services/subnet/subnet.go
lxiaopei: still subnetSet?

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -242,5 +244,58 @@ func (r *SubnetReconciler) Start(mgr ctrl.Manager) error {
...
+		if len(crdSubnetSetIDs) == 0 {
+			continue
+		}
+
+		for _, elem := range nsxSubnetList {
```
lxiaopei, line 282: we do not need to iterate all the nsxSubnetList again, since the above already checked all the NSX subnets created for Subnet CRs. 
Could we remove those NSX subnets from nsxSubnetList if found they have corresponding Subnet CR?
Then in the left NSX Subnet list, check they have corresponding SubnetSet CR.  
Instead of checking NSX Subnet tags, could we use the store SubnetIndexFunc?

TaoZou1: split the function into subnet_controller/subnetset_controller gc

pkg/nsx/services/subnet/store.go
```diff
@@ -37,6 +38,16 @@ func subnetIndexFunc(obj interface{}) ([]string, error) {
 	}
 }
 
+// subnetsetIndexFunc is used to filter out NSX Subnets which are tagged with subnetset UID.
+func subnetTypeIndexFunc(obj interface{}) ([]string, error) {
+	switch o := obj.(type) {
+	case model.VpcSubnet:
+		return filterTag(o.Tags, common.TagScopeSubnetSetCRId), nil
```
zhengxiexie, line 45: I think it should be common.TagScopeSubnetCRType

TaoZou1: you're right.

pkg/nsx/services/subnet/store.go
```diff
@@ -37,6 +38,16 @@ func subnetIndexFunc(obj interface{}) ([]string, error) {
 	}
 }
 
+// subnetTypeIndexFunc is used to filter out NSX Subnets which are tagged with subnetcr type.
+func subnetTypeIndexFunc(obj interface{}) ([]string, error) {
+	switch o := obj.(type) {
+	case model.VpcSubnet:
+		return filterTag(o.Tags, common.TagScopeSubnetCRType), nil
```
dantingl, line 45: I don't think we should add cr type in Tag, how about checking "nsx-op/subnetset_cr_uid" and "nsx-op/subnet_cr_uid"

dantingl (dismissed):

dantingl (approved):

PR #294, opened 2023-08-18
Support passing user/pass/ca from parameter

TaoZou1:
Default nsx-operator reloads user/name/ca from file. Support to pass user/pass/ca from parameter


zhengxiexie (approved):
LGTM

PR #300, opened 2023-08-24
Refactor vpc service to remove mediator dependencies

seanpang-vmware:


seanpang-vmware:
this way seems not very convenient either, will refactor again.

seanpang-vmware:
1. we must use service interface to limit the exposed method instead of using service instance, this part is the same.
2. if we still keep mediator, the code using sequence would be like follow: (using vpc service for example)
   s := initService()  s is of type vpcserviceinterface
   ....
   mediator.registerService(s)
   ... 
   in vpc controller code
   mediator.PublicMethodsInInterface()

   pros: use central mediator as api lib. controller can use it without knowing what is inside the mediator
   cons:  expose all method to controller even if it is not needed.

3. if we do not use mediator:
   s := initService()
   ...
  typedef NamespaceReconciler stuct {
   ...
   S *VPCServiceInterface
  }
  ...
    
  namespaceReconciler := &namespacecontroller.NamespaceReconciler{
		...
               S:  vpcserviceInterface
  }

pros: previous cons
cons: previous pros, controllers need to know which interfaces to use and include them into its own struct.

cmd/main.go
```diff
@@ -71,9 +77,41 @@ func init() {
 	if metrics.AreMetricsExposed(cf) {
 		metrics.InitializePrometheusMetrics()
 	}
+
+	mgr, err = ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
```
timdengyun, line 81: we had better not move these initialization code in init function, we should keep it in the main function.

cmd/main.go
```diff
@@ -134,29 +172,26 @@ func StartIPPoolController(mgr ctrl.Manager, commonService common.Service) {
...
+	vpcReconciler.Service = vpcService
+
 	if err := vpcReconciler.Start(mgr); err != nil {
 		log.Error(err, "failed to create vpc controller", "controller", "VPC")
 		os.Exit(1)
 	}
 }
 
-func StartNamespaceController(mgr ctrl.Manager, commonService common.Service) {
+func StartNamespaceController() {
```
timdengyun, line 189: https://github.com/vmware-tanzu/nsx-operator/pull/280 in this PR, we are tarting to move start controller into controller, so do we consider this method, or we still use PR 280 to do it?

cmd/main.go
```diff
@@ -209,17 +216,17 @@ func main() {
 		staticroutecontroller.StartStaticRouteController(mgr, commonService)
 		subnetport.StartSubnetPortController(mgr, commonService)
 
-		StartNamespaceController(mgr, commonService)
-		StartVPCController(mgr, commonService)
-		StartIPPoolController(mgr, commonService)
+		StartNamespaceController()
```
timdengyun, line 219: I still suggest that we pass commonService as a argument to startcontroller instead of using one global variable.

pkg/controllers/common/services.go
```diff
@@ -0,0 +1,11 @@
+package common
```
timdengyun, line 1: I suggest that we move this service interface definiation to  nsx/service/common module, so controller module don't need have any service related code.

pkg/controllers/namespace/namespace_controller.go
```diff
@@ -34,9 +34,10 @@ var (
 
 // NamespaceReconciler process namespace create/delete event
 type NamespaceReconciler struct {
-	Client    client.Client
-	Scheme    *apimachineryruntime.Scheme
-	NSXConfig *config.NSXOperatorConfig
+	Client     client.Client
+	Scheme     *apimachineryruntime.Scheme
+	NSXConfig  *config.NSXOperatorConfig
+	VPCService common.IVPCService
```
timdengyun, line 40: It's good to add comments that why we only use IVPCService here instead of using VPCService

timdengyun: It's good rename VPCService as a more meaningful name to help reader understand this purpose of this interface service.

pkg/nsx/services/mediator/mediator.go
```diff
@@ -49,7 +49,7 @@ func (serviceMediator *ServiceMediator) ListVPCInfo(ns string) []common.VPCResou
 
 // This method is used for subnet service since vpc network config contains default subnet size
 // and default subnet access mode.
-func (m *ServiceMediator) GetVPCNetworkConfigByNamespace(ns string) *vpc.VPCNetworkConfigInfo {
+func (m *ServiceMediator) GetVPCNetworkConfigByNamespace(ns string) *common.VPCNetworkConfigInfo {
```
timdengyun, line 52: As we discussed, we can consider remove this mediator interface.

pkg/nsx/services/common/types.go
```diff
@@ -122,3 +122,14 @@ type VPCResourceInfo struct {
 	ID       string
 	ParentID string
 }
+
+type VPCNetworkConfigInfo struct {
```
timdengyun, line 126: Dose this VPCNetworkConfigInfo struct will be used other modules except VPC, if not, we shoud not put it in the common service file.

pkg/nsx/services/vpc/vpc.go
```diff
@@ -51,15 +51,15 @@ type VPCService struct {
 	IpblockStore *IPBlockStore
 }
 
-func (s *VPCService) RegisterVPCNetworkConfig(ncCRName string, info VPCNetworkConfigInfo) {
+func (s *VPCService) RegisterVPCNetworkConfig(ncCRName string, info common.VPCNetworkConfigInfo) {
```
timdengyun, line 54: It's good to put some comments to say this implementation of which interface function.

pkg/controllers/common/services.go
```diff
@@ -0,0 +1,11 @@
...
+
+import "github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
+
+type IVPCService interface {
+	RegisterNamespaceNetworkconfigBinding(ns string, ncCRName string)
```
timdengyun, line 6: Please Add some description to tell users what's the purpose of this interface function.

PR #301, opened 2023-08-24
Avoid duplicate port creation/update for the subnet port CR

heypnus:
When the operator handle the subnetport CR event, it will try to get NSX subnet every time even though the NSX port has been created. That may cause the duplicate NSX port update with different parent NSX subnet if the subnet port CR references the subnetset with multiple NSX subnets.

This path will avoid the duplicate request if the NSX port already existed. It will also enhance the comparation function to avoid unnecessary update.

### DIFF
--- a/pkg/controllers/subnetport/subnetport_controller.go
+++ b/pkg/controllers/subnetport/subnetport_controller.go
@@ -309,7 +309,11 @@ func deleteSuccess(r *SubnetPortReconciler, _ *context.Context, _ *v1alpha1.Subn
 }
 
 func (r *SubnetPortReconciler) GetSubnetPathForSubnetPort(obj *v1alpha1.SubnetPort) (string, error) {
-	subnetPath := ""
+	subnetPath := r.Service.GetSubnetPathFromStoreByKey(string(obj.UID))
+	if len(subnetPath) > 0 {
+		log.V(1).Info("NSX subnet port had been created, returning the existing NSX subnet path", "subnetPort.UID", obj.UID, "subnetPath", subnetPath)
+		return subnetPath, nil
+	}
 	if len(obj.Spec.Subnet) > 0 {
 		subnet := &v1alpha1.Subnet{}
 		namespacedName := types.NamespacedName{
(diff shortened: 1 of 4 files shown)


pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -309,7 +309,11 @@ func deleteSuccess(r *SubnetPortReconciler, _ *context.Context, _ *v1alpha1.Subn
 }
 
 func (r *SubnetPortReconciler) GetSubnetPathForSubnetPort(obj *v1alpha1.SubnetPort) (string, error) {
-	subnetPath := ""
+	subnetPath := r.Service.GetSubnetPathFromStoreByKey(string(obj.UID))
+	if len(subnetPath) > 0 {
+		log.Info("NSX subnet port had been created, returning the existing NSX subnet path", "subnetPort.UID", obj.UID, "subnetPath", subnetPath)
```
dantingl, line 314: V1 log level?

pkg/nsx/services/subnetport/builder.go
```diff
@@ -49,7 +49,7 @@ func (service *SubnetPortService) buildSubnetPort(obj *v1alpha1.SubnetPort, nsxS
 			// AppId:             common.String("nsx.ns-3.pod-5%"),
 			// ContextId:         common.String("95ccaad4-0dfb-469e-a1e6-27d815826382"),
 			Id:         common.String(nsxCIFID.String()),
-			TrafficTag: nil,
+			TrafficTag: common.Int64(0),
```
dantingl, line 52: Why cannot it be nil?

pkg/nsx/services/subnetport/subnetport.go
```diff
@@ -111,7 +112,11 @@ func (service *SubnetPortService) CreateOrUpdateSubnetPort(obj *v1alpha1.SubnetP
 		log.Error(err, "check and update NSX subnet port state failed, would retry exponentially", "subnetport", obj.UID)
 		return err
 	}
-	log.Info("successfully created or updated subnetport", "subnetport", obj.UID)
+	if isChanged {
+		log.Info("successfully created or updated subnetport", "subnetport", obj.UID)
+	} else {
+		log.Info("successfully checked subnetport", "subnetport", obj.UID)
```
dantingl, line 118: How about subnetport already exists?

pkg/nsx/services/subnetport/builder.go
heypnus: If we write nil, the created NSX resource's will be 0 automatically. Then when we compare the existing resource (get via NSX API) and the desired resource, the result is not equal (0 != nil). So I set it to 0 to avoid this kind of inequality and avoid a duplicate patch request.

dantingl (approved):

PR #304, opened 2023-08-28
Add some utils func

TaoZou1:
Add GenerateID, GenerateName, BuildBasicTags

### DIFF
--- a/pkg/nsx/services/staticroute/builder.go
+++ b/pkg/nsx/services/staticroute/builder.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
+	"github.com/vmware-tanzu/nsx-operator/pkg/util"
 )
 
 func validateStaticRoute(obj *v1alpha1.StaticRoute) error {
@@ -41,31 +42,12 @@ func (service *StaticRouteService) buildStaticRoute(obj *v1alpha1.StaticRoute) (
 		nexthop.IpAddress = &obj.Spec.NextHops[index].IPAddress
 		sr.NextHops = append(sr.NextHops, nexthop)
 	}
-	sr.Id = String(fmt.Sprintf("sr_%s", fmt.Sprint(obj.UID)))
-	sr.DisplayName = String(fmt.Sprintf("sr-%s-%s", obj.Namespace, obj.Name))
+	sr.Id = String(util.GenerateID(string(obj.UID), "sr", "", ""))
+	sr.DisplayName = String(util.GenerateTruncName(common.MaxNameLength, obj.Name, "sr", "", "", ""))
 	sr.Tags = service.buildBasicTags(obj)
 	return sr, nil
 }
 
 func (service *StaticRouteService) buildBasicTags(obj *v1alpha1.StaticRoute) []model.Tag {
-	tags := []model.Tag{
-		{
-			Scope: String(common.TagScopeCluster),
-			Tag:   String(service.NSXConfig.Cluster),
-		},
-		{
-			Scope: String(common.TagScopeNamespace),
-			Tag:   String(obj.ObjectMeta.Namespace),
-		},
-		// TODO: get namespace uid
-		{
-			Scope: String(common.TagScopeStaticRouteCRName),
-			Tag:   String(obj.ObjectMeta.Name),
-		},
-		{
-			Scope: String(common.TagScopeStaticRouteCRUID),
-			Tag:   String(string(obj.UID)),
-		},
-	}
-	return tags
+	return util.BuildBasicTags(service.Service.NSXConfig.Cluster, obj, "")
 }
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -14,16 +14,38 @@ import (
 
 	"github.com/apparentlymart/go-cidr/cidr"
 	mapset "github.com/deckarep/golang-set"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
+	"github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
 )
 
 const wcpSystemResource = "vmware-system-shared-t1"
 const HashLength int = 8
+
+var (
+	String    = common.String
+	basicTags = []string{common.TagScopeNamespace, common.TagScopeNamespaceUID,
+		common.TagScopeStaticRouteCRName, common.TagScopeStaticRouteCRUID,
+		common.TagScopeSecurityPolicyCRName, common.TagScopeSecurityPolicyCRUID,
+		common.TagScopeSubnetCRName, common.TagScopeSubnetCRUID,
+		common.TagScopeSubnetPortCRName, common.TagScopeSubnetPortCRUID,
+		common.TagScopeVPCCRName, common.TagScopeVPCCRUID,
+		common.TagScopeIPPoolCRName, common.TagScopeIPPoolCRUID,
+		common.TagScopeSubnetSetCRName, common.TagScopeSubnetSetCRUID}
+	tagsScopeSet = sets.New[string]()
+)
+
+func init() {
+	for _, tag := range basicTags {
+		tagsScopeSet.Insert(tag)
+	}
+}
 
 var log = logf.Log.WithName("pkg").WithName("utils")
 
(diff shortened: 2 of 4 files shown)


pkg/util/utils.go
```diff
@@ -244,3 +250,104 @@ func UpdateK8sResourceAnnotation(client client.Client, ctx *context.Context, k8s
...
+		log.Info("unknow obj type", "obj", obj)
+	}
+
+	if len(namespaceID) > 0 {
+		tags = append(tags, model.Tag{Scope: String(common.TagScopeNamespaceUID), Tag: String(string(namespaceID))})
+	}
+	return tags
+}
+
+func AppendTags(basicTags, tags []model.Tag) []model.Tag {
```
zhengxiexie, line 348: no usage.  Besides, I think it's better to replace append with append_if_not_exist which is implemented by a new function, in case the upper level code already adds the tag by mistake.

pkg/util/utils.go
```diff
@@ -244,3 +250,104 @@ func UpdateK8sResourceAnnotation(client client.Client, ctx *context.Context, k8s
...
+	}
+	name.WriteString(res_name)
+	if len(suffix) > 0 {
+		name.WriteString("-")
+		name.WriteString(suffix)
+	}
+	return name.String()
+}
+
+func GenerateTruncName(limit int, res_name, prefix, suffix, project, cluster string) string {
```
zhengxiexie, line 297: where to use it?

pkg/util/utils.go
```diff
@@ -244,3 +250,104 @@ func UpdateK8sResourceAnnotation(client client.Client, ctx *context.Context, k8s
...
+
+	}
+	if len(suffix) > 0 {
+		id.WriteString("_")
+		id.WriteString(suffix)
+	}
+	return id.String()
+}
+
+func GenerateDisplayName(res_name, prefix, suffix, project, cluster string) string {
```
zhengxiexie, line 275: only used in GenerateTruncName, which has no usage.

pkg/util/utils.go
```diff
@@ -244,3 +250,104 @@ func UpdateK8sResourceAnnotation(client client.Client, ctx *context.Context, k8s
 	}
 	return nil
 }
+
+// some resources has complex index, so set it type to string
```
zhengxiexie, line 254: only used in test code.

pkg/util/utils.go
```diff
@@ -244,3 +250,104 @@ func UpdateK8sResourceAnnotation(client client.Client, ctx *context.Context, k8s
...
+	if len(index) > 0 {
+		id.WriteString("_")
+		id.WriteString(index)
+
+	}
+	if len(suffix) > 0 {
+		id.WriteString("_")
+		id.WriteString(suffix)
+	}
+	return id.String()
```
timdengyun, line 272: we have NormalizeId func, 
https://github.com/vmware-tanzu/nsx-operator/blob/vpc_dev/pkg/util/utils.go#L57
do we need to generate ID suing this func?

pkg/util/utils.go
```diff
@@ -244,3 +250,104 @@ func UpdateK8sResourceAnnotation(client client.Client, ctx *context.Context, k8s
...
+	if len(suffix) > 0 {
+		name.WriteString("-")
+		name.WriteString(suffix)
+	}
+	return name.String()
+}
+
+func GenerateTruncName(limit int, res_name, prefix, suffix, project, cluster string) string {
+	name := GenerateDisplayName(res_name, prefix, suffix, project, cluster)
+	return NormalizeName(name)
```
timdengyun, line 299: NSX resources names have the below limit for display_name length:
1. segment display_name 80 chars
2. group name + domain name 245 chars
3. other NSX resources display_name 255 chars
We use truncated name to limit the display_name
when creating nsx resources.

so, NormalizeName can handle this limit?

pkg/util/utils.go
TaoZou1: NormalizeId only used for truncate ID.

pkg/util/utils.go
TaoZou1: NormalizeName could only handle the third, the 1,2 could not be handled by that.

pkg/util/utils.go
TaoZou1: add a sample

pkg/util/utils.go
TaoZou1: yes, if caller who insure name length is not beyond the limit could call this one

pkg/util/utils.go
TaoZou1: add a sample

pkg/util/utils.go
TaoZou1: add checking

pkg/util/utils.go
TaoZou1: in ncp, they used the same function like NormalizeName, just limit is different.

pkg/util/utils.go
TaoZou1: it just used for those cr which needs more tags than basic tags, I added checking. but in NCP, there is also no such checking.

pkg/util/utils.go
```diff
@@ -244,3 +270,106 @@ func UpdateK8sResourceAnnotation(client client.Client, ctx *context.Context, k8s
 	}
 	return nil
 }
+
+// some resources has complex index, so set it type to string
+func GenerateID(res_id, prefix, suffix string, index string) string {
```
seanpang-vmware, line 275: func GenerateID(res_id string, prefix string, suffix string, index string) string {

pkg/util/utils.go
```diff
@@ -244,3 +270,106 @@ func UpdateK8sResourceAnnotation(client client.Client, ctx *context.Context, k8s
...
+
+	}
+	if len(suffix) > 0 {
+		id.WriteString("_")
+		id.WriteString(suffix)
+	}
+	return id.String()
+}
+
+func GenerateDisplayName(res_name, prefix, suffix, project, cluster string) string {
```
seanpang-vmware, line 295: same as above, we should use explicit typedef for param res_name, prefix, suffix, project as string

pkg/util/utils.go
seanpang-vmware: same with param type

pkg/util/utils.go
```diff
@@ -244,3 +270,106 @@ func UpdateK8sResourceAnnotation(client client.Client, ctx *context.Context, k8s
...
+	}
+	return name.String()
+}
+
+func GenerateTruncName(limit int, res_name, prefix, suffix, project, cluster string) string {
+	name := GenerateDisplayName(res_name, prefix, suffix, project, cluster)
+	return normalizeNamebyLimit(name, limit)
+}
+
+func BuildBasicTags(service common.Service, obj interface{}, namespaceID types.UID) []model.Tag {
```
seanpang-vmware, line 322: if we only need cluster, i suggest not to use the whole comon.service as param, only a "cluster string" should be enough

pkg/util/utils.go
```diff
@@ -244,3 +270,106 @@ func UpdateK8sResourceAnnotation(client client.Client, ctx *context.Context, k8s
...
+		log.Info("unknown obj type", "obj", obj)
+	}
+
+	if len(namespaceID) > 0 {
+		tags = append(tags, model.Tag{Scope: String(common.TagScopeNamespaceUID), Tag: String(string(namespaceID))})
+	}
+	return tags
+}
+
+func AppendTags(basicTags, tags []model.Tag) []model.Tag {
```
seanpang-vmware, line 368: if we only plan to use this method in BuildBasicTags, i suggest to change it lower case appendTags.
if we plan to use it as a common method, i suggest to add validation for basicTags, if it is nil, we should init a blank list for it.

pkg/util/utils.go
zhengxiexie: image
it's better to comply with go doc format.

pkg/util/utils.go
zhengxiexie: +1, there is getCluster() func elsewhere.

pkg/nsx/services/staticroute/builder.go
```diff
@@ -41,31 +42,12 @@ func (service *StaticRouteService) buildStaticRoute(obj *v1alpha1.StaticRoute) (
 		nexthop.IpAddress = &obj.Spec.NextHops[index].IPAddress
 		sr.NextHops = append(sr.NextHops, nexthop)
 	}
-	sr.Id = String(fmt.Sprintf("sr_%s", fmt.Sprint(obj.UID)))
-	sr.DisplayName = String(fmt.Sprintf("sr-%s-%s", obj.Namespace, obj.Name))
+	sr.Id = String(util.GenerateID(string(obj.UID), "", "", ""))
```
zhengxiexie, line 45: so all the other CR's id and displayname comply with this  and leave the last parameters blank?

pkg/util/utils.go
TaoZou1: done

pkg/util/utils.go
TaoZou1: it should be the same.

pkg/nsx/services/staticroute/builder.go
TaoZou1: yes

pkg/util/utils.go
TaoZou1: done

pkg/util/utils.go
TaoZou1: It's better to check basicTags. I prefer to return nil instead of init a blank list for it.
the basicTags should come from BuildBasicTags()

pkg/util/utils.go
```diff
@@ -244,3 +270,110 @@ func UpdateK8sResourceAnnotation(client client.Client, ctx *context.Context, k8s
...
+	if len(suffix) > 0 {
+		name.WriteString("-")
+		name.WriteString(suffix)
+	}
+	return name.String()
+}
+
+func GenerateTruncName(limit int, res_name, prefix, suffix, project, cluster string) string {
+	name := GenerateDisplayName(res_name, prefix, suffix, project, cluster)
+	return normalizeNamebyLimit(name, limit)
```
heypnus, line 319: For now, when you generate the truncated name, the steps are generate full name with prefix/suffix/project/cluster -> limit the length, right? But that may lose some info when it is truncated, e.g. the suffix. I think the better sequence should be limit the length of every part firstly, then generate the full name, like what we did in NCP. How do you think?

dantingl (approved):

pkg/util/utils.go
TaoZou1: done

dantingl (approved):

PR #307, opened 2023-08-28
Fix staticroute bugs

TaoZou1:
StaticRoute uses one copy of ServiceMediator which is not initialized. Use ServiceMediator directly
StaticRoute GC uses the Path, so using Get API to get the sr and store it


TaoZou1:
wrong base commit

TaoZou1:
wrong base commit

PR #310, opened 2023-08-29
Add cleanup for ippool and subnet

zhengxiexie:
Add cleanup for ippool and subnet as securitypolicy does.

### DIFF
--- a/pkg/clean/types.go
+++ b/pkg/clean/types.go
@@ -3,3 +3,29 @@ package clean
 type cleanup interface {
 	Cleanup() error
 }
+
+type cleanupFunc func() (cleanup, error)
+
+type CleanupService struct {
+	cleans []cleanup
+	err    error
+}
+
+func NewCleanupService() *CleanupService {
+	return &CleanupService{}
+}
+
+func (c *CleanupService) AddCleanupService(f cleanupFunc) *CleanupService {
+	var clean cleanup
+	if c.err != nil {
+		return c
+	}
+
+	clean, c.err = f()
+	if c.err != nil {
+		return c
+	}
+
+	c.cleans = append(c.cleans, clean)
+	return c
+}
--- a/pkg/clean/clean.go
+++ b/pkg/clean/clean.go
@@ -10,7 +10,9 @@ import (
 	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/ippool"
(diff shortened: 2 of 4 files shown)


cmd_clean/main.go
```diff
@@ -53,7 +53,9 @@ func main() {
 	cf.SsoDomain = vcSsoDomain
 	cf.HttpsPort = vcHttpsPort
 	cf.Thumbprint = []string{thumbprint}
```
dantingl, line 55: thumbprint could also be nil, thumbprint and nsx ca cert are two options

pkg/clean/types.go
```diff
@@ -3,3 +3,27 @@ package clean
...
+	cleans []cleanup
+	err    error
+}
+
+func NewCleanupService() *CleanupService {
+	return &CleanupService{}
+}
+
+func (c *CleanupService) AddCleanupService(f func() (cleanup, error)) *CleanupService {
+	var clean cleanup
```
TaoZou1, line 17: the parameter looks complex, how about using type to define that ?

pkg/clean/types.go
```diff
@@ -3,3 +3,27 @@ package clean
...
+	if c.err != nil {
+		return c
+	}
+
```
TaoZou1, line 21: it seems that not necessary to use chain pattern, check the err in line 18 seems strange

pkg/clean/types.go
seanpang-vmware, line 17: seems no need to define a var here, just use := in line 22?

pkg/clean/types.go
```diff
@@ -3,3 +3,27 @@ package clean
...
+	clean, c.err = f()
+	if c.err != nil {
+		return c
+	}
```
TaoZou1, line 25: does that mean if one service fail to init, the cleanup could still continue without report a error? or we should break the cleanup procedure and quit? or retry some time here to quit?

cmd_clean/main.go
zhengxiexie: Oh, there is 
	nsxConfig.Thumbprint = removeEmptyItem(nsxConfig.Thumbprint)
	nsxConfig.CaFile = removeEmptyItem(nsxConfig.CaFile)
in validateCert, can remove the `if` here.

pkg/clean/types.go
zhengxiexie: good

zhengxiexie: image
no , since I use  `c.err = f()` not `err:=f()`

pkg/clean/types.go
zhengxiexie: It's important since it would return the error and skip the following steps in advance.
An elegant pattern.

pkg/clean/types.go
zhengxiexie: ditto

seanpang-vmware (approved):

PR #311, opened 2023-08-30
[NSXServiceAccount] Support CA validation and client cert rotation

gran-vmv:
GC will run immediately after start.
Add EnableCertRotation property in NSXServiceAccount.Spec.
Cache CA content in memory.
Populate CA content to Secret when processing NSXServiceAccount create event.
Check all Secrets' ca.crt are up-to-date when first NSXServiceAccount GC finished. Update if needed.
Check all Secrets' client cert need renew every 720 GC interval.
Renew client cert if cert will expire in 7d when rotation is supported (NSXT >=4.1.3 and EnableCertRotation is true).
Use 365d instead of 3650d for new client cert when rotation is supported (NSXT >=4.1.3 and EnableCertRotation is true).

### DIFF
--- a/build/yaml/crd/nsx.vmware.com_nsxserviceaccounts.yaml
+++ b/build/yaml/crd/nsx.vmware.com_nsxserviceaccounts.yaml
@@ -35,6 +35,10 @@ spec:
           spec:
             description: NSXServiceAccountSpec defines the desired state of NSXServiceAccount
             properties:
+              enableCertRotation:
+                description: EnableCertRotation enables cert rotation feature in this
+                  cluster when NSXT >=4.1.3
+                type: boolean
               vpcName:
                 type: string
             type: object
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -38,6 +38,7 @@ type NSXOperatorConfig struct {
 	*K8sConfig
 	*VCConfig
 	*HAConfig
+	configCache configCache
 }
 
 func init() {
@@ -50,6 +51,29 @@ func (operatorConfig *NSXOperatorConfig) HAEnabled() bool {
 		return true
 	}
 	return false
+}
+
+func (operatorConfig *NSXOperatorConfig) GetCACert() []byte {
+	ca := operatorConfig.configCache.nsxCA
+	if ca == nil {
+		ca = []byte{}
+		for _, caFile := range operatorConfig.CaFile {
+			caCert, err := os.ReadFile(caFile)
+			if err != nil || len(caCert) == 0 {
+				configLog.Errorf("Failed to read CA file %s, err=%v, skip", caFile, err)
+				continue
+			}
+			ca = append(ca, caCert...)
+			ca = append(ca, []byte("\n")...)
+		}
+		operatorConfig.configCache.nsxCA = ca
+	}
+	return ca
+}
+
+type configCache struct {
+	// nsxCA stores all file contents of NsxConfig.CaFile in a byte slice
+	nsxCA []byte
 }
 
 type DefaultConfig struct {
@@ -181,6 +205,7 @@ func NewNSXOpertorConfig() *NSXOperatorConfig {
 		&K8sConfig{},
 		&VCConfig{},
 		&HAConfig{},
+		configCache{},
 	}
 	return defaultNSXOperatorConfig
 }
--- a/pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
+++ b/pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
@@ -38,7 +38,18 @@ var (
 	MetricResType           = common.MetricResTypeNSXServiceAccount
 )
 
-// NSXServiceAccountReconciler reconciles a NSXServiceAccount object
+// NSXServiceAccountReconciler reconciles a NSXServiceAccount object.
+// Requires NSXT 4.0.1
+//
+// create/delete event will be processed by Reconcile
+//
+// update event with realized resource missing NSX resources will be processed by Reconcile since NSXT 4.1.2
+//
+// # GarbageCollector will clean up stale NSX resources and Secret on every GCInterval
+//
+// GarbageCollector will check and make all Secrets' CA up-to-date on first GC run
+//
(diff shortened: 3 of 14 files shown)


pkg/util/crypto.go
```diff
@@ -18,6 +18,7 @@ const (
 	DefaultRSABits = 2048
 	// For now the ClusterControlPlane API doesn't support rotating certificate. We set long valid time to avoid certificate expiration.
 	DefaultValidDays          = 3650
```
edwardbadboy, line 20: We can add a DefaultValidDaysWithRotation and set it to 1 year. When NSX is >= 4.2, we can use DefaultValidDaysWithRotation for generating the cert, otherwise, use DefaultValidDays .

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -176,6 +178,22 @@ func (r *NSXServiceAccountReconciler) GarbageCollector(cancel chan bool, timeout
 		}
 		gcSuccessCount, gcErrorCount := r.garbageCollector(nsxServiceAccountUIDSet, nsxServiceAccountList)
 		log.V(1).Info("gc collects NSXServiceAccount CR", "success", gcSuccessCount, "error", gcErrorCount)
+		// Validate ca at first time
+		// Validate client cert every GCValidationInterval
+		if count == 0 {
+			for _, nsxServiceAccount := range nsxServiceAccountList.Items {
+				if nsxserviceaccount.IsNSXServiceAccountRealized(&nsxServiceAccount.Status) {
+					if err := r.Service.ValidateAndUpdateRealizedNSXServiceAccount(context.TODO(), &nsxServiceAccount, ca); err != nil {
```
edwardbadboy, line 186: GarbageCollector runs in a different Goroutine than Manager. Do we need to introduce lock between ValidateAndUpdateRealizedNSXServiceAccount and other methods of NSXServiceAccountService?

pkg/config/config.go
```diff
@@ -47,6 +48,28 @@ func (operatorConfig *NSXOperatorConfig) HAEnabled() bool {
 	return false
 }
 
+func (operatorConfig *NSXOperatorConfig) GetCACert() []byte {
+	ca := operatorConfig.configCache.nsxCA
```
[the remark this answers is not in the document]
gran-vmv: Yes.
Discussed with @TaoZou1 , it will restart in ncp+wcp env.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
gran-vmv: Currently no. I'll add more comments on this.
The same NSXServiceAccount won't be processed by different goroutine.
The PI/CCP cache is thread safe.

edwardbadboy: I also notice that setupWithManager passes runtime.NumCPU() to MaxConcurrentReconciles. I don't find locks being used in NSXServiceAccountService, do we need any locks?

gran-vmv: No, this property works as `numWorkers`, we don't have the non-threadSafe resources accessed in parallel.

pkg/util/crypto.go
gran-vmv: Fixed. Currently we use `NSXCheckVersionForNSXServiceAccountRestore`(>=4.1.2), and added `TODO` marks on this part.

gran-vmv: Update the NSX version requirement to >=4.1.3

timdengyun (commented):
LGTM

timdengyun (approved):

build/yaml/crd/nsx.vmware.com_nsxserviceaccounts.yaml
```diff
@@ -35,6 +35,10 @@ spec:
           spec:
             description: NSXServiceAccountSpec defines the desired state of NSXServiceAccount
             properties:
+              enableCertRotation:
+                description: EnableCertRotation enables cert rotation feature in this
```
liu4480, line 39: do you mean this will go into nsx-keeper-413, is nsx code merged into the branch?

gran-vmv: Yes. Since NSXT 4.1.3 and 4.2.0, CCP cert can be updated.
Then we can create a new NSXServiceAccount with enableCertRotation=true to enable the cert rotation feature.

gran-vmv: The addon-controller can set enableCertRotation based on deployed interworking version. It is TBD now.

liu4480 (approved):
LGTM

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -177,7 +190,28 @@ func (r *NSXServiceAccountReconciler) GarbageCollector(cancel chan bool, timeout
 		}
 		gcSuccessCount, gcErrorCount := r.garbageCollector(nsxServiceAccountUIDSet, nsxServiceAccountList)
 		log.V(1).Info("gc collects NSXServiceAccount CR", "success", gcSuccessCount, "error", gcErrorCount)
+		count, ca = r.validateRealized(count, ca, nsxServiceAccountList)
```
edwardbadboy, line 193: It seems it needs to wait for GCInterval (60s) before the first GC action. I think 60s delay is OK for client cert rotation. On the other hand, when NSX ca cert is updated, is it better to propagate the new ca cert to all NSXSAs as soon as possible?

edwardbadboy (commented):
Looks good in general. A minor comment.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
gran-vmv: Fixed code and updated PR description.

edwardbadboy (approved):

gran-vmv:
Manually verified with feature disabled and enabled.

TaoZou1 (approved):

PR #312, opened 2023-08-31
Add cleanup for staticroute

TaoZou1:


pkg/nsx/services/staticroute/staticroute.go
```diff
@@ -133,3 +133,17 @@ func (service *StaticRouteService) ListStaticRoute() []model.StaticRoutes {
 	}
 	return staticRouteSet
 }
+
+func (service *StaticRouteService) Cleanup() error {
+	staticRouteSet := service.ListStaticRoute()
+	log.Info("cleaning up staticroute", "count", len(staticRouteSet))
+	for _, staticRoute := range staticRouteSet {
+		path := strings.Split(*staticRoute.Path, "/")
```
dantingl, line 141: Do we need to print every route path when deleting it

TaoZou1: done

pkg/nsx/services/staticroute/staticroute.go
```diff
@@ -133,3 +133,19 @@ func (service *StaticRouteService) ListStaticRoute() []model.StaticRoutes {
 	}
 	return staticRouteSet
 }
+
+func (service *StaticRouteService) Cleanup() error {
+	staticRouteSet := service.ListStaticRoute()
+	description := "cleanup staticroute"
```
seanpang-vmware, line 139: using the same log only print different parameter seems a little strange, could we still print what it is doing in log, eg: "removing static route" in 140

TaoZou1: done

dantingl (approved):

PR #325, opened 2023-09-05
Add nsx operator version tag into util basic tags

timdengyun:


pkg/util/utils.go
```diff
@@ -35,14 +35,16 @@ const (
...
+		common.TagScopeNamespace, common.TagScopeNamespaceUID,
 		common.TagScopeStaticRouteCRName, common.TagScopeStaticRouteCRUID,
 		common.TagScopeSecurityPolicyCRName, common.TagScopeSecurityPolicyCRUID,
 		common.TagScopeSubnetCRName, common.TagScopeSubnetCRUID,
 		common.TagScopeSubnetPortCRName, common.TagScopeSubnetPortCRUID,
 		common.TagScopeVPCCRName, common.TagScopeVPCCRUID,
 		common.TagScopeIPPoolCRName, common.TagScopeIPPoolCRUID,
-		common.TagScopeSubnetSetCRName, common.TagScopeSubnetSetCRUID}
+		common.TagScopeSubnetSetCRName, common.TagScopeSubnetSetCRUID,
+	}
```
TaoZou1, line 47: I forget to add TagScopeCluster, TagScopeVersion here, could  you help to add them?

pkg/util/utils.go
```diff
@@ -344,6 +346,10 @@ func BuildBasicTags(cluster string, obj interface{}, namespaceID types.UID) []mo
 			Scope: String(common.TagScopeCluster),
 			Tag:   String(cluster),
 		},
+		{
+			Scope: String(common.TagScopeVersion),
+			Tag:   String(common.TagValueVersion10),
```
dantingl, line 351: where is TagValueVersion10, can we set value with major/minor/patch version like 1.0.0

timdengyun: TagValueVersion10 is 1.0 defined in 
https://github.com/vmware-tanzu/nsx-operator/blob/vpc_dev/pkg/nsx/services/common/types.go#L64
yes, we can set 1 as major version, .x as minor version, the last .x as patch version.

pkg/util/utils.go
timdengyun: Done

pkg/nsx/services/common/types.go
```diff
@@ -61,7 +61,9 @@ const (
 	TagValueGroupScope              string = "scope"
 	TagValueGroupSrc                string = "source"
 	TagValueGroupDst                string = "destination"
-	TagValueVersion10               string = "1.0"
+	TagValueMajorVersion            string = "1"
+	TagValueMinorVersion            string = ".0"
+	TagValuePatchVersion            string = ".0"
 	AnnotationVPCNetworkConfig      string = "nsx.vmware.com/vpc_network_config"
```
TaoZou1, line 67: how about:
TagValueMinorVersion  string = "0"
TagValuePatchVersion  string = "0"
define a var
	TagValueVersion10               = [3]string{TagValueMajorVersion, TagValueMinorVersion, TagValuePatchVersion}
when you need version string, 
 strings.Join(TagValueVersion10, ".")

timdengyun: fixed

TaoZou1 (approved):

PR #326, opened 2023-09-05
Update MaxNameLength and use GenerateTruncName to build subnet name

zhengxiexie:
nsx-t limits the name to 80-byte.


pkg/nsx/services/common/types.go
```diff
@@ -18,7 +18,7 @@ const (
 	HashLength                      int    = 8
 	MaxTagLength                    int    = 256
 	MaxIdLength                     int    = 255
-	MaxNameLength                   int    = 255
+	MaxNameLength                   int    = 80
```
dantingl, line 21: Better to use specific max length for Subnet display name

zhengxiexie: ok

PR #327, opened 2023-09-05
Fix issue when building ip block

seanpang-vmware:
using string(ip) will return an empty value, should use ip.String()


lxiaopei (approved):
Approve if changed the typo "stroe"

heypnus (approved):

PR #331, opened 2023-09-06
Fix the concurrency issue in subnet port creation

heypnus:
The patch will add a lock when creating a new NSX subnet for the
subnetport event to avoid creating multiple NSX subnet for different
subnetports in the same time.

It also changes the search index for the existing NSX subnets because
the NSX subnet tag for subnetset CR was changed from nsx-op/subnet_cr_uid
to nsx-op/subnetset_cr_uid:<subnetset CR uid>.

Tests done:

1. create 20 subnetport CRs in a short time and ensure that all of
   the NSX subnet ports are created in the same subnet (the subnet has
   enough capacity for the ports)

### DIFF
--- a/pkg/controllers/common/utils.go
+++ b/pkg/controllers/common/utils.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"sync"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


dantingl (approved):

torresergio (approved):
LGTM

jwsui (approved):
LGTM

PR #332, opened 2023-09-08
Fix default-pod-subnetset having no nsx-op/namespace

zhengxiexie:
When default-pod-subnetset of a vpc is created, it misses the tag scope
nsx-op/namespace and nsx-op/namespace_uid, and then it leads to the
problem which failed to sync the labels of namespace updating event
to subnetset. The basic reason is that the function
`util.AppendTags` filters nsx-op/namespace and nsx-op/namespace_uid scope,
replace it with `append` should fix this issue.

### DIFF
--- a/pkg/nsx/services/subnet/builder.go
+++ b/pkg/nsx/services/subnet/builder.go
@@ -41,7 +41,7 @@ func (service *SubnetService) buildSubnetSetName(subnetset *v1alpha1.SubnetSet, 
 }
 
 func (service *SubnetService) buildSubnet(obj client.Object, tags []model.Tag) (*model.VpcSubnet, error) {
-	tags = util.AppendTags(service.buildBasicTags(obj), tags)
+	tags = append(service.buildBasicTags(obj), tags...)
 	var nsxSubnet *model.VpcSubnet
 	var staticIpAllocation bool
 	switch o := obj.(type) {
--- a/pkg/nsx/services/subnet/subnet.go
+++ b/pkg/nsx/services/subnet/subnet.go
@@ -337,7 +337,7 @@ func (service *SubnetService) UpdateSubnetSetTags(ns string, vpcSubnetSets []mod
 		for _, tag := range vpcSubnetSet.Tags {
 			if *tag.Scope == common.TagScopeSubnetSetCRName {
 				name = *tag.Tag
-			} else if *tag.Scope == common.TagScopeNamespace {
+			} else if *tag.Scope == common.TagScopeNamespace || *tag.Scope == common.TagScopeVMNamespace{
 				namespace = *tag.Tag
 				if namespace != ns {
 					break


pkg/nsx/services/subnet/builder.go
```diff
@@ -41,7 +41,7 @@ func (service *SubnetService) buildSubnetSetName(subnetset *v1alpha1.SubnetSet,
 }
 
 func (service *SubnetService) buildSubnet(obj client.Object, tags []model.Tag) (*model.VpcSubnet, error) {
-	tags = util.AppendTags(service.buildBasicTags(obj), tags)
+	tags = append(service.buildBasicTags(obj), tags...)
```
timdengyun, line 44: In the base functions of util:
https://github.com/vmware-tanzu/nsx-operator/blob/vpc_dev/pkg/util/utils.go#L40
we add common.TagScopeNamespace, common.TagScopeNamespaceUID, however, these two tags should not be included in vm subnet, we need to revise this to make sure all resource except vm subnet to add these two.

timdengyun:
> util.AppendTags filters nsx-op/namespace and nsx-op/namespace_uid scope, this is not right.

This message is just spoken statement, not literature or formal writings. 

please refer to  commit msg: https://github.com/vmware-tanzu/nsx-operator/pull/331
In this PR msg, we know what is the issue, what the patch is trying to fix.

The patch will add a lock when creating a new NSX subnet for the
subnetport event to avoid creating multiple NSX subnet for different
subnetports in the same time.

It also changes the search index for the existing NSX subnets because
the NSX subnet tag for subnetset CR was changed from nsx-op/subnet_cr_uid
to nsx-op/subnetset_cr_uid:.

Tests done:

    create 20 subnetport CRs in a short time and ensure that all of
    the NSX subnet ports are created in the same subnet (the subnet has
    enough capacity for the ports)

zhengxiexie:
> > util.AppendTags filters nsx-op/namespace and nsx-op/namespace_uid scope, this is not right.
> 
> This message is just spoken statement, not literature or formal writings.
> 
> please refer to commit msg: #331 In this PR msg, we know what is the issue, what the patch is trying to fix.
> 
> The patch will add a lock when creating a new NSX subnet for the subnetport event to avoid creating multiple NSX subnet for different subnetports in the same time.
> 
> It also changes the search index for the existing NSX subnets because the NSX subnet tag for subnetset CR was changed from nsx-op/subnet_cr_uid to nsx-op/subnetset_cr_uid:.
> 
> Tests done:
> 
> ```
> create 20 subnetport CRs in a short time and ensure that all of
> the NSX subnet ports are created in the same subnet (the subnet has
> enough capacity for the ports)
> ```

ok

pkg/nsx/services/subnet/builder.go
timdengyun: So, you mean there is a bug in util.AppendTags for subnetset, is it?
if yes, can we fix it by changing util.AppendTags funcs to handle subnet/subnetset in proper way? I think util.AppendTags is common func provided for resources.

zhengxiexie: refer https://github.com/vmware-tanzu/nsx-operator/pull/336

timdengyun (approved):

PR #335, opened 2023-09-08
Fix cleanup for subnet and ippool

zhengxiexie:
1. Should delete subnet and subnetset, ListSubnetID should get them all.
2. Since ippool and staticroute rely on vpcService of serviceMediator, init vpc first, otherwise, it would report nil pointer when deleting ippool or staticroute.

### DIFF
--- a/pkg/clean/clean.go
+++ b/pkg/clean/clean.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 
 	"github.com/vmware-tanzu/nsx-operator/pkg/config"
+	commonctl "github.com/vmware-tanzu/nsx-operator/pkg/controllers/common"
 	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
@@ -58,6 +59,9 @@ func InitializeCleanupService(cf *config.NSXOperatorConfig) (*CleanupService, er
 		NSXConfig: cf,
 	}
 
+	vpcService, vpcErr := vpc.InitializeVPC(commonService)
+	commonctl.ServiceMediator.VPCService = vpcService
+
 	// initialize all the CR services
 	// Use Fluent Interface to escape error check hell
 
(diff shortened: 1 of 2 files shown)


pkg/clean/clean.go
```diff
@@ -58,6 +59,9 @@ func InitializeCleanupService(cf *config.NSXOperatorConfig) (*CleanupService, er
 		NSXConfig: cf,
 	}
 
+	vpcService, vpcErr := vpc.InitializeVPC(commonService)
+	commonctl.ServiceMediator.VPCService = vpcService
+
```
seanpang-vmware, line 64: just to confirm, only pre-init vpc service in mediator is ok, right? since in mediator there are also other services

zhengxiexie: I’m not sure, at least currently there is no fatal error.

seanpang-vmware (approved):

pkg/nsx/services/subnet/subnet.go
```diff
@@ -308,8 +308,9 @@ func (service *SubnetService) UpdateSubnetSetStatus(obj *v1alpha1.SubnetSet) err
 }
 
 func (service *SubnetService) ListSubnetID() sets.String {
-	subnetSet := service.SubnetStore.ListIndexFuncValues(common.TagScopeSubnetCRUID)
-	return subnetSet
+	subnets := service.SubnetStore.ListIndexFuncValues(common.TagScopeSubnetCRUID)
+	subnetSets := service.SubnetStore.ListIndexFuncValues(common.TagScopeSubnetSetCRUID)
+	return subnets.Union(subnetSets)
 }
```
heypnus, line 314: Can we remove `ListSubnetID()` and change `Cleanup()` to something like:
```
func (service *SubnetService) Cleanup() error {
	subnets := service.SubnetStore.List()
	log.Info("cleaning up subnet", "count", len(subnets))
	for _, subnet := range subnets {
		nsxSubnet, _ := subnet.(model.VpcSubnet)
		err := service.DeleteSubnet(nsxSubnet)
		if err != nil {
			return err
		}
	}
	return nil
}
```
because I think it's strange to firstly get the ids from store and secondly get objects from the store by ids...

PR #336, opened 2023-09-08
Remove nsx-op/namespace and nsx-op/namespace_uid tags for subnet/subn…

timdengyun:
This patch is to remove nsx-op/namespace and nsx-op/namespace_uid
when adding the basic tags for subnet/subnetset, since subnet/subnetset
need distinguish Pod and VM subnet. For VM subnet/subnetset, it will use
nsx-op/vm_namespace and nsx-op/vm_namespace_uid.
The subnet/subnetset service will handle adding nsx-op/namespace
and nsx-op/namespace_uid tags for pod and vm subnet.

### DIFF
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -37,7 +37,6 @@ var (
 	String    = common.String
 	basicTags = []string{
 		common.TagScopeCluster, common.TagScopeVersion,
-		common.TagScopeNamespace, common.TagScopeNamespaceUID,
 		common.TagScopeStaticRouteCRName, common.TagScopeStaticRouteCRUID,
 		common.TagScopeSecurityPolicyCRName, common.TagScopeSecurityPolicyCRUID,
 		common.TagScopeSubnetCRName, common.TagScopeSubnetCRUID,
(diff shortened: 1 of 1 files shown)


pkg/util/utils.go
```diff
@@ -355,10 +354,12 @@ func BuildBasicTags(cluster string, obj interface{}, namespaceID types.UID) []mo
 	switch i := obj.(type) {
 	case *v1alpha1.StaticRoute:
 		tags = append(tags, model.Tag{Scope: String(common.TagScopeNamespace), Tag: String(i.ObjectMeta.Namespace)})
+		tags = append(tags, model.Tag{Scope: String(common.TagScopeNamespaceUID), Tag: String(string(namespaceID))})
```
heypnus, line 357: This may cause a different behavior, e.g. the static route will have a empty tag `{"nsx-op/namespace_uid": ""}`, I'm not sure whether it's ok...

timdengyun: why we can not get namespaceID from i.ObjectMeta.Namespace instead of passing it from caller?

heypnus: i.ObjectMeta.Namespace is name, not uid

seanpang-vmware (approved):

heypnus (approved):

PR #337, opened 2023-09-08
Add NSX Operator apis in a module

lxiaopei:
This patch generates nsx-operator apis in a separate module, to be imported by other components.


ihgann:
Is all of this required? A lot of this looks like general cleanup, but maybe it is required? In either case - I'm excited for this!

PR #340, opened 2023-09-12
Fix error when search subnet by key

jwsui:
NSX subnet ID had been reformatted with prefix. To be compatible with the format, update parameter of searching NSX subnet by key.


zhengxiexie (approved):
LGTM

PR #343, opened 2023-09-14
Support insecure for testing purpose

TaoZou1:
If insecure=True, validate cert will be passed. but createTransport doesn't support that option. Support this option for testing purpose


zhengxiexie (approved):
LGTM

PR #344, opened 2023-09-14
Fix bugs of updating NSX subnet managed by SubnetSet CR

jwsui:
A random index is added to NSX subnet ID when updating NSX subnet managed by SubnetSet CR. This will cause the issue that nsx-operator will create a new NSX subnet instead of updating the existing one as there is a different ID for each request. This patch fixes this issue with below case passed:
1. nsx-operator updates existing NSX subnet tags when related namespace labels are added/updated/deleted instead of creating new NSX subnet.


zhengxiexie (dismissed):
LGTM

zhengxiexie (approved):

PR #351, opened 2023-09-20
[NSXServiceAccount] Fix incorrect GC in restore case

gran-vmv:
Add UID check when creating/deleting NSXServiceAccount to avoid unexpected PI/CCP/Secret operation.
After NSXT restore finished, PI/CCP will be restored to old state, and it may contains same cluster name if a cluster is deleted and re-created after backup point. UID validation is required on this case.

### DIFF
--- a/pkg/nsx/services/nsxserviceaccount/cluster.go
+++ b/pkg/nsx/services/nsxserviceaccount/cluster.go
@@ -208,7 +208,8 @@ func (s *NSXServiceAccountService) UpdateRealizedNSXServiceAccount(ctx context.C
 
 func (s *NSXServiceAccountService) createPIAndCCP(normalizedClusterName string, vpcPath string, cert string, existingClusterId *string, obj *v1alpha1.NSXServiceAccount) (string, error) {
 	// create PI
-	if piObj := s.PrincipalIdentityStore.GetByKey(normalizedClusterName); piObj == nil {
+	hasPI := len(s.PrincipalIdentityStore.GetByIndex(common.TagScopeNSXServiceAccountCRUID, string(obj.UID))) > 0
+	if piObj := s.PrincipalIdentityStore.GetByKey(normalizedClusterName); !hasPI && piObj == nil {
 		pi, err := s.NSXClient.WithCertificateClient.Create(mpmodel.PrincipalIdentityWithCertificate{
 			IsProtected: &isProtectedTrue,
 			Name:        &normalizedClusterName,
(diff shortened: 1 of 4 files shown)


pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -284,27 +290,45 @@ func (s *NSXServiceAccountService) getProxyEndpoints(ctx context.Context) (v1alp
...
+	if !isDeleteSecret {
+		isDeletePI = len(s.PrincipalIdentityStore.GetByIndex(common.TagScopeNSXServiceAccountCRUID, string(uid))) > 0
+		isDeleteCCP = len(s.ClusterControlPlaneStore.GetByIndex(common.TagScopeNSXServiceAccountCRUID, string(uid))) > 0
+	}
 	// delete ClusterControlPlane
-	cascade := true
-	if err := s.NSXClient.ClusterControlPlanesClient.Delete(siteId, enforcementpointId, normalizedClusterName, &cascade); err != nil {
-		log.Error(err, "failed to delete", "ClusterControlPlane", normalizedClusterName)
-		return err
+	if isDeleteCCP {
```
liu4480, line 321: looks like isDeleteCCP is always true, may I know the reason to have a if here?

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -284,27 +290,45 @@ func (s *NSXServiceAccountService) getProxyEndpoints(ctx context.Context) (v1alp
...
+			log.Error(err, "failed to delete", "ClusterControlPlane", normalizedClusterName)
+			return err
+		}
+		s.ClusterControlPlaneStore.Delete(model.ClusterControlPlane{Id: &normalizedClusterName})
 	}
-	s.ClusterControlPlaneStore.Delete(model.ClusterControlPlane{Id: &normalizedClusterName})
 
 	// delete PI
-	if piobj := s.PrincipalIdentityStore.GetByKey(normalizedClusterName); piobj != nil {
+	if piobj := s.PrincipalIdentityStore.GetByKey(normalizedClusterName); isDeletePI && (piobj != nil) {
```
liu4480, line 331: so does isDeletePI

pkg/nsx/services/nsxserviceaccount/cluster.go
gran-vmv: No, when `isDeleteSecret` is false, we need to check the UID of CCP/PI we want to delete to avoid unexpected deletion.

edwardbadboy:
Suggest to write a summary of the problem and the solution in PR and commit message.

gran-vmv:
> Suggest to write a summary of the problem and the solution in PR and commit message.

Updated.

timdengyun (approved):

liu4480 (approved):
LGTM

lxiaopei (approved):
LGTM.

PR #355, opened 2023-09-28
Support SecurityPolicy cleanup in VPC

timdengyun:
In Nsx Operator, we support SecurityPolicy in both VPC netowrk and non-VPC network. Nsx Operator cleanup only supports cleanup in VPC network.
So, we need to add one more argument to call delete SecurityPolicy interface to differentiate in VCP cleanup or not.


zhengxiexie (approved):
LGTM

PR #357, opened 2023-10-10
Cherry-pick NSXCheckVersion refactoring from vpc_dev

gran-vmv:
Cherry-pick `NSXCheckVersion` refactoring in #260 to avoid mass conflicts.

### DIFF
--- a/pkg/nsx/client.go
+++ b/pkg/nsx/client.go
@@ -26,10 +26,15 @@ import (
 )
 
 const (
-	FeatureSecurityPolicy           string = "SECURITY_POLICY"
-	FeatureNSXServiceAccount        string = "NSX_SERVICE_ACCOUNT"
-	FeatureNSXServiceAccountRestore string = "NSX_SERVICE_ACCOUNT_RESTORE"
+	VPC = iota
+	SecurityPolicy
+	ServiceAccount
+	ServiceAccountRestore
+	StaticRoute
+	AllFeatures
 )
+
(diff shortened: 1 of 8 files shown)


pkg/nsx/client.go
```diff
@@ -26,11 +26,16 @@ import (
 )
 
 const (
-	FeatureSecurityPolicy           string = "SECURITY_POLICY"
-	FeatureNSXServiceAccount        string = "NSX_SERVICE_ACCOUNT"
-	FeatureNSXServiceAccountRestore string = "NSX_SERVICE_ACCOUNT_RESTORE"
+	VPC = iota
```
timdengyun, line 29: It's good to remove VPC and StaticRoute here, since we don't use it yet.

pkg/nsx/client.go
```diff
@@ -26,11 +26,16 @@ import (
...
+	SecurityPolicy
+	ServiceAccount
+	ServiceAccountRestore
+	StaticRoute
+	AllFeatures
 )
 
+var FeaturesName = [AllFeatures]string{"VPC", "SECURITY_POLICY", "NSX_SERVICE_ACCOUNT", "NSX_SERVICE_ACCOUNT_RESTORE", "STATIC_ROUTE"}
```
timdengyun, line 37: ditto

pkg/nsx/client.go
timdengyun: also, we already remove VPC and StaticRoute version check in 
func (nsxVersion *NsxVersion) featureSupported(feature int) bool

gran-vmv: Since we may add new features here on both main and vpc_dev branch, we'd better to keep an union here to avoid many conflicts when cherry-pick changes to vpc_dev and to main.

timdengyun (approved):

PR #366, opened 2023-10-23
Use fixed value to set number of reconciles

TaoZou1:
On Linux, runtime.NumCPU() function relies on the sched_getaffinity system call which only considers CPU core bindings and not considner CPU resource restrictions imposed by cgroups in container environments. Automaxprocs automatically set GOMAXPROCS to match Linux container CPU quota.  
The number of reconciles uses the fixed value now

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	vmv1alpha1 "github.com/vmware-tanzu/vm-operator/api/v1alpha1"
+	_ "go.uber.org/automaxprocs"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
--- a/pkg/controllers/common/utils.go
+++ b/pkg/controllers/common/utils.go
@@ -106,3 +106,8 @@ func NodeIsMaster(node *v1.Node) bool {
 	}
 	return false
 }
+
+// NumReconcile now uses the fix number of concurrency
+func NumReconcile() int {
+	return MaxConcurrentReconciles
+}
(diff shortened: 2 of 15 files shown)


pkg/controllers/common/utils.go
```diff
@@ -106,3 +107,14 @@ func NodeIsMaster(node *v1.Node) bool {
 	}
 	return false
 }
+
+// NumReconcile depends on the https://github.com/uber-go/automaxprocs which will set the number of CPUs from cgroups information.
+// automaxprocs only support linxu platform.
```
zhengxiexie, line 112: typo linux

pkg/controllers/common/utils.go
```diff
@@ -106,3 +107,14 @@ func NodeIsMaster(node *v1.Node) bool {
...
+func NumReconcile() int {
+	numCPU := runtime.GOMAXPROCS(0)
+	if numCPU <= 1 {
```
zhengxiexie, line 115: Why hard code with 4 or 8 instead of return NumCPU or GOMAXPROCS?

pkg/controllers/common/utils.go
TaoZou1: done

pkg/controllers/common/utils.go
TaoZou1: nsx-operator is a network-intensive task， so each CPU should have more than 1 thread to utilize CPU resource. 4 refers to NCP setting

cmd/main.go
```diff
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	vmv1alpha1 "github.com/vmware-tanzu/vm-operator/api/v1alpha1"
+	_ "go.uber.org/automaxprocs"
```
dantingl, line 11: Do we need this module?

pkg/controllers/common/utils.go
```diff
@@ -17,6 +17,10 @@ import (
 	servicecommon "github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
 )
 
+const (
+	NumberConciler = 8
```
dantingl, line 21: We can define some constant values(maybe in common/types.go?) like MaxConcurrentReconciles in general, and specific ones for SubnetPort/Pod? Or we can wait for system test result before we distinguish between different resources.

TaoZou1:
create a new pull reques for main branch. close this one.

PR #367, opened 2023-10-24
Fix type issue of ippool

zhengxiexie:
Move the logic of deciding type of ippool before add and delete,
or else it would fail to delete ippool if user doesn't specify
type field.

### DIFF
--- a/pkg/controllers/ippool/ippool_controller.go
+++ b/pkg/controllers/ippool/ippool_controller.go
@@ -124,6 +124,17 @@ func (r *IPPoolReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 
 	// TODO: Xiaopei's suggestions: is there possibility that IPPool was deleted from nsx store but NSX block subnet was not deleted?
 
+	if obj.Spec.Type == "" {
+		vpcNetworkConfig := common.ServiceMediator.GetVPCNetworkConfigByNamespace(obj.Namespace)
+		if vpcNetworkConfig == nil {
+			err := fmt.Errorf("operate failed: cannot get configuration for IPPool CR")
+			log.Error(err, "failed to find VPCNetworkConfig for IPPool CR", "ippool", req.NamespacedName, "namespace %s", obj.Namespace)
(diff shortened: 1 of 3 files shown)


build/yaml/crd/nsx.vmware.com_ippools.yaml
```diff
@@ -173,9 +173,6 @@ spec:
                 type: array
               type:
                 description: Type defines the type of this IPPool, Public or Private.
-                enum:
```
wyike, line 176: Suggest not to remove the validations on public and private enum. It would be useful.

Defining omitempty to Type field will allow user doesn't specify type field and will avoid the "spec.type: Unsupported value: "":" error.  In the meanwhile unmarshal the json to golang struct will got zero value of Type as well.

wyike:
Let' remove the change on CRD and API? They are not needed.

build/yaml/crd/nsx.vmware.com_ippools.yaml
zhengxiexie: ok

wyike (approved):

TaoZou1 (approved):

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -124,6 +124,17 @@ func (r *IPPoolReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 
 	// TODO: Xiaopei's suggestions: is there possibility that IPPool was deleted from nsx store but NSX block subnet was not deleted?
 
+	if obj.Spec.Type == "" {
```
wyike, line 127: sorry for the late awareness.

If spec.Type is set to a value at the beginning, I notice

```
			controllerutil.AddFinalizer(obj, servicecommon.IPPoolFinalizerName)
			if err := r.Client.Update(ctx, obj); err != nil {
				log.Error(err, "add finalizer", "ippool", req.NamespacedName)
				return resultRequeue, err
}
```
there will be an add finalizer later.

Will it patch the Type to ippool object? Is this what you want? Have you tested the change locally first?

PR #368, opened 2023-10-25
Change network config crd id to path

seanpang-vmware:
NsxtProject and externalIPBlocks now uses resource path instead of resource id

Test Passed on mocked CI testbed with VPC testcases

### DIFF
--- a/pkg/controllers/vpc/vpc_utils.go
+++ b/pkg/controllers/vpc/vpc_utils.go
@@ -2,13 +2,16 @@ package vpc
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
+	"strings"
 
 	"github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
 	"github.com/vmware-tanzu/nsx-operator/pkg/config"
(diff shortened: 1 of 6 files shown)


pkg/controllers/vpc/vpc_utils.go
```diff
@@ -107,3 +110,33 @@ func getExistingConditionOfType(conditionType v1alpha1.ConditionType, existingCo
...
+// example /orgs/default/projects/nsx_operator_e2e_test
+func NsxtProjectPathToId(path string) (string, string, error) {
+	parts := strings.Split(path, "/")
+	if len(parts) < 4 {
+		return "", "", errors.New("Invalid NSXT project path")
+	}
+	return parts[2], parts[len(parts)-1], nil
+}
+
+func BuildNetworkConfigInfo(vpcConfigCR v1alpha1.VPCNetworkConfiguration) (*vpc.VPCNetworkConfigInfo, error) {
```
lxiaopei, line 124: can this be a private func?

lxiaopei (approved):
LGTM。Thanks for the change @seanpang-vmware Sean

PR #373, opened 2023-10-27
Add go module to pkg/apis, also bump packages of controller runtime

lubronzhan:
Make pkg/apis as sub go module, so it could be easily consume by other projects

At the same time, also bumped controller-runtime and other packages as part of go mod tidy.

Built all sub module using below commands
```
$ find . -name go.mod -print0 | \
  xargs -0 dirname | \
  xargs -I% -n1 bash -c 'cd %; echo; echo "building $(pwd)"; time go build ./...; cd - 2>&1 >/dev/null'


building /Users/lzhan/work/src/github.com/lubronzhan/nsx-operator

real    0m2.947s
user    0m4.571s
sys     0m2.254s

building /Users/lzhan/work/src/github.com/lubronzhan/nsx-operator/pkg/apis

real    0m0.204s
user    0m0.414s
sys     0m0.607s
```

Release graph
```
Changed the pkg/apis packages to be their own Go modules so other projects can import those APIs without pulling in this project's entire dependency graph.
```


lubronzhan:
Should I commit to vpc_dev instead?

lubronzhan:
Close the PR since it's done on vpc branch

PR #379, opened 2023-11-01
Add virtualmachines CRD to avoid error in e2e running

zhengxiexie:
This file test/e2e/manifest/common/virtualmachine.yaml is dumped from WCP env Since this CRD file is needed for sunetport controller, so adding it in the e2e test.

### DIFF
--- a/test/e2e/manifest/common/virtualmachine.yaml
+++ b/test/e2e/manifest/common/virtualmachine.yaml
@@ -1,0 +1,596 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: virtualmachines.vmoperator.vmware.com
+spec:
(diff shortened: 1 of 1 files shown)


test/e2e/manifest/testIPPool/ippool.yaml
```diff
@@ -3,7 +3,7 @@ kind: IPPool
 metadata:
   name: guestcluster-ippool-2
 spec:
-  type: public
+  type: Public
```
timdengyun, line 6: please remove ip pool related test code and only add VM CRD for this PR. It's good to trace incremental change in one individual change, also follow agile methodology.

zhengxiexie: ok

timdengyun (approved):

PR #380, opened 2023-11-01
Fix typo of ippool e2e test case

zhengxiexie:
Fix ippool e2e test case

    Rename public to Public in ippool yaml.
    Update log in ippool e2e test.


timdengyun:
please change PR title and description to tell what exactly this PR is doing. It looks like  this PR is not for Ippool controller or service update, it's doing update for IPPool e2e test related  update. If that, we should make statement more precise.

timdengyun (approved):

PR #381, opened 2023-11-01
Fix ippool logic bug

zhengxiexie:
If namespace created and VPC not finished creating, ippool would report all ip blocks are all exhausted.
1. Modify it, if VPC not finished, just retry.
2. If all blocks are all exhausted, clear the state which saves the exhausted ipblock and retry exponentially.

### DIFF
--- a/pkg/controllers/ippool/ippool_controller.go
+++ b/pkg/controllers/ippool/ippool_controller.go
@@ -172,8 +172,10 @@ func (r *IPPoolReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 			updateFail(r, &ctx, obj, &err)
 			// if all ip blocks are exhausted, we should not retry
 			if errors.As(err, &util.IPBlockAllExhaustedError{}) {
-				log.Error(err, "ip blocks are all exhausted, would not retry", "ippool", req.NamespacedName)
-				return resultNormal, nil
+				log.Error(err, "ip blocks are all exhausted, would retry exponentially", "ippool", req.NamespacedName)
+				r.Service.ExhaustedIPBlock = []string{}
+				log.Info("Clear ExhaustedIPBlock: ", "ExhaustedIPBlock", r.Service.ExhaustedIPBlock)
+				return common.ResultRequeueAfter10sec, err
 			}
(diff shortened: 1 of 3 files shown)


pkg/controllers/ippool/ippool_controller.go
```diff
@@ -172,8 +172,10 @@ func (r *IPPoolReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 			updateFail(r, &ctx, obj, &err)
 			// if all ip blocks are exhausted, we should not retry
 			if errors.As(err, &util.IPBlockAllExhaustedError{}) {
-				log.Error(err, "ip blocks are all exhausted, would not retry", "ippool", req.NamespacedName)
-				return resultNormal, nil
+				log.Error(err, "ip blocks are all exhausted, would retry exponentially", "ippool", req.NamespacedName)
+				r.Service.ExhaustedIPBlock = []string{}
+				log.Info("Clear ExhaustedIPBlock: ", "ExhaustedIPBlock", r.Service.ExhaustedIPBlock)
+				return resultRequeue, err
```
seanpang-vmware, line 178: maybe we need a little time before requeue, maybe 5 seconds?

pkg/nsx/services/ippool/builder.go
```diff
@@ -94,7 +94,7 @@ func (service *IPPoolService) buildIPSubnet(IPPool *v1alpha2.IPPool, subnetReque
 	VPCInfo := commonctl.ServiceMediator.GetVPCsByNamespace(IPPool.Namespace)
 	if len(VPCInfo) == 0 {
 		log.Error(nil, "failed to find VPCInfo for IPPool CR", "IPPool", IPPool.Name, "namespace %s", IPPool.Namespace)
```
seanpang-vmware, line 96: is %s needed?

zhengxiexie: ok

pkg/controllers/ippool/ippool_controller.go
zhengxiexie: OK

seanpang-vmware (approved):
LGTM

PR #385, opened 2023-11-08
Fix dumping http response error

TaoZou1:
Dump http response body by io.ReadAll instead of print it directly


dantingl (approved):

zhengxiexie:
run e2e

PR #386, opened 2023-11-09
Add retry to cleanup NSX resources

zhengxiexie:
Add retry to Cleanup func, and summarize the errors returned to the upper.
 1. failed to validate config
 2. failed to get nsx client
 3. failed to initialize cleanup service
 4. failed to clean up specific resource We would retry 5 times if the 4th error occurs.

### DIFF
--- a/pkg/clean/clean.go
+++ b/pkg/clean/clean.go
@@ -6,6 +6,8 @@ package clean
 import (
 	"fmt"
 
+	"k8s.io/client-go/util/retry"
+
 	"github.com/vmware-tanzu/nsx-operator/pkg/config"
 	commonctl "github.com/vmware-tanzu/nsx-operator/pkg/controllers/common"
 	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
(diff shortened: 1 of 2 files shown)


pkg/clean/clean.go
```diff
@@ -25,19 +27,35 @@ var log = logger.Log
 // including security policy, static route, subnet, subnet port, subnet set, vpc, ip pool, nsx service account
 // it is usually used when nsx-operator is uninstalled and remove all the resources created by nsx-operator
 // return error if any, return nil if no error
+// the error roughly are:
```
dantingl, line 30: Do we need to add error code for these cases, or do they return static error messages, especially should be user friendly for the above 3 cases

zhengxiexie:
e2e

zhengxiexie:
e2e

pkg/clean/clean.go
ihgann: The error code should not be here - they should be in `main.go`. In this case, specific error types should be returned.

pkg/clean/clean.go
```diff
@@ -25,19 +27,35 @@ var log = logger.Log
...
+// 3. failed to initialize cleanup service
+// 4. failed to clean up specific resource
 func Clean(cf *config.NSXOperatorConfig) error {
 	log.Info("starting NSX cleanup")
 	if err := cf.ValidateConfigFromCmd(); err != nil {
 		return fmt.Errorf("failed to validate config: %w", err)
 	}
 	if cleanupService, err := InitializeCleanupService(cf); err != nil {
-		return fmt.Errorf("failed to initialize cleanup service: %w", err)
+		return err // failed to get nsx client
```
ihgann, line 41: Why remove the context of the previous note?

pkg/clean/clean.go
```diff
@@ -25,19 +27,35 @@ var log = logger.Log
...
 	} else if cleanupService.err != nil {
 		return fmt.Errorf("failed to initialize cleanup service: %w", cleanupService.err)
 	} else {
```
ihgann, line 44: nit: It's cleaner to remove this `else` clause. `else` is implicit in this section.

pkg/clean/clean.go
zhengxiexie: ok

pkg/clean/clean.go
zhengxiexie: I think it is duplicated with line 43, so remove it.

seanpang-vmware (approved):

PR #390, opened 2023-11-15
[NSXServiceAccount] Support CA validation and client cert rotation(#311)

timdengyun:
CP PR: https://github.com/vmware-tanzu/nsx-operator/pull/311

Add EnableCertRotation property in NSXServiceAccount.Spec. Cache CA content in memory.
Populate CA content to Secret when processing NSXServiceAccount create event. Check all Secrets' ca.crt are up-to-date when first NSXServiceAccount GC finished. Update if needed. Check all Secrets' client cert need renew every 720 GC interval. Renew client cert if cert will expire in 7d when rotation is supported (NSXT >=4.1.3 and EnableCertRotation is true). Use 365d instead of 3650d for new client cert when rotation is supported (NSXT >=4.1.3 and EnableCertRotation is true).

### DIFF
--- a/pkg/controllers/vpc/vpc_utils.go
+++ b/pkg/controllers/vpc/vpc_utils.go
@@ -7,13 +7,14 @@ import (
 	"reflect"
 	"strings"
 
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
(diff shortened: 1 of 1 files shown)


gran-vmv:
Could you also CP #351 to vpc_dev?

Issue #394, opened 2023-11-20
SecurityPolicy can not delete a rule when this rule index is not last one. 


timdengyun:
In certain case, SecurityPolicy can not delete a rule  when this rule index is not last one. 
For example, for the SP below:
```
apiVersion: nsx.vmware.com/v1alpha1
kind: SecurityPolicy
metadata:
  name: sp-app-access222
  namespace: ns-1
spec:
  priority: 0
  appliedTo:
    - podSelector:
        matchLabels:
          role: db
  rules:
    - direction: in ----->>rule1
      action: allow
      sources:
        - podSelector:
            matchLabels:
              app: coffee
      ports:
        - protocol: TCP
          port: 6001

    - direction: in ----->>rule2
      action: allow
      sources:
        - vmSelector:
            matchLabels:
              app: coco
      ports:
        - protocol: TCP
          port: 8282

    - direction: out. ----->>rule3
      action: allow
      destinations:
        - namespaceSelector:
            matchLabels:
              ns-name: ns-2
          vmSelector:
            matchLabels:
              app: web
      ports:
        - protocol: TCP
          port: 3306
```
When user tries to delete the second rule2, it will fail. The error is showing the source group in the rule 2 is being used and can not be deleted. 
```
    - direction: in
      action: allow
      sources:
        - vmSelector:
            matchLabels:
              app: coco
      ports:
        - protocol: TCP
          port: 8282
```

This issue is caused by the original rule2 ID: sp_spUID_1_0_0 will be regenerated by the original rule3, since we don't use the unique ID for the each rule generated.

timdengyun:
The issue has been resolved with PR#395

PR #395, opened 2023-11-20
Generate hash ID to make security policy rule ID unique

timdengyun:
In certain case, SecurityPolicy cannot delete a rule
when this rule is not last one.
the rule ID is sp_spUID_0_0_0, sp_spUID_1_0_0, and sp_spUID_2_0_0, respectively.

When user is trying to delete rule 2, the deleting will fail.
This issue is caused by the original rule2 ID: sp_spUID_1_0_0 will be regenerated
by the original rule3, because rule3 becomes the new rule2, the new rule2 id will
be sp_spUID_2_0_0. NSX will complain that groups in the deleted rule2 are still
being referenced by sp_spUID_2_0_0.

Hence, this patch is to
1.Generate hash ID for security policy rule content and append this
hash ID to the final generated rule ID to make each rule's ID unique.
So, the final generated rule ID is: sp_spUID_rulehashID_ruleIndex_portIndex_portAddressIndex.

The testing Done:
3. Creating namedport SP with rule1 and rule2, and then adding rule2 in
the middle to make the original rule2 becomes rule3.
4. Creating SPs without this patch, and staring NSX Operator again with this
patch to verify that new generated rules with unique rule ID are
created successfully. This is to test upgrade existing SPs.

Fix the issue: https://github.com/vmware-tanzu/nsx-operator/issues/394

### DIFF
--- a/pkg/nsx/services/common/types.go
+++ b/pkg/nsx/services/common/types.go
@@ -34,6 +34,9 @@ const (
 	TagScopeNCPVNETInterface        string = "ncp/vnet_interface"
 	TagScopeVPCCRName               string = "nsx-op/vpc_cr_name"
 	TagScopeVPCCRUID                string = "nsx-op/vpc_cr_uid"
+	TagValueGroupScope              string = "scope"
+	TagValueGroupSource             string = "source"
+	TagValueGroupDestination        string = "destination"
 
 	GCInterval    = 60 * time.Second
 	FinalizerName = "securitypolicy.nsx.vmware.com/finalizer"
--- a/pkg/nsx/services/securitypolicy/builder.go
+++ b/pkg/nsx/services/securitypolicy/builder.go
@@ -9,7 +9,7 @@ import (
 
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
-	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
@@ -95,16 +95,15 @@ func (service *SecurityPolicyService) buildSecurityPolicy(obj *v1alpha1.Security
 }
 
 func (service *SecurityPolicyService) buildPolicyGroup(obj *v1alpha1.SecurityPolicy) (*model.Group, string, error) {
-	policyGroup := model.Group{}
-
-	policyGroup.Id = String(service.buildPolicyGroupID(obj))
+	policyAppliedGroup := model.Group{}
+	policyAppliedGroup.Id = String(service.buildAppliedGroupID(obj, -1))
 
 	// TODO: have a common function to generate ID and Name with parameters like prefix, suffix
-	policyGroup.DisplayName = String(fmt.Sprintf("%s-%s-scope", obj.ObjectMeta.Namespace, obj.ObjectMeta.Name))
+	policyAppliedGroup.DisplayName = String(service.buildAppliedGroupName(obj, -1))
 
 	appliedTo := obj.Spec.AppliedTo
-	targetTags := service.buildTargetTags(obj, &appliedTo, -1)
-	policyGroup.Tags = targetTags
+	targetTags := service.buildTargetTags(obj, &appliedTo, nil, -1)
+	policyAppliedGroup.Tags = targetTags
 	if len(appliedTo) == 0 {
 		return nil, "ANY", nil
 	}
@@ -117,7 +116,7 @@ func (service *SecurityPolicyService) buildPolicyGroup(obj *v1alpha1.SecurityPol
 		criteriaCount, totalExprCount, err = service.updateTargetExpressions(
 			obj,
 			&target,
-			&policyGroup,
+			&policyAppliedGroup,
 			i,
 		)
 		if err == nil {
@@ -146,11 +145,13 @@ func (service *SecurityPolicyService) buildPolicyGroup(obj *v1alpha1.SecurityPol
 		return nil, "", err
 	}
 
-	policyGroupPath := service.buildPolicyGroupPath(obj)
-	return &policyGroup, policyGroupPath, nil
+	policyAppliedGroupPath := service.buildAppliedGroupPath(obj, -1)
+	return &policyAppliedGroup, policyAppliedGroupPath, nil
 }
 
-func (service *SecurityPolicyService) buildTargetTags(obj *v1alpha1.SecurityPolicy, targets *[]v1alpha1.SecurityPolicyTarget, idx int) []model.Tag {
+func (service *SecurityPolicyService) buildTargetTags(obj *v1alpha1.SecurityPolicy, targets *[]v1alpha1.SecurityPolicyTarget,
+	rule *v1alpha1.SecurityPolicyRule, ruleIdx int,
+) []model.Tag {
 	basicTags := service.buildBasicTags(obj)
 	sort.Slice(*targets, func(i, j int) bool {
 		k1, _ := json.Marshal((*targets)[i])
@@ -161,7 +162,7 @@ func (service *SecurityPolicyService) buildTargetTags(obj *v1alpha1.SecurityPoli
 	targetTags := []model.Tag{
 		{
 			Scope: String(common.TagScopeGroupType),
-			Tag:   String("scope"),
+			Tag:   String(common.TagValueGroupScope),
 		},
 		{
 			Scope: String(common.TagScopeSelectorHash),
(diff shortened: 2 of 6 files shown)


dantingl (commented):
I don't think we should use index in rule id, otherwise sometimes the id needs to be changed if the index is changed

timdengyun:
> I don't think we should use index in rule id, otherwise sometimes the id needs to be changed if the index is changed

I agree on this point. I did a rough test, we can remove rule index from rule id part, it can resolve duplicated rule issue as well. However, there is another issue, we also use rule index in the rule group id. we need to refactor this part as well once excluding rule index from rule ID. I suggest we can do it in the vpc dev story.

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -564,114 +590,100 @@ func (service *SecurityPolicyService) buildRuleAppliedGroupByRule(obj *v1alpha1.
...
-		criteriaCount, totalExprCount, err = service.updatePeerExpressions(
-			obj,
-			&peer,
-			&ruleSrcGroup,
-			i,
-		)
-		if err == nil {
-			ruleSrcGroupCount += criteriaCount
-			ruleSrcGroupTotalExprCount += totalExprCount
+func (service *SecurityPolicyService) buildRulePeerGroupPath(obj *v1alpha1.SecurityPolicy, groupID string) (string, error) {
```
zhengxiexie, line 601: Unused parameter 'obj *v1alpha1.SecurityPolicy'

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -564,114 +590,100 @@ func (service *SecurityPolicyService) buildRuleAppliedGroupByRule(obj *v1alpha1.
...
 	} else {
-		ruleSrcGroupName = fmt.Sprintf("%s-%d-src", obj.ObjectMeta.Name, idx)
-	}
-	ruleSrcGroupPath := fmt.Sprintf("/infra/domains/%s/groups/%s", getDomain(service), ruleSrcGroupID)
-	peerTags := service.BuildPeerTags(obj, &sources, idx)
-	ruleSrcGroup := model.Group{
-		Id:          &ruleSrcGroupID,
-		DisplayName: &ruleSrcGroupName,
-		Tags:        peerTags,
+		return fmt.Sprintf("sp_%s_%d_dst", obj.UID, ruleIdx)
```
zhengxiexie, line 597: ditto

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -309,12 +310,34 @@ func (service *SecurityPolicyService) buildExpressionsMatchExpression(matchExpre
...
+		return fmt.Sprintf("sp_%s_%d_scope", obj.UID, ruleIdx)
+	}
+
 	return fmt.Sprintf("sp_%s_scope", obj.UID)
 }
 
-func (service *SecurityPolicyService) buildPolicyGroupPath(obj *v1alpha1.SecurityPolicy) string {
-	return fmt.Sprintf("/infra/domains/%s/groups/%s", getDomain(service), service.buildPolicyGroupID(obj))
+// build appliedTo group path for both policy and rule levels.
+func (service *SecurityPolicyService) buildAppliedGroupPath(obj *v1alpha1.SecurityPolicy, groupID string) string {
```
heypnus, line 323: How do you think about unifying the styles for these 3 functions `buildAppliedGroupID`, `buildAppliedGroupName` and `buildAppliedGroupPath`? i.e.
```
func (service *SecurityPolicyService) buildAppliedGroupId(obj *v1alpha1.SecurityPolicy, ruleIdx int) string {
    ...
}
...
func (service *SecurityPolicyService) buildAppliedGroupName(obj *v1alpha1.SecurityPolicy, ruleIdx int) string {
    ...
}
...
func (service *SecurityPolicyService) buildAppliedGroupPath(obj *v1alpha1.SecurityPolicy, ruleIdx int) string {
    ...
}
```
And the similar question for `buildRulePeerGroupXXX`

pkg/nsx/services/common/types.go
```diff
@@ -34,6 +34,9 @@ const (
 	TagScopeNCPVNETInterface        string = "ncp/vnet_interface"
 	TagScopeVPCCRName               string = "nsx-op/vpc_cr_name"
 	TagScopeVPCCRUID                string = "nsx-op/vpc_cr_uid"
+	TagValueGroupScope              string = "scope"
+	TagValueGroupSrc                string = "source"
+	TagValueGroupDst                string = "destination"
```
heypnus, line 39: I prefer to `TagValueGroupSource` and `TagValueGroupDestination`.

pkg/nsx/services/securitypolicy/builder.go
timdengyun: Thanks, after refactoring, we removed parameter: obj *v1alpha1.SecurityPolicy

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -564,114 +590,100 @@ func (service *SecurityPolicyService) buildRuleAppliedGroupByRule(obj *v1alpha1.
...
 
-func (service *SecurityPolicyService) buildRuleSrcGroup(obj *v1alpha1.SecurityPolicy, rule *v1alpha1.SecurityPolicyRule, idx int) (*model.Group, string, error) {
-	var ruleSrcGroupName string
-	sources := rule.Sources
-	ruleSrcGroupID := fmt.Sprintf("sp_%s_%d_src", obj.UID, idx)
-	if len(rule.Name) > 0 {
-		ruleSrcGroupName = fmt.Sprintf("%s-src", rule.Name)
+func (service *SecurityPolicyService) buildRulePeerGroupID(obj *v1alpha1.SecurityPolicy, ruleIdx int, isSource bool) string {
+	if isSource == true {
+		return fmt.Sprintf("sp_%s_%d_src", obj.UID, ruleIdx)
```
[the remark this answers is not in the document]
timdengyun: For this patch, we don't want to change groupID with hash value. 
Moreover, we need to generate group peer contents hash if we want to add hash value for groupID, instead of passing rule obj hash.

pkg/nsx/services/securitypolicy/builder.go
timdengyun: Same as above.

pkg/nsx/services/securitypolicy/builder.go
timdengyun: Thanks Qian. That's a good idea.  I refactored the code to unify the buildAppliedGroupID, buildAppliedGroupName and buildAppliedGroupPath as you suggested.

pkg/nsx/services/common/types.go
timdengyun: Fixed.

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -564,114 +595,100 @@ func (service *SecurityPolicyService) buildRuleAppliedGroupByRule(obj *v1alpha1.
 	return &ruleAppliedGroup, ruleAppliedGroupPath, nil
 }
 
-func (service *SecurityPolicyService) buildRuleSrcGroup(obj *v1alpha1.SecurityPolicy, rule *v1alpha1.SecurityPolicyRule, idx int) (*model.Group, string, error) {
-	var ruleSrcGroupName string
-	sources := rule.Sources
-	ruleSrcGroupID := fmt.Sprintf("sp_%s_%d_src", obj.UID, idx)
-	if len(rule.Name) > 0 {
-		ruleSrcGroupName = fmt.Sprintf("%s-src", rule.Name)
+func (service *SecurityPolicyService) buildRulePeerGroupID(obj *v1alpha1.SecurityPolicy, ruleIdx int, isSource bool) string {
```
timdengyun, line 598: @heypnus  I also refactored the code to unify the buildRulePeerGroupID, buildRulePeerGroupName and buildRulePeerGroupPath as you suggested.

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -309,12 +310,34 @@ func (service *SecurityPolicyService) buildExpressionsMatchExpression(matchExpre
...
+	if groupID == "" {
+		groupID = service.buildAppliedGroupID(obj, -1)
+	}
+
+	return fmt.Sprintf("/infra/domains/%s/groups/%s", getDomain(service), groupID)
+}
+
+// build appliedTo group display name for both policy and rule levels.
+func (service *SecurityPolicyService) buildAppliedGroupName(obj *v1alpha1.SecurityPolicy, rule *v1alpha1.SecurityPolicyRule, ruleIdx int) string {
+	if ruleIdx != -1 && rule != nil {
```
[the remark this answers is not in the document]
timdengyun: We don't have case that there are ruleIdx != -1 && rule == nil.
We only have ruleIdx != -1 && rule != nil or ruleIdx == -1 && rule == nil.
Anyway, we removed the rule argument from this func.So, we don't need use this condition.
There is only place to use this one
https://github.com/vmware-tanzu/nsx-operator/pull/395/files#diff-878f5907665c91bbee664334052881facd7f44a8617ab92547f46a730c8a97d9R175

PR #397, opened 2023-11-21
Generate hash ID to make for security policy rule ID unique

timdengyun:
In certain case, SecurityPolicy cannot delete a rule when this rule is not last one.
the rule ID is sp_spUID_0, sp_spUID_1, and sp_spUID_2, respectively.

When user is trying to delete rule 2, the deleting will fail. This issue is caused by the original rule2 ID: sp_spUID_1_0_0 will be regenerated by the original rule3, because rule3 becomes the new rule2, the new rule2 id will be sp_spUID_2. NSX will complain that groups in the deleted rule2 are still being referenced by sp_spUID_2.

Hence, this patch is to
1.Generate hash ID for security policy rule content and append this hash ID to the final generated rule ID to make each rule's ID unique. So, the final generated rule ID is: sp_<spUID>_<rule hash ID>_ruleIndex.

The testing Done:
3. Creating SPs without this patch, and staring NSX Operator again with this patch to verify that new generated rules with unique rule ID are created successfully. This is to test upgrade existing SPs.

### DIFF
--- a/pkg/controllers/securitypolicy_controller_test.go
+++ b/pkg/controllers/securitypolicy_controller_test.go
@@ -199,7 +199,7 @@ func TestSecurityPolicyReconciler_Reconcile(t *testing.T) {
 	defer patches.Reset()
 
 	//  DeletionTimestamp.IsZero = false, Finalizers doesn't include util.FinalizerName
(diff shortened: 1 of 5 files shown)


timdengyun:
closing this PR and recreate a new one: https://github.com/vmware-tanzu/nsx-operator/pull/398

PR #398, opened 2023-11-21
Generate hash ID to make for security policy rule ID unique

timdengyun:
In certain case, SecurityPolicy cannot delete a rule when this rule is not the last one.
the rule ID is sp_spUID_0, sp_spUID_1, and sp_spUID_2, respectively.

When user is trying to delete rule 2, the deleting will fail. This issue is caused by the original rule2 ID: sp_spUID_1_0_0 will be regenerated by the original rule3, because rule3 becomes the new rule2, the new rule2 id will be sp_spUID_2. NSX will complain that groups in the deleted rule2 are still being referenced by sp_spUID_2.

Hence, this patch is to
1.Generate hash ID for security policy rule content and append this hash ID to the final generated rule ID to make each rule's ID unique. 
So, the final generated rule ID is: sp_spUID_rulehashID_ruleIndex.

The testing Done:
3. Creating SPs without this patch, and staring NSX Operator again with this patch to verify that new generated rules with unique rule ID are created successfully. This is to test upgrade existing SPs.

### DIFF
--- a/pkg/nsx/services/builder.go
+++ b/pkg/nsx/services/builder.go
@@ -9,7 +9,7 @@ import (
 
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
-	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
 	"github.com/vmware-tanzu/nsx-operator/pkg/util"
@@ -28,10 +28,11 @@ const (
 )
 
 func (service *SecurityPolicyService) buildSecurityPolicy(obj *v1alpha1.SecurityPolicy) (*model.SecurityPolicy,
-	*[]model.Group, error) {
+	*[]model.Group, error,
+) {
 	var nsxRules []model.Rule
 	var nsxGroups []model.Group
-	log.V(1).Info("building the model SecurityPolicy from CR SecurityPolicy", "object", *obj)
+	log.Info("building the model SecurityPolicy from CR SecurityPolicy", "object", *obj)
 	nsxSecurityPolicy := &model.SecurityPolicy{}
 
 	policyID := fmt.Sprintf("sp_%s", obj.UID)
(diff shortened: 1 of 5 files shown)


pkg/nsx/services/builder.go
```diff
@@ -294,21 +307,46 @@ func (service *SecurityPolicyService) buildExpressionsMatchExpression(matchExpre
...
+}
+
+// build appliedTo group path for both policy and rule levels.
+func (service *SecurityPolicyService) buildAppliedGroupPath(obj *v1alpha1.SecurityPolicy, ruleIdx int) string {
+	var groupID string
+	if ruleIdx == -1 {
+		groupID = service.buildAppliedGroupID(obj, -1)
+	} else {
+		groupID = service.buildAppliedGroupID(obj, ruleIdx)
+	}
```
heypnus, line 340: It looks that L335-L340 can be simplified to:
`groupID = service.buildAppliedGroupID(obj, ruleIdx)`

pkg/nsx/services/builder.go
```diff
@@ -374,33 +408,24 @@ func (service *SecurityPolicyService) buildRuleAndGroups(obj *v1alpha1.SecurityP
...
+	)
+	if err != nil {
+		return nil, nil, err
 	}
+	ruleGroups = append(ruleGroups, nsxRuleAppliedGroup)
 	nsxRule.Scope = []string{nsxRuleAppliedGroupPath}
 
-	log.V(1).Info("built rule and groups", "nsxRuleAppliedGroup", nsxRuleAppliedGroup, "nsxRuleSrcGroup", nsxRuleSrcGroup, "nsxRuleDstGroup", nsxRuleDstGroup, "action", *nsxRule.Action, "direction", *nsxRule.Direction)
+	log.V(1).Info("built rule and groups", "nsxRuleAppliedGroup", nsxRuleAppliedGroup,
+		"~", nsxRuleSrcGroup, "nsxRuleDstGroup", nsxRuleDstGroup,
```
heypnus, line 425: `"~"`?

pkg/nsx/services/builder.go
```diff
@@ -839,12 +845,13 @@ func (service *SecurityPolicyService) mergeSelectorMatchExpression(matchExpressi
...
-	var mexprInOpCount = 0
-	var mexprInValueCount = 0
+	matchLabels map[string]string,
+) (int, error) {
+	mexprInOpCount := 0
+	mexprInValueCount := 0
 	var err error = nil
-	var errorMsg = ""
-	var exists = false
+	errorMsg := ""
```
heypnus, line 853: Why we changed all `var errorMsg = ""` to `errorMsg := ""`?

pkg/nsx/services/builder.go
timdengyun: Thanks, good point. I've changed the code as advised.

pkg/nsx/services/builder.go
timdengyun: Thanks for bringing out the accident change in the previous patch.  I've changed the code to fix this error.

pkg/nsx/services/builder.go
timdengyun: This sort of changes were done by golangci-lint tool in vscode. I think the tool prefers this code style way.

heypnus (dismissed):

heypnus (approved):

PR #405, opened 2023-11-23
Support pass external http.Client to cleanup

TaoZou1:
Cleanup needs to support using external http.Client


TaoZou1:
closed it since the PR is on the vpc_dev branch.

PR #408, opened 2023-11-27
Re-implemented DLB cleanup in nsx-operator

zhengxiexie:
Delete DLB resources.
The size of load balancer service can be, SMALL, MEDIUM, LARGE, XLARGE, or DLB. The first four sizes are realized on Edge node as a centralized load balancer. DLB is realized on each ESXi hypervisor as a distributed load balancer. Previously, this cleanup function was implemented in NCP nsx_policy_cleanup.py. Now, it is re-implemented in nsx-operator pkg/clean/.


pkg/clean/clean_dlb.go
```diff
@@ -0,0 +1,147 @@
...
+	for _, item := range resp["results"].([]interface{}) {
+		if item.(mapInterface)["size"].(string) == "DLB" && checkTagsExist(item.(mapInterface)["tags"].([]interface{})) {
+			dlbServicesPath = append(dlbServicesPath, item.(mapInterface)["path"].(string))
+		}
+	}
+	return dlbServicesPath, nil
+}
+
+func httpGetVirtualServers(cf *config.NSXOperatorConfig, dlbServicesPath []string) ([]string, []string, error) {
+	url := "https://" + cf.NsxApiManagers[0] + ":443/policy/api/v1/infra/lb-virtual-servers/"
```
lxiaopei, line 105: can use cluster tag as query when getting lb-virtual-servers to only retrieve virtual servers for DLB.

pkg/clean/clean_dlb.go
```diff
@@ -0,0 +1,147 @@
...
+		return err
+	}
+
+	allPaths := append(dlbVirtualServersPath, dlbServicesPath...)
+	allPaths = append(allPaths, dlbPoolsPath...)
+	log.Info("Deleting DLB resources", "paths", allPaths)
+	for _, path := range allPaths {
+		url := "https://" + cf.NsxApiManagers[0] + ":443/policy/api/v1" + path
+		if err = httpDelete(cf, url); err != nil {
+			return err
```
lxiaopei, line 143: Besides DLB virtual servers, DLB service, we also need to delete DLB group,  persistence profile for DLB virtual server.

pkg/clean/clean_dlb.go
```diff
@@ -0,0 +1,147 @@
...
+		return err
+	}
+	dlbVirtualServersPath, dlbPoolsPath, err := httpGetVirtualServers(cf, dlbServicesPath)
+	if err != nil {
+		return err
+	}
+
+	allPaths := append(dlbVirtualServersPath, dlbServicesPath...)
+	allPaths = append(allPaths, dlbPoolsPath...)
+	log.Info("Deleting DLB resources", "paths", allPaths)
```
lxiaopei, line 139: there should be an order when deleting DLB resources. I think the order should be delete DLB group -> delete virtual servers -> DLB services -> DLB pools -> persistent profiles for DLB.

zhengxiexie: done

pkg/clean/clean_dlb.go
zhengxiexie: done

pkg/clean/clean_dlb.go
zhengxiexie: ok

pkg/clean/clean_dlb.go
```diff
@@ -0,0 +1,163 @@
...
+	for key, value := range params {
+		encodedKey := neturl.QueryEscape(key)
+		encodedValue := neturl.QueryEscape(value)
+		encodedPairs = append(encodedPairs, fmt.Sprintf("%s:%s", encodedKey, encodedValue))
+	}
+
+	encodedQuery := strings.Join(encodedPairs, "%20AND%20")
+	url := "https://" + cf.NsxApiManagers[0] + ":443/policy/api/v1/search/query?query=" + encodedQuery
+
+	resp, err := httpGet(cf, url)
```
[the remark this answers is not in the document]
zhengxiexie: Others resources use Hierarchical api, perhaps not adapt to here.
They are all encapsulated like 
`err = service.NSXClient.InfraClient.Patch(*infraIPPool, &EnforceRevisionCheckParam)`

TaoZou1, line 106: it's better that not create the httpClient, since wcp want to controll the number of http.Client. you could refer to cluster.GetVersion. (but using the ep.client instead of ep.noBalancerClient
The other problem here is that httpClient not support envoy, not support switch nsxt. if you using the ep.client, the transport will handle that.

pkg/clean/clean_dlb.go
```diff
@@ -0,0 +1,163 @@
...
+		resourcePath = appendIfNotExist(resourcePath, item.(mapInterface)["path"].(string))
+	}
+	return resourcePath, nil
+}
+
+func CleanDLB(cf *config.NSXOperatorConfig) error {
+	log.Info("Deleting DLB resources started")
+
+	dlbGroupsPath, err := httpQueryDLBResources(cf, "Group")
+	if err != nil {
```
TaoZou1, line 121: it seems that you could range []string("Group", "LBVirtualServer", "LBService", "LBPool", "LBCookiePersistenceProfile") since line120-153 has the same logic

TaoZou1 (approved):

PR #369, opened 2023-10-27
Add AVI allow rule when VPC created

TaoZou1:
Add AVI allow rule when VPC created so VIP of AVI lb could be accessed outside the VPC

PR #410, opened 2023-11-28
Add AVI allow rule when VPC created(#369)

TaoZou1:
cherry pick from  #369 
Add AVI allow rule when VPC created so VIP of AVI lb could be accessed outside the VPC
It will check nsxt version to enable it

Test Done:
Create lb service in one VPC, create pod in other VPC
1. Run the nsx-operator, check if the rule has been added
2. Visit lb service ip from other VPC pod
3. Restart the nsx-operator, check if the rule skip to update since CIDR is the same


timdengyun:
Please state which PR you picked from vpc_dev

timdengyun (approved):

PR #418, opened 2023-11-30
Support external http.Client

TaoZou1:
Cleanup needs to support using external http.Client While using external http.Client, it will disable keepAlive.

It also give a sample for how to create a http.Client for cleanup. The sample uses insecure mode to make sample simple. external http.Client should be response for authentication. It should also be in charge of log for debug

Test Done:
non-wcp env by running:
./bin/clean -cluster='k8scl-one'  -mgr-ip="nsxt-mgr-1.vmware.com"  -nsx-user="admin" -nsx-passwd='xxxx'

### DIFF
--- a/cmd_clean/main.go
+++ b/cmd_clean/main.go
@@ -4,8 +4,11 @@
 package main
 
 import (
+	"crypto/tls"
 	"flag"
+	"net/http"
 	"os"
+	"time"
 
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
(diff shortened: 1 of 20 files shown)


cmd_clean/main.go
```diff
@@ -60,9 +79,20 @@ func main() {
...
 	cf.CaFile = []string{caFile}
 	cf.Cluster = cluster
+
 	logf.SetLogger(logger.ZapLogger(cf))
 
-	err := clean.Clean(cf)
+	tr := &http.Transport{
+		IdleConnTimeout: 30 * time.Second,
+	}
+	tr.TLSClientConfig = &tls.Config{
```
zhengxiexie, line 88: Better comment to explain why use insecure mode.

zhengxiexie, line 88: why not put TLSClientConfig in line 86.

cmd_clean/main.go
```diff
@@ -33,6 +36,22 @@ var (
...
+func (t *Transport) base() http.RoundTripper {
+	if t.Base != nil {
+		return t.Base
+	}
+	return http.DefaultTransport
+}
+
 func main() {
 	flag.StringVar(&vcEndpoint, "vc-endpoint", "", "nsx manager ip")
 	flag.StringVar(&vcSsoDomain, "vc-sso-domain", "", "nsx manager ip")
```
zhengxiexie, line 57: Better to add a flag use-external-http and set default false, if false, use previous logic, else use httpclient, thus not affecting the existed logic.

cmd_clean/main.go
TaoZou1: done

zhengxiexie (approved):
LGTM

PR #422, opened 2023-12-01
Subnet e2e main

jwsui:


dantingl (commented):
Can we add another test case to create a specific SubnetSet CR

jwsui:
> Can we add another test case to create a specific SubnetSet CR

Sure, done.

dantingl (approved):

PR #423, opened 2023-12-03
Update VPCNetworkConfiguration CRD client

jwsui:
For cluster-wide resources, `// +genclient:nonNamespaced` is required to generate CRD client. Otherwise, it will be treated as namespace scope resource.


zhengxiexie (approved):
LGTM

PR #440, opened 2023-12-11
Bump Golang to 1.21.5.

jwsui:


pkg/third_party/retry/retry_test.go
```diff
@@ -159,8 +159,9 @@ func TestMaxDelay(t *testing.T) {
 	)
 	dur := time.Since(start)
 	assert.Error(t, err)
-	assert.True(t, dur > 170*time.Millisecond, "5 times with maximum delay retry is longer than 170ms")
-	assert.True(t, dur < 200*time.Millisecond, "5 times with maximum delay retry is shorter than 200ms")
+	fmt.Print(dur)
+	assert.True(t, dur > 50*time.Millisecond, "5 times with maximum delay retry is longer than 170ms", dur)
```
timdengyun, line 163: why we change 170*time.Millisecond to 50*time.Millisecond?

pkg/third_party/retry/retry_test.go
```diff
@@ -159,8 +159,9 @@ func TestMaxDelay(t *testing.T) {
...
+	assert.True(t, dur < 250*time.Millisecond, "5 times with maximum delay retry is shorter than 200ms", dur)
```
timdengyun, line 164: Same here

test/e2e/nsx_subnet_test.go
```diff
@@ -62,7 +62,7 @@ func TestDefaultSubnetSet(t *testing.T) {
 
 	portPath, _ := filepath.Abs("./manifest/testSubnet/subnetport_1.yaml")
 	err = applyYAML(portPath, E2ENamespace)
-	time.Sleep(10 * time.Second)
+	time.Sleep(30 * time.Second)
```
timdengyun, line 65: why we change 10 * time.Second to to 30 * time.Second?

jwsui: In Subnet e2e, we'd better to make this longer to ensure that NSX subnet is created and subnet info is updated to SubnetSet CR status. Otherwise, we may easily hit the subnet e2e failure.

pkg/third_party/retry/retry_test.go
jwsui: There are 5 times retry(without exponential backoff) with delay in range [10, 50], so that the expected should be in [10*5, 50*5].

pkg/third_party/retry/retry_test.go
jwsui: Replied above.

pkg/third_party/retry/retry_test.go
jwsui: align with upstream.

https://github.com/avast/retry-go/blob/master/retry_test.go#L264

TaoZou1 (approved):

PR #446, opened 2023-12-14
Increase backoff to wait for subnet realized

jwsui:
`retry.DefaultRetry` waits for 5 interations with time duration as 10ms. This is too short to wait for subnet being realized. Increase the backoff to fix this.


jwsui:
Close as this had been merged with another PR.

PR #450, opened 2023-12-18
Disable cgo to solve glibc dep issue

seanpang-vmware:


timdengyun:
Please squash two commits into one, Thanks

PR #451, opened 2023-12-18
Change go version in go.mod

lxiaopei:
To let other components easily depend on nsx-operator module, use go version 1.21 instead of 1.21.5 in go.mod.


seanpang-vmware (approved):

pkg/apis/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/vmware-tanzu/nsx-operator/pkg/apis
 
-go 1.21.5
```
jwsui, line 3: I think we need 1.21.5 for CVE fix.

lxiaopei: We have 1.21.5 in Dockerfile.

PR #453, opened 2023-12-18
Refactor VPC service

seanpang-vmware:


cmd/main.go
```diff
@@ -130,29 +130,26 @@ func StartIPPoolController(mgr ctrl.Manager, commonService common.Service) {
 	}
 }
 
-func StartVPCController(mgr ctrl.Manager, commonService common.Service) {
+func StartVPCController(mgr ctrl.Manager, commonService common.Service, vpcService *vpc.VPCService) {
```
jwsui, line 133: Do we still need common.Service?

zhengxiexie: +1, could remove it in both StartVPCController StartNamespaceController

pkg/nsx/services/common/types.go
```diff
@@ -156,3 +156,17 @@ type VPCResourceInfo struct {
 	ID       string
 	ParentID string
 }
+
+type VPCNetworkConfigInfo struct {
+	IsDefault               bool
```
TaoZou1, line 161: does VPCNetworkConfigInfo need to move here? or just keep it in vpc/types.go? if we won't hit the cycle refer issue

seanpang-vmware: I moved here per Yun's comment in another patch which is one vpc_dev. it makes sense, because this struct is used for both VPC service and NamespaceController, so it is a common resource.

jwsui (approved):
LGTM.

PR #461, opened 2023-12-21
Update vpc/tag/subnetset name

TaoZou1:
1. vpc name: <cluster>-<vpc>-<ns>-<vpc cr name> ->
<ns>-<vpc cr name>-<cluster>
2. tags: nsx-op/subnetset_cr_uid  -> nsx-op/subnetset_uid nsx-op/subnetset_cr_name -> nsx-op/subnetset_name
3. subnetset CR name: remove subnetset suffix, change to pod-default, vm-default


pkg/controllers/subnetset/vpc_handler.go
```diff
@@ -3,26 +3,24 @@ package subnetset
...
 )
 
 // VPCHandler handles VPC event for SubnetSet:
 // - VPC creation: create default SubnetSet for the VPC.
 // - VPC deletion: delete all SubnetSets under the VPC.
 
 var defaultSubnetSets = map[string]string{
-	"default-vm-subnetset":  common.LabelDefaultVMSubnetSet,
-	"default-pod-subnetset": common.LabelDefaultPodSubnetSet,
+	"vm-default":  common.LabelDefaultVMSubnetSet,
```
jwsui, line 22: To align with this, I think the name [here](https://github.com/vmware-tanzu/nsx-operator/blob/main/test/e2e/nsx_subnet_test.go#L16) also needs to be updated. So maybe it's better to move this to common.types like `common.XXXXXX`.

pkg/nsx/services/common/types.go
```diff
@@ -57,8 +57,8 @@ const (
 	TagScopeSubnetCRType            string = "nsx-op/subnet_cr_type"
```
jwsui, line 57: Should we rename all CR tags here? eg. `vpc_cr_name`/`ippool_cr_name`/...

pkg/controllers/subnetset/vpc_handler.go
TaoZou1: done

pkg/nsx/services/common/types.go
TaoZou1: done

jwsui (approved):
LGTM.

PR #463, opened 2023-12-22
Change pkg/apis and pkg/client tag in go.mod

lxiaopei:
In go.mod, change pkg/apis and pkg/client tag to v0.0.1.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/prometheus/client_golang v1.16.0
 	github.com/sirupsen/logrus v1.9.0
 	github.com/stretchr/testify v1.8.2
-	github.com/vmware-tanzu/nsx-operator/pkg/apis v1.0.0
+	github.com/vmware-tanzu/nsx-operator/pkg/apis v0.0.1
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -22,8 +22,8 @@ require (
 	github.com/prometheus/client_golang v1.16.0
 	github.com/sirupsen/logrus v1.9.0
 	github.com/stretchr/testify v1.8.2
-	github.com/vmware-tanzu/nsx-operator/pkg/apis v1.0.0
```
jwsui, line 25: I think the replace in line #5 and #6 could be removed.

wyike: +1

lxiaopei: for nsx-operator build, do you want to use the local file from pkg/apis and pkg/client folder or a specific version?

jwsui: v0.0.1 would be good enough.

lxiaopei: change to use the local files for pkg/apis and pkg/client instead of tags, to avoid changes when tags change.

PR #467, opened 2024-01-03
Use default tag in Networkconfig

seanpang-vmware:
Use default tag on networkconfig CRD to identify default config


pkg/controllers/vpc/vpc_utils.go
```diff
@@ -123,13 +125,29 @@ func nsxtProjectPathToId(path string) (string, string, error) {
...
+	annos := vpcConfigCR.GetAnnotations()
+	val, exist := annos[types.AnnotationDefaultNetworkConfig]
+	if exist {
+		boolVar, err := strconv.ParseBool(val)
+		if err != nil {
+			log.Error(err, "failed to parse annotation to check default NetworkConfig", "Annotation", annos[types.AnnotationDefaultNetworkConfig])
+			return false
+		}
+		return boolVar
+	} else {
```
lxiaopei, line 138: nit:  no need for this "else".

pkg/nsx/services/vpc/vpc.go
```diff
@@ -69,31 +59,40 @@ type AVIAllowRule struct {
 	PubIpblockStore     *PubIPblockStore
 }
 
+func (s *VPCService) GetDefaultNetworkConfig() (bool, *VPCNetworkConfigInfo) {
+	for _, v := range s.VPCNetworkConfigMap {
```
lxiaopei, line 63: in case of 200 VPCNetworkConfig CR, each time need to loop to find the default one, is it possible we use variable for the default one?

seanpang-vmware: @zhengxiexie  the default nc name is not always default, the purpose of this patch is to use annotations on nc cr to get the default nc name.
@lxiaopei , agree, that is my first thought. i will change it according to your suggestion.

lxiaopei (approved):
LGTM.

PR #468, opened 2024-01-03
Add make clean library in Dockerfile

zhengxiexie:
The background is we want to run clean process in e2e CI pipeline, so make the binary in Dockerfile and it would be triggered in the container of testbed when running e2e pipeline.


seanpang-vmware (approved):
LGTM

PR #473, opened 2024-01-08
Merge multiple rules into one rule if they have the same destinations

zhengxiexie:
When expand rules for named-port and non-named-port, we should merge the multiple service entries into one rule.
e.g.
          Src  Dst Services
ruleA: Any Any TCP (Source: Any | Destination: 80-8888) 
ruleB: Any Any TCP (Source: Any | Destination: 9999)
 ----->
          Src  Dst Services
rule:   Any Any TCP (Source: Any | Destination: 80-8888)
************** TCP (Source: Any | Destination: 9999)

However,
          Src  Dst Services
ruleA: Any IPSetGroup1 TCP (Source: Any | Destination: 80-8888) 
ruleB: Any IPSetGroup2 TCP (Source: Any | Destination: 9999) 
Can't merge since the Destinations are different.

### DIFF
--- a/pkg/nsx/services/securitypolicy/expand.go
+++ b/pkg/nsx/services/securitypolicy/expand.go
@@ -41,6 +41,27 @@ func (service *SecurityPolicyService) expandRule(obj *v1alpha1.SecurityPolicy, r
 		nsxGroups = append(nsxGroups, nsxGroups2...)
 		nsxRules = append(nsxRules, nsxRules2...)
 	}
+
+	// Merge multiple rules into one rule if they have the same destinations,
+	// let the rule's services be the union of all rules' services.
+	// Implement merge rules here, using hash
+	var mapRules = make(map[string]*model.Rule)
+	for _, rule := range nsxRules {
+		destination := ""
+		if len(rule.DestinationGroups) > 0 {
+			destination = rule.DestinationGroups[0]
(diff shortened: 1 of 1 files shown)


zhengxiexie:
Tests Done
1. non-named-port security policy
apiVersion: nsx.vmware.com/v1alpha1
kind: SecurityPolicy
metadata:
  name: named-port-policy
  namespace: web
spec:
  priority: 10
  appliedTo:
    - podSelector:
        matchLabels:
          role: web
  rules:
    - direction: in
      action: allow
      ports:
        - protocol: TCP
          port: 80
          endPort: 8888
        - protocol: TCP
          port: 9999
image
2. named-port policy with different destinations
test/e2e/testSecurityPolicy/rule-out-rule-applied-to-with-dst-with-dup-port-multi.yaml
image
2. named-port policy with same destinations
test/e2e/testSecurityPolicy/rule-out-rule-applied-to-with-dst-with-dup-port.yaml
image

timdengyun:
For this case, named-port policy with same destinations test/e2e/testSecurityPolicy/rule-out-rule-applied-to-with-dst-with-dup-port.yaml 
There is only one port number after parsing named port, so there is only one ip set group generated.

timdengyun (approved):

PR #477, opened 2024-01-11
Set staticIPAllocation enable true for e2e subnetset test yaml

zhengxiexie:
When realize subnet port, if enable is not true, nsx would report error LogicalSwitch 1c104366-7266-4035-856c-27a0e10500ce does not have IpPoolId needed by LogicalPort 'StaticIpAllocationDto{enabled='false'}'}'


jwsui (approved):
LGTM.

PR #478, opened 2024-01-12
Bump golang 1.21 for v3.2.1 to avoid CVEs

zhengxiexie:


build/image/photon/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.19.10 as golang-build
+FROM golang:1.21.5 as golang-build
```
timdengyun, line 1: After upgrade golang to 1.21, you should be care is there is glibc issue.
https://github.com/vmware-tanzu/nsx-operator/pull/448
if yes, you should disable CGO.

timdengyun:
https://github.com/vmware-tanzu/nsx-operator/blob/v3.2.1/.github/workflows/makefile.yml#L5
please also change github action from main branch to v3.2.1, so that we can see every golang update could make unit test run successfully.

build/image/photon/Dockerfile
zhengxiexie: ok

zhengxiexie:
> In addition, we should change https://github.com/vmware-tanzu/nsx-operator/blob/v3.2.1/.github/workflows/makefile.yml#L5 github action from main branch to v3.2.1, so that we can see every golang update could make unit test run successfully.

ok

timdengyun (approved):

PR #479, opened 2024-01-15
Subnet service refactor

jwsui:
Initialize subnet service once then inject it to multiple reconcilers instead of initializing service in different reconcilers with a lock ensuring the initialization is invoked once.

https://github.com/tnqn/nsx-operator/pull/1#discussion_r1293271402


seanpang-vmware (approved):
LGTM

PR #483, opened 2024-01-16
Refactor e2e test

jwsui:
1. Use camel case instead of snake case to align with nsx-operator code style.
2. Use t.Cleanup to register clean functions to ensure that cleanup can be performed when child test is executed in another goroutine.


zhengxiexie:
It's better to remove time.sleep(), and replace with waitForCRReadyOrDeleted etc., and TestSubnetSet is also redundant.

jwsui:
> It's better to remove time.sleep(), and replace with waitForCRReadyOrDeleted etc., and TestSubnetSet is also redundant.

I‘ve replaced sleep with waitForCRReady. But there are still some time.sleep left as they are needed for nsx to update tags.  TestSubnetSet is not redundant, child-case is used in this case to resuse setup/teardown for all cases.

zhengxiexie (approved):
LGTM

PR #485, opened 2024-01-18
Revert "Support external http.Client"

TaoZou1:
This reverts commit 2d7f460c9124d79558e035a885fcfde3915fc03a.
Remove http client from external
Using external http client will make it difficult to switch NSXT.
After discussed with wcp team, they agreed to use http client created by nsx-operator


zhengxiexie (approved):
LGTM

PR #487, opened 2024-01-18
Update log variant

TaoZou1:
Update log to 'log = logger.Log'
change  ZapLogger interface to ZapLogger(cfDebug bool, cfLogLevel int) to avoid import cycle


zhengxiexie (approved):
LGTM

PR #489, opened 2024-01-19
Support timeout in Context in clean.go

zhengxiexie:
Extend it to support passing in a context so the caller themselves can define a timeout.

### DIFF
--- a/cmd_clean/main.go
+++ b/cmd_clean/main.go
@@ -4,8 +4,10 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"os"
+	"time"
 
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
@@ -77,10 +79,12 @@ func main() {
 	cf.EnvoyPort = envoyPort
 
 	logf.SetLogger(logger.ZapLogger(cf.DefaultConfig.Debug, config.LogLevel))
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*5)
+	defer cancel()
 
-	err := clean.Clean(cf)
+	err := clean.Clean(ctx, cf)
 	if err != nil {
-		log.Error(err, "failed to clean nsx resources", "status", err.Error())
+		log.Error(err, "failed to clean nsx resources")
 		os.Exit(1)
 	}
 	os.Exit(0)
--- a/pkg/clean/clean.go
+++ b/pkg/clean/clean.go
@@ -4,6 +4,7 @@
 package clean
 
 import (
+	"context"
 	"errors"
 	"fmt"
 
@@ -19,6 +20,7 @@ import (
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/subnet"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/subnetport"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/vpc"
+	nsxutil "github.com/vmware-tanzu/nsx-operator/pkg/nsx/util"
 )
 
 var log = logger.Log
@@ -32,39 +34,45 @@ var log = logger.Log
 // GetNSXClientFailed  			indicate that could not retrieve nsx client to perform cleanup operation
 // InitCleanupServiceFailed 	indicate that error happened when trying to initialize cleanup service
 // CleanupResourceFailed    	indicate that the cleanup operation failed at some services, the detailed will in the service logs
-func Clean(cf *config.NSXOperatorConfig) error {
+func Clean(ctx context.Context, cf *config.NSXOperatorConfig) error {
 	log.Info("starting NSX cleanup")
 	if err := cf.ValidateConfigFromCmd(); err != nil {
-		return errors.Join(ValidationFailed, err)
+		return errors.Join(nsxutil.ValidationFailed, err)
 	}
 	nsxClient := nsx.GetClient(cf)
 	if nsxClient == nil {
-		return GetNSXClientFailed
+		return nsxutil.GetNSXClientFailed
 	}
 	if cleanupService, err := InitializeCleanupService(cf); err != nil {
-		return errors.Join(InitCleanupServiceFailed, err)
+		return errors.Join(nsxutil.InitCleanupServiceFailed, err)
 	} else if cleanupService.err != nil {
-		return errors.Join(InitCleanupServiceFailed, cleanupService.err)
+		return errors.Join(nsxutil.InitCleanupServiceFailed, cleanupService.err)
 	} else {
 		for _, clean := range cleanupService.cleans {
-			if err := retry.OnError(retry.DefaultRetry, func(err error) bool {
-				if err != nil {
-					log.Info("retrying to clean up NSX resources", "error", err)
-					return true
-				}
-				return false
-			}, func() error {
-				if err := clean.Cleanup(); err != nil {
-					return err
-				}
-				return nil
-			}); err != nil {
-				return errors.Join(CleanupResourceFailed, err)
+			if err := retry.OnError(retry.DefaultRetry, retriable, wrapCleanFunc(ctx, clean)); err != nil {
+				return errors.Join(nsxutil.CleanupResourceFailed, err)
 			}
 		}
 	}
 	log.Info("cleanup NSX resources successfully")
 	return nil
+}
+
+func retriable(err error) bool {
+	if err != nil && !errors.As(err, &nsxutil.TimeoutFailed) {
+		log.Info("retrying to clean up NSX resources", "error", err)
+		return true
+	}
+	return false
+}
+
+func wrapCleanFunc(ctx context.Context, clean cleanup) func() error {
+	return func() error {
+		if err := clean.Cleanup(ctx); err != nil {
+			return err
+		}
+		return nil
+	}
 }
 
 // InitializeCleanupService initializes all the CR services
--- a/pkg/clean/types.go
+++ b/pkg/clean/types.go
@@ -1,7 +1,9 @@
 package clean
 
+import "context"
+
 type cleanup interface {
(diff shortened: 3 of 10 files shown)


cmd_clean/main.go
```diff
@@ -62,8 +64,20 @@ func main() {
 	cf.Cluster = cluster
 
 	logf.SetLogger(logger.ZapLogger(cf.DefaultConfig.Debug, config.LogLevel))
+	// just a demo to show how to use customer http client
```
ihgann, line 67: Is this a demo, or can it actually be run this way?

If just a demo, maybe this should be moved to a `demo` package or something so user's would not intentionally run this as a script/binary?

pkg/clean/clean.go
```diff
@@ -27,39 +28,44 @@ var log = logger.Log
...
-			if err := retry.OnError(retry.DefaultRetry, func(err error) bool {
-				if err != nil {
-					log.Info("retrying to clean up NSX resources", "error", err)
-					return true
-				}
-				return false
-			}, func() error {
-				if err := clean.Cleanup(); err != nil {
-					return err
+	select {
```
ihgann, line 33: This isn't really doing anything, because it will only check on invocation of this function. We want the logic actually passed down the stack. Essentially, `clean.Cleanup()` should accept a `ctx` and further down the line.

pkg/clean/types.go
```diff
@@ -41,4 +41,5 @@ var (
 	GetNSXClientFailed       = Status{Code: 2, Message: "failed to get nsx client"}
 	InitCleanupServiceFailed = Status{Code: 3, Message: "failed to initialize cleanup service"}
 	CleanupResourceFailed    = Status{Code: 4, Message: "failed to clean up"}
+	TimeoutFailed            = Status{Code: 5, Message: "failed because of timeout"}
```
ihgann, line 44: Slightly related - why not have `Status` implement the `error` interface? For consumers that want to use `error` handling, they can do so easily, while your `main.go` package can still interpret the error as a `Status`. e.g.

```
func Clean() error  {
    return TimeoutFailed
}
```

in `main.go`, it could then do something like:

```
err := clean.Clean()

var statusErr Status
if errors.As(err, &statusErr) {
    return statusErr.Code, Message, // etc.
}

// Otherwise, it's some generic standard processing error
```

Consumer's can then cast this way to make localizable errors too if they need to and it reduces the need for extra return values (which I would argue are confusing since the error responses are essentially duplicated).

ihgann (changes requested):
Also, please include unit tests!

cmd_clean/main.go
zhengxiexie: It actually be run this way, removed the `demo` comment

zhengxiexie:
> Also, please include unit tests!

OK, unit tests would be added later.

pkg/clean/clean.go
zhengxiexie: Actually, you want to support timeout in context, it can satisfy your need https://github.com/vmware-tanzu/nsx-operator/pull/489/files#diff-be6b68a2cf4b036867ce2ddc12ed2a9c862555110e0265a06b7d42ced61ec345R77, we think there is no need to accept a `ctx` in `clean.Cleanup()`, `clean.Cleanup() ` is our inner logic, you don't need to concern it.

pkg/clean/types.go
zhengxiexie: @seanpang-vmware What's your opinion？

seanpang-vmware: I'm OK with it.

pkg/clean/clean.go
zhengxiexie: ok

pkg/clean/types.go
zhengxiexie: @ihgann pls help review https://github.com/vmware-tanzu/nsx-operator/pull/497, and I'll rebase it to continue modifying `context`

pkg/nsx/services/subnet/subnet.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/realizestate"
+	util2 "github.com/vmware-tanzu/nsx-operator/pkg/nsx/util"
```
ihgann, line 21: tiny nit: Maybe a different import name?

pkg/nsx/services/staticroute/staticroute.go
```diff
@@ -146,16 +148,21 @@ func (service *StaticRouteService) ListStaticRoute() []*model.StaticRoutes {
...
 	for _, staticRoute := range staticRouteSet {
 		path := strings.Split(*staticRoute.Path, "/")
 		log.Info("removing staticroute", "staticroute path", *staticRoute.Path)
-		err := service.DeleteStaticRouteByPath(path[2], path[4], path[6], *staticRoute.Id)
-		if err != nil {
-			log.Error(err, "remove staticroute failed", "staticroute id", *staticRoute.Id)
-			return err
+		select {
+		case <-ctx.Done():
+			return nsxutil.TimeoutFailed
```
ihgann, line 159: To be robust, we can probably enhance this with even stronger error types - e.g. `TimeoutDeletingStaticRoute`, but I don't know how valuable going this granular is. Not a blocker, for sure, just a thought.

pkg/clean/clean.go
```diff
@@ -33,41 +35,47 @@ var log = logger.Log
...
 	if nsxClient == nil {
-		return GetNSXClientFailed
+		return nsxutil.GetNSXClientFailed
 	}
 	if cleanupService, err := InitializeCleanupService(cf); err != nil {
-		return errors.Join(InitCleanupServiceFailed, err)
+		return errors.Join(nsxutil.InitCleanupServiceFailed, err)
 	} else if cleanupService.err != nil {
-		return errors.Join(InitCleanupServiceFailed, cleanupService.err)
+		return errors.Join(nsxutil.InitCleanupServiceFailed, cleanupService.err)
```
ihgann, line 50: Not your change: Why does `InitializeCleanupService` both return an error, but also have an error attached to `cleanupService` object? This is an anti-pattern - would be better to just return the error in the one spot. Maybe a follow-up change can address this?

pkg/nsx/services/vpc/vpc.go
```diff
@@ -549,21 +550,31 @@ func (s *VPCService) CreateorUpdateVPC(obj *v1alpha1.VPC) (*model.Vpc, *common.V
...
 
 	ipblocks := s.IpblockStore.List()
 	log.Info("cleaning up ipblocks", "Count", len(ipblocks))
 	for _, ipblock := range ipblocks {
 		ipb := ipblock.(*model.IpAddressBlock)
-		if err := s.deleteIPBlock(*ipb.Path); err != nil {
-			return err
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("failed to clean up ippool, context cancelled")
```
ihgann, line 573: nit: Any reason this one doesn't return TimeoutFailed?

pkg/nsx/services/vpc/vpc.go
```diff
@@ -549,21 +550,31 @@ func (s *VPCService) CreateorUpdateVPC(obj *v1alpha1.VPC) (*model.Vpc, *common.V
...
-func (s *VPCService) Cleanup() error {
+func (s *VPCService) Cleanup(ctx context.Context) error {
 	vpcs := s.ListVPC()
 	log.Info("cleaning up vpcs", "Count", len(vpcs))
 	for _, vpc := range vpcs {
-		if err := s.DeleteVPC(*vpc.Path); err != nil {
-			return err
+		select {
+		case <-ctx.Done():
+			return util2.TimeoutFailed
```
ihgann, line 559: For all instances like this - can you actually at the very least wrap `ctx.Err()`? It will allow clients to check if the errors is [`context.DeadlineExceeded`](https://pkg.go.dev/context#pkg-variables).

Keeping your own error is fine too - but maybe something like:

```
return errors.Join(nsxutil.TimeoutFailed, ctx.Err())
```

?

ihgann (changes requested):
Overall, LGTM. A few comments.

pkg/nsx/services/securitypolicy/firewall.go
```diff
@@ -552,14 +554,19 @@ func (service *SecurityPolicyService) ListSecurityPolicyID() sets.Set[string] {
...
 	// Delete all the security policies in store
 	uids := service.ListSecurityPolicyID()
 	log.Info("cleaning up security policies", "count", len(uids))
 	for uid := range uids {
-		err := service.DeleteSecurityPolicy(types.UID(uid), true)
-		if err != nil {
-			return err
+		select {
+		case <-ctx.Done():
+			return nsxutil.TimeoutFailed
```
timdengyun, line 564: SP parts looks ok!

timdengyun:
since the PR  https://github.com/vmware-tanzu/nsx-operator/pull/392 merged, please rebase and fix conflicts

pkg/nsx/services/subnet/subnet.go
zhengxiexie: sure

pkg/nsx/services/vpc/vpc.go
zhengxiexie: fixed

pkg/nsx/services/vpc/vpc.go
zhengxiexie: OK

timdengyun (approved):

ihgann:
I don't have approval permissions, but this looks good to me.

PR #490, opened 2024-01-19
Support podSelector namespaceSelector conversion for SP in VPC mode

timdengyun:
This patch is to:
1. Support changing SecurityPolicy Rule peer(source/destination) group
from podSelector to namespaceSelector in VPC mode, and vice versa.
2. Remove duplicated project shares and project groups when expanding rules.
3. Fix the bug that rule store is not cleaned in security policy GC or
cleanup process.
4. Split project group store from group store which is for storing groups
in VPC level.
5. Wrap project groups and shares with SecurityPolicy, groups, rules in
VPC level into one hierarchy resource tree for OrgRootClient API call.
So, there is only one time API call for SecurityPolicy
create/update/delete.
6. Enhance log output to use one word for log info as the key index.

Test Done:
1. Create a SecurityPolicy, and then apply a change from podSelector to namespaceSelector in VPC mode.
2. Create a SecurityPolicy, and then apply a change from namespaceSelector to podSelector in VPC mode.
3. Create a SecurityPolicy, and then delete a rule with namespaceSelector.
4. Create a SecurityPolicy, and then delete a port entry from the rule with namespaceSelector.
5. Create a SecurityPolicy, and  then delete a rule with namespaceSelector and add a rule with namespaceSelector
at the same time.
7.  Create a SecurityPolicy with namespaceSelector and podSelector in non-VPC mode with v4.1.2 branch, 
and then restart NSX operator with patch change to see if operator upgrade from v4.1.2 to current 
patch version is ok. 
8. Create/update/delete a SecurityPolicy with namedport.
9. Create/update/delete a NetworkPolicy with namedport.
8. GC in VPC mode.
10. GC in non-VPC mode.


pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -784,7 +790,7 @@ func (service *SecurityPolicyService) buildRulePeerGroup(obj *v1alpha1.SecurityP
 		}
 	}
 	log.V(2).Info(fmt.Sprintf("build rule %s group criteria", ruleDirection),
-		"total criteria", rulePeerGroupCriteriaCount, "total expressions of criteria", rulePeerGroupTotalExprCount)
+		"TotalCriteria", rulePeerGroupCriteriaCount, "totalExprsOfCriteria", rulePeerGroupTotalExprCount)
```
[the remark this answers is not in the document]
timdengyun: Thanks for correction, fixed

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -816,18 +822,17 @@ func (service *SecurityPolicyService) buildRulePeerGroup(obj *v1alpha1.SecurityP
...
-			log.Error(err, "failed to build SharedWith path", "rule group Name", rulePeerGroupName)
+			log.Error(err, "failed to build SharedWith path", "ruleGroupName", rulePeerGroupName)
 			return nil, "", nil, err
 		}
 		// Build a nsx share resource in project level
-		share, err := service.buildProjectShare(obj, &rulePeerGroup, []string{rulePeerGroupPath}, *sharedWith, createdFor)
+		nsxShare, err := service.buildProjectShare(obj, &rulePeerGroup, []string{rulePeerGroupPath}, *sharedWith, createdFor)
 		if err != nil {
-			log.Error(err, "failed to build project share", "rule group Name", rulePeerGroupName)
+			log.Error(err, "failed to build project nsx share", "ruleGroupName", rulePeerGroupName)
```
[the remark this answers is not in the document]
timdengyun: Thanks for correction, fixed

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -76,38 +68,57 @@ func (service *SecurityPolicyService) buildSecurityPolicy(obj *v1alpha1.Security
...
 	nsxSecurityPolicy.Rules = nsxRules
 	nsxSecurityPolicy.Tags = service.buildBasicTags(obj, createdFor)
-	log.V(1).Info("built nsxSecurityPolicy", "nsxSecurityPolicy", nsxSecurityPolicy, "nsxGroups", nsxGroups)
-	return nsxSecurityPolicy, &nsxGroups, &projectShares, nil
+	// nsxRules info are included in nsxSecurityPolicy obj
+	log.Info("built nsxSecurityPolicy", "nsxSecurityPolicy", nsxSecurityPolicy, "nsxGroups", nsxGroups)
+
+	for i := len(nsxProjectShares) - 1; i >= 0; i-- {
+		log.Info("built nsxSecurityPolicy", "projectGroups", *(nsxProjectShares[i].shareGroup),
+			"projectShares", *(nsxProjectShares[i].share))
```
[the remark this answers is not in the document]
timdengyun: I tried to directly print nsxProjectShares for project group and project share info, however, it print nothing.So I have to use loop iteration to print one by one. Is there some idea to print in one line?

pkg/nsx/services/securitypolicy/firewall.go
```diff
@@ -96,6 +99,13 @@ func InitializeSecurityPolicy(service common.Service, vpcService common.VPCServi
 		}),
 		BindingType: model.RuleBindingType(),
 	}}
+
+	securityPolicyService.projectGroupStore = &GroupStore{ResourceStore: common.ResourceStore{
```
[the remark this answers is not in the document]
timdengyun: I introduce one more group store specific for project group is to simplify the logic and save running time.Space consumption should be the same for using one store and two stores. 
If without project store, in the logic blow:
https://github.com/vmware-tanzu/nsx-operator/pull/490/files/a2d5bf2cfe22b4cc87339447bcf8a2eacd1d839a#diff-53bc6d3b74f0d34b4187204c1bcf7ce4afe93c7f73e60b5de43c40dc44b88af0R417

and in the firstly, we need to build new project groups with buildSecurityPolicy
nsxSecurityPolicy, nsxGroups, nsxProjectShares, err := service.buildSecurityPolicy(obj, createdFor).

and then,  we have to find all existing project groups from group store using groupShared tag, like below loop:

for i := len(groups) - 1; i >= 0; i-- {
	for j := len(groups[i].Tags) - 1; j >= 0; j-- {
			if *(groups[i].Tags[j].Scope) == common.TagScopeProjectGroupShared {
				if *(groups[i].Tags[j].Tag) == "true" {
					nsxProjectGroups = append(nsxProjectGroups, *groups[i])
                               }
                      }
           }
}

finally, we can compare these project groups if there is any change:
changed, stale := common.CompareResources(GroupsPtrToComparable(existingProjectGroups), GroupsToComparable(projectGroups))

This project is taking more time and more complicated, that is why I add one more project group store. Of course, bringing one more store also brings more store management code, but, it helps to simply the project group update logic because we need to handle project group in the  API other than the groups saved in VPC level.

pkg/nsx/services/securitypolicy/firewall.go
```diff
@@ -353,52 +381,97 @@ func (service *SecurityPolicyService) createOrUpdateSecurityPolicy(obj *v1alpha1
...
+		changedProjectGroups, staleProjectGroups := ComparableToGroups(changed), ComparableToGroups(stale)
+		if len(changedProjectGroups) == 0 && len(staleProjectGroups) == 0 {
+			log.Info("project groups are not changed, skip updating them")
+		}
+
+		// 1.Remove stale nsxRules
+		if len(staleProjectGroups) != 0 {
+			// Before deleting the stale project groups and project shares, we must delete the nsxRules to which the project shares are referencing
+			// The deleted stale nsxRules might also include other stale rules to which the project shares are not referencing,
+			// However, we can delete these rules together because the stale rule deletion process is an intermediate step that causes temporary inconsistence
```
heypnus, line 428: `inconsistency`

pkg/nsx/services/securitypolicy/firewall.go
```diff
@@ -353,52 +381,97 @@ func (service *SecurityPolicyService) createOrUpdateSecurityPolicy(obj *v1alpha1
...
+		changed, stale := common.CompareResources(GroupsPtrToComparable(existingProjectGroups), GroupsToComparable(projectGroups))
+		changedProjectGroups, staleProjectGroups := ComparableToGroups(changed), ComparableToGroups(stale)
+		if len(changedProjectGroups) == 0 && len(staleProjectGroups) == 0 {
+			log.Info("project groups are not changed, skip updating them")
+		}
+
+		// 1.Remove stale nsxRules
+		if len(staleProjectGroups) != 0 {
+			// Before deleting the stale project groups and project shares, we must delete the nsxRules to which the project shares are referencing
+			// The deleted stale nsxRules might also include other stale rules to which the project shares are not referencing,
```
heypnus, line 427: You mean, we will delete ALL stale rules in this step, no matter whether it's linked to any project share?

pkg/nsx/services/securitypolicy/firewall.go
```diff
@@ -353,52 +381,97 @@ func (service *SecurityPolicyService) createOrUpdateSecurityPolicy(obj *v1alpha1
...
+		}
+		for i := len(staleProjectShares) - 1; i >= 0; i-- {
+			staleProjectShares[i].MarkedForDelete = &MarkedForDelete
+		}
+		finalProjectShares = append(finalProjectShares, staleProjectShares...)
+		finalProjectShares = append(finalProjectShares, changedProjectShares...)
+
+		// 2.Create/update project groups and shares
+		if len(finalProjectGroups) != 0 || len(finalProjectShares) != 0 {
+			projectInfra, err := service.WrapHierarchyProjectResources(finalProjectShares, finalProjectGroups)
```
heypnus, line 464: Can we wrap securitypolicy & projectshares & project groups into one object?

pkg/nsx/services/securitypolicy/firewall.go
timdengyun: Thanks for corrected.

pkg/nsx/services/securitypolicy/firewall.go
timdengyun: yes, it is. I think the final result is to delete all stale rules, so, it doesn't matter we delete them in the beginning or in the middle or at last.  But for the stale rules specific for group share, we must delete them before deleting project share and project group.

timdengyun: Of course, we can also identify the exact stale rules for deleted project share, if that, it needs more computation, we need use buildRuleID tag in the stale project group to compare each stale rule list to selected the final stale project rules.

pkg/nsx/services/securitypolicy/firewall.go
timdengyun: fixed

pkg/nsx/services/securitypolicy/firewall.go
timdengyun: Unfortunately, we can not warp securitypolicy & project shares & project groups into obj, because securitypolicy is put in VPC level, so we need to use OrgRootClient to wrap securitypolicy, rules and groups of VPC level into one API Obj. As for project shares & project groups, which are put in project level, we use ProjectInfraClient to wrap these two into one API obj.

pkg/nsx/services/securitypolicy/firewall.go
timdengyun: we have merged rules in VPC level and project level shares,groups together into one H-API call, so, there is no need to use this intermediate step any more.

pkg/nsx/services/securitypolicy/firewall.go
timdengyun: Much appreciation for bringing up this critical point for discussion. Ultimately, we have find a solution that we can wrap ecuritypolicy, rules, groups in VPC level & projectshares & project groups into one Hierarchy tree object. This approach  help us to manipulate all sort of objects in one API call, which absolutely helps the code quality in term of performance and security.  The only words I want to say, Thank you.

heypnus (approved):

PR #492, opened 2024-01-19
Fix the pointer error during comparing the security policies

heypnus:


TaoZou1 (approved):

timdengyun (approved):
I found the same issue, and address it in my patch
https://github.com/vmware-tanzu/nsx-operator/pull/490
Thanks for quick fix in a small patch.

PR #495, opened 2024-01-22
Fix bug not using pointer

TaoZou1:
The object return by store list is pointer instead of instance. Update ListStaticRoute to return array of pointer

### DIFF
--- a/pkg/nsx/services/staticroute/staticroute.go
+++ b/pkg/nsx/services/staticroute/staticroute.go
@@ -137,11 +137,11 @@ func (service *StaticRouteService) DeleteStaticRoute(namespace string, uid strin
 	return service.DeleteStaticRouteByPath(path[2], path[4], *vpc[0].Id, uid)
 }
 
-func (service *StaticRouteService) ListStaticRoute() []model.StaticRoutes {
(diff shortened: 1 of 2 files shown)


pkg/nsx/services/staticroute/staticroute.go
```diff
@@ -137,11 +137,11 @@ func (service *StaticRouteService) DeleteStaticRoute(namespace string, uid strin
 	return service.DeleteStaticRouteByPath(path[2], path[4], *vpc[0].Id, uid)
 }
 
-func (service *StaticRouteService) ListStaticRoute() []model.StaticRoutes {
+func (service *StaticRouteService) ListStaticRoute() []*model.StaticRoutes {
```
timdengyun, line 140: For store, we use ListIndexFuncValues instead of list() function from strore, seemly this way is better. Of course, using list() also might be ok.

timdengyun (approved):

pkg/nsx/services/staticroute/staticroute.go
TaoZou1: ListIndexFuncValues will return key or id? for staticroute, the delete needs model.StaticRoutes.Path/Id, so here use List()

PR #497, opened 2024-01-23
Use error type instead of error code in cleanup

seanpang-vmware:
1. Status impl error interface, so the return value of cleanup changed from status, error to error only.
2. Provide go doc on Cleanup func to provide more details about returned error type.
3. Remove OK status, when cleanup success, just return nil error.

### DIFF
--- a/cmd_clean/main.go
+++ b/cmd_clean/main.go
@@ -63,9 +63,9 @@ func main() {
 
 	logf.SetLogger(logger.ZapLogger(cf.DefaultConfig.Debug, config.LogLevel))
 
-	status, err := clean.Clean(cf)
+	err := clean.Clean(cf)
 	if err != nil {
-		log.Error(err, "failed to clean nsx resources", "status", status)
+		log.Error(err, "failed to clean nsx resources", "status", err.Error())
 		os.Exit(1)
 	}
 	os.Exit(0)
--- a/pkg/clean/clean.go
+++ b/pkg/clean/clean.go
@@ -4,6 +4,7 @@
 package clean
 
 import (
+	"errors"
 	"fmt"
 
 	"k8s.io/client-go/util/retry"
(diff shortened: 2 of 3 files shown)


cmd_clean/main.go
```diff
@@ -63,9 +63,9 @@ func main() {
 
 	logf.SetLogger(logger.ZapLogger(cf.DefaultConfig.Debug, config.LogLevel))
 
-	status, err := clean.Clean(cf)
+	err := clean.Clean(cf)
 	if err != nil {
-		log.Error(err, "failed to clean nsx resources", "status", status)
+		log.Error(err, "failed to clean nsx resources", "status", err.Error())
 		os.Exit(1)
```
ihgann, line 69: Question - was the intention of the status code to also have `os.Exit()` return the code number?

pkg/clean/clean.go
```diff
@@ -27,19 +27,25 @@ var log = logger.Log
 // including security policy, static route, subnet, subnet port, subnet set, vpc, ip pool, nsx service account
 // it is usually used when nsx-operator is uninstalled and remove all the resources created by nsx-operator
 // return error if any, return nil if no error
-func Clean(cf *config.NSXOperatorConfig) (Status, error) {
+// the error type include followings:
+//
+//	 	ValidationFailed 			indicate that the config is incorrect and failed to pass validation
```
ihgann, line 32: nit: spacing is off compared with the rest

pkg/clean/clean.go
```diff
@@ -27,19 +27,25 @@ var log = logger.Log
...
 	}
 	nsxClient := nsx.GetClient(cf)
 	if nsxClient == nil {
-		return GetNSXClientFailed, fmt.Errorf("failed to get nsx client")
+		return GetNSXClientFailed
 	}
 	if cleanupService, err := InitializeCleanupService(cf); err != nil {
-		return InitCleanupServiceFailed, fmt.Errorf("failed to initialize cleanup service: %w", err)
+		return InitCleanupServiceFailed
 	} else if cleanupService.err != nil {
```
ihgann, line 47: Not related to your change, but - why does `cleanupService` have a property called `err`? This is pretty uncommon to see.

pkg/clean/clean.go
```diff
@@ -27,19 +27,25 @@ var log = logger.Log
...
+		return ValidationFailed
 	}
 	nsxClient := nsx.GetClient(cf)
 	if nsxClient == nil {
-		return GetNSXClientFailed, fmt.Errorf("failed to get nsx client")
+		return GetNSXClientFailed
 	}
 	if cleanupService, err := InitializeCleanupService(cf); err != nil {
-		return InitCleanupServiceFailed, fmt.Errorf("failed to initialize cleanup service: %w", err)
+		return InitCleanupServiceFailed
```
ihgann, line 46: I think it would be valuable to wrap the original error. You know better about this than I with respect to the internals of these packages,  but Go supports error wrapping so we can validate the specific type returned, but also keep context of deeper stack errors.

See https://pkg.go.dev/errors. In particular, you can make easy use of `errors.Join` - see https://pkg.go.dev/errors#example-Join.

ex:

```go
return errors.Join(InitCleanupServiceFailed, err)
```

Consumers can then use this for logging or deeper inspection if they desire.

cmd_clean/main.go
seanpang-vmware: we do not intent to use error code as exit code since log message already contains enough information.

seanpang-vmware:
> Overall, LGTM. A few additional suggestions. Can we add unit tests as well?

Xie will add Unit test for cleanup in another patch

pkg/clean/clean.go
zhengxiexie: @ihgann refer https://github.com/vmware-tanzu/nsx-operator/blob/main/pkg/clean/clean.go#L82, the purpose is to avoid error check hell.

zhengxiexie (approved):
LGTM

PR #498, opened 2024-01-23
Temporarily disable traffic check in SP

timdengyun:
This patch is to temporarily disable traffic check and named ports
check in Security Policy e2e test.

In VPC mode, currently, it's unable to create a subnetport in k8s env,
so, we can not check pod running and pod traffic.

Also, creating named port related security policy will fail as well
because there are no running pods created for the named port security policy.

We still check securitypolicy creation, update and deletion for e2e test.

### DIFF
--- a/test/e2e/framework.go
+++ b/test/e2e/framework.go
@@ -8,20 +8,16 @@ import (
 	"net"
 	"os/exec"
 	"regexp"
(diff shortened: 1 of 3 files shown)


zhengxiexie (approved):
LGTM

PR #500, opened 2024-01-24
Add marked_for_delete in search query param

timdengyun:
This patch is to add marked_for_delete:false in the nsx client search query param to filter the deleted resource in the NSXT side.


zhengxiexie (approved):
LGTM

PR #518, opened 2024-02-07
Support caFile as raw content of a cert

TaoZou1:
caFile parameter is either or existed cert file name or raw content of a cert
cleanup will pass raw content of a cert as string


zhengxiexie (approved):
Also update usage of cmd_clean/main.go ?

PR #519, opened 2024-02-08
Add record.EventRecorder for recording events when reconciling

zhengxiexie:


pkg/controllers/ippool/ippool_controller.go
```diff
@@ -47,24 +48,29 @@ type IPPoolReconciler struct {
 	Scheme     *apimachineryruntime.Scheme
 	Service    *ippool.IPPoolService
 	VPCService servicecommon.VPCServiceProvider
+	Recorder   record.EventRecorder
 }
 
-func deleteSuccess(r *IPPoolReconciler, _ *context.Context, _ *v1alpha2.IPPool) {
+func deleteSuccess(r *IPPoolReconciler, _ *context.Context, o *v1alpha2.IPPool) {
+	r.Recorder.Event(o, v1.EventTypeNormal, "SuccessfulDelete", "SuccessfulDelete: IPPool CR has been successfully deleted")
```
dantingl, line 55: If CR is deleted successful, we do not need to create event as CR is deleted, right?

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -116,6 +122,7 @@ func (r *IPPoolReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		log.Error(err, "unable to fetch ippool CR", "req", req.NamespacedName)
 		return resultNormal, client.IgnoreNotFound(err)
 	}
+	r.Recorder.Event(obj, v1.EventTypeNormal, "ReconcilingIPPool", "Reconciling IPPool")
```
dantingl, line 125: I think we can just post event after operation is done

pkg/controllers/ippool/ippool_controller.go
zhengxiexie: I searched other pages saying that there are normal events below we could record:

ObjectCreated: Record when a new object is created.
ObjectUpdated: Record when an object is updated.
ObjectDeleted: Record when an object is deleted.
ObjectFinalized: Record when an object is finalized.
ObjectInitialized: Record when an object is initialized.

 Also Custom Events:

> You can also define custom events based on the specific logic of your controller. For example, if there are certain conditions that need to be highlighted during reconciliation, you can create custom events for those conditions.

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -47,24 +48,29 @@ type IPPoolReconciler struct {
...
 
 func updateSuccess(r *IPPoolReconciler, c *context.Context, o *v1alpha2.IPPool) {
 	r.setReadyStatusTrue(c, o, metav1.Now())
+	r.Recorder.Event(o, v1.EventTypeNormal, "SuccessfulUpdate", "SuccessfulUpdate: IPPool CR has been successfully updated")
 	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerUpdateSuccessTotal, MetricResType)
 }
 
 func updateFail(r *IPPoolReconciler, c *context.Context, o *v1alpha2.IPPool, e *error) {
 	r.setReadyStatusFalse(c, o, metav1.Now(), e)
+	r.Recorder.Event(o, v1.EventTypeWarning, "FailUpdate", fmt.Sprintf("FailUpdate: %v", *e))
```
TaoZou1, line 73: just "Error: %v" in message or "%v", you could align to Line61

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -47,24 +48,29 @@ type IPPoolReconciler struct {
...
-func deleteSuccess(r *IPPoolReconciler, _ *context.Context, _ *v1alpha2.IPPool) {
+func deleteSuccess(r *IPPoolReconciler, _ *context.Context, o *v1alpha2.IPPool) {
+	r.Recorder.Event(o, v1.EventTypeNormal, "SuccessfulDelete", "SuccessfulDelete: IPPool CR has been successfully deleted")
 	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteSuccessTotal, MetricResType)
 }
 
 func deleteFail(r *IPPoolReconciler, c *context.Context, o *v1alpha2.IPPool, e *error) {
 	r.setReadyStatusFalse(c, o, metav1.Now(), e)
+	r.Recorder.Event(o, v1.EventTypeWarning, "FailDelete", fmt.Sprintf("FailError: %v", *e))
 	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteFailTotal, MetricResType)
```
TaoZou1, line 62: could you put reason in const?  in Message, just "Error:" since the reason is Faildelete

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -47,24 +48,29 @@ type IPPoolReconciler struct {
...
 }
 
 func updateSuccess(r *IPPoolReconciler, c *context.Context, o *v1alpha2.IPPool) {
 	r.setReadyStatusTrue(c, o, metav1.Now())
+	r.Recorder.Event(o, v1.EventTypeNormal, "SuccessfulUpdate", "SuccessfulUpdate: IPPool CR has been successfully updated")
```
TaoZou1, line 67: no needs to include 'SuccessfulUpdate' in message

pkg/controllers/ippool/ippool_controller.go
TaoZou1: The other concern is that Reconcile call updateFail even the request will be retried. I think it should not record event if the request will be retried. since it's not the final status. The error should be divided into retriable/not retriable error.

TaoZou1 (approved):

PR #524, opened 2024-02-23
check license from nsxt side

TaoZou1:
While init, it will check the license from nsxt side. If CONTAINER license is disable, it will reboot. If DFW license is disable, security policy will be only response for DELETE operation. It will run a routine to check license periodically. If there is no DFW license, it will check license more frequently

SecurityPolicy controller will check if error is invalid license error.

Test Done:
no CONTAINER license
1. if no CONTAINER license, nsx-operator should reset CONTAINER license enable, DFW disable
1. nsx-operator could bootup
2. security policy failed to create or update
3. security policy could be deleted CONTAINER license enable, DFW enable -> CONTAINER/DFW disable
1. nsx-operator restart due to DFW changed no DFW license, but nsx-operator try to create security policy
1. nsx-operator restart due to invalid license error

### DIFF
--- a/pkg/third_party/retry/retry_test.go
+++ b/pkg/third_party/retry/retry_test.go
@@ -159,8 +159,8 @@ func TestMaxDelay(t *testing.T) {
 	)
 	dur := time.Since(start)
 	assert.Error(t, err)
-	assert.True(t, dur > 170*time.Millisecond, "5 times with maximum delay retry is longer than 170ms")
-	assert.True(t, dur < 200*time.Millisecond, "5 times with maximum delay retry is shorter than 200ms")
+	assert.Greater(t, dur, 120*time.Millisecond, "5 times with maximum delay retry is greater than 120ms")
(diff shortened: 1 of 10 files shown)


pkg/third_party/retry/retry_test.go
```diff
@@ -159,8 +159,8 @@ func TestMaxDelay(t *testing.T) {
 	)
 	dur := time.Since(start)
 	assert.Error(t, err)
-	assert.True(t, dur > 170*time.Millisecond, "5 times with maximum delay retry is longer than 170ms")
-	assert.True(t, dur < 200*time.Millisecond, "5 times with maximum delay retry is shorter than 200ms")
+	assert.Greater(t, dur, 120*time.Millisecond, "5 times with maximum delay retry is less than 120ms")
```
timdengyun, line 162: Maybe it should be:
greater than 120ms?
less than 250ms?

TaoZou1: fixed

timdengyun (approved):

Issue #460, opened 2023-12-20
Local replacements causes downstream builds to fail


brakthehack:
Hi, we have another similar occurrence to #309 . This time the offending commits have packages with local code:

https://github.com/vmware-tanzu/nsx-operator/commit/c0c1d166de5d47f85db0be9c9f8bd09187942991
https://github.com/vmware-tanzu/nsx-operator/commit/585b431da40e1c571be1c40a8477abdb868a581b

This breaks consumers because they too must make the replacement. However, the replacement is actually impossible in this case. Please stop using replacements to solve dependency issues since they are not transitive.

cc @jwsui @timdengyun

timdengyun:
This issue is resolved after merging PR: https://github.com/vmware-tanzu/nsx-operator/pull/527 
Thanks @jwsui

PR #527, opened 2024-03-11
Remove local replacement

jwsui:
Use commit id of pkg/client to specify version, and remove local replacement of pkg/client and pkg/apis.


build/yaml/crd/nsx.vmware.com_subnetports.yaml
```diff
@@ -48,6 +48,44 @@ spec:
           spec:
             description: SubnetPortSpec defines the desired state of SubnetPort.
             properties:
+              attachmentRef:
```
timdengyun, line 51: it looks like this change is from crd, not from this PR? is it?

timdengyun: I recall this portion of code has been removed in PR: https://github.com/vmware-tanzu/nsx-operator/pull/520/files#diff-1c96cd7788a18229fd17ec928994f98a437dfedf84c0ad7fecb4a345435250f5L51, since we don't have attachment in subnet port but in annotation, do you debase the latest code

jwsui: Thanks, this is because we are still using some old commit of pkg/apis in go.mod, so I upgrade the commit in go.mod as well.

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -253,8 +253,6 @@ func (r *NSXServiceAccountReconciler) updateNSXServiceAccountStatus(ctx *context
 	obj := o
 	if e != nil && *e != nil {
 		obj = o.DeepCopy()
-		obj.Status.Phase = nsxvmwarecomv1alpha1.NSXServiceAccountPhaseFailed
```
timdengyun, line 256: why we remove status field from serveracount? is there CRD change? we has better to double confirm with @gran-vmv

timdengyun:
once the patch merged, please close the issue:https://github.com/vmware-tanzu/nsx-operator/issues/460

build/yaml/crd/nsx.vmware.com_subnetports.yaml
timdengyun: One question, how we generated the new commit for pkg/apis? if we merged this patch, will the commit of pkg/apis will be changed again?

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
gran-vmv: I think we should not remove this since it will block the consumer logic.

gran-vmv: Maybe we can add a comment that it is being used and requires change on both side.

jwsui: Done, thanks

build/yaml/crd/nsx.vmware.com_subnetports.yaml
timdengyun: talk offline, we use go get xx@commitid to generate pseudo-randmon-uid here for pkg/apis

go.mod
```diff
@@ -22,8 +16,8 @@ require (
 	github.com/prometheus/client_golang v1.16.0
 	github.com/sirupsen/logrus v1.9.0
 	github.com/stretchr/testify v1.8.2
-	github.com/vmware-tanzu/nsx-operator/pkg/apis v0.0.0-20231227030711-34390278f3cd
-	github.com/vmware-tanzu/nsx-operator/pkg/client v0.0.0-00010101000000-000000000000
+	github.com/vmware-tanzu/nsx-operator/pkg/apis v0.0.0-20240305035435-c992c623aad3
```
timdengyun, line 19: It's good that we clearly say this uid is in accordance with which exact commit id of pkg/apis in the PR commit msg, it will help us to track this lib change history and what exact commit we provide to users.

go.mod
```diff
@@ -22,8 +16,8 @@ require (
...
+	github.com/vmware-tanzu/nsx-operator/pkg/client v0.0.0-20240102061654-537b080e159f
```
timdengyun, line 20: Ditto.

timdengyun (approved):

go.mod
timdengyun: c992c623aad3 is commit id of pkg/apis lib.

jwsui: Synced offline.

go.mod
jwsui: Same here.

PR #537, opened 2024-03-28
Fix bug which misses k8s Client when cleaning network policy

zhengxiexie:
When cleanup network policy, it will complain
"cleaning up security policies created for network policy	{"count": 6} E0328 02:51:16.193184 3574675 runtime.go:79]
 Observed a panic: "invalid memory address or nil pointer dereference"
 (runtime error: invalid memory address or nil pointer dereference)
 Because it use Client in getSharedVPCNamespaceFromNS of vpc.go,
 we haven't initialized k8s client in cleanup library.

### DIFF
--- a/pkg/clean/clean.go
+++ b/pkg/clean/clean.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 
 	"k8s.io/client-go/util/retry"
+	ctrl "sigs.k8s.io/controller-runtime"
 
 	"github.com/vmware-tanzu/nsx-operator/pkg/config"
 	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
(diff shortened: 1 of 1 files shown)


pkg/clean/clean.go
```diff
@@ -96,7 +97,16 @@ func wrapCleanFunc(ctx context.Context, clean cleanup) func() error {
...
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{})
+	if err != nil {
+		return nil, err
+	}
+
+	// ensure we can get mgr.GetClient() and the cache already started
+	go mgr.Start(context.Background())
+
 	var commonService = common.Service{
+		Client:    mgr.GetClient(),
```
heypnus, line 109: 1. The dependency on K8s client is introduced by https://github.com/vmware-tanzu/nsx-operator/pull/435/files#diff-f8c6c91e11a26fcc30a56d7531f369fe0367ff19108a9b22891cbc0ac7e8e0e8L197 , not network policy feature. In the above change, the code invoke `s.Client.Get` instead of `s.VpcStore.GetVPCsByNamespace`.
2. I think the cleanup should not depends on the K8s client. It only needs the NSX client. If we introduce the K8s client, then if the cleanup is invoked after the K8s cluster being destroyed, the cleanup also will fail, is it right? If it's right, I think we'd better avoid to the K8s query , do you think it's feasible?

zhengxiexie: Yes, it's better to avoid importing k8s client, @seanpang-vmware do you have any suggestion?

seanpang-vmware: I think we need to init a k8s client in cleanup process, we can not avoid using k8s client. even if we can find a way to work around this issue, we may still hit other problem in the future.
From operator logic, it should be a common case to use k8s client to get resource info when cleaning up.

PR #539, opened 2024-04-06
Change securitypolicy GC and cleanup to get VPC info by NSX path

timdengyun:
In GC or VCP Cleanup, it might be that there is k8s env available, so, it's not possible to get VPC info from k8s side when vpcService is trying to list VPC info of securitypolicy namespace by k8s client.

This patch is to get VPC info of the securitypolicy from the securitypolicy store where the securitypolicy path saves its NSX path.


pkg/nsx/services/securitypolicy/firewall.go
```diff
@@ -656,10 +656,18 @@ func (service *SecurityPolicyService) deleteSecurityPolicy(obj interface{}, isVp
...
+		if isGcOrVpcCleanup == false {
+			vpcInfo, err = service.getVpcInfo(spNameSpace)
+			if err != nil {
+				return err
+			}
+		} else {
+			// In GC or cleanup process, vpcInfo should be listed directly from security policy store to avoid calling VPC service.
+			// Get orgId, projectId, vpcId from security policy path "/orgs/<orgId>/projects/<projectId>/vpcs/<vpcId>/security-policies/<spId>"
+			vpcInfo = new(common.VPCResourceInfo)
+			vpcInfo.OrgID, vpcInfo.ProjectID, vpcInfo.VPCID, _ = nsxutil.ParseVPCPath(*(nsxSecurityPolicy.Path))
```
zhengxiexie, line 670: Better check if Path is valid, or it may crash. Better to refactor ParseVPCPath to return err.

pkg/nsx/services/securitypolicy/firewall.go
```diff
@@ -656,10 +656,18 @@ func (service *SecurityPolicyService) deleteSecurityPolicy(obj interface{}, isVp
 	finalSecurityPolicyCopy := *nsxSecurityPolicy
 	finalSecurityPolicyCopy.Rules = nsxSecurityPolicy.Rules
 
-	if isVpcEnabled(service) || isVpcCleanup {
-		vpcInfo, err := service.getVpcInfo(spNameSpace)
-		if err != nil {
-			return err
+	if isVpcEnabled(service) || isGcOrVpcCleanup {
```
zhengxiexie, line 659: Perhaps Gc can't gurantee the Path must synced, we'd better to move it as we discussed.

pkg/controllers/securitypolicy/securitypolicy_controller.go
```diff
@@ -163,7 +163,7 @@ func (r *SecurityPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	} else {
 		if controllerutil.ContainsFinalizer(obj, servicecommon.SecurityPolicyFinalizerName) {
 			metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerDeleteTotal, MetricResType)
-			if err := r.Service.DeleteSecurityPolicy(obj, false, servicecommon.ResourceTypeSecurityPolicy); err != nil {
+			if err := r.Service.DeleteSecurityPolicy(obj.UID, false, servicecommon.ResourceTypeSecurityPolicy); err != nil {
```
heypnus, line 166: Why here changes `obj` to `obj.UID`? I have this question because in `networkpolicy_controller.go` we also have the similar usage `r.Service.DeleteSecurityPolicy(networkPolicy, false, servicecommon.ResourceTypeNetworkPolicy)`

timdengyun: good point. In the very beginning of SP implementation, we use obj.UID to delete securitypolicy, the purpose of passing UID instead is trying to get SP from store not recreate it in deleteSecurityPolicy function for the
case *v1alpha1.SecurityPolicy: 

Later, I changed it to obj from obj.UID, because I want to leverage the processing buildSecurityPolicy to handle all situation for project related resources. 
case *v1alpha1.SecurityPolicy:
nsxSecurityPolicy, nsxGroups, projectShares, err = service.buildSecurityPolicy(sp, createdFor)

By now, I realized that we don't need passing obj anymore since we we have project store to save project related resource and send one API call. the whole process is simplified. 

But for r.Service.DeleteSecurityPolicy(networkPolicy, false, servicecommon.ResourceTypeNetworkPolicy)

I also checked the code,and I found that we might not be able to pass  networkPolicy.UID in DeleteSecurityPolicy because networkpolicy will call convertNetworkPolicyToInternalSecurityPolicies in DeleteSecurityPolicy to do conversion:
case *networkingv1.NetworkPolicy:
		internalSecurityPolicies, err := service.convertNetworkPolicyToInternalSecurityPolicies(obj.(*networkingv1.NetworkPolicy))

pkg/nsx/services/securitypolicy/firewall.go
timdengyun: Thanks for advice, It's good to check nsxSecurityPolicy.Path. but I feel that there is no need to check ParseVPCPath since we don't know what info the user needs. Also, if there is invalid info from VPC path, 
and later, API call or something will pop up error.

pkg/nsx/services/securitypolicy/firewall.go
timdengyun: Agreed, I revert changes for GC part.

pkg/controllers/securitypolicy/securitypolicy_controller.go
timdengyun: I'd like to listen your advice.

heypnus: > By now, I realized that we don't need passing obj anymore since we we have project store to save project related resource and send one API call. the whole process is simplified.

Is this change necessary to the topic `Change securitypolicy cleanup to get VPC info by NSX path`? If not, I can try to unify the whole process for both sp and np in another patch and we can keep your change in this patch.

timdengyun: sounds good,  it's good that we can rethink to try to unify the both SP and NP deletion. I can remove this change since it's not absolute necessary for this patch.

heypnus: lgtm

pkg/nsx/services/securitypolicy/firewall.go
```diff
@@ -656,9 +656,23 @@ func (service *SecurityPolicyService) deleteSecurityPolicy(obj interface{}, isVp
...
+		var vpcInfo *common.VPCResourceInfo
+		if isVpcCleanup == false {
+			vpcInfo, err = service.getVpcInfo(spNameSpace)
+			if err != nil {
+				return err
+			}
+		} else {
+			// In cleanup process, vpcInfo should be listed directly from security policy store to avoid calling VPC service.
+			// Get orgId, projectId, vpcId from security policy path "/orgs/<orgId>/projects/<projectId>/vpcs/<vpcId>/security-policies/<spId>"
+			if nsxSecurityPolicy.Path == nil {
```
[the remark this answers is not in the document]
timdengyun: I'm not sure I understand this correct. If NSX security policy is created in NSX successfully by API call.  For cleanup, there must be a valid path by which we can delete it. 
For normal case, I think it means there is SP created in NSX side, and after operator restated from crash, it will re-initialized NSX security policy from nsx side and save it in the store. 
Meanwhile, opreator will rebuild a NSX security policy from CR and compare it with the existing one in the store. but the SP UID should be the same not changed since CR is not changed. At last, operator will re-sync with NSX to make both sides consistency.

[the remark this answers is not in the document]
heypnus: My bad, so the path is the NSX security policy object's path, not the path in security policy CR status. Ignore the previous comment. It's ok.

heypnus (approved):
LGTM

PR #557, opened 2024-04-29
Dump SDK API error

TaoZou1:
Dump detail of SDK API return error

Test Done:
1. change PageSize=2000 to trigger error
2. check if message is dumped

2024-05-10 07:28:22.635 INFO    common/store.go:105     translate error {"apierror": {"Details":"Field level validation errors: {value 1,100 of property page_size has exceeded its maximum valid value 1000}","ErrorCode":255,"ErrorData":null,"ErrorMessage":"Field level validation errors: {value 1,100 of property page_size has exceeded its maximum valid value 1000}","ModuleName":"common-services","RelatedErrors":null}, "error type": "INVALID_REQUEST"}

### DIFF
--- a/pkg/controllers/staticroute/staticroute_controller.go
+++ b/pkg/controllers/staticroute/staticroute_controller.go
@@ -31,6 +31,7 @@ import (
 	_ "github.com/vmware-tanzu/nsx-operator/pkg/nsx/ratelimiter"
 	commonservice "github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/staticroute"
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/util"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


pkg/controllers/staticroute/staticroute_controller.go
```diff
@@ -97,6 +98,10 @@ func (r *StaticRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		if err := r.Service.CreateOrUpdateStaticRoute(req.Namespace, obj); err != nil {
 			updateFail(r, &ctx, obj, &err)
 			// TODO: if error is not retriable, not requeue
+			apierror, _ := util.DumpAPIError(err)
```
dantingl, line 101: Is this test code?

pkg/controllers/staticroute/staticroute_controller.go
```diff
@@ -97,6 +98,10 @@ func (r *StaticRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		if err := r.Service.CreateOrUpdateStaticRoute(req.Namespace, obj); err != nil {
 			updateFail(r, &ctx, obj, &err)
 			// TODO: if error is not retriable, not requeue
+			apierror, errortype := util.DumpAPIError(err)
```
zhengxiexie, line 101: A nit suggestion, if apierror is not nil, could we write it to status, so we can debug from CR status more straight forward?

zhengxiexie:
squash into 1 commit.

TaoZou1:
There is two commits here, close it and create a new one

PR #560, opened 2024-05-06
Add lock for vpc networkconfig storage

seanpang-vmware:
Add locker for VPC network config storage map object to avoid concurrent modification

### DIFF
--- a/pkg/nsx/services/vpc/vpc.go
+++ b/pkg/nsx/services/vpc/vpc.go
@@ -45,13 +45,23 @@ var (
 	enableAviAllowRule = false
 )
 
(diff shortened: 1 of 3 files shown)


pkg/nsx/services/vpc/vpc.go
```diff
@@ -45,13 +45,23 @@ var (
 	enableAviAllowRule = false
 )
 
+type VPCNetworkInfoStore struct {
```
zhengxiexie, line 48: Directly embed sync.Mutex and s.Lock() instead of s.VPCNetworkConifgStore.m.Lock(), this way is better?

zhengxiexie (approved):

PR #565, opened 2024-05-08
Fix the issue that VM cannot get DHCP IP

heypnus:
In DHCP subnet, the static ip pool size should be zero.

Testing done:

1. Patch the fix in a live testbed and ceate a DHCP subnet
2. Create a VM under the DHCP subnet and check the VM can get the DHCP IP:
>     # kubectl get vm -n yiyi-ns  -o wide test-vm-dhcp-cloudinit-jammy-fixed
>     NAME                                 POWER-STATE   CLASS               IMAGE                   PRIMARY-IP4    AGE
>     test-vm-dhcp-cloudinit-jammy-fixed   PoweredOn     best-effort-small   vmi-d8b7c860a0eab5c84   172.26.1.131   5m47s


jwsui (approved):
LGTM.

PR #567, opened 2024-05-08
Add timeout for clean initialize

TaoZou1:
The cleanup should support timeout for whole procedure. Currently initializing service doesn't support timeout since nsx-operator doesn't need it. Add timeout control during clean initialize

Test Done:
1. add time.Sleep(20*time.Second) in InitializeCleanupService
2. set the timeout to 10*time.Second in ctx, check if clean procedure could be timeout


dantingl (approved):

zhengxiexie (approved):
LGTM

PR #568, opened 2024-05-08
Specify the release branches the triggering workflow must run on in order to trigger related workflow

wenqiq:
Specify the release branches the triggering workflow must run on in order to trigger related workflow

ref: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

    Signed-off-by: Wenqi Qiu <wenqiq@vmware.com>

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -2,9 +2,13 @@ name: Makefile CI
 
 on:
   push:
-    branches: [ main ]
+    branches:
+    - main
+    - 'v[0-9].[0-9]+.[0-9]+'
   pull_request:
-    branches: [ main ]
+    branches:
+    - main
+    - 'v[0-9].[0-9]+.[0-9]+'
 
 jobs:
   build:


.github/workflows/makefile.yml
```diff
@@ -2,9 +2,15 @@ name: Makefile CI
 
 on:
   push:
-    branches: [ main ]
+    branches:
+    - main
+    - release-*
```
dantingl, line 7: We use vx.x.x for product branch, like 4.1.2

dantingl: Coud we remove release* and feature*, it may be private branches.
And for v*, is there any way we can identify wildcard with format v[0-9]\\.[0-9]\\.[0-9]?

wenqiq: Done.
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

dantingl (approved):

PR #569, opened 2024-05-10
Fix the invalid isolation section issue for networkpolicy without specific ingress/egress

heypnus:
When a network policy without any ingress/egress is created, e.g.

```
kind: NetworkPolicy
metadata:
  name: networkpolicy-deny-ingress
spec:
  podSelector: {}
  policyTypes:
    - Ingress
```

the NSX operator didn't create the correct isolation section because
the section doesn't have any rule. This patch will fix the issue.

Testing done:
```
Step 1. Create the above network policy "networkpolicy-deny-ingress" in
        namespace default.
Step 2. Check the NSX and ensure that there's a valid rule "ingress-isolation"
        in the isolation section "np-default-networkpolicy-deny-ingress-isolation".
```

### DIFF
--- a/pkg/nsx/services/securitypolicy/firewall.go
+++ b/pkg/nsx/services/securitypolicy/firewall.go
@@ -188,108 +188,139 @@ func (service *SecurityPolicyService) CreateOrUpdateSecurityPolicy(obj interface
 	return err
 }
 
-func (service *SecurityPolicyService) convertNetworkPolicyToInternalSecurityPolicies(networkPolicy *networkingv1.NetworkPolicy) ([]*v1alpha1.SecurityPolicy, error) {
-	securityPolicies := []*v1alpha1.SecurityPolicy{}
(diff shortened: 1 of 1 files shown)


pkg/nsx/services/securitypolicy/firewall.go
```diff
@@ -209,6 +211,58 @@ func (service *SecurityPolicyService) convertNetworkPolicyToInternalSecurityPoli
 			},
 		},
 	}
+	// 1.1 Generating ingress allow rules in allow section
```
zhengxiexie, line 214: A nit suggestion, convertNetworkPolicyToInternalSecurityPolicies func is too long, could you break it into smaller functions?

heypnus: done

zhengxiexie (approved):
LGTM

PR #570, opened 2024-05-10
Add clean avi subnet ports before deleting vpc

zhengxiexie:
Avi controller will create subnetport on avi subnet of VPC, when there are stale subnetports left behind, it will fail to cleanup VPC complaining "IpAddressSubnet cannot be deleted as IP from subnet range is being used.", if delete avi subnet ports before deleting VPC, it will succeed to cleanup vpc even though there is avi subnet.

### DIFF
--- a/pkg/nsx/services/vpc/clean_avi.go
+++ b/pkg/nsx/services/vpc/clean_avi.go
@@ -1,0 +1,58 @@
+package vpc
+
+import (
+	"context"
+	"errors"
+
+	mapset "github.com/deckarep/golang-set"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx"
+	nsxutil "github.com/vmware-tanzu/nsx-operator/pkg/nsx/util"
+)
+
+type (
+	mapInterface = map[string]interface{}
+)
+
+const (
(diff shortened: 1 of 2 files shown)


pkg/nsx/services/vpc/clean_avi.go
```diff
@@ -0,0 +1,65 @@
...
+func appendIfNotExist(slice []string, s string) []string {
+	for _, item := range slice {
+		if item == s {
+			return slice
+		}
+	}
+	return append(slice, s)
+}
+
+func httpGetAviResources(cluster *nsx.Cluster, vpcPath string) ([]string, error) {
```
TaoZou1, line 29: httpGetAviResources -> getAviPortsPath?

TaoZou1, line 29: How about use a sets.NewString() instead of []string to avoid calling appendIfNotExist?
BTW, will the path has the same value?

zhengxiexie: ok

zhengxiexie: ok, vpcPath different value.

TaoZou1 (approved):

PR #571, opened 2024-05-11
Dump api error

TaoZou1:
Dump detail of SDK API return error

Test Done:
    change PageSize=1100 to trigger error
    check if message is dumped

2024-05-10 07:28:22.635 INFO common/store.go:105 translate error {"apierror": {"Details":"Field level validation errors: {value 1,100 of property page_size has exceeded its maximum valid value 1000}","ErrorCode":255,"ErrorData":null,"ErrorMessage":"Field level validation errors: {value 1,100 of property page_size has exceeded its maximum valid value 1000}","ModuleName":"common-services","RelatedErrors":null}, "error type": "INVALID_REQUEST"}


zhengxiexie (approved):
LGTM

PR #579, opened 2024-05-17
Fix bug when updateFail in networkinfo

zhengxiexie:
Update setNetworkInfoVPCStatus function to remove VPC if createdVPC is empty

### DIFF
--- a/pkg/controllers/networkinfo/networkinfo_utils.go
+++ b/pkg/controllers/networkinfo/networkinfo_utils.go
@@ -16,15 +16,13 @@ import (
 )
 
 func deleteFail(r *NetworkInfoReconciler, c *context.Context, o *v1alpha1.NetworkInfo, e *error, client client.Client) {
-	createdVpc := &v1alpha1.VPCState{}
-	setNetworkInfoVPCStatus(c, o, client, createdVpc)
+	setNetworkInfoVPCStatus(c, o, client, nil)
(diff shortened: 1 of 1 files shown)


pkg/controllers/networkinfo/networkinfo_utils.go
```diff
@@ -51,6 +51,12 @@ func deleteSuccess(r *NetworkInfoReconciler, _ *context.Context, o *v1alpha1.Net
 }
 
 func setNetworkInfoVPCStatus(ctx *context.Context, networkInfo *v1alpha1.NetworkInfo, client client.Client, createdVPC *v1alpha1.VPCState) {
+	// if createdVPC is empty, remove the VPC from networkInfo
+	if reflect.DeepEqual(*createdVPC, v1alpha1.VPCState{}) {
```
seanpang-vmware, line 55: this is a specific scenario for update fail to remove all state, if so, we could give a nil value directly instead of comparing 2 empty obj.

dantingl:
I think we should not keep the deleted VPC in store, and do not even add it in NetworkInfo CR in the first time

pkg/controllers/networkinfo/networkinfo_utils.go
zhengxiexie: ok

zhengxiexie:
> I think we should not keep the deleted VPC in store, and do not even add it in NetworkInfo CR in the first time

Already deleted in store. https://github.com/vmware-tanzu/nsx-operator/pull/579

seanpang-vmware (approved):

PR #580, opened 2024-05-20
Workaround for DATA RACE

TaoZou1:
There is data race while running cleanup
The root cause is that lazy creation without lock
1. switch to use NewConnector instead of NewRestConnector
2. call connector.NewExecutionContext to avoid lazy creation

Test Done:
1. use go run -race cmd_clean/main.go to run and check if there is DATA RACE
2. no data race reported


zhengxiexie (approved):
LGTM

PR #584, opened 2024-05-22
Fix NetworkInfo gc issue

seanpang-vmware:
1. refactor update NetworkInfo CRD logic
2. do not insert vpc into store if we hit nsx issue that vpc is created but nsx api return error
3. in networkinfo gc logic, only compare if the nsx-vpc's namespace still exist.

### DIFF
--- a/pkg/controllers/networkinfo/networkinfo_controller.go
+++ b/pkg/controllers/networkinfo/networkinfo_controller.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	apimachineryruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
@@ -17,7 +18,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
-
 	"github.com/vmware-tanzu/nsx-operator/pkg/controllers/common"
 	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
 	"github.com/vmware-tanzu/nsx-operator/pkg/metrics"
@@ -56,7 +56,7 @@ func (r *NetworkInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			controllerutil.AddFinalizer(obj, commonservice.NetworkInfoFinalizerName)
 			if err := r.Client.Update(ctx, obj); err != nil {
 				log.Error(err, "add finalizer", "NetworkInfo", req.NamespacedName)
-				updateFail(r, &ctx, obj, &err, r.Client)
+				updateFail(r, &ctx, obj, &err, r.Client, nil)
 				return common.ResultRequeue, err
 			}
 			log.V(1).Info("added finalizer on NetworkInfo CR", "NetworkInfo", req.NamespacedName)
@@ -64,8 +64,8 @@ func (r *NetworkInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 		createdVpc, nc, err := r.Service.CreateOrUpdateVPC(obj)
 		if err != nil {
-			log.Error(err, "operate failed, would retry exponentially", "VPC", req.NamespacedName)
-			updateFail(r, &ctx, obj, &err, r.Client)
+			log.Error(err, "create vpc failed, would retry exponentially", "VPC", req.NamespacedName)
+			updateFail(r, &ctx, obj, &err, r.Client, nil)
 			return common.ResultRequeueAfter10sec, err
 		}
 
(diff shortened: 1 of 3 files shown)


pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -179,8 +218,13 @@ func (r *NetworkInfoReconciler) Start(mgr ctrl.Manager) error {
 	return nil
 }
 
-// GarbageCollector collect vpc which has been removed from networkInfo crd.
-// cancel is used to break the loop during UT
+// The GC logic for nsx-vpc is that:
```
zhengxiexie, line 221: Comment should have the following format 'GarbageCollector ...' (with an optional leading article)

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -179,8 +218,13 @@ func (r *NetworkInfoReconciler) Start(mgr ctrl.Manager) error {
...
+// 1. list all current existing namespace in kubenetes
```
zhengxiexie, line 222: kubernetes

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -190,31 +234,35 @@ func (r *NetworkInfoReconciler) GarbageCollector(cancel chan bool, timeout time.
...
-				crdNetworkInfoVPCStateSet.Insert(vpcState.VPCPath)
-			}
+		nsSet := sets.NewString()
+		for _, ns := range namespaces.Items {
+			nsSet.Insert(ns.Namespace)
 		}
 
 		for _, elem := range nsxVPCList {
-			if crdNetworkInfoVPCStateSet.Has(*elem.Path) {
+			// for go lint Implicit memory aliasing in for loop
```
zhengxiexie, line 257: Can `for i := len(nsxVPCList) - 1; i >= 0; i-- ` avoid this problem?

pkg/nsx/services/vpc/vpc.go
```diff
@@ -563,19 +563,15 @@ func (s *VPCService) CreateOrUpdateVPC(obj *v1alpha1.NetworkInfo) (*model.Vpc, *
...
 			// failed to read, but already created, we consider this scenario as success, but store may not sync with nsx
 			log.Info("confirmed VPC is not created", "VPC", createdVpc.Id)
 			return nil, nil, err
 		} else {
-			// vpc created anyway, update store, and in this scenario, we condsider creating successfully
-			log.Info("read VPCs from NSX after creation failed, still update VPC store", "VPC", *createdVpc.Id)
-			s.VpcStore.Add(&failedVpc)
-			return &failedVpc, &nc, nil
+			// vpc created anyway, in this case, we consider this vpc is created successfully and continue realize process
+			log.Info("vpc crearted although nsx return error, continue realization check", "VPC", *failedVpc.Id)
```
lxiaopei, line 574: crearted -> created

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -77,8 +77,15 @@ func (r *NetworkInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 			err = r.Service.CreateOrUpdateAVIRule(createdVpc, obj.Namespace)
 			if err != nil {
-				log.Error(err, "operate failed, would retry exponentially", "NetworkInfo", req.NamespacedName)
-				updateFail(r, &ctx, obj, &err, r.Client)
+				state := &v1alpha1.VPCState{
+					Name:                    *createdVpc.DisplayName,
+					VPCPath:                 *createdVpc.Path,
+					DefaultSNATIP:           "",
+					LoadBalancerIPAddresses: "",
+					PrivateIPv4CIDRs:        []string{},
```
lxiaopei, line 85: privateCIDRs need be set

dantingl: For VPC update case, we need get DefaultSNATIP/LoadBalancerIPAddresses/PrivateIPv4CIDRs here, right?

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -179,8 +209,13 @@ func (r *NetworkInfoReconciler) Start(mgr ctrl.Manager) error {
...
 
-// GarbageCollector collect vpc which has been removed from networkInfo crd.
-// cancel is used to break the loop during UT
+// GarbageCollector logic for nsx-vpc is that:
+// 1. list all current existing namespace in kubernetes
+// 2. list all the nsx-vpc in vpcStore
+// 3. loop all the nsx-vpc to get its namespace, check if the namespace still exist
+// 4. if ns do not exist anymore, delete the nsx-vpc resource
+// Here we did not check the nsx-vpc path with the NetworkInfo CRD status, because the CRD status is updated
+// after nsx-vpc created, so there maybe race condition that gc happens when nsx-vpc added to store but CRD not updated yet.
```
dantingl, line 218: Do we need the comment line 218?

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -190,31 +225,35 @@ func (r *NetworkInfoReconciler) GarbageCollector(cancel chan bool, timeout time.
...
+			nsSet.Insert(ns.Namespace)
 		}
 
 		for _, elem := range nsxVPCList {
-			if crdNetworkInfoVPCStateSet.Has(*elem.Path) {
+			// for go lint Implicit memory aliasing in for loop
+			// this limitation is fixed after golang 1.22, should remove the temp var after upgrading to 1.22
+			tempElem := elem
+			nsxVPCNamespace := getNamespaceFromNSXVPC(&tempElem)
+			if nsSet.Has(nsxVPCNamespace) {
```
dantingl, line 252: I think for NSX Operator restart case, Namespace cannot be deleted as we have the finalizer. So it's better we also check whether Namespace is in deletion?

pkg/controllers/networkinfo/networkinfo_controller.go
seanpang-vmware: yes, this part is for failed case, we updated what we can get to NetworkInfo CR and put the event back to queue for next loop.

pkg/controllers/networkinfo/networkinfo_controller.go
seanpang-vmware: I think this case need to be handled in NetworkInfo CRD reconciler, not in gc logic

lxiaopei (approved):

PR #588, opened 2024-05-27
Move GC into Reconcile to be called only once

zhengxiexie:
When nsx-operator turns on HA mode, two pods can run GC simultaneously, this may lead to a problem that the resources are inconsistent. Moving GC into reconcile, ensuring only one pod running GC.

### DIFF
--- a/pkg/controllers/ippool/ippool_controller.go
+++ b/pkg/controllers/ippool/ippool_controller.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"regexp"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
@@ -39,6 +40,7 @@ var (
 	resultNormal  = common.ResultNormal
 	resultRequeue = common.ResultRequeue
 	MetricResType = common.MetricResTypeIPPool
+	once          sync.Once
 )
 
 // IPPoolReconciler reconciles a IPPool object
@@ -114,6 +116,8 @@ func (r *IPPoolReconciler) setReadyStatusTrue(ctx *context.Context, ippool *v1al
 }
 
 func (r *IPPoolReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	// Use once.Do to ensure gc is called only once
+	once.Do(func() { go r.IPPoolGarbageCollector(make(chan bool), servicecommon.GCInterval) })
 	obj := &v1alpha2.IPPool{}
 	log.Info("reconciling ippool CR", "ippool", req.NamespacedName)
 	metrics.CounterInc(r.Service.NSXConfig, metrics.ControllerSyncTotal, MetricResType)
@@ -260,7 +264,6 @@ func (r *IPPoolReconciler) Start(mgr ctrl.Manager) error {
 	if err != nil {
 		return err
 	}
-	go r.IPPoolGarbageCollector(make(chan bool), servicecommon.GCInterval)
 	return nil
 }
 
--- a/pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
+++ b/pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
@@ -38,6 +39,7 @@ var (
 	ResultRequeue           = common.ResultRequeue
 	ResultRequeueAfter5mins = common.ResultRequeueAfter5mins
 	MetricResType           = common.MetricResTypeNSXServiceAccount
+	once                    sync.Once
 )
 
 // NSXServiceAccountReconciler reconciles a NSXServiceAccount object.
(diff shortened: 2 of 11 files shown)


pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -65,6 +67,8 @@ type NSXServiceAccountReconciler struct {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.0/pkg/reconcile
 func (r *NSXServiceAccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	// Use once.Do to ensure gc is called only once
+	once.Do(func() { go r.GarbageCollector(make(chan bool), servicecommon.GCInterval) })
```
gran-vmv, line 71: GC won't work until at least one resource is reconciled. Is it acceptable?

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller_test.go
```diff
@@ -513,6 +513,12 @@ func TestNSXServiceAccountReconciler_Reconcile(t *testing.T) {
 				defer patches.Reset()
 			}
 
+			patches2 := gomonkey.ApplyMethod(reflect.TypeOf(r.Service), "ListNSXServiceAccountRealization",
```
gran-vmv, line 516: Mock `GarbageCollector` here. Otherwise you need to close `cancel chan bool` for every test case.

zhengxiexie: ok Thanks~~

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
zhengxiexie: Yes

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -114,6 +116,8 @@ func (r *IPPoolReconciler) setReadyStatusTrue(ctx *context.Context, ippool *v1al
 }
 
 func (r *IPPoolReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
heypnus, line 118: So you mean that in the active pod, the NSX operator runs the Reconcile; in the standby pod, the NSX operator doesn't run the Reconcile, right? Could you please point out the logic to control that? I want to check it also.

zhengxiexie: Yes, from the log
```
2024-06-14 05:24:29.170	INFO	starting server	{"kind": "health probe", "addr": "[::]:8385"}
I0614 05:24:29.171301  340277 leaderelection.go:250] attempting to acquire leader lease default/nsx-operator...
I0614 05:24:29.201179  340277 leaderelection.go:260] successfully acquired lease default/nsx-operator
2024-06-14 05:24:29.201	DEBUG	sc2-hs2-nsbu-ncp-dhcp-83-146_31a75034-77ea-4c8c-a087-a868ea856207 became leader	{"type": "Normal", "object": {"kind":"Lease","namespace":"default","name":"nsx-operator","uid":"d571880d-fcfc-4aca-b832-81e0a21199cc","apiVersion":"coordination.k8s.io/v1","resourceVersion":"10266376"}, "reason": "LeaderElection"}
2024-06-14 05:24:29.201	INFO	Starting EventSource	{"controller": "networkinfo", "controllerGroup": "nsx.vmware.com", "controllerKind": "NetworkInfo", "source": "kind source: *v1alpha1.NetworkInfo"}
2024-06-14 05:24:29.201	INFO	Starting EventSource	{"controller": "networkinfo", "controllerGroup": "nsx.vmware.com", "controllerKind": "NetworkInfo", "source": "kind source: *v1alpha1.VPCNetworkConfiguration"}
2024-06-14 05:24:29.201	INFO	Starting Controller	{"controller": "networkinfo", "controllerGroup": "nsx.vmware.com", "controllerKind": "NetworkInfo"}
```
leaderelection.go controllers it.

timdengyun: So, if the pod is running from active to standby, do we see logs to show Reconcile is stopped? 
From code logic, we see GC is started in start manger function as a goroutine. So, gorouting is still running even the pod in standby mod? Do we have log to show how the pod behaves in standby mode?

pkg/controllers/ippool/ippool_controller.go
```diff
@@ -114,6 +116,8 @@ func (r *IPPoolReconciler) setReadyStatusTrue(ctx *context.Context, ippool *v1al
...
+	// Use once.Do to ensure gc is called only once
+	once.Do(func() { go r.IPPoolGarbageCollector(make(chan bool), servicecommon.GCInterval) })
```
timdengyun, line 120: once.Do can only ensure GC is called once, but once pod is switched from master to standby, this go routing will be stopped?

zhengxiexie: refer https://github.com/vmware-tanzu/nsx-operator/pull/601

pkg/controllers/ippool/ippool_controller.go
zhengxiexie: As we discussed, the process would crash if it transitions to standby mode.

TaoZou1 (approved):

timdengyun (approved):

PR #593, opened 2024-05-29
Initialize NSXLB controllers and services

gran-vmv:


pkg/controllers/endpoints/endpoints_lb_controller.go
```diff
@@ -0,0 +1,4 @@
+/* Copyright © 2024 Broadcom, Inc. All Rights Reserved.
```
luolanzone, line 1: Is there any plan to update all license files in this repo? I saw existing files are still using VMware license.

gran-vmv: @lxiaopei

PR #598, opened 2024-06-12
Remove logger global setting for cleanup

TaoZou1:
Using separate instance for each cleanup instance
    
    Cleanup should support multiple instances which each should have its
    own logger instance.
    Remove glogal SetLogger calling for cleanup, but nsx-operator still have it
    Using logger.InitLog to create initialize log
    
    Test Done:
    1. run cleanup, check if log output is normal
    2. run manager, check if log output is normal

### DIFF
--- a/pkg/clean/clean.go
+++ b/pkg/clean/clean.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sync"
 	"time"
 
 	"github.com/vmware-tanzu/nsx-operator/pkg/config"
@@ -25,11 +24,7 @@ import (
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
-
-var log = logger.Log
-var once sync.Once
 
 var Backoff = wait.Backoff{
 	Steps:    12,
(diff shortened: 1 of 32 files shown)


pkg/clean/clean.go
```diff
@@ -52,9 +51,9 @@ func Clean(ctx context.Context, cf *config.NSXOperatorConfig, logg *logr.Logger,
 	// Clean may be called many times, only init once
 	once.Do(func() {
 		if logg != nil {
-			logf.SetLogger(*logg)
+			log = *logg
 		} else {
-			logf.SetLogger(logger.ZapLogger(debug, logLevel))
+			log = logger.ZapLogger(debug, logLevel)
 		}
 	})
```
ihgann, line 58: Actually, I think the premise of this is incorrect. Suppose multiple clusters invoke `Clean`, they should each operate with independent loggers, right?

Instead of a package-level:

```go
var log = logger.Log
```

What if you just used:

```
if logg == nil {
    logg = logger.ZapLogger(debug, logLevel)
}

logg.Info...
```

and pass `logg` down to each consuming function that needs it as an added required argument? You can then remove the package global as it's no longer necessary.

The package scope is pretty small, and this means you have a contained logger per instance of `Clean`. This also helps in case retries offer contextual data.

TaoZou1: it's correct.
so using passing log inside clean package.
but controller runtime still needs SetLogger.
it complained while I removed the SetLogger.
[controller-runtime] log.SetLogger(...) was never called; logs will not be displayed.

zhengxiexie (dismissed):
LGTM

pkg/clean/clean.go
```diff
@@ -48,15 +44,14 @@ var Backoff = wait.Backoff{
...
-			logf.SetLogger(logger.ZapLogger(debug, logLevel))
-		}
-	})
+func Clean(ctx context.Context, cf *config.NSXOperatorConfig, log *logr.Logger, debug bool, logLevel int) error {
+	// Clean needs to support many instances which each have its own logger
+	if log == nil {
+		logg := logger.ZapLogger(debug, logLevel)
+		log = &logg
+	}
+	logf.SetLogger(*log)
```
ihgann, line 53: I feel this is strange for a library, unrelated to logging, to set a global logger. I think this is better set in `cmd_clean` and omitted here.

TaoZou1: remove it.

zhengxiexie (approved):
LGTM

ihgann (approved):
LGTM from consumer perspective.

seanpang-vmware (approved):

PR #601, opened 2024-06-14
Refactor GC

zhengxiexie:
Refactor GC

Add a common GenericGarbageCollector.
Replace sets.Set[string] with sets.String, since it supports Difference method, avoiding iterating in GC.

### DIFF
--- a/pkg/nsx/services/common/store.go
+++ b/pkg/nsx/services/common/store.go
@@ -27,7 +27,7 @@ type Store interface {
 	// to specific nsx-t side resource and then add it to the store.
 	TransResourceToStore(obj *data.StructValue) error
 	// ListIndexFuncValues is the method to list all the values of the index
-	ListIndexFuncValues(key string) sets.Set[string]
+	ListIndexFuncValues(key string) sets.String //nolint:staticcheck // Ignore the deprecation warning for sets.String
 	// Apply is the method to create, update and delete the resource to the store based
 	// on its tag MarkedForDelete.
 	Apply(obj interface{}) error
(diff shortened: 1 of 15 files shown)


pkg/nsx/services/common/types.go
```diff
@@ -167,6 +168,12 @@ type Service struct {
 	Client    client.Client
 	NSXClient *nsx.Client
 	NSXConfig *config.NSXOperatorConfig
+	Cancel    context.Context
```
timdengyun, line 171: where we use Cancel?

pkg/nsx/services/common/types.go
```diff
@@ -167,6 +168,12 @@ type Service struct {
...
+}
+
+type ServiceAccount struct {
```
timdengyun, line 174: ditto

pkg/nsx/services/common/types.go
zhengxiexie: oh, should removed.

TaoZou1 (dismissed):

pkg/nsx/services/common/types.go
timdengyun: do we need this struct?

zhengxiexie: fixed

pkg/nsx/services/common/store.go
```diff
@@ -27,7 +27,7 @@ type Store interface {
 	// to specific nsx-t side resource and then add it to the store.
 	TransResourceToStore(obj *data.StructValue) error
 	// ListIndexFuncValues is the method to list all the values of the index
-	ListIndexFuncValues(key string) sets.Set[string]
+	ListIndexFuncValues(key string) sets.String //nolint:staticcheck // Ignore the deprecation warning for sets.String
```
timdengyun, line 30: I think the current sets.Set[string] is right way, instead of reverting to deprecated sets.String

PR #602, opened 2024-06-17
Add IPAddressUsage resoruce type

timdengyun:
Adding a new IPAddressUsage resource type which is used
for an aggregated API server to get IPAddressUsage from the VPC
that is associated with the VPC.

The namespace scoped IPAddressUsage resource is put in the group: vpc.nsx.vmware.com, version: v1alpha1. for the aggregated API service. 

The final exposed API from aggregated API service is:
/apis/vpc.nsx.vmware.com/v1alpha1/namespaces/<namespace>/ipaddressusages
{

The initial test:
Creating a IPAddressUsage resource in the ns-1 namespace.

```
apiVersion: vpc.nsx.vmware.com/v1alpha1
kind: IPAddressUsage
metadata:
  name: the-ips-1
ipBlocks:
- path: "/infra/ip-blocks/934e516a-733d-47c6-bd82-738b4fa69a9b"
  available: 1023
  allocatedByVpc:
    count: 51
    ipAddresses:
    - address: 10.0.10.1-10.0.10.4
      subnet: "/orgs/default/projects/<project-id>/vpcs/vpc_1/subnets/subnet_1"
    cidr: 10.0.10.0/16
    visibility: EXTERNAL
```

and using API: kubectl get  ipaddressusages -nns-1 -v6 -o json

```
I0617 13:11:20.455809 2471515 loader.go:395] Config loaded from file:  /etc/kubernetes/admin.conf
I0617 13:11:20.497285 2471515 round_trippers.go:553] GET **https://172.10.0.2:6443/apis/vpc.nsx.vmware.com/v1alpha1/namespaces/ns-1/ipaddressusages?limit=500** 200 OK in 32 milliseconds
{
    "apiVersion": "v1",
    "items": [
        {
            "apiVersion": "vpc.nsx.vmware.com/v1alpha1",
            "ipBlocks": [
                {
                    "allocatedByVpc": {
                        "cidr": "10.0.10.0/16",
                        "count": 51,
                        "ipAddresses": [
                            {
                                "address": "10.0.10.1-10.0.10.4",
                                "subnet": "/orgs/default/projects/\u003cproject-id\u003e/vpcs/vpc_1/subnets/subnet_1"
                            }
                        ],
                        "visibility": "EXTERNAL"
                    },
                    "available": 1023,
                    "path": "/infra/ip-blocks/934e516a-733d-47c6-bd82-738b4fa69a9b"
                }
            ],
            "kind": "IPAddressUsage",
            "metadata": {
                "annotations": {
                    "kubectl.kubernetes.io/last-applied-configuration": "{\"apiVersion\":\"nsx.vmware.com/v1alpha3\",\"ipBlocks\":[{\"allocatedByVpc\":{\"cidr\":\"10.0.10.0/16\",\"count\":51,\"ipAddresses\":[{\"address\":\"10.0.10.1-10.0.10.4\",\"subnet\":\"/orgs/default/projects/\\u003cproject-id\\u003e/vpcs/vpc_1/subnets/subnet_1\"}],\"visibility\":\"EXTERNAL\"},\"available\":1023,\"path\":\"/infra/ip-blocks/934e516a-733d-47c6-bd82-738b4fa69a9b\"}],\"kind\":\"IPBlockUsage\",\"metadata\":{\"annotations\":{},\"name\":\"the-ips-1\",\"namespace\":\"ns-1\"}}\n"
                },
                "creationTimestamp": "2024-06-17T13:06:58Z",
                "name": "the-ips-1",
                "namespace": "ns-1",
                "resourceVersion": "2",
                "uid": "01394d1a-f886-4435-997b-a49ce3d3b6ab"
            }
        },
}
```


pkg/apiserver/apis/v1alpha3/ipblockusage_types.go
```diff
@@ -0,0 +1,68 @@
...
+
+package v1alpha3
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	GroupName    = "nsx.vmware.com"
+	Version      = "v1alpha3"
```
jianjuns, line 12: Why it should be v1alpha3?

pkg/apiserver/apis/v1alpha3/ipblockusage_types.go
```diff
@@ -0,0 +1,68 @@
...
+   SPDX-License-Identifier: Apache-2.0 */
+
+package v1alpha3
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	GroupName    = "nsx.vmware.com"
```
jianjuns, line 11: No issue to share the same API Group with CRDs?

pkg/apiserver/apis/v1alpha3/ipblockusage_types.go
```diff
@@ -0,0 +1,68 @@
...
+	Count       int         `json:"count,omitempty" protobuf:"bytes,1,opt,name=count"`
+	IPAddresses []IPAddress `json:"ipAddresses,omitempty" protobuf:"bytes,2,opt,name=ipAddresses"`
+	CIDR        string      `json:"cidr,omitempty" protobuf:"bytes,3,opt,name=cidr"`
+	Visibility  Visibility  `json:"visibility,omitempty" protobuf:"bytes,4,opt,name=visibility"`
+}
+
+// IPBlocks contains the used IPAddresses and CIDR.
+type IPBlocks struct {
+	Path      string `json:"path,omitempty" protobuf:"bytes,1,rep,name=path"`
+	Available int    `json:"available,omitempty" protobuf:"bytes,2,rep,name=available"`
```
jianjuns, line 39: There is no "available count" per Block? It does not make sense to have a single available count given each Block has different visibilities. Please monitor NSX API changes.

pkg/apiserver/apis/v1alpha3/ipblockusage_types.go
```diff
@@ -0,0 +1,68 @@
...
+type AllocatedByVPC struct {
+	Count       int         `json:"count,omitempty" protobuf:"bytes,1,opt,name=count"`
+	IPAddresses []IPAddress `json:"ipAddresses,omitempty" protobuf:"bytes,2,opt,name=ipAddresses"`
+	CIDR        string      `json:"cidr,omitempty" protobuf:"bytes,3,opt,name=cidr"`
+	Visibility  Visibility  `json:"visibility,omitempty" protobuf:"bytes,4,opt,name=visibility"`
+}
+
+// IPBlocks contains the used IPAddresses and CIDR.
+type IPBlocks struct {
+	Path      string `json:"path,omitempty" protobuf:"bytes,1,rep,name=path"`
```
jianjuns, line 38: Question is the IPBlock name unique? Finally, we may expose IPBlock as a K8s resource from CCI apiserver, and so using the K8s IPBlock resource name makes more sense.

pkg/apiserver/apis/v1alpha3/ipblockusage_types.go
```diff
@@ -0,0 +1,68 @@
...
+	// AllocatedByVPC contains the cidr and used IP Range.
+	Allocated AllocatedByVPC `json:"allocatedByVpc,omitempty" protobuf:"bytes,3,rep,name=allocatedByVpc"`
+}
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// IPBlockUsage
+// +k8s:openapi-gen=true
+// +genclient:Namespaced
```
jianjuns, line 49: If we expose all Tenant scope resources from CCI apiserver, then they should be cluster scoped. Probably let us hold this change until the plan is clear.

jianjuns (commented):
Let us hold

pkg/apiserver/apis/v1alpha3/ipblockusage_types.go
timdengyun: Yes, that is true. From NSX IPBlock API, there should only be Path and CIDR fields. 

Here, the IPBlocks definition just an internal data type specific for IPBlockUsage to align
NSX VPC IPBlockUsage API:

 GET /policy/api/v1/orgs/default/projects/project-1/vpcs/vpc-1/ip-address-usage
  example_response: |
    {
    "ip_blocks": [
        {
            "path": "/infra/ip-blocks/934e516a-733d-47c6-bd82-738b4fa69a9c",
            "available": 1024,
            "allocated_by_vpc": {
                "count": 512,
                "ip-addresses": [
                    {
                        "address": "10.0.0.1-10.0.0.4",
                        "subnet": "/orgs/default/projects/<project-id>/vpcs/vpc_1/subnets/subnet_1"
                    }
                ],
                "cidr": "10.0.0.0/16",
                "visibility": "EXTERNAL"
            }
        }
    ]
    }
    
In this API, is to retrives the usage information for IP addresses within a specific VPC. The response will tell user each IPBlock info, including available cout, allocated ipaddress by which subnet, cidr, and visibilities inside one VPC.

timdengyun:
> Let us hold

sure.

pkg/apiserver/apis/v1alpha3/ipblockusage_types.go
jianjuns: I do not think this version trick is the right approach, let us use a different APIGroup. I am thinking about the following:
1. Use an APIGroup like vpc.nsx.vmware.com or a name that can indicate it is aggregated API not CRD.
2. If we believe we will finally remove or rename CRD APIGroup in future, e.g. to crd.nsx.vmware.com, then we can go your way of using nsx.vmware.com for API aggregation and use a new version.

What you guys think? @dantingl @lxiaopei

pkg/apiserver/apis/v1alpha3/ipblockusage_types.go
jianjuns: Ok, I think the confusion is your struct is named "IPBlocks", but it is actually for a single IPBlock info. Please use a better name. What is the type name in NSX API?

pkg/apiserver/apis/v1alpha3/ipblockusage_types.go
jianjuns: I do not mean your patch, but finally we need to expose IPBlocks in CCI apiserver too, so Tenant Admin can for example view External IPBlocks shared to the Project, and create Project IPBlocks.
In such cases, do all IPBlocks inc. shared External IPBlocks, Project IPBlocks, and Private IPBlocks have a unique name in NSX API inside a Project? What you would use for IPBlock resource names in CCI apiserver? We should use that name here, not NSX path, which is useless for CCI/Supervisor users.

pkg/apiserver/apis/v1alpha3/ipblockusage_types.go
timdengyun: I see. It should be IPBlock for a single IPBlock.
in NSX API: ip_blocks is a list for each single ip_block. I will rename my code to align this.

pkg/apiserver/apis/v1alpha3/ipblockusage_types.go
jianjuns: "apiservice.nsx.vmware.com" can be one way.
Or we change CRD group to "crd.nsx.vmware.com" in future, and use "nsx.vmware.com" or "vpc.nsx.vmware.com" for APIService resources.
Let us see if @tnqn has an opinion.

pkg/apiserver/apis/v1alpha3/ipblockusage_types.go
jianjuns: I would consider adding the API to CCI first, not a Namespace scoped resource in Supervisor.

timdengyun: Agreed, if we could expose all IPBlocks(External, Project and Private) info from CCI side, there is no need to do double effort to expose these info in SV any longer.

pkg/apiserver/apis/v1alpha3/ipblockusage_types.go
tnqn: You may also consider using "extensions.nsx.vmware.com" to hold the new API, using "extensions" to represent this is served by extension apiserver.
Later if it's decided CRDs should go to "crd.nsx.vmware.com", this API may be kept in this group or be moved to "nsx.vmware.com".

jianjuns: Do you still prefer not to add "extensions" to APIService group, if we start from scratch with no existing group? If so, another choice is to use "vpc.nsx.vmware.com".

timdengyun: Yes, it's good to move the existing CRD  group nsx.vmware.com to crd.nsx.vmware.com. We use extensions.nsx.vmware.com for aggregated APIService group. Does it make sense?

tnqn: For me it depends on whether the API is supposed to do currently and in the future.

From other threads, it seems the API's backend API is "/policy/api/v1/orgs/default/projects/project-1/vpcs/vpc-1/ip-address-usage", which seems to get all usages of all IPBlocks used by this VPC. I assume the API will not evolve to represent IPBlocks themselves, then it's specific to VPC and "vpc.nsx.vmware.com" would make sense to me. (BTW, why the API doesn't follow the API name to be "IPAddressUsage"?)

But if it's the opposite, i.e., it's supposed to CRUD IPBlocks in the future, then it seems not proper to be specific to VPC?

tnqn: If my understanding it correct, I think "IPBlockUsage" may be a bit confusing as it sounds the usage of an IPBlock and user may expect the resource name to be IPBlock, but it's actually about the IP usages of a single VPC on multiple IPBlocks, the resource name would be VPC name, right? If yes, I feel the API name should be VPCIPAddressUsage, or IPAddressUsage since it's already in VPC API group.

timdengyun: Thanks for Quan. This API is designed to show IpAddress usage inside one namesapce. Of course, current, there is only VPC created for one VPC. so, vpc.nsx.vmware.com would make more sense. 

Also, IPAddressUsage name is more appropriate to align with NSX API side, Thanks for advice.

timdengyun: Sure, Let's use vpc.nsx.vmware.com as for now.

tnqn: vpc.nsx.vmware.com makes sense to me.

timdengyun:
move to CCI

PR #604, opened 2024-06-18
Refactor subnet subnetset CRD status

timdengyun:
1.Remove subnet CRD unused DHCP options since these fields are not
used, so not expose to uesers.
2.Add gatewayAddresses and DHCPServerAddresses into the subnet CRD status
field, also adding them in subnet field of subnetset CRD.
3.Rename status ipAddress to networkAddresses


jwsui (approved):
LGTM.

pkg/apis/nsx.vmware.com/v1alpha1/subnet_types.go
```diff
@@ -78,20 +80,6 @@ type StaticIPAllocation struct {
 type DHCPConfig struct {
 	// +kubebuilder:default:=false
 	EnableDHCP bool `json:"enableDHCP,omitempty"`
```
jianjuns, line 82: Why we remove all DHCP settings? I feel we should keep DHCPV4PoolSize and DNSClientConfig, and also add "ntp_servers".

BTW, what is the behavior when DHCP is enabled on the Subnet? Do we disable IPAM and MAC allocation for all Subnet Ports, or there is another flag on NetworkInterface to control DHCP or IPAM? @lxiaopei

pkg/apis/nsx.vmware.com/v1alpha1/subnet_types.go
```diff
@@ -30,9 +30,11 @@ type SubnetSpec struct {
 
 // SubnetStatus defines the observed state of Subnet.
 type SubnetStatus struct {
-	NSXResourcePath string      `json:"nsxResourcePath,omitempty"`
-	IPAddresses     []string    `json:"ipAddresses,omitempty"`
-	Conditions      []Condition `json:"conditions,omitempty"`
+	NSXResourcePath   string      `json:"nsxResourcePath,omitempty"`
+	NetworkAddress    []string    `json:"networkAddress,omitempty"`
+	GatewayAddress    []string    `json:"gatewayAddress,omitempty"`
+	DhcpServerAddress []string    `json:"dhcpServerAddress,omitempty"`
```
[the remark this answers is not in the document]
timdengyun: Fixed.

pkg/apis/nsx.vmware.com/v1alpha1/subnet_types.go
timdengyun: Since, as of now, we only use the EnableDHCP field in DHCPConfig. So, We think at this moment, we don't need to expose other properties. 
When DHCP is enabled on subnet, we will have the corresponding flag setting on the 
subnetports to control DHCP or IPAM.

https://github.com/vmware-tanzu/nsx-operator/blob/main/pkg/nsx/services/subnetport/builder.go#L36
https://github.com/vmware-tanzu/nsx-operator/blob/main/pkg/nsx/services/subnetport/builder.go#L70

@heypnus  please keep me honest. 

For DHCPV4PoolSize and DNSClientConfig, ntp_servers, we don't use it, do we still put them in CRD? @jwsui @heypnus

jianjuns: I feel we need DNS and NTP settings in 9.0, and better to have DHCPV4PoolSize too. Could we add them?

Have we talked to VM Operator folks to make sure they will

timdengyun: Sure, I will check with Qian for this part.

jianjuns: There is another possibility - if we can expose VPCServiceProfile and make it modifiable in CCI apiserver, users can configure VPC level DHCP settings inc. DNS and NTP there, and then we can skip the settings on Subnet. Would let you decide which way to go.

timdengyun: Sure. Another thing, I checked NSX API: DHCPV4PoolSize is not supported from NSX 4.2, so, we remove this field? 
For DNS and NTP settings in 9.0, we can see how to support it.

jianjuns: Sure, we can remove deprecated fields in VPC 2.0.

For DNS/NTP, I am fine you want to remove them in this PR, and we can add them back after concluding the 9.0 plan.

pkg/apis/v1alpha1/subnet_types.go
```diff
@@ -30,9 +30,11 @@ type SubnetSpec struct {
 
 // SubnetStatus defines the observed state of Subnet.
 type SubnetStatus struct {
-	NSXResourcePath string      `json:"nsxResourcePath,omitempty"`
-	IPAddresses     []string    `json:"ipAddresses,omitempty"`
-	Conditions      []Condition `json:"conditions,omitempty"`
+	NSXResourcePath   string      `json:"nsxResourcePath,omitempty"`
+	NetworkAddress    []string    `json:"networkAddress,omitempty"`
```
jianjuns, line 34: So, in NSX API, these fields is a single string or a string array?

In K8s API naming convention, they should be in plural ("XYZAddresses") if there can be multiple addresses?

pkg/apis/nsx.vmware.com/v1alpha1/subnet_types.go
timdengyun: Sure, we can remove DHS/NTP, and will add back if necessary.

pkg/apis/v1alpha1/subnet_types.go
timdengyun: Yes, in NSX API, these fields are a single string not a string array.So, it's good to keep it as string instead of array to align with NSX API?

jianjuns: Then we should use string, rather than []string.

The content of a string is one address or can be multiple addresses? If latter, we should name the fields as "xyzAddresses".

timdengyun: Sorry, my bad, NSX API return status list which includes multiple NetworkAddress. We should name this field to NetworkAddresses as the plural type.

jianjuns (approved):

jwsui (approved):

PR #611, opened 2024-06-24
Add support for VPC mixed mode

wenyingd:
1. Add controller for CR Network to maintain the default Network resource in system Namespaces.
2. Add watch in Pod/Namepsace/Service/NetworkPolicy controller to requeue resources when Network type is switch to VPC.
3. Add precheck on if the default Network type is VPC or not in resource create events.
4. Add webhook in NCP owned resource creations to check if the default Network type is VPC or not in a Namespace.


zhengxiexie:

PR #632, opened 2024-07-10
Get path info from store while deleting staticroute

TaoZou1:
In previous version, controller tried to get org/project/vpc info from vpc service. But vpc service may delete vpc before staticroute.

Read org/project/vpc from staticroute path

### DIFF
--- a/pkg/nsx/services/staticroute/staticroute.go
+++ b/pkg/nsx/services/staticroute/staticroute.go
@@ -14,6 +14,7 @@ import (
 	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
 	nsxutil "github.com/vmware-tanzu/nsx-operator/pkg/nsx/util"
+	"github.com/vmware-tanzu/nsx-operator/pkg/util"
 )
 
 type StaticRouteService struct {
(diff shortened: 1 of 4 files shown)


timdengyun (dismissed):

zhengxiexie (approved):
LGTM

pkg/nsx/services/staticroute/staticroute.go
```diff
@@ -135,11 +136,16 @@ func (service *StaticRouteService) GetUID(staticroute *model.StaticRoutes) *stri
 }
 
 func (service *StaticRouteService) DeleteStaticRoute(namespace string, uid string) error {
-	vpc := service.VPCService.ListVPCInfo(namespace)
-	if len(vpc) == 0 {
+	id := String(util.GenerateID(uid, "sr", "", ""))
```
timdengyun, line 139: we don't need namespace as an argument any more.
shall we use uid passed from the argument instead of generating a new id?

pkg/nsx/services/staticroute/staticroute.go
```diff
@@ -134,12 +135,17 @@ func (service *StaticRouteService) GetUID(staticroute *model.StaticRoutes) *stri
 
 }
 
-func (service *StaticRouteService) DeleteStaticRoute(namespace string, uid string) error {
-	vpc := service.VPCService.ListVPCInfo(namespace)
-	if len(vpc) == 0 {
+func (service *StaticRouteService) DeleteStaticRoute(uid string) error {
+	id := String(util.GenerateID(uid, "sr", "", ""))
```
timdengyun, line 139: I see we already pass UID to DeleteStaticRoute
https://github.com/vmware-tanzu/nsx-operator/pull/632/files#diff-98a294a67dfe8f3714cd9405624de61b68e401d40cc3c4c26e4c172cc4da2287R116

Why we re generate uid here by calling util.GenerateID, I think they are the same.
So why not just using uid passed in?

timdengyun (approved):

zhengxiexie (approved):
LGTM

PR #638, opened 2024-07-17
Adapt to new apiVersion for controllers

zhengxiexie:
Since apiVersion `nsx.vmware.com` conflicts with other teams, this patch supports the new apiVersion: `crd.nsx.vmware.com`.
All vpc controllers will watch the new apiVersion, however, there is a special case, securitypolicy.
T1 mode securitypolicy has been released to the market, so we should check if the user is using T1, if yes, we should go on supporting it. The implementation behind it is to reuse the previous logic underlay, `convert.go` is to converse the structs of the new and the old between, in the securitypolicy controller, we check whether to watch the old or the latest.
On the NCP side, it will apply all the yamls in dir `build/yaml/crd`, besides, since we will go on supporting T1 mode, we must check vpc FSS, if it is vpc mode, then also apply the yaml in dir `build/yaml/crd/t1/`, or in dir `build/yaml/crd/vpc/`.

For VPC mode securitypolicy, it can be created successfully

image

So does T1 mode
image

New apiVersion also applies to other CRs.


pkg/apis/crd.nsx.vmware.com/v1alpha2/ippool_types.go
```diff
@@ -0,0 +1,82 @@
+/* Copyright © 2023 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: Apache-2.0 */
+
+package v1alpha2
```
timdengyun, line 4: we also don't need v1alpha2 ipool, please remove it.

build/yaml/crd/new/crd.nsx.vmware.com_securitypolicies.yaml
```diff
@@ -0,0 +1,660 @@
+---
+apiVersion: apiextensions.k8s.io/v1
```
timdengyun, line 2: I feel this new one should be put in crd folder.

build/yaml/crd/old/nsx.vmware.com_securitypolicies.yaml
```diff
@@ -0,0 +1,651 @@
+---
+apiVersion: apiextensions.k8s.io/v1
```
timdengyun, line 2: crd/old might not be a good name, how about crd/T1/ or crd/legacy

pkg/nsx/services/common/types.go
```diff
@@ -87,21 +87,22 @@ const (
...
-	SecurityPolicyFinalizerName    = "securitypolicy.nsx.vmware.com/finalizer"
-	NetworkPolicyFinalizerName     = "networkpolicy.nsx.vmware.com/finalizer"
-	StaticRouteFinalizerName       = "staticroute.nsx.vmware.com/finalizer"
-	NSXServiceAccountFinalizerName = "nsxserviceaccount.nsx.vmware.com/finalizer"
-	SubnetFinalizerName            = "subnet.nsx.vmware.com/finalizer"
-	SubnetSetFinalizerName         = "subnetset.nsx.vmware.com/finalizer"
-	SubnetPortFinalizerName        = "subnetport.nsx.vmware.com/finalizer"
-	NetworkInfoFinalizerName       = "networkinfo.nsx.vmware.com/finalizer"
-	PodFinalizerName               = "pod.nsx.vmware.com/finalizer"
+	OldSecurityPolicyFinalizerName = "securitypolicy.nsx.vmware.com/finalizer"
```
timdengyun, line 96: OldSecurityPolicyFinalizerName ->T1SecurityPolicyFinalizerName

pkg/nsx/services/common/types.go
```diff
@@ -87,21 +87,22 @@ const (
...
+	NewSecurityPolicyFinalizerName = "securitypolicy.crd.nsx.vmware.com/finalizer"
```
timdengyun, line 97: NewSecurityPolicyFinalizerName -> SecurityPolicyFinalizerName

pkg/controllers/securitypolicy/securitypolicy_controller.go
```diff
@@ -100,22 +109,36 @@ func (r *SecurityPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ResultNormal, client.IgnoreNotFound(err)
 	}
 
+	isZero := false
+	finalizerName := servicecommon.OldSecurityPolicyFinalizerName
+	realObj := &v1alpha1.SecurityPolicy{}
+	switch obj.(type) {
```
timdengyun, line 115: we already make sure obj type is either &crdv1alpha1.SecurityPolicy{} when VPC enabled or 
&v1alpha1.SecurityPolicy{} for T1.
Why we still do version conversion here. It looks like we assume v1alpha1.SecurityPolicy can be exixst in VPC mode, but this is impossible. Since VPC is greenfield, there is no existing CR with crdv1alpha1.SecurityPolicy. So, in VPC mode, we can always check &crdv1alpha1.SecurityPolicy{}
and when VPC disabled, we only check T1 &v1alpha1.SecurityPolicy{}.
One thing is if we also install T1 v1alpha1.SecurityPolicy CRD in VPC mode. I think the answer is no. 
For T1, we only install v1alpha1.SecurityPolicy CRD  and install crdv1alpha1.SecurityPolicy in VPC.

timdengyun:
squash two commits into one commit.

build/yaml/crd/crd.nsx.vmware.com_ippools.yaml
```diff
@@ -0,0 +1,251 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.14.0
+  name: ippools.crd.nsx.vmware.com
```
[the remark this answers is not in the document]
zhengxiexie: OK，in next PR.

pkg/apis/crd.nsx.vmware.com/v1alpha2/ippool_types.go
zhengxiexie: ditto

build/yaml/crd/new/crd.nsx.vmware.com_securitypolicies.yaml
zhengxiexie: ok

build/yaml/crd/old/nsx.vmware.com_securitypolicies.yaml
zhengxiexie: ok

pkg/nsx/services/common/types.go
zhengxiexie: ok

pkg/nsx/services/common/types.go
zhengxiexie: ok

cmd/main.go
```diff
@@ -18,11 +18,11 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
+	crdv1alpha1 "github.com/vmware-tanzu/nsx-operator/pkg/apis/crd.nsx.vmware.com/v1alpha1"
```
tnqn, line 21: It uses the whole domain as the package name? I think it's uncommon, why doesn't it just name the package as "crd"?

In my mind the new group can be "crd" and the old can be "legacy".

pkg/controllers/securitypolicy/securitypolicy_controller.go
```diff
@@ -46,6 +47,7 @@ var (
 	ResultRequeueAfter5mins = common.ResultRequeueAfter5mins
 	MetricResType           = common.MetricResTypeSecurityPolicy
 	once                    sync.Once
+	EnableVPCNetwork        = false
```
tnqn, line 50: Make it a member variable of `SecurityPolicyReconciler`, declaring a global variable for something that can be updated at runtime could cause troubles to maintenance and tests and is bad for encapsulation.

pkg/controllers/securitypolicy/securitypolicy_controller.go
```diff
@@ -91,7 +93,14 @@ func (r *SecurityPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	// Use once.Do to ensure gc is called only once
 	common.GcOnce(r, &once)
 
-	obj := &v1alpha1.SecurityPolicy{}
+	var obj client.Object
+	if r.Service.NSXConfig.EnableVPCNetwork {
+		obj = &crdv1alpha1.SecurityPolicy{}
+		EnableVPCNetwork = true
```
tnqn, line 99: Is it guaranteed that other code relies on `EnableVPCNetwork` are always executed after `SecurityPolicyReconciler.Reconcile` is called?

pkg/controllers/securitypolicy/securitypolicy_controller.go
```diff
@@ -91,7 +93,14 @@ func (r *SecurityPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	} else {
+		obj = &v1alpha1.SecurityPolicy{}
+	}
```
tnqn, line 102: Does it need to duplicate the definition of the structs?

I thought you could declare the structs in one package and reference it in another package, then you only need to deal with one struct. See how https://github.com/kubernetes/kubernetes/blob/v1.20.0/pkg/apis/extensions/register.go shares the struct with https://github.com/kubernetes/kubernetes/blob/v1.20.0/pkg/apis/apps/register.go

But not sure if `r.Client` can have some ways to differentiate which API we want to access.

pkg/nsx/services/securitypolicy/convert.go
```diff
@@ -0,0 +1,285 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	crdv1alpha1 "github.com/vmware-tanzu/nsx-operator/pkg/apis/crd.nsx.vmware.com/v1alpha1"
+	"github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
+)
+
+func DeepCopyInto(in *v1alpha1.SecurityPolicy, out *crdv1alpha1.SecurityPolicy) {
```
tnqn, line 14: Did you write these manually? deepcopy-gen can be used to generate such code and is safer to use it to avoid unexpected copy issues.

cmd/main.go
zhengxiexie: Yes, github.com/vmware-tanzu/nsx-operator/pkg/apis is the module name, and it contains `nsx.vmware.com` dir which is named as discussed, now we add the new name `crd.nsx.vmware.com`, that is better to be distinguished.

pkg/controllers/securitypolicy/securitypolicy_controller.go
zhengxiexie: It is used by https://github.com/vmware-tanzu/nsx-operator/blob/main/pkg/controllers/securitypolicy/securitypolicy_controller.go#L380-L380, a bit not convenient to put it as a member variable.

pkg/controllers/securitypolicy/securitypolicy_controller.go
zhengxiexie: r.Service.NSXConfig.EnableVPCNetwork is initialized after loading config, EnableVPCNetwork is used only in https://github.com/vmware-tanzu/nsx-operator/blob/main/pkg/controllers/securitypolicy/securitypolicy_controller.go#L380-L380

pkg/nsx/services/securitypolicy/convert.go
zhengxiexie: Yes, I manually updated some code based on deepcopy-gen, since it only supports the same type copy.

cmd/main.go
tnqn: Can you share where it was discussed for me to understand why it's made so? It's neither common in K8s nor Go to have a domain as a package name.

pkg/controllers/securitypolicy/securitypolicy_controller.go
tnqn: The link doesn't point to its usage. I found it myself but the reason doesn't really justify it, you can't make everything global variables to worsen the encapsulation to adapt code that may not have been encapsulated well.

In this specific case, I see at least 3 ways to make `reconcileSecurityPolicy` the member method of `SecurityPolicyReconciler`, by which it could access both `EnableVPCNetwork` and `Client` directly.
1. Leverage `handler.Funcs` when calling `Watches`
2. Leverage `handler.EnqueueRequestsFromMapFunc` when calling `Watches`
3. Pass the function `reconcileSecurityPolicy` or an interface that implements `reconcileSecurityPolicy` to `EnqueueRequestForNamespace`

1 and 2 make more sense.

pkg/controllers/securitypolicy/securitypolicy_controller.go
tnqn: It doesn't really answer the question. I'm asking `EnableVPCNetwork`, not `r.Service.NSXConfig.EnableVPCNetwork`. 

In another word, is it guaranteed `reconcileSecurityPolicy` will only be executed after `SecurityPolicyReconciler.Reconcile`?

pkg/nsx/services/securitypolicy/convert.go
tnqn: I'm not confident on manual code on this and people would just forgot updating this file when the struct changes. I'm not sure what you mean "it only supports the same type copy", I believe it supports deepcopy of different types of same structs, which is its main use case.

pkg/controllers/securitypolicy/securitypolicy_controller.go
zhengxiexie: Yes, I don't find ways for `r.Client` to differentiate, so I must duplicate it.

pkg/controllers/securitypolicy/securitypolicy_controller.go
zhengxiexie: OK, got it! Thanks

pkg/controllers/securitypolicy/securitypolicy_controller.go
zhengxiexie: Fixed.

pkg/nsx/services/securitypolicy/convert.go
zhengxiexie: I tried conversion-gen, but it is a bit unfriendly to the user. I think twice since the structure is the same, only in a different package, it is safe to use 
```
func T1ToVPC(in *v1alpha1.SecurityPolicy, out *crdv1alpha1.SecurityPolicy) {
	out = (*crdv1alpha1.SecurityPolicy)(unsafe.Pointer(in))
	out.APIVersion = "crd.nsx.vmware.com/v1alpha1"
	controllerutil.RemoveFinalizer(out, common.T1SecurityPolicyFinalizerName)
	controllerutil.AddFinalizer(out, common.SecurityPolicyFinalizerName)
}

func VPCToT1(in *crdv1alpha1.SecurityPolicy, out *v1alpha1.SecurityPolicy) {
	out = (*v1alpha1.SecurityPolicy)(unsafe.Pointer(in))
	out.APIVersion = "nsx.vmware.com/v1alpha1"
	controllerutil.RemoveFinalizer(out, common.SecurityPolicyFinalizerName)
	controllerutil.AddFinalizer(out, common.T1SecurityPolicyFinalizerName)
}

```
and it is more concise for us, what do you think?

cmd/main.go
zhengxiexie: @dantingl @timdengyun What's your opinion? How about arranging it in the next PR? It is not urgent.

dantingl: Please create a github issue to address it in a following PR.

dantingl (commented):
Please check in after local e2e test is done

dantingl (approved):

cmd/main.go
zhengxiexie: ok

timdengyun (approved):

PR #639, opened 2024-07-17
[TenantManager][ExternalAddressBinding] Add CRD

gran-vmv:


pkg/apis/nsx.vmware.com/v1alpha1/externaladdressbinding_types.go
```diff
@@ -0,0 +1,39 @@
+package v1alpha1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+type ExternalAddressBindingSpec struct {
```
jianjuns, line 5: How about call it AddressBinding, in case later we will support Project (Private_TGW) IP too?

pkg/apis/nsx.vmware.com/v1alpha1/externaladdressbinding_types.go
```diff
@@ -0,0 +1,39 @@
...
+	VMName        string `json:"vmName"`
+	InterfaceName string `json:"interfaceName"`
```
jianjuns, line 7: Could we make it optional, and if not set use the first interface of the VM (given most VMs have only a single interface)?

pkg/apis/nsx.vmware.com/v1alpha1/externaladdressbinding_types.go
gran-vmv: Fixed.

pkg/apis/nsx.vmware.com/v1alpha1/externaladdressbinding_types.go
gran-vmv: Added `omitempty` for this.

pkg/apis/nsx.vmware.com/v1alpha1/addressbinding_types.go
```diff
@@ -0,0 +1,41 @@
+package v1alpha1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+type AddressBindingSpec struct {
+	// VMName contains the VM's name
+	VMName string `json:"vmName"`
```
jianjuns, line 7: Question - when later we support address binding for non-VM ports, e.g. PodVM, then we may have to make this one optional; is that an incompatible API change and we may have to update the API version? If yes, I would add omitempty for this one too.

jianjuns (approved):
LGTM.

pkg/apis/nsx.vmware.com/v1alpha1/addressbinding_types.go
gran-vmv: Fixed. Thanks.

pkg/apis/nsx.vmware.com/v1alpha1/addressbinding_types.go
```diff
@@ -0,0 +1,41 @@
...
+	Status AddressBindingStatus `json:"status"`
+}
+
+//+kubebuilder:object:root=true
+
+// AddressBindingList contains a list of AddressBinding.
+type AddressBindingList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []NetworkInfo `json:"items"`
```
jianjuns, line 36: NetworkInfo?

gran-vmv: Fixed. Thanks.

pkg/apis/nsx.vmware.com/v1alpha1/addressbinding_types.go
tnqn: I think it's compatible to change a required field to optional as the change won't break users who use the old client that requires the field. And it doesn't need to increase API version.

jianjuns: Ok, in this case, let us keep this field required for now. @gran-vmv

gran-vmv: Removed omitempty for VMName

jianjuns (approved):

dantingl (approved):

PR #640, opened 2024-07-18
Add support for PrincipalIdentity

TaoZou1:
Add support for PrincipalIdentity.
Log unsupported type.
Return error if CasttoPointer result is nil


zhengxiexie (approved):
LGTM

gran-vmv (approved):

PR #646, opened 2024-07-23
Chang LabelLbIngressIpMode label value

timdengyun:
This patch is to change LabelLbIngressIpMode label from  tanzu.vmware.com/ingress-ip-mode to nsx.vmware.com/ingress-ip-mode

### DIFF
--- a/pkg/nsx/services/common/types.go
+++ b/pkg/nsx/services/common/types.go
@@ -59,7 +59,7 @@ const (
 	LabelDefaultSubnetSet              string = "nsxoperator.vmware.com/default-subnetset-for"
 	LabelDefaultVMSubnetSet            string = "VirtualMachine"
 	LabelDefaultPodSubnetSet           string = "Pod"
-	LabelLbIngressIpMode               string = "tanzu.vmware.com/ingress-ip-mode"
+	LabelLbIngressIpMode               string = "nsx.vmware.com/ingress-ip-mode"
 	LabelLbIngressIpModeVipValue       string = "vip"
 	LabelLbIngressIpModeProxyValue     string = "proxy"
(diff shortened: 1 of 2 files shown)


TaoZou1 (approved):

pkg/nsx/services/common/types.go
```diff
@@ -59,7 +59,7 @@ const (
 	LabelDefaultSubnetSet              string = "nsxoperator.vmware.com/default-subnetset-for"
```
jianjuns, line 59: Why this one alone uses "nsxoperator.vmware.com"?

ihgann (approved):
Looks good. I think @jianjuns' comment is relevant, but possibly a separate change?

pkg/nsx/services/common/types.go
timdengyun: I think this label we used it to indicate the default subnetset is created for VM or POD,
like
labels:
    nsxoperator.vmware.com/default-subnetset-for: VirtualMachine
Do we also need to change it:
    nsx.vmware.com/default-subnetset-for?

timdengyun:
> Looks good. I think @jianjuns' comment is relevant, but possibly a separate change?

Thanks @ihgann, Jianjuan's comments is not directly relevant. We might put change in another patch.

pkg/nsx/services/common/types.go
jianjuns: If we follow the way to use nsxoperator.vmware.com for NSX Operator specific logic (not exposed to end users), I am fine with that too.

timdengyun: Thanks.  We keep this specific label as it is for internal use.

PR #649, opened 2024-07-25
Fix nil panic when an unexpected NSX response is received

wenqiq:
Fix the panic caused by invalid memory address or nil pointer dereference when an unexpected NSX response is received.


timdengyun (approved):
LGTM

timdengyun (approved):
LGTM

zhengxiexie (approved):

PR #653, opened 2024-07-30
Update code-generator for new apiVersion

zhengxiexie:


dantingl (dismissed):

pkg/client/clientset/versioned/typed/nsx.vmware.com/v1alpha1/nsxserviceaccount.go
```diff
@@ -1,182 +0,0 @@
-/* Copyright © 2024 VMware, Inc. All Rights Reserved.
```
gran-vmv, line 1: Is this file no longer needed? Who is the consumer of this clientset/informer/lister?

dantingl (approved):

pkg/controllers/subnetset/subnetset_webhook.go
```diff
@@ -29,7 +29,7 @@ var NSXOperatorSA = "system:serviceaccount:vmware-system-nsx:ncp-svc-account"
 
 type SubnetSetValidator struct {
 	Client  client.Client
-	decoder *admission.Decoder
+	decoder admission.Decoder
```
timdengyun, line 32: why need change this file?

zhengxiexie: When building
go: toolchain upgrade needed to resolve k8s.io/client-go/discovery
go: k8s.io/client-go@v0.30.3 requires go >= 1.22.0; switching to go1.22.5

And then controller-runtime/pkg/webhook/admission is not compatible with the previous, so I updated it.

timdengyun (approved):

PR #658, opened 2024-07-31
Fix nil pointer in subnet status

zhengxiexie:


wenqiq:
It seems that this PR is duplicated with https://github.com/vmware-tanzu/nsx-operator/pull/649?

zhengxiexie:
> It seems that this PR is duplicated with #649?

OK，closed

PR #659, opened 2024-07-31
Fix subnet/subneset ipv4_subnet_size setting

timdengyun:
Fix the bug that the NSX subnet IPv4SubnetSize is not set by user-defined CR, either from subnet CR or subnetset CR.
The value of IPv4SubnetSize is only set from vpcnetworkconfiguration if there is no value defined in subnet CR or subnetset CR.


wenqiq (approved):
LGTM

wenqiq:
Should we update the defaultSubnetSize field in the e2e test manifest in this PR?

timdengyun:
> Should we upda

I put change in another PR.

yanjunz97 (approved):
LGTM

PR #661, opened 2024-08-01
Change VPCNetworkconfiguration DefaultIPv4SubnetSize to 32

timdengyun:
NSX only allows the subnet IPv4SubnetSize to be power of 2, i.e. 16, 32, 64 etc.
So, this patch is to change VPCNetworkconfiguration DefaultIPv4SubnetSize to 32.


wenqiq (approved):
lgtm

jianjuns:
@lxiaopei : please confirm what is the default size for WCP. NSX is using 64; if no special reason we should keep it consistent.

jianjuns:
I found WCP with T1 is just using 16 as default size. In this case, probably let us do 32 for now.

jianjuns (approved):

PR #662, opened 2024-08-01
Remove NSXApiError wrapping for non-NSX client

heypnus:
If the error is not returned by NSX API, e.g. K8s client, we should not wrap the error by NSXApiError.


timdengyun (approved):

zhengxiexie (approved):
LGTM

PR #663, opened 2024-08-01
Check loadbalancer type

TaoZou1:
Check loadbalancer type
    
    if use_avi_lb is true in the config, nsx-operator will also check
    1. if alb endpoint existed
    2. no nsx lbs resources found
    if all conditions met, nsx-operator will return 'avi', else return
    'nsx-lb'
    
    Test Done:
    1. set use_avi_lb true in config, get avi endpoint, no nsx lbs , return "avi"
    2. set use_avi_lb true in config, no avi endpoint, return "nsx-lb"
    3. set use_avi_lb false in config, return "nsx-lb"

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -411,13 +411,6 @@ func (nsxConfig *NsxConfig) ValidateConfigFromCmd() error {
 	return nsxConfig.validate(true)
 }
 
-func (nsxConfig *NsxConfig) NSXLBEnabled() bool {
-	if nsxConfig.UseAVILoadBalancer == false && (nsxConfig.UseNSXLoadBalancer == nil || *nsxConfig.UseNSXLoadBalancer == true) {
-		return true
-	}
-	return false
-}
-
 func (nsxConfig *NsxConfig) GetNSXLBSize() string {
 	lbsSize := nsxConfig.NSXLBSize
 	if lbsSize == "" {
--- a/pkg/nsx/services/vpc/vpc.go
+++ b/pkg/nsx/services/vpc/vpc.go
@@ -12,6 +12,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/retry"
 
 	"github.com/vmware-tanzu/nsx-operator/pkg/apis/vpc/v1alpha1"
 	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
@@ -30,14 +31,18 @@ const (
 	GroupKey                   = "/orgs/%s/projects/%s/vpcs/%s/groups/%s"
 	SecurityPolicyKey          = "/orgs/%s/projects/%s/vpcs/%s/security-policies/%s"
 	RuleKey                    = "/orgs/%s/projects/%s/vpcs/%s/security-policies/%s/rules/%s"
+	albEndpointPath            = "policy/api/v1/infra/sites/default/enforcement-points/alb-endpoint"
+	LBProviderNSX              = "nsx-lb"
+	LBProviderAVI              = "avi"
 )
 
 var (
-	log             = &logger.Log
-	ctx             = context.Background()
-	ResourceTypeVPC = common.ResourceTypeVpc
-	NewConverter    = common.NewConverter
-
+	log                       = &logger.Log
+	ctx                       = context.Background()
+	ResourceTypeVPC           = common.ResourceTypeVpc
+	NewConverter              = common.NewConverter
+	lbProvider                = ""
+	lbProviderMutex           = &sync.Mutex{}
 	MarkedForDelete           = true
 	enableAviAllowRule        = false
 	EnforceRevisionCheckParam = false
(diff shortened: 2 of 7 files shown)


pkg/nsx/services/vpc/vpc.go
```diff
@@ -974,3 +976,39 @@ func (s *VPCService) GetNSXLBSPath(lbsId string) string {
 	}
 	return *vpcLBS.Path
 }
+
+func GetAlbEndpoint(cluster *nsx.Cluster) error {
+	_, err := cluster.HttpGet(albEndpointPath)
+	return err
+}
+
+func (vpcService *VPCService) GetLbProvider() string {
```
gran-vmv, line 985: Do you want to change existing lb type check to this method? Check `NSXLBEnabled()` and `UseAVILoadBalancer` in whole project.

TaoZou1: it could be used to replace the logic inside NSXLBEnabled(), NSXLBEnabled could call GetLbProvider() == "nsx_lb"

pkg/config/config.go
```diff
@@ -384,7 +384,6 @@ func (nsxConfig *NsxConfig) validateCert() error {
 	}
 	return nil
 }
-
```
yuntanghsu, line 387: I think we need this empty line here?

yuntanghsu (approved):
LGTM, just one nit

pkg/config/config.go
TaoZou1: done

pkg/nsx/services/common/types.go
```diff
@@ -126,6 +126,8 @@ const (
 	DstGroupSuffix          = "dst"
 	IpSetGroupSuffix        = "ipset"
 	SharePrefix             = "share"
+	NSX_LB                  = "nsx-lb"
```
dantingl, line 129: Let's use a function to identify whether we use AVI or not, return true/false instead of types

pkg/nsx/services/vpc/vpc.go
dantingl: I agree, please replace NSXLBEnabled() or UseAVILoadBalancer()

pkg/nsx/services/common/types.go
TaoZou1: done

pkg/nsx/services/vpc/vpc.go
TaoZou1: move NSXLBEnabled to vpc

pkg/nsx/services/vpc/vpc.go
```diff
@@ -585,7 +588,7 @@ func (s *VPCService) CreateOrUpdateVPC(obj *v1alpha1.NetworkInfo) (*model.Vpc, *
```
dantingl, line 588: line 577: the we should also call NSXLBEnabled or getLBProvider to set AVI flag in VPC API

pkg/nsx/services/vpc/vpc.go
```diff
@@ -970,3 +973,49 @@ func (s *VPCService) GetNSXLBSPath(lbsId string) string {
...
+	lbProviderMutex.Lock()
+	defer lbProviderMutex.Unlock()
+
+	if lbProvider == "" {
+		lbProvider = vpcService.getLBProvider()
+	}
+	return lbProvider == common.NSX_LB
+}
+
+func (vpcService *VPCService) getLBProvider() string {
```
dantingl, line 992: Can we rename it like UseAVILoadBalancer, so the return value will always be true or false, no need to define constant type names

TaoZou1: it copied the NCP implement.
I have moved definition from common to vpc internal since only NSXLBEnabled  provided. There is no need to define constant type names

TaoZou1: and is it possible that we could other lb ? use type name inside vpc seems ok.

heypnus:
Please polish your title to provide enough info.

pkg/nsx/services/vpc/vpc.go
```diff
@@ -30,14 +31,18 @@ const (
 	GroupKey                   = "/orgs/%s/projects/%s/vpcs/%s/groups/%s"
 	SecurityPolicyKey          = "/orgs/%s/projects/%s/vpcs/%s/security-policies/%s"
 	RuleKey                    = "/orgs/%s/projects/%s/vpcs/%s/security-policies/%s/rules/%s"
+	albEndpointPath            = "policy/api/v1/infra/sites/default/enforcement-points/alb-endpoint"
+	NSX_LB                     = "nsx-lb"
+	AVI                        = "avi"
```
heypnus, line 36: `NSX_LB` - > `LBProviderNSX`
`AVI` -> `LBPrivoderAVI`
or some other names like that. Otherwise, `NSX_LB` doesn't match the naming style, and these 2 var name doesn't include their purpose.

pkg/nsx/services/vpc/vpc.go
```diff
@@ -728,3 +733,49 @@ func (s *VPCService) GetNSXLBSPath(lbsId string) string {
 	}
 	return *vpcLBS.Path
 }
+
+func GetAlbEndpoint(cluster *nsx.Cluster) error {
+	_, err := cluster.HttpGet(albEndpointPath)
+	return err
+}
+
+func (vpcService *VPCService) NSXLBEnabled() bool {
```
heypnus, line 742: This func should have a UT?

pkg/nsx/services/vpc/vpc.go
TaoZou1: fixed

pkg/nsx/services/vpc/vpc.go
TaoZou1: added.

heypnus (dismissed):

dantingl (approved):

ifelse2021 (approved):

heypnus (approved):

PR #668, opened 2024-08-05
check license from nsxt side

TaoZou1:
While init, it will check the license from nsxt side. If CONTAINER license is disable, it will reboot. If DFW license is disable, security policy will be only response for DELETE operation. It will run a routine to check license periodically. If there is no DFW license, it will check license more frequently

SecurityPolicy controller will check if error is invalid license error.

Test Done:
case:no CONTAINER license
1. if no CONTAINER license, nsx-operator should reset 

case:CONTAINER license enable, DFW disable
1. nsx-operator could bootup
2. security policy failed to create or update
3. security policy could be deleted 

case:CONTAINER license enable, DFW enable -> CONTAINER disable, DFW disable
1. nsx-operator restart due to DFW changed 

case:no DFW license, but nsx-operator try to create security policy
1. nsx-operator restart due to invalid license error

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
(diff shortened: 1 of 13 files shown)


dantingl (approved):

timdengyun (approved):
LGTM

PR #675, opened 2024-08-08
Rename crd.nsx.vmware.com dir to vpc +1

zhengxiexie:
Before rename dir name `nsx.vmware.com` to t1, `crd.nsx.vmware.com` to vpc, should add this dir in advance to let `go mod tidy` find this module.


timdengyun (approved):

dantingl (commented):
Can we call the directory crd instead of vpc? Which way is better?

zhengxiexie:
> Can we call the directory crd instead of vpc? Which way is better?

I think keeping consistent with build/yaml/crd/vpc(t1) is better.

TaoZou1 (approved):

PR #676, opened 2024-08-08
Support Subnet/SubnetSet AccessMode immutable

timdengyun:
For NSX subnet, once it is created, it's not allowed to change accessMode.
To support this restriction, for the Subnet/Subnetset CR, we also
add a method to make accessMode immutable once created.

This patch is to:
Change Subnet and SubnetSet CR definition to make Subnet/SubnetSet
accessMode is immutable once the CR is created.

Aslo, this patch is to
1. Remove checking pod-default subnetset in subnet controller.
2. Fix the issue that subnet CR spec is not updated after creating CR.

**Test done**:
subnet Test yaml:
```
cat vpctest-subnet-dhcp.yaml
kind: Subnet
metadata:
  name: user-pod-subnet-dhcp1
  namespace: vpc-test
spec:
  DHCPConfig:
    enableDHCP: true
  advancedConfig:
    staticIPAllocation:
      enable: false
  ipv4SubnetSize: 64
  #accessMode: Private
  accessMode: Publicd
```
 
1. Use invalid accessMode "Publicd"
kubectl apply -f vpctest-subnet-dhcp.yaml
output:
```
The Subnet "user-pod-subnet-dhcp1" is invalid:
* spec.accessMode: Unsupported value: "Publicd": supported values: "Private", "Public", "PrivateTGW"
* <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation
```

2. Create subnet with accessMode "Public"
accessMode: Public
kubectl apply -f vpctest-subnet-dhcp.yaml
output: subnet user-pod-subnet-dhcp1 CR is created correctly.
```
subnet.crd.nsx.vmware.com/user-pod-subnet-dhcp1 created
kubectl get Subnet -A
NAMESPACE   NAME                    ACCESSMODE   IPV4SUBNETSIZE   IPADDRESSES
vpc-test    user-pod-subnet-dhcp1   Public       64
```

3. Change subnet CR user-pod-subnet-dhcp1 accessMode to Private
accessMode is not allowed to change any more.
kubectl apply -f vpctest-subnet-dhcp.yaml
output: 
```
`The Subnet "user-pod-subnet-dhcp1" is invalid: spec.accessMode: Invalid value: "string": Value is immutable`
 kubectl get Subnet -A
NAMESPACE   NAME                    ACCESSMODE   IPV4SUBNETSIZE   IPADDRESSES
vpc-test    user-pod-subnet-dhcp1   Public       64
```
4.  Apply a subnet test yaml without accessMode set.
subnet Test yaml:
```
cat user-pod-subnet-dhcp2.yaml
kind: Subnet
metadata:
  name: user-pod-subnet-dhcp2
  namespace: vpc-test
spec:
  DHCPConfig:
    enableDHCP: true
  advancedConfig:
    staticIPAllocation:
      enable: false
  ipv4SubnetSize: 64
  #accessMode: Public
```
kubectl apply -f vpctest-subnet-dhcp.yaml, get
```
kubectl get subnet -nvpc-test    user-pod-subnet-dhcp2
NAME                    ACCESSMODE   IPV4SUBNETSIZE   IPADDRESSES
user-pod-subnet-dhcp2   Private      64
```
accessMode is set as Private by subnet controller since it's empty.
And then change accessMode, and apply this yaml again.
`accessMode: Public`
we get accessMode is not allowed to change.
```
kubectl apply -f ./vpctest-subnet-dhcp.yaml
The Subnet "user-pod-subnet-dhcp2" is invalid: spec.accessMode: Invalid value: "string": Value is immutable
```

5. Do the similar test for the SubnetSet yaml

### DIFF
--- a/pkg/apis/vpc/v1alpha1/subnet_types.go
+++ b/pkg/apis/vpc/v1alpha1/subnet_types.go
@@ -17,6 +17,7 @@ type SubnetSpec struct {
 	IPv4SubnetSize int `json:"ipv4SubnetSize,omitempty"`
 	// Access mode of Subnet, accessible only from within VPC or from outside VPC.
 	// +kubebuilder:validation:Enum=Private;Public;PrivateTGW
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
 	AccessMode AccessMode `json:"accessMode,omitempty"`
 	// Subnet CIDRS.
 	// +kubebuilder:validation:MinItems=0
--- a/build/yaml/crd/vpc/crd.nsx.vmware.com_subnets.yaml
+++ b/build/yaml/crd/vpc/crd.nsx.vmware.com_subnets.yaml
@@ -67,6 +67,9 @@ spec:
                 - Public
                 - PrivateTGW
                 type: string
+                x-kubernetes-validations:
+                - message: Value is immutable
+                  rule: self == oldSelf
               advancedConfig:
                 description: Subnet advanced configuration.
                 properties:
--- a/build/yaml/crd/vpc/crd.nsx.vmware.com_subnetsets.yaml
+++ b/build/yaml/crd/vpc/crd.nsx.vmware.com_subnetsets.yaml
@@ -67,6 +67,9 @@ spec:
                 - Public
                 - PrivateTGW
                 type: string
+                x-kubernetes-validations:
+                - message: Value is immutable
+                  rule: self == oldSelf
               advancedConfig:
                 description: Subnet advanced configuration.
                 properties:
--- a/go.mod
+++ b/go.mod
@@ -26,7 +26,7 @@ require (
 	github.com/prometheus/client_golang v1.16.0
 	github.com/sirupsen/logrus v1.9.0
 	github.com/stretchr/testify v1.8.4
-	github.com/vmware-tanzu/nsx-operator/pkg/apis v0.0.0-20240812065223-970ecf0e07ab
+	github.com/vmware-tanzu/nsx-operator/pkg/apis v0.0.0-20240813023528-cb525458c6ee
 	github.com/vmware-tanzu/nsx-operator/pkg/client v0.0.0-20240102061654-537b080e159f
 	github.com/vmware-tanzu/vm-operator/api v1.8.2
 	github.com/vmware/govmomi v0.27.4
--- a/pkg/apis/vpc/v1alpha1/subnetset_types.go
+++ b/pkg/apis/vpc/v1alpha1/subnetset_types.go
@@ -15,6 +15,7 @@ type SubnetSetSpec struct {
 	IPv4SubnetSize int `json:"ipv4SubnetSize,omitempty"`
 	// Access mode of Subnet, accessible only from within VPC or from outside VPC.
 	// +kubebuilder:validation:Enum=Private;Public;PrivateTGW
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
 	AccessMode AccessMode `json:"accessMode,omitempty"`
 	// Subnet advanced configuration.
 	AdvancedConfig AdvancedConfig `json:"advancedConfig,omitempty"`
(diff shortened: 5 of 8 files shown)


pkg/apis/v1alpha1/subnetset_types.go
```diff
@@ -14,7 +14,10 @@ type SubnetSetSpec struct {
 	// +kubebuilder:validation:Minimum:=16
 	IPv4SubnetSize int `json:"ipv4SubnetSize,omitempty"`
 	// Access mode of Subnet, accessible only from within VPC or from outside VPC.
-	// +kubebuilder:validation:Enum=Private;Public
+	// +kubebuilder:default:=Private
+	// +kubebuilder:validation:Enum=Private;Public;PrivateTGW
+	// +kubebuilder:validation:Required
```
dantingl, line 19: Is it required if we have default type

timdengyun: kubebuilder:default:=Private, this is not required, we can remove it if we don't want a default value.
but I think we want Private as default value

wenqiq: I think we don't need a default value here, in the subnet controller, the `AccessMode` will be set a default value: https://github.com/vmware-tanzu/nsx-operator/blob/2cc0f8de2fc1615b517c25b5e32bebfa51aeaa71/pkg/controllers/subnet/subnet_controller.go#L81

timdengyun: https://github.com/vmware-tanzu/nsx-operator/pull/664/files#diff-30b5978138c76f4191e9f5dbdc5c051570ab228c98ab270b9f7e30db1ae42d9eR66
we  will not have defaultSubnetAccessMode field any more in VPCNetworkConfiguration CR, which is being changed to podSubnetAccessMode(for POD subnetset only).
Also possibly, podSubnetAccessMode will be removed as well. 
That is to say there is no any subnet accessmode setting in VPCNetworkConfiguration CR any more. 
We can remove the default value: private from CRD definition at this moment.
If necessary, we can set default subnet/subnetset accessmode in the controller to the value we want in the future.

timdengyun: remove default value for both subnet and subnetset.

wenqiq (approved):
lgtm. Just a small question. Under what circumstances will this value `AccessMode` will be set to ""? Because I see that there is a condition in the code(`obj.Spec.AccessMode == ""`).

timdengyun:
> if obj.Spec.AccessMode == ""

for this condition, 
`if obj.Spec.AccessMode == ""`
This will happen when user don't set AccessMode in the subnet CR or subnetset CR. 

So, we set default value for AccessMode even if user don't set AccessMode value in the CR,
the AccessMode filed will be set as default value.

pkg/apis/vpc/v1alpha1/subnet_types.go
```diff
@@ -17,6 +17,8 @@ type SubnetSpec struct {
 	IPv4SubnetSize int `json:"ipv4SubnetSize,omitempty"`
 	// Access mode of Subnet, accessible only from within VPC or from outside VPC.
 	// +kubebuilder:validation:Enum=Private;Public;PrivateTGW
+	// +kubebuilder:validation:Required
```
dantingl, line 20: This `Required` in label does not matter, right? If we do not set it as required in CRD yaml definition.
To me I do not want to make AccessMode as required, we should have a default value either in definition or in NSX Operator code.

timdengyun: If you always want this field to have a value, that means, it's Required. The difference putting  > Required

dantingl (approved):

wenqiq (approved):
lgtm

wenqiq:
I still have a concern that is not related to this PR.       
I don't know what our API definition for users is. I feel a little strange about using both `omitempty` and `enum` at the same time, because `enum` indicates choosing one of three values(`Private;Public;PrivateTGW`), while `omitempty` indicates that users can leave this value blank(just not set the `AccessMode` field). It means that the user can choose not to set this value or set it to one of the three options.

PR #626, opened 2024-07-03
Implement ipaddressallocation controller

zhengxiexie:
IPAddressAllocation with Private ip_address_block_visibility can be realized successfully.
```
apiVersion: nsx.vmware.com/v1alpha1
kind: IPAddressAllocation
metadata:
  name: guestcluster-workers-a
  namespace: zx-test
spec:
  ip_address_block_visibility: Private
  allocation_size: 32
```
image

IPAddressAllocation with External ip_address_block_visibility can be realized successfully

```
apiVersion: nsx.vmware.com/v1alpha1
kind: IPAddressAllocation
metadata:
  name: guestcluster-workers-b
  namespace: zx-test
spec:
  ip_address_block_visibility: External
  allocation_size: 32
```
image

The default ip_address_block_visibility for IPAddressAllocation will be Private if ip_address_block_visibility is not defined.
```
apiVersion: nsx.vmware.com/v1alpha1
kind: IPAddressAllocation
metadata:
  name: guestcluster-workers-b
  namespace: zx-test
spec:
  allocation_size: 32
```
image

Both could be deleted successfully.

If the first ipblcok is exhausted, will try next one
```
apiVersion: nsx.vmware.com/v1alpha1
kind: IPAddressAllocation
metadata:
  name: guestcluster-workers-c
  namespace: zx-test-2
spec:
  ip_address_block_visibility: Private
  allocation_size: 131072
```
image

image

If all the blocks are exhausted, will return error
```
apiVersion: nsx.vmware.com/v1alpha1
kind: IPAddressAllocation
metadata:
  name: guestcluster-workers-d
  namespace: zx-test-2
spec:
  ip_address_block_visibility: Private
  allocation_size: 33554432
```
image

PR #685, opened 2024-08-13
Add staticroute e2e test

TaoZou1:
staticroute e2e test


zhengxiexie (commented):
Better to add 3 cases, staticroute for ip block visibility -- Private, External, None
Refer https://github.com/vmware-tanzu/nsx-operator/pull/626

PR #686, opened 2024-08-13
Fix a subnetport ID issue during check the port state

heypnus:
We changed the subnetport ID to `<pod/vm name>-<uid>` but when getting
the NSX port state, the operator still tried to get the resource via
object uid. This patch will fix the issue.


dantingl (approved):

wenyingd (approved):
LGTM

PR #704, opened 2024-08-19
Move VPCConnectivityProfilesClient from controller to services

wenqiq:
Move the VPCConnectivityProfilesClient call from controller to services

### DIFF
--- a/pkg/nsx/services/vpc/vpc.go
+++ b/pkg/nsx/services/vpc/vpc.go
@@ -459,6 +459,20 @@ func (s *VPCService) GetAVISubnetInfo(vpc model.Vpc) (string, string, error) {
 	return path, cidr, nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/nsx/services/vpc/vpc.go
```diff
@@ -459,6 +459,20 @@ func (s *VPCService) GetAVISubnetInfo(vpc model.Vpc) (string, string, error) {
 	return path, cidr, nil
 }
 
+func (s *VPCService) GetVpcConnectivityProfile(nc *common.VPCNetworkConfigInfo, vpcConnectivityProfilePath string) (*model.VpcConnectivityProfile, error) {
```
yanjunz97, line 462: Could we get VPCConnectivityProfile only by path like https://github.com/vmware-tanzu/nsx-operator/blob/8f09829755082af31326a653ea0affc2d1ddc786/pkg/nsx/services/vpc/vpc.go#L806 Should vpcConnectivityProfilePath always contain org and project?

yanjunz97 (approved):

PR #706, opened 2024-08-20
Fix the attachment id generation issue

heypnus:
If we use the NSX subnet port ID to generate the attachment id,
there may be the same id generated because the UUID generation logic
only read the first 16 bytes of the port ID. Then the NSX will report
the error "Port attachment ID xxx used by another segment port ...".

This patch will use the subnetport/pod UID as the input parameter.

Testing done: patch the fix in the problematic testbed and confirmed
that the issue can be fixed.


dantingl (approved):

TaoZou1:
could you explain why need change the id and remove id from compare?

dantingl (approved):

TaoZou1 (approved):

PR #716, opened 2024-08-22
Fix delete ipaddressallocation bug

zhengxiexie:
Fix this issue that if failed to create ipaddressallocation, to delete the stale also failed, for the reason that can't find the resource. Besides, also update patching error including getting error info to status.

### DIFF
--- a/pkg/nsx/services/ipaddressallocation/compare.go
+++ b/pkg/nsx/services/ipaddressallocation/compare.go
@@ -13,12 +13,15 @@ type (
 
 type Comparable = common.Comparable
 
-func (iap *IpAddressAllocation) Key() string {
-	return *iap.Id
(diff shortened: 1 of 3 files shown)


pkg/nsx/services/ipaddressallocation/compare.go
```diff
@@ -18,6 +18,9 @@ func (iap *IpAddressAllocation) Key() string {
 }
 
 func (iap *IpAddressAllocation) Value() data.DataValue {
+	if iap == nil {
```
dantingl, line 21: What does 'iap' mean?

zhengxiexie: ok, fix with ipa.

dantingl (approved):

wenqiq (approved):
lgtm

PR #720, opened 2024-08-23
Remove advancedConfig in subnet/subnetset CR

heypnus:
Testing done:
1. create the subnet CR with `DHCPConfig.enableDHCP=false`, check the NSX subnet has properties `dhcp_config.enable_dhcp=false` and `advanced_config.static_ip_allocation.enabled=true`
2. create the subnet CR with `DHCPConfig.enableDHCP=true`, check the NSX subnet has properties `dhcp_config.enable_dhcp=true` and `advanced_config.static_ip_allocation.enabled=false`
3. create the subnet CR without explict DHCPConfig, check the NSX subnet has properties `dhcp_config.enable_dhcp=false` and `advanced_config.static_ip_allocation.enabled=true`
4. create a new namespace and check the subnetset `vm-default` has spec `DHCPConfig.enableDHCP=false`, then create a VM in the namespace, check the new created NSX subnet has properties `dhcp_config.enable_dhcp=false` and `advanced_config.static_ip_allocation.enabled=true`


dantingl (dismissed):

pkg/nsx/services/subnet/builder.go
```diff
@@ -79,7 +79,7 @@ func (service *SubnetService) buildSubnet(obj client.Object, tags []model.Tag) (
 			DhcpConfig:     service.buildDHCPConfig(o.Spec.DHCPConfig.EnableDHCP, int64(o.Spec.IPv4SubnetSize-4)),
 			DisplayName:    String(service.buildSubnetSetName(o, index)),
 		}
-		staticIpAllocation = o.Spec.AdvancedConfig.StaticIPAllocation.Enable
+		staticIpAllocation = !o.Spec.DHCPConfig.EnableDHCP
```
seanpang-vmware, line 82: not sure about the logic here, if this is design intent, then all LGTM

seanpang-vmware (approved):

pkg/nsx/services/subnet/builder.go
timdengyun, line 82: we might need to handle subnets allocated from default subnetset. 
For the default-pod subnetset, we should always set StaticIPAllocation as True, since container can not support DHCP case.  DHCP value is also default false. Maybe we can also explicitly set DHCP as false for the default subnetset. 

As for default-vm subnetset, this is for WCP case, the default value had better to be true as well.

dantingl: it means if DHCPConfig is unset, we also need to enable staticIpAllocation, right?

heypnus: Yes, the `EnableDHCP` has the default value `false`, so if it's unset, it will be `false` and staticIpAllocation will be enabled.
```
type DHCPConfig struct {
	// +kubebuilder:default:=false
	EnableDHCP bool `json:"enableDHCP,omitempty"`
}
```
Here's the testing done:
Apply the new CRD, then
1. create the subnet CR with `DHCPConfig.enableDHCP=false`, check the NSX subnet has properties `dhcp_config.enable_dhcp=false` and `advanced_config.static_ip_allocation.enabled=true`
2. create the subnet CR with `DHCPConfig.enableDHCP=true`, check the NSX subnet has properties `dhcp_config.enable_dhcp=true` and `advanced_config.static_ip_allocation.enabled=false`
3. create the subnet CR without explict DHCPConfig, check the NSX subnet has properties `dhcp_config.enable_dhcp=false` and `advanced_config.static_ip_allocation.enabled=true`

dantingl (approved):

timdengyun (approved):

PR #723, opened 2024-08-26
Using reflect to get pointer

TaoZou1:
It's easy to forget to add type support in CasttoPointer. 
Change CasttoPointer to use reflect. 
The performance will be worse but no need to add each new type for it

### DIFF
--- a/pkg/nsx/util/utils.go
+++ b/pkg/nsx/util/utils.go
@@ -26,7 +26,6 @@ import (
 	apierrors "github.com/vmware/vsphere-automation-sdk-go/lib/vapi/std/errors"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/bindings"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
-	mpmodel "github.com/vmware/vsphere-automation-sdk-go/services/nsxt-mp/nsx/model"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 
 	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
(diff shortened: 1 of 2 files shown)


pkg/nsx/util/utils_test.go
```diff
@@ -534,6 +534,21 @@ func TestCasttoPointer(t *testing.T) {
 		{
 			name: "UnsupportedType",
 			obj:  model.Tag{},
+			want: &model.Tag{},
+		},
+		{
+			name: "PrincipalIdentit",
```
gran-vmv, line 540: PrincipalIdentity

pkg/nsx/util/utils_test.go
```diff
@@ -534,6 +534,21 @@ func TestCasttoPointer(t *testing.T) {
...
+			obj:  mpmodel.PrincipalIdentity{},
+			want: &mpmodel.PrincipalIdentity{},
+		},
+		{
+			name: "LBs",
```
gran-vmv, line 545: LBS ptr or LBService ptr

pkg/nsx/util/utils_test.go
```diff
@@ -534,6 +534,21 @@ func TestCasttoPointer(t *testing.T) {
...
+			obj:  &model.LBService{},
+			want: &model.LBService{},
+		},
+		{
+			name: "nil",
```
gran-vmv, line 550: Also add a case for "typed nil"?

TaoZou1: Done

pkg/nsx/util/utils_test.go
TaoZou1: done

pkg/nsx/util/utils_test.go
TaoZou1: done

pkg/nsx/util/utils.go
```diff
@@ -553,46 +552,16 @@ func FindTag(tags []model.Tag, tagScope string) string {
...
+	if obj == nil {
 		return nil
 	}
+	if reflect.TypeOf(obj).Kind() == reflect.Ptr {
+		return obj
+	}
+
+	pointer := reflect.New(reflect.TypeOf(obj))
+	pointer.Elem().Set(reflect.ValueOf(obj))
+	return pointer.Interface()
```
wenqiq, line 564: Go 1.18 and later versions support generics, I think it will make this function more type-safe and reduce its reliance on reflect.

gran-vmv: Maybe you can check `k8s.io/utils/ptr` to simplify this code.

gran-vmv (approved):

PR #733, opened 2024-08-29
Add day2 lb provider checking

TaoZou1:
Add day2 lb provider checking
    
    For day0, if there is no lb provider, only VPC will be created.
    For day2, lb provider should be ready.
    For day2, nsx-operator will check lb provider to create
    lbs for nsx-lb  or set endpoint for avi-lb
    
    While get avi port in cleanup, return nil if not found
    
   Test Done:
    Case 1:
    1. set use_avi_lb = true in ncp.ini
       alb_endpoint is absent
       *vpcConnectivityProfile.ServiceGateway.Enable is true
    2. run the manager, check if lb provider is nsx-lb
    Case 2:
    1. set use_avi_lb = false
       *vpcConnectivityProfile.ServiceGateway.Enable is false
       check if lb provider is none-lb
       create a ns, check if only vpc created
    2. update the *vpcConnectivityProfile.ServiceGateway.Enable to true
        check if lbs has been created in system ns
    Case 3:
    1. set use_avi_lb = true in ncp.ini
       alb_endpoint is created
       check if lb provider type is avi-lb

### DIFF
--- a/pkg/nsx/services/vpc/builder.go
+++ b/pkg/nsx/services/vpc/builder.go
@@ -28,16 +28,20 @@ func generateIPBlockKey(block model.IpAddressBlock) string {
 }
 
 func buildNSXVPC(obj *v1alpha1.NetworkInfo, nsObj *v1.Namespace, nc common.VPCNetworkConfigInfo, cluster string,
-	nsxVPC *model.Vpc, useAVILB bool) (*model.Vpc,
+	nsxVPC *model.Vpc, useAVILB bool, lbProviderChanged bool) (*model.Vpc,
 	error) {
 	vpc := &model.Vpc{}
 	if nsxVPC != nil {
 		// for upgrade case, only check public/private ip block size changing
-		if !IsVPCChanged(nc, nsxVPC) {
+		if !IsVPCChanged(nc, nsxVPC) && !lbProviderChanged {
 			log.Info("no changes on current NSX VPC, skip updating", "VPC", nsxVPC.Id)
 			return nil, nil
 		}
 		// for updating vpc case, use current vpc id, name
+		if useAVILB && lbProviderChanged {
+			loadBalancerVPCEndpointEnabled := true
+			nsxVPC.LoadBalancerVpcEndpoint = &model.LoadBalancerVPCEndpoint{Enabled: &loadBalancerVPCEndpointEnabled}
+		}
 		*vpc = *nsxVPC
 	} else {
 		// for creating vpc case, fill in vpc properties based on networkconfig
--- a/pkg/nsx/services/vpc/vpc.go
+++ b/pkg/nsx/services/vpc/vpc.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"sync"
 
+	apierrors "github.com/vmware/vsphere-automation-sdk-go/lib/vapi/std/errors"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -22,6 +23,8 @@ import (
 	"github.com/vmware-tanzu/nsx-operator/pkg/util"
 )
 
+type LBProvider string
+
 const (
 	VpcDefaultSecurityPolicyId = "default-layer3-section"
 	VPCKey                     = "/orgs/%s/projects/%s/vpcs/%s"
@@ -29,8 +32,10 @@ const (
 	SecurityPolicyKey          = "/orgs/%s/projects/%s/vpcs/%s/security-policies/%s"
 	RuleKey                    = "/orgs/%s/projects/%s/vpcs/%s/security-policies/%s/rules/%s"
 	albEndpointPath            = "policy/api/v1/infra/sites/default/enforcement-points/alb-endpoint"
-	LBProviderNSX              = "nsx-lb"
-	LBProviderAVI              = "avi"
+	edgeClusterPath            = "api/v1/edge-clusters"
+	NSXLB                      = LBProvider("nsx-lb")
+	AVILB                      = LBProvider("avi")
+	NoneLB                     = LBProvider("none")
 )
 
 var (
@@ -38,7 +43,7 @@ var (
 	ctx                       = context.Background()
 	ResourceTypeVPC           = common.ResourceTypeVpc
 	NewConverter              = common.NewConverter
-	lbProvider                = ""
+	globalLbProvider          = NoneLB
 	lbProviderMutex           = &sync.Mutex{}
 	MarkedForDelete           = true
 	EnforceRevisionCheckParam = false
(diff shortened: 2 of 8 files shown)


zhengxiexie (dismissed):
LGTM

pkg/nsx/services/vpc/builder.go
```diff
@@ -28,16 +28,20 @@ func generateIPBlockKey(block model.IpAddressBlock) string {
...
 	vpc := &model.Vpc{}
 	if nsxVPC != nil {
 		// for upgrade case, only check public/private ip block size changing
-		if !IsVPCChanged(nc, nsxVPC) {
+		if !IsVPCChanged(nc, nsxVPC) && !lbProviderChanged {
 			log.Info("no changes on current NSX VPC, skip updating", "VPC", nsxVPC.Id)
 			return nil, nil
 		}
 		// for updating vpc case, use current vpc id, name
+		if useAVILB && lbProviderChanged {
```
dantingl, line 41: no need to check lbProviderChanged, if useAVILB is true, then enable loadBalancerVPCEndpointEnabled?

pkg/nsx/services/vpc/vpc.go
```diff
@@ -627,7 +631,35 @@ func (s *VPCService) GetVpcConnectivityProfile(nc *common.VPCNetworkConfigInfo,
...
+*/
+func (s *VPCService) IsLBProviderChanged(existingVPC *model.Vpc, lbProvider LBProvider) bool {
+	if existingVPC == nil {
+		return false
+	}
+	if lbProvider == AVILB {
+		if existingVPC.LoadBalancerVpcEndpoint.Enabled == nil {
+			return true
+		} else {
+			if !*existingVPC.LoadBalancerVpcEndpoint.Enabled {
```
dantingl, line 648: Why not use:
if existingVPC.LoadBalancerVpcEndpoint.Enabled == nil ||
  !*existingVPC.LoadBalancerVpcEndpoint.Enabled {
    return true
}

pkg/nsx/services/vpc/builder.go
```diff
@@ -28,16 +28,20 @@ func generateIPBlockKey(block model.IpAddressBlock) string {
...
 
 func buildNSXVPC(obj *v1alpha1.NetworkInfo, nsObj *v1.Namespace, nc common.VPCNetworkConfigInfo, cluster string,
-	nsxVPC *model.Vpc, useAVILB bool) (*model.Vpc,
+	nsxVPC *model.Vpc, useAVILB bool, lbProviderChanged bool) (*model.Vpc,
 	error) {
 	vpc := &model.Vpc{}
 	if nsxVPC != nil {
 		// for upgrade case, only check public/private ip block size changing
-		if !IsVPCChanged(nc, nsxVPC) {
+		if !IsVPCChanged(nc, nsxVPC) && !lbProviderChanged {
```
dantingl, line 36: I think you should put lbProviderChanged logic in IsVPCChanged, otherwise IsVPCChanged should rename to IsVPCPrivateCIDRChanged

pkg/nsx/services/vpc/builder.go
TaoZou1: if the lbProviderChanged is false, that means is AVILB and loadBalancerVPCEndpointEnabled is true.

pkg/nsx/services/vpc/builder.go
TaoZou1: vpc doesn't include lb provider, LBS created with VPC is just due to using HAPI.  without lb Provider, VPC could also be created.

pkg/nsx/services/vpc/vpc.go
TaoZou1: updated

pkg/nsx/services/vpc/vpc.go
```diff
@@ -627,7 +631,31 @@ func (s *VPCService) GetVpcConnectivityProfile(nc *common.VPCNetworkConfigInfo,
...
 
-func (s *VPCService) CreateOrUpdateVPC(obj *v1alpha1.NetworkInfo, nc *common.VPCNetworkConfigInfo) (*model.Vpc, error) {
+/*
+IsLBProviderChanged is used to judge if the lb provider is changed from day0 to day2
+
+	The lb provider is allowed to be NoneLB in day0, and changed to AIVLB or NSXLB in day2
+	return true if the lb provider is changed
+*/
+func (s *VPCService) IsLBProviderChanged(existingVPC *model.Vpc, lbProvider LBProvider) bool {
+	if existingVPC == nil {
```
wenyingd, line 641: Is it easier to call this function with the condition that existing VPC is not nil? Just feel it is not necessary to add such check if we can have some pre-check or guidance in the caller.

pkg/nsx/services/vpc/vpc.go
```diff
@@ -856,13 +884,14 @@ func (s *VPCService) ValidateGatewayConnectionStatus(nc *common.VPCNetworkConfig
...
 	log.Info("cleaning up vpcs", "Count", len(vpcs))
+	lbProvider := s.GetLBProvider()
 	for _, vpc := range vpcs {
 		select {
 		case <-ctx.Done():
 			return errors.Join(nsxutil.TimeoutFailed, ctx.Err())
 		default:
 			// first clean avi subnet ports, or else vpc delete will fail
-			if !s.NSXLBEnabled() {
+			if lbProvider == AVILB {
```
wenyingd, line 894: Will it cause stale Avi settings with this case: In round 1, Avi is detected, then nsx-operator restarted; In round 2 (after nsx-operator restarts), Avi is down, then NSX-LB is detected. An old vpc (created in round 1) is deleted. Then we lost the chance to clean-up the Avi subnet ports?

pkg/nsx/services/vpc/builder.go
dantingl: Add TODO to put lbProviderChanged into IsVPCChanged, and move IsVPCChanged out of buildNSXVPC

pkg/nsx/services/vpc/vpc.go
```diff
@@ -233,204 +238,6 @@ func (s *VPCService) DeleteVPC(path string) error {
 	return nil
 }
 
-func (s *VPCService) addClusterTag(query string) string {
```
dantingl, line 236: Why removed all these new changes?

pkg/nsx/services/vpc/vpc.go
```diff
@@ -784,7 +615,7 @@ func (s *VPCService) CreateOrUpdateVPC(obj *v1alpha1.NetworkInfo, nc *common.VPC
 }
 
 func (s *VPCService) GetGatewayConnectionTypeFromConnectionPath(connectionPath string) (string, error) {
-	/* examples of connection_path:
+	/* examples of connection_pat
```
dantingl, line 618: connection_path

pkg/nsx/services/vpc/vpc.go
```diff
@@ -887,86 +719,6 @@ func (s *VPCService) Cleanup(ctx context.Context) error {
 		}
 	}
 
-	// Delete NCP created resources (share/sharedResources/cert/LBAppProfile/LBPersistentProfile
```
dantingl, line 890: same here

pkg/nsx/services/vpc/vpc.go
```diff
@@ -1005,50 +757,82 @@ func (s *VPCService) GetNSXLBSPath(lbsId string) string {
...
 	}
-	return lbProvider == LBProviderNSX
-}
 
-func (vpcService *VPCService) getLBProvider() string {
+	ncName := common.SystemVPCNetworkConfigurationName
+	netConfig, _ := vpcService.GetVPCNetworkConfig(ncName)
+	nc := &netConfig
+
+	edgeEnable := vpcService.EdgeClusterEnabled(nc)
```
dantingl, line 802: Just need to check CR status about AutoSnatEnabled

pkg/nsx/services/vpc/vpc.go
TaoZou1: remove checking.

pkg/nsx/services/vpc/vpc.go
TaoZou1: it's due to wrong operation while switch branch. fixed.

pkg/nsx/services/vpc/vpc.go
TaoZou1: Done

pkg/nsx/services/vpc/vpc.go
TaoZou1: add checking

wenyingd (dismissed):
LGTM

zhengxiexie (approved):
LGTM

PR #744, opened 2024-09-04
Change the LBS name to default

wenqiq:
The LBS name will be changed from `<NetworkInfo-name>-<NetworkInfo-uid>` to `default`

### DIFF
--- a/pkg/nsx/services/vpc/builder.go
+++ b/pkg/nsx/services/vpc/builder.go
@@ -92,7 +92,6 @@ func buildNSXVPC(obj *v1alpha1.NetworkInfo, nsObj *v1.Namespace, nc common.VPCNe
 func buildNSXLBS(obj *v1alpha1.NetworkInfo, nsObj *v1.Namespace, cluster, lbsSize, vpcPath string, relaxScaleValidation *bool) (*model.LBService, error) {
 	lbs := &model.LBService{}
 	lbsName := defaultLBSName
-	// Use VPC id for auto-created LBS id
 	lbs.Id = common.String(defaultLBSName)
 	lbs.DisplayName = &lbsName
 	lbs.Tags = util.BuildBasicTags(cluster, obj, nsObj.GetUID())
--- a/pkg/nsx/services/vpc/builder.go
+++ b/pkg/nsx/services/vpc/builder.go
@@ -1,6 +1,9 @@
 package vpc
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 	v1 "k8s.io/api/core/v1"
 
@@ -10,8 +13,8 @@ import (
 )
 
 var (
-	DefaultVPCIPAddressType               = "IPV4"
-	DefaultLoadBalancerVPCEndpointEnabled = true
+	DefaultVPCIPAddressType = "IPV4"
+	defaultLBSName          = "default"
 )
 
 // private ip block cidr is not unique, there maybe different ip blocks using same cidr, but for different vpc cr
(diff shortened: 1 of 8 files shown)


dantingl (approved):

pkg/nsx/services/vpc/builder.go
```diff
@@ -65,9 +66,9 @@ func buildNSXVPC(obj *v1alpha1.NetworkInfo, nsObj *v1.Namespace, nc common.VPCNe
 
 func buildNSXLBS(obj *v1alpha1.NetworkInfo, nsObj *v1.Namespace, cluster, lbsSize, vpcPath string, relaxScaleValidation *bool) (*model.LBService, error) {
 	lbs := &model.LBService{}
-	lbsName := util.GenerateIDByObjectByLimit(obj, common.MaxNameLength)
```
wenyingd, line 68: Please also update LBStore's keyFunc. If we use "default" as the Id for the lbs created in any VPC, we can't use it as the key since it is not unique any more. 
Besides, you may also need to check the consumption logic which now uses the vpc.Id (previously the same as lbs's id) as the key to get lbs in a VPC.

I would prefer to add these two changes in this patch,
1. Use "VPC.ID-lbs.ID" as the key in LBStore,
2. Introduce a new public function in LBStore to get the default LBS by VPC, then constructs the lbs Key as "VPC.ID-default", and call GetByKey to find the lbs. As for the origianl caller with GetByKey(VPC.id), update to consume the new function.

gran-vmv: Agree with wenying.

gran-vmv: If one VPC has only one LBS, it should be OK. Not sure if we can guarantee this.

gran-vmv: If one VPC has only one LBS, it should be OK. Not sure if we can guarantee this.
@TaoZou1 @dantingl Any suggestion?

pkg/nsx/services/vpc/vpc.go
```diff
@@ -997,8 +997,8 @@ func (service *VPCService) ListVPCInfo(ns string) []common.VPCResourceInfo {
 	return VPCInfoList
 }
 
-func (s *VPCService) GetNSXLBSPath(lbsId string) string {
```
wenyingd, line 1000: Add some comments here to explain why the caller shall provide the VPC's ID but not the LBS's ID?

Personally, I would more like to rename this function as `GetDefaultNSXLBSPathByVPC`, then we can have other functions to support getting NSX lb path directly by LBS's ID in future if multiple LBS are required in one VPC. Not strong opinion at this moment because we now only have one lbs per VPC.

wenyingd (commented):
Overall LGTM, a minor comment.

pkg/nsx/services/vpc/builder.go
TaoZou1: currently, one VPC has only one LBS

wenyingd (approved):

pkg/nsx/services/vpc/builder.go
```diff
@@ -65,9 +81,9 @@ func buildNSXVPC(obj *v1alpha1.NetworkInfo, nsObj *v1.Namespace, nc common.VPCNe
...
+	lbsName := defaultLBSName
 	// Use VPC id for auto-created LBS id
-	lbs.Id = common.String(util.GenerateIDByObject(obj))
+	lbs.Id = common.String(defaultLBSName)
```
TaoZou1, line 86: the id is also changed to 'default'? 
lbs.Id = &defaultLBSName

pkg/nsx/services/vpc/builder.go
```diff
@@ -10,8 +13,8 @@ import (
 )
 
 var (
-	DefaultVPCIPAddressType               = "IPV4"
-	DefaultLoadBalancerVPCEndpointEnabled = true
+	DefaultVPCIPAddressType = "IPV4"
+	defaultLBSName          = "default"
```
TaoZou1, line 17: use const instead of var?

pkg/nsx/services/vpc/builder.go
```diff
@@ -27,6 +30,19 @@ func generateIPBlockKey(block model.IpAddressBlock) string {
 	return *cidr + "_" + nsUID
 }
 
+func generateLBSKey(lbs model.LBService) string {
+	if lbs.ConnectivityPath == nil {
+		return ""
```
TaoZou1, line 35: which scenario the ConnectivityPath is a nil?  needs to log a error for LBService?

pkg/nsx/services/vpc/builder.go
wenqiq: It will be used as  `&defaultLBSName`.

dantingl (approved):

timdengyun (approved):

PR #749, opened 2024-09-05
Improve readability of subnetset/subnetport status message

yanjunz97:


pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -92,15 +93,18 @@ func (r *SubnetPortReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		nsxSubnetPath, err := r.GetSubnetPathForSubnetPort(ctx, subnetPort)
 		if err != nil {
 			log.Error(err, "failed to get NSX resource path from subnet", "subnetport", subnetPort)
+			updateFail(r, ctx, subnetPort, &err)
 			return common.ResultRequeue, err
 		}
 		labels, err := r.getLabelsFromVirtualMachine(ctx, subnetPort)
 		if err != nil {
 			log.Error(err, "failed to get labels from virtualmachine", "subnetPort.Name", subnetPort.Name, "subnetPort.UID", subnetPort.UID)
+			updateFail(r, ctx, subnetPort, &err)
```
dantingl, line 102: This should be fine to skip updating VM labels, right? The SubnetPort should still be ready=true?

yanjunz97: It seems getLabelsFromVirtualMachine only returns error when the VirtualMachine CR is not available. We may not move on SubnetPort creation if the corresponding VM is not available?

https://github.com/vmware-tanzu/nsx-operator/blob/72baaea6f4fc96ab3858d6d81f1ead929c20bde1/pkg/controllers/subnetport/subnetport_controller.go#L479-L494

heypnus (approved):

dantingl (approved):

PR #753, opened 2024-09-06
Fix delete static route using ID

TaoZou1:
While deleting static route, operator used the Id as VPCID. 
Fix that issue

Test Done:
1. replace nsx-operator binary in the live env 
2. check if the orphan static route has been deleted.


dantingl (approved):

zhengxiexie (approved):
LGTM

PR #755, opened 2024-09-06
Delete subnetport when the associated subnet is deleted

yanjunz97:
Sometimes the subnet may be deleted while the subnetport is created.
This PR adds a check on subnet for subnetport and re-create the subnetport when
the associated subnet does not exists.


dantingl (approved):

timdengyun (approved):
LGTM

PR #758, opened 2024-09-09
Add detailed error message for subnet/subnetset failures

yanjunz97:


wenqiq (commented):
Should detailed error messages be added in other places as well? I noticed that `updateFail` in several places just have empty strings.

yanjunz97:
> Should detailed error messages be added in other places as well? I noticed that `updateFail` in several places just have empty strings.

Updated, thanks. It makes sense to me to return all of those errors when updateFail. Please let me know if you think any of them are not necessary to be exposed to users.

wenqiq (approved):
LGTM. Do you think the updateFail in the subnetset controller should also be updated? @dantingl

dantingl:
> LGTM. Do you think the updateFail in the subnetset controller should also be updated? @dantingl

Yes, is there a way to update SubnetSet CR status as well?

dantingl (approved):

yanjunz97:
Included the error message for SubnetSet failures. And also updated for deleteFail to propagate meaningful error message. Please take another look.

wenqiq (approved):
lgtm

dantingl (approved):

PR #643, opened 2024-07-19
Add K8s CR name and uuid into NSX resource ID

wenyingd:
A pre-created VPC is possibly shared in mutiple K8s Namespaces even K8s
clusters, which required the corresponding NSX resources created in the same VPC
must use unique id.

To avoid NSX resources created by the K8s CRs in different Namespaces use the
same ID and to improve the readability of the reource ID, this change uses the
format ${cr_name}-${cr_uuid} to generate the NSX resource ID. For the scenario
that one K8s CR is translated to multiple NSX resources, e.g., Groups, a suffix
is added in the ID field to ensure it is unique. In the meanwhile, the prefix
which represents the resource type is removed from both the NSX resource id and
display_name.

Another change is for the VpcSubnet Id generated by SubnetSet CR. The index is
changed from a uuid to the hash of a uuid and its length is 8.

For the NSX resource's display_name, the corresponding CR's name is used to
improve its readability. The exceptions include Subnet and VPC (NSX LBs is also
changed), which uses format ${cr_name}-${cr_uuid}. This is because vCenter
create folders for subnet and VPC using its display_name, we use a UUID as a
suffix in the display_names field to ensure it is unique.
    
These NSX resources created because of K8s CRs are impacted in this change.
- Subnet and Subnetset
- SubnetPort
- SecurityPolicy and NetworkPolicy
- StaticRoute
- Group
- IPAllocation
    
Test Done:
Unit test is added.
e2e test
      - K8s CR: subnet/subnetset
      - K8s CR: staticroute
      - K8s CR: securitypolicy and networkpolicy (including Groups and Rule)     
      - K8s CR: subnetport

```
---------- kubectl get subnets ----------
NAME      Namespace        UID
subnet1   vpc2-namespace   d578ef4b-e1d2-471b-88f7-f6885077efae
********** NSX API: get VpcSubnet, tag: nsx-op/subnet_name: subnet1 **********
{
  "id": "subnet1-d578ef4b-e1d2-471b-88f7-f6885077efae",
  "display_name": "3d7b2c3f-1b87-454f-9f38-5bdebd9d8214-subnet1"
}
---------- kubectl get staticroutes ----------
NAME     Namespace        UID
test-2   vpc2-namespace   50fd5059-55bf-4016-8c05-a195693644e0
********** NSX API: get StaticRoutes, tag: nsx-op/static_route_name: test-2 **********
{
  "id": "test-2-50fd5059-55bf-4016-8c05-a195693644e0",
[...]

PR #763, opened 2024-09-14
Unify NSX resoruce ID and name connector as underline

timdengyun:
Previsouly, we use hyphen "-" to connect strings when building NSX resource name, and use underline "_" to connect strings when builindg NSX resource ID.

This patch is to unify NSX resource ID and name connecotr as underline when building ID and name from K8s CR.

For the NSX resoruce ID and name convention,
this patch is follow the standard in PR:https://github.com/vmware-tanzu/nsx-operator/pull/643

These NSX resources name are impacted in this change:
VPC
Subnet
SubnetPort
SecurityPolicy and NetworkPolicy
NSGroup and IPSetGroup
NSRule
Share
StaticRoute
IPAllocation

This patch is aslo to add two flags for GC and vpcCleanup in order to
distinguish these two cases.

Test Done:
1. Create ingress Networkpolicy
```
kind: NetworkPolicy
metadata:
  name: np-app-access-ing
  namespace: ns-2
  annotations:
    ncp/log_traffic: "OFF"
spec:
  podSelector:
    matchLabels:
      role: db
  ingress:
  - from:
    - podSelector:
        matchLabels:
          app: coffee
    - namespaceSelector:
        matchLabels:
          ns-name: ns-3
      podSelector:
        matchLabels:
          app: tea
    - ipBlock:
        cidr: 172.26.0.8/29
        except:
        - 172.26.0.11/32
    ports:
    - protocol: TCP
      port: 6001
    - protocol: TCP
      port: 7000
      endPort: 7010
```

The generated ingress allow SecurityPolicy:
```
SecurityPolicy name: np-app-access
SecurityPolicy ID: np-app-access_d9dabc6b-9668-4cd1-9c52-264ff60be3f8_allow
```
Rule:
```
Name: TCP.6001_ingress_allow
ID: np-app-access_d9dabc6b-9668-4cd1-9c52-264ff60be3f8_allow_0_6c2a026ca143812daa72699fb924ee36b33b5cdc_0_0
```

The generated ingress isolation SecurityPolicy:
```
SecurityPolicy name:np-app-access
SecurityPolicy ID: np-app-access_d9dabc6b-9668-4cd1-9c52-264ff60be3f8_isolation
```
Rule:
```
Name: ingress_isolation
ID:np-app-access_d9dabc6b-9668-4cd1-9c52-264ff60be3f8_isolation_0_114fed106ef3b5eae2a583f312435e84c02ca97f_0_0
```

Project Share
```
name: dy-project1_group_np-app-access_0_src_share 
ID: dy-project1_group_np-app-access_d9dabc6b-9668-4cd1-9c52-264ff60be3f8_allow_0_src_share
```

Project Group
```
Name: np-app-access_0_src
ID: np-app-access_d9dabc6b-9668-4cd1-9c52-264ff60be3f8_allow_0_src
```

VPC NSGroup:
```
Name: test_np-app-access_scope
ID: np-app-access_d9dabc6b-9668-4cd1-9c52-264ff60be3f8_isolation_scope
```

2. Create/delete SecurityPolicy in VPC mode.

3.  Create/delete SecurityPolicy with namedport
The generated IPSet group:
```
Name: TCP.db-port.TCP.3366_egress_allow_ipset
ID: sp-namedport-web_90fc82ff-395e-47d0-bd02-534d71b2e434_2_6227e9f64a8ca9ab0e34891f3725a4357bc8d917_0_1_ipset
Name: TCP.db-port.TCP.80_egress_allow_ipset
ID: sp-namedport-web_90fc82ff-395e-47d0-bd02-534d71b2e434_2_6227e9f64a8ca9ab0e34891f3725a4357bc8d917_0_0_ipset
```

4. T1 upgrade case:
Create a SecurityPolicy in v4.2.1 code,

image

and start nsx operator with this patch to see if name are changed from "-" to "_".

image

5. Create SecurityPolicy in T1 and do GC

6. Create SecurityPolicy in VPC and do GC

7. Create NetworkPolicy in VPC and do GC


pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -1883,25 +1885,19 @@ func (service *SecurityPolicyService) buildRulePortsString(ports *[]v1alpha1.Sec
...
-
-func (service *SecurityPolicyService) BuildNetworkPolicyAllowPolicyID(uid string) string {
-	return fmt.Sprintf("%s_allow", uid)
-}
-
-func (service *SecurityPolicyService) BuildNetworkPolicyIsolationPolicyName(uid string) string {
-	return fmt.Sprintf("%s-isolation", uid)
+func (service *SecurityPolicyService) BuildNetworkPolicyAllowPolicyID(obj *networkingv1.NetworkPolicy) string {
+	suffix := "allow"
+	return util.GenerateIDByObjectWithSuffix(obj, suffix)
```
wenyingd, line 1897: This looks incorrect, it will generate the NetworkPolicy's ID as "$networkpolicy.name_$networkpolicy.uid_allow". If we use this as the id of the internal SecurityPolicy's UID, then the final generated NSX Policy ID should be "$networkpolicy.name_$networkpolicy.name_$networkpolicy.uid_allow" ?

pkg/util/utils.go
```diff
@@ -82,7 +82,7 @@ func NormalizeName(name string) string {
 }
 
 func normalizeNameByLimit(name string, suffix string, limit int) string {
```
wenyingd, line 84: This function is also called by `NormalizeName` and `NormalizeLabelKey` which are finally called by `updateExpressionsMatchLabels`, could you double check if the change has any impact on that function?

pkg/nsx/services/securitypolicy/firewall.go
```diff
@@ -304,12 +304,11 @@ func (service *SecurityPolicyService) populateRulesForIsolationSection(spIsolati
 }
 
 func (service *SecurityPolicyService) generateSectionForNetworkPolicy(networkPolicy *networkingv1.NetworkPolicy, sectionType string) (*v1alpha1.SecurityPolicy, error) {
```
wenyingd, line 306: Do you think we can add a test case to verify the final NSX security policy resources' ID and Display Name with an input of type NetworkPolicy CR?

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -417,13 +419,13 @@ func (service *SecurityPolicyService) buildAppliedGroupName(obj *v1alpha1.Securi
 	var rule *v1alpha1.SecurityPolicyRule
 	if ruleIdx != -1 {
 		rule = &(obj.Spec.Rules[ruleIdx])
-		ruleName := fmt.Sprintf("%s-%d", obj.Name, ruleIdx)
+		ruleName := fmt.Sprintf("%s_%d", obj.Name, ruleIdx)
```
wenyingd, line 422: Do you plan use another change to remove ruleIdx from the NSX rule id/name ?

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -1710,7 +1712,7 @@ func (service *SecurityPolicyService) updatePeerExpressions(obj *v1alpha1.Securi
 }
 
 func (service *SecurityPolicyService) buildShareName(nsxResourceID, groupName string) string {
-	nsxShareName := util.GenerateTruncName(common.MaxNameLength, fmt.Sprintf("%s-group-%s", nsxResourceID, groupName), "", common.ShareSuffix, "", "")
+	nsxShareName := util.GenerateTruncName(common.MaxNameLength, fmt.Sprintf("%s_group_%s", nsxResourceID, groupName), "", common.ShareSuffix, "", "")
```
wenyingd, line 1715: Is it better if we publish function `connectStrings`, and call it here other places where we manually use connect string like `fmt.Sprintf(%s_%s)`? And we can define a constant/variable for the default connector which doesn't need to expose.

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -663,11 +665,11 @@ func (service *SecurityPolicyService) buildRuleDisplayName(rule *v1alpha1.Securi
...
 		// For the security policy rule with namedPort, it will be expanded to the multiple security policy rules based on resolution of named port.
-		// e.g. input: security policy's rule name: TCP.http-UDP.1234-ingress-allow,
-		// expand to NSX security policy rules with name TCP.http-UDP.1234.TCP.80-ingress-allow and TCP.http-UDP.1234.UDP.1234-ingress-allow.
-		// in case that user defined input security policy's rule name: sp-namedport-rule,
-		// expand to NSX security policy rules with name sp-namedport-rule.TCP.80-ingress-allow and sp-namedport-rule.UDP.1234-ingress-allow.
-		index := strings.Index(ruleName, "-"+suffix)
+		// e.g. input: security policy's rule name: TCP.http_UDP.1234_ingress-allow,
+		// expand to NSX security policy rules with name TCP.http_UDP.1234_TCP.80_ingress-allow and TCP.http_UDP.1234_UDP.1234_ingress-allow.
+		// in case that user defined input security policy's rule name: sp_namedport_rule,
+		// expand to NSX security policy rules with name sp_namedport_rule.TCP.80_ingress-allow and sp_namedport_rule.UDP.1234_ingress-allow.
```
wenyingd, line 671: Why do we still use connector "-" between `TCP.80_ingress` an `allow` ?

pkg/nsx/services/securitypolicy/builder.go
timdengyun: Thanks so much for pointing this wrong change. I reverted my change.

pkg/util/utils.go
timdengyun: Yes, there is no impact to updateExpressionsMatchLabels. If there is label change, the controller will sync the new generated labels to nsx tags, so the resource will be updated.

pkg/nsx/services/securitypolicy/builder.go
timdengyun: Yes, I'm planning to do it once this PR is merged. So, we can narrow the scope of this PR.

pkg/nsx/services/securitypolicy/builder.go
timdengyun: Good suggestion!  We expose ConnectStrings as a common function for connecting the strings. 
Also, we define a var for common usage.
`ConnectorUnderline string = "_"`

pkg/nsx/services/securitypolicy/builder.go
timdengyun: Yes, that's right, it should be `TCP.80_ingress_allow`

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -38,18 +39,15 @@ var (
...
-	prefix := common.SecurityPolicyPrefix
-	if createdFor != common.ResourceTypeSecurityPolicy {
-		prefix = common.NetworkPolicyPrefix
-	}
-	// For T1 scenario, we use ns-name as the key resource name for SecurityPolicy, it is to be consistent with the
+	// For T1 scenario, we use ns_name as the key resource name for SecurityPolicy, it is to be consistent with the
 	// previous solutions.
-	return util.GenerateTruncName(common.MaxNameLength, fmt.Sprintf("%s-%s", obj.Namespace, obj.Name), prefix, "", "", "")
+	return util.GenerateTruncName(common.MaxNameLength, util.ConnectStrings(common.ConnectorUnderline, obj.Namespace, obj.Name),
+		common.SecurityPolicyPrefix, "", "", "")
```
wenyingd, line 50: I feel it is more effecient if we use `strings.Join([]string{obj.Namespace, obj.Name}, common.ConnectorUnderline)` here when the string contents are known and not empty

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -378,7 +376,7 @@ func (service *SecurityPolicyService) buildAppliedGroupID(obj *v1alpha1.Security
 	if IsVPCEnabled(service) {
 		suffix := common.TargetGroupSuffix
 		if ruleIdx != -1 {
-			suffix = fmt.Sprintf("%d_%s", ruleIdx, suffix)
+			suffix = util.ConnectStrings(common.ConnectorUnderline, strconv.Itoa(ruleIdx), suffix)
```
wenyingd, line 379: ditto

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -628,20 +626,20 @@ func (service *SecurityPolicyService) buildRuleDisplayName(rule *v1alpha1.Securi
 	if ruleDirection == "IN" {
```
wenyingd, line 626: Is this more clean?

```
	var ruleAct string
	switch ruleAction {
	case util.ToUpper(v1alpha1.RuleActionAllow):
		ruleAct = common.RuleActionAllow
	case util.ToUpper(v1alpha1.RuleActionDrop):
		ruleAct = common.RuleActionDrop
	case util.ToUpper(v1alpha1.RuleActionReject):
		ruleAct = common.RuleActionRejec
	}
	ruleDir := common.RuleEgress
	if ruleDirection == "IN" {
		ruleDir = common.RuleIngress
	}
	suffix := strings.Join([]string{ruleDir, ruleAct}, common.ConnectorUnderline)
```

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -650,10 +648,10 @@ func (service *SecurityPolicyService) buildRuleDisplayName(rule *v1alpha1.Securi
 		// if it has its own name generated, usually, it's for the internal isolation security policy rule created for network policy.
 		ruleName = rule.Name
 		if createdFor != common.ResourceTypeNetworkPolicy {
-			// If user defines the rule name, the generated NSX security policy rule will also be added with the same suffix: "-direction-action" as building rulePortsString
-			// e.g. input security policy's rule name: sp-rule,
-			// the generated NSX security policy rule name: sp-rule-ingress-allow
-			ruleName = ruleName + "-" + suffix
+			// If user defines the rule name, the generated NSX security policy rule will also be added with the same suffix: "_direction-action" as building rulePortsString
```
wenyingd, line 651:
```suggestion
			// If user defines the rule name, the generated NSX security policy rule will also be added with the same suffix: "_direction_action" as building rulePortsString
```

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -1709,13 +1708,15 @@ func (service *SecurityPolicyService) updatePeerExpressions(obj *v1alpha1.Securi
...
+func (service *SecurityPolicyService) buildShareName(nsxProjectID, groupName string) string {
+	resName := util.ConnectStrings(common.ConnectorUnderline, nsxProjectID, "group", groupName)
+	nsxShareName := util.GenerateTruncName(common.MaxNameLength, resName, "", common.ShareSuffix, "", "")
 	return nsxShareName
 }
 
-func (service *SecurityPolicyService) buildShareID(nsxResourceID, groupID string) string {
-	nsxShareId := util.GenerateID(fmt.Sprintf("%s_group_%s", nsxResourceID, groupID), "", common.ShareSuffix, "")
+func (service *SecurityPolicyService) buildShareID(nsxProjectID, groupID string) string {
+	resID := util.ConnectStrings(common.ConnectorUnderline, nsxProjectID, "group", groupID)
```
wenyingd, line 1718: ditto

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -1875,33 +1876,25 @@ func (service *SecurityPolicyService) buildRulePortString(port *v1alpha1.Securit
...
-	return fmt.Sprintf("%s-allow", uid)
-}
-
 func (service *SecurityPolicyService) BuildNetworkPolicyAllowPolicyID(uid string) string {
-	return fmt.Sprintf("%s_allow", uid)
-}
-
-func (service *SecurityPolicyService) BuildNetworkPolicyIsolationPolicyName(uid string) string {
-	return fmt.Sprintf("%s-isolation", uid)
+	return util.ConnectStrings(common.ConnectorUnderline, uid, common.RuleActionAllow)
```
wenyingd, line 1895: ditto

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -1875,33 +1876,25 @@ func (service *SecurityPolicyService) buildRulePortString(port *v1alpha1.Securit
...
 }
 
 func (service *SecurityPolicyService) BuildNetworkPolicyIsolationPolicyID(uid string) string {
-	return fmt.Sprintf("%s_isolation", uid)
+	return util.ConnectStrings(common.ConnectorUnderline, uid, common.RuleActionDrop)
```
wenyingd, line 1899: ditto

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -38,18 +39,15 @@ var (
...
 	}
-	prefix := common.SecurityPolicyPrefix
-	if createdFor != common.ResourceTypeSecurityPolicy {
-		prefix = common.NetworkPolicyPrefix
-	}
-	// For T1 scenario, we use ns-name as the key resource name for SecurityPolicy, it is to be consistent with the
+	// For T1 scenario, we use ns_name as the key resource name for SecurityPolicy, it is to be consistent with the
 	// previous solutions.
-	return util.GenerateTruncName(common.MaxNameLength, fmt.Sprintf("%s-%s", obj.Namespace, obj.Name), prefix, "", "", "")
+	return util.GenerateTruncName(common.MaxNameLength, util.ConnectStrings(common.ConnectorUnderline, obj.Namespace, obj.Name),
```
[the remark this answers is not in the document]
timdengyun: I'd like to unify the name T1 and VPC, which will save our code maintenance work, also good for users. 
This is no rick for upgrade case if only chaing display name. we will not change IDs format for T1, so, it will not
change created NSX resources in T1 just update time. 

I will unify the name in the next patch when removing rule idx, so this patch will revert this change at moment.

pkg/nsx/services/securitypolicy/builder.go
timdengyun: Thanks so much for suggestion, I refine the code as suggested, please take a look.

pkg/nsx/services/securitypolicy/builder.go
timdengyun: Good advice, it makes sense to me. I changed the code as suggested.

pkg/nsx/services/securitypolicy/builder.go
timdengyun: Fixed, thanks for advice.

pkg/nsx/services/securitypolicy/builder.go
timdengyun: Fixed, Thanks.

pkg/nsx/services/securitypolicy/builder.go
timdengyun: Fixed, Thanks

pkg/nsx/services/securitypolicy/builder.go
timdengyun: Fixed, Thanks

pkg/nsx/services/securitypolicy/builder.go
timdengyun: Fixed, Thanks.

pkg/nsx/services/securitypolicy/firewall.go
```diff
@@ -597,21 +597,22 @@ func (service *SecurityPolicyService) createOrUpdateVPCSecurityPolicy(obj *v1alp
 	return nil
 }
 
-func (service *SecurityPolicyService) DeleteSecurityPolicy(obj interface{}, isVPCCleanupOrGC bool, createdFor string) error {
+func (service *SecurityPolicyService) DeleteSecurityPolicy(obj interface{}, isGC, isVPCCleanup bool, createdFor string) error {
```
wenyingd, line 600: I find that all consumers in this change has set parameter "isGC" as `false`, is it really needed in the public function `DeleteSecurityPolicy`?

pkg/util/utils.go
```diff
@@ -421,24 +421,25 @@ func connectStrings(sep string, parts ...string) string {
 	return strings.Join(strParts, sep)
 }
 
-func GenerateDisplayName(res_name, prefix, suffix, project, cluster string) string {
-	// Return a string in this format: prefix-cluster-res_name-project-suffix.
-	return connectStrings("-", prefix, cluster, res_name, project, suffix)
+func GenerateDisplayName(connector, resName, prefix, suffix, project, cluster string) string {
```
wenyingd, line 424: Not introduced by you, but it looks we can make this function as private since it is only called in `GenerateTruncName`?

wenyingd (commented):
Overall LGTM, the last two comments are left.

pkg/nsx/services/securitypolicy/firewall.go
timdengyun: We set `isGC` as `true` in the networkpolicy and securitypolicy GC function:

https://github.com/vmware-tanzu/nsx-operator/pull/763/files#diff-816bade9860c7cd5eb294711ae56296dc1f400e3c4f546b52368482f01956b00R206

https://github.com/vmware-tanzu/nsx-operator/pull/763/files#diff-f45f721ba06ab477de36df3522bb24718a5f69d5747f3d6b5ae543fe98d68877R380

wenyingd: Got it, thanks for helping point it out.

pkg/util/utils.go
timdengyun: Good suggestion, Thanks. I made this function as a private function since not called by external world.

wenyingd (approved):
LGTM

TaoZou1 (approved):

PR #765, opened 2024-09-14
[WIP]Support dhcp config change at runtime

yanjunz97:
Pending subnet status fix from NSX to finish the testing


yanjunz97:
Closed as NSX does not support dhcp switch any more

PR #767, opened 2024-09-18
Unify the printer columns for VPCNetworkConfiguration 

wenyingd:
This change has removed the printer columns of private IPs and NSXProject from the additional printer columns of VPCNetworkConfiguration CR, and added the VPC path in the print by consuming field ".status.vpcs[0].vpcPath".

This is because,
1. the field ".spec.privateIPs" is not set in the case of using a pre-created VPC, which leads to the printer is empty in the corresponding CR
2. the VPC path has included the path of NSX project, which makes an additional column of NSXProject looks duplicated.

Test Done:
With this change, the output of command "kubectl get vpcnetworkconfigurations" looks as this,

```
# kubectl get vpcnetworkconfigurations
NAME                                                                  VPCPATH
default                                                               /orgs/default/projects/project-quality/vpcs/ft-dact-testns-user11-4-f0c83f65-38e8-4b36-b0fc-2cc2e522aaa3
imgpullbackoff-ns-d649d069-561c-4f85-90dd-2caeac1b3036                /orgs/default/projects/project-quality/vpcs/imgpullbackoff-ns-d3e17fc0-576f-4b56-a1ef-31a35df27fb5
largelayers-ns-57dc07f5-3b2e-467d-a26d-2f6c9d22102c                   /orgs/default/projects/project-quality/vpcs/largelayers-ns-f066181b-3a89-4303-9949-f994c261e58b
ood-ns-40742079-a383-4101-95cf-11faa0a362a7                           /orgs/default/projects/project-quality/vpcs/ood-ns-f8300c53-d890-4bad-8f72-1cb9e8ce2044
popularimages-ns-8c3f9232-348c-4643-a3f3-cb1d46266660                 /orgs/default/projects/project-quality/vpcs/popularimages-ns-7b948a6c-47f1-4ad2-b224-aabbd3500852
pre-vpc-c5852a8b-e633-4bb8-9821-2f78504fb801                          /orgs/default/projects/project-quality/vpcs/pre-vpc
pre-vpc3-0f5db49a-8d54-41ad-9b1c-3779f6c0702b                         /orgs/default/projects/project-quality/vpcs/pre-vpc3
sa-rolebindings-test-8be83728-0412-40db-bd2e-d598081a7651             /orgs/default/projects/project-quality/vpcs/sa-rolebindings-test-45a88b66-a833-49af-865c-b997138fd354
[...]

PR #768, opened 2024-09-18
Unify the printer columns for VPCNetworkConfiguration (#767)

wenyingd:
This change has removed "PrivateIPs" and "NSXProject" from the additional printer columns of VPCNetworkConfiguration CR, and added the VPC path in the print by consuming field ".status.vpcs[0].vpcPath". This is because,
1.the field ".spec.privateIPs" is not set in the case of using a pre-created
  VPC, which leads to the printer is empty in the corresponding CR
2.the VPC path has included the path of NSX project, which makes the contents
  in column of NSXProject looks duplicated.


TaoZou1 (approved):

wenqiq (approved):
LGTM

PR #769, opened 2024-09-19
Remove Subnet/SubnetSet Finalizer

wenqiq:
This PR tries to fix the issue that Subnet CR deletion takes about 2 minutes (due to the NSX IP release issue), resulting in a prolonged Namespace deletion time.
Remove NetworkInfo Finalizer in https://github.com/vmware-tanzu/nsx-operator/pull/778 

TestDone:
1. Create a Subnet and check its status; update the Subnet and check its status; delete the Subnet and check for any stale Subnet in NSX.(included in e2e test cases)
2. Create a SubnetSet and check its status; update the SubnetSet and check its status; delete the SubnetSet and check for any stale Subnet in NSX.(included in e2e test cases)
3. Rebase #778 and test the deletion of a Namespace, then delete the Pod, the Namespace will not stuck in terminating status.


pkg/controllers/common/utils.go
```diff
@@ -40,7 +40,7 @@ func AllocateSubnetFromSubnetSet(subnetSet *v1alpha1.SubnetSet, vpcService servi
 			return *nsxSubnet.Path, nil
 		}
 	}
-	tags := subnetService.GenerateSubnetNSTags(subnetSet, subnetSet.Namespace)
+	tags := subnetService.GenerateSubnetNSTags(subnetSet, subnetSet.Name, subnetSet.Namespace)
```
yanjunz97, line 43: It looks like the name/namespace are all fetched from the object, why we pass them separately?

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -49,94 +49,94 @@ type SubnetSetReconciler struct {
...
+		log.Info("Finished reconciling SubnetSet", "SubnetSet", req.NamespacedName, "time", time.Since(startTime))
+	}()
+
+	subnetsetCR := &v1alpha1.SubnetSet{}
 	metrics.CounterInc(r.SubnetService.NSXConfig, metrics.ControllerSyncTotal, MetricResTypeSubnetSet)
 
-	if err := r.Client.Get(ctx, req.NamespacedName, obj); err != nil {
-		log.Error(err, "unable to fetch subnetset CR", "req", req.NamespacedName)
+	if err := r.Client.Get(ctx, req.NamespacedName, subnetsetCR); err != nil {
+		if err := r.deleteSubnetBySubnetSetName(req.Name); err != nil {
```
yanjunz97, line 61: Do we need to also use namespace?

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -288,33 +288,62 @@ func (r *SubnetSetReconciler) CollectGarbage(ctx context.Context) {
 	}
 }
 
-func (r *SubnetSetReconciler) DeleteSubnetForSubnetSet(obj v1alpha1.SubnetSet, updataStatus bool) (bool, error) {
-	nsxSubnets := r.SubnetService.SubnetStore.GetByIndex(servicecommon.TagScopeSubnetSetCRUID, string(obj.GetUID()))
-	hitError := false
-	hasStaleSubnetPorts := false
+func (r *SubnetSetReconciler) cleanStaleSubnetsForSubnetSet(subnetSetName, subnetsetID string) error {
+	nsxStaleSubnets := r.SubnetService.SubnetStore.GetByIndex(servicecommon.TagScopeSubnetSetCRName, subnetSetName)
```
yanjunz97, line 292: Similar here, do we need to add namespace when getting the stale subnets?

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -49,94 +49,94 @@ type SubnetSetReconciler struct {
...
 		return ResultNormal, client.IgnoreNotFound(err)
 	}
+	if err := r.cleanStaleSubnetsForSubnetSet(subnetsetCR.Name, string(subnetsetCR.UID)); err != nil {
+		return ResultRequeue, err
+	}
+	if !subnetsetCR.ObjectMeta.DeletionTimestamp.IsZero() {
+		metrics.CounterInc(r.SubnetService.NSXConfig, metrics.ControllerDeleteTotal, MetricResTypeSubnetSet)
+		err := r.deleteSubnetForSubnetSet(*subnetsetCR, false)
+		if err != nil {
+			log.Error(err, "deletion NSX Subnet failed, would retry exponentially", "SubnetSet", req.NamespacedName)
```
yanjunz97, line 74: deletion NSX Subnet -> deleting NSX Subnet or NSX Subnet deletion?

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -49,94 +49,94 @@ type SubnetSetReconciler struct {
...
+			deleteFail(r, ctx, subnetsetCR, err.Error())
+			return ResultRequeue, err
+		}
+		if err := r.Client.Delete(ctx, subnetsetCR); err != nil {
+			log.Error(err, "deletion Subnet CR failed, would retry exponentially", "SubnetSet", req.NamespacedName)
```
yanjunz97, line 79: ditto

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -50,102 +47,130 @@ type SubnetReconciler struct {
...
-		tags := r.SubnetService.GenerateSubnetNSTags(obj, obj.Namespace)
-		if tags == nil {
-			return ResultRequeue, errors.New("failed to generate subnet tags")
+	if err := r.cleanStaleSubnets(subnetCR.Name, subnetCR.Namespace, string(subnetCR.UID)); err != nil {
+		return ResultRequeue, err
+	}
+	if !subnetCR.DeletionTimestamp.IsZero() {
+		metrics.CounterInc(r.SubnetService.NSXConfig, metrics.ControllerDeleteTotal, MetricResTypeSubnet)
+		if err := r.DeleteSubnetByID(*subnetCR); err != nil {
+			log.Error(err, "deletion NSX Subnet failed, would retry exponentially", "Subnet", req.NamespacedName)
```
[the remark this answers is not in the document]
wenqiq: done

pkg/controllers/common/utils.go
wenqiq: Updated.

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -50,102 +47,130 @@ type SubnetReconciler struct {
...
+		log.Error(err, "update Subnet status failed, would retry exponentially", "Subnet", req.NamespacedName)
+		updateFail(r, ctx, subnetCR, err.Error())
+		return ResultRequeue, err
 	}
+	updateSuccess(r, ctx, subnetCR)
 	return ctrl.Result{}, nil
 }
 
-func (r *SubnetReconciler) DeleteSubnet(obj v1alpha1.Subnet) error {
+func (r *SubnetReconciler) DeleteSubnetByID(obj v1alpha1.Subnet) error {
```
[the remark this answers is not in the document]
wenqiq: Good catch. Updated.

pkg/controllers/subnetset/subnetset_controller.go
wenqiq: Updated.

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -49,94 +49,100 @@ type SubnetSetReconciler struct {
...
+			if err := r.deleteSubnetBySubnetSetName(req.Name, req.Namespace); err != nil {
+				log.Error(err, "failed to delete NSX Subnet", "SubnetSet", req.NamespacedName)
+				return ResultRequeue, err
+			}
+			return ResultNormal, nil
+		}
+		log.Error(err, "unable to fetch Subnet CR", "req", req.NamespacedName)
+		return ResultRequeue, err
+	}
+	if err := r.cleanStaleSubnetsForSubnetSet(subnetsetCR.Name, subnetsetCR.Namespace, string(subnetsetCR.UID)); err != nil {
```
yanjunz97, line 72: It may have the same issue as cleanStaleSubnets in subnet controller?

pkg/nsx/services/subnet/subnet_test.go
```diff
@@ -0,0 +1,91 @@
...
+			Namespace: "test-ns",
+		},
+	}
+
+	// Generate tags for the Subnet
+	tags := service.GenerateSubnetNSTags(subnet)
+
+	// Validate the tags
+	assert.NotNil(t, tags)
+	assert.Equal(t, 4, len(tags)) // 3 tags should be generated
```
yanjunz97, line 64: 3 tags -> 4 tags?

pkg/nsx/services/subnet/subnet_test.go
```diff
@@ -0,0 +1,91 @@
...
+			},
+		},
+	}
+
+	// Generate tags for the SubnetSet
+	tagsSet := service.GenerateSubnetNSTags(subnetSet)
+
+	// Validate the tags for SubnetSet
+	assert.NotNil(t, tagsSet)
+	assert.Equal(t, 4, len(tagsSet)) // 3 tags should be generated
```
yanjunz97, line 88: ditto

yanjunz97 (commented):
Would be better if adding some testing done in the description.

By the way, would you know why the coverage report link does not show up in this PR?

yanjunz97:
Also remove the unused variable?

https://github.com/vmware-tanzu/nsx-operator/blob/c1ef61a708167a678154e2539e19c41ee0e6da7b/pkg/nsx/services/common/types.go#L110-L111

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -50,102 +48,144 @@ type SubnetReconciler struct {
...
+			return ResultNormal, nil
 		}
-		tags := r.SubnetService.GenerateSubnetNSTags(obj, obj.Namespace)
-		if tags == nil {
-			return ResultRequeue, errors.New("failed to generate subnet tags")
+		log.Error(err, "unable to fetch Subnet CR", "req", req.NamespacedName)
+		return ResultRequeue, err
+	}
+	// Clean stale subnets, clean all the NSX Subnets from store that has the same name but has the different UID
+	if err := r.cleanStaleSubnets(subnetCR.Name, subnetCR.Namespace, string(subnetCR.UID)); err != nil {
```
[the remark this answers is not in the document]
wenqiq: Actually, in the `cleanStaleSubnets` function, it has skipped deleting the `subnetport` with the same subnet id. So all the subnet in NSX with the same name would be deleted, and only preserve the NSX subnet with the same ID.
`if *nsxSubnet.Id == id`
```
func (r *SubnetReconciler) cleanStaleSubnets(name, ns, id string) error {
	subnetsToDelete := []*model.VpcSubnet{}
	nsxSubnets := r.SubnetService.SubnetStore.GetByIndex(servicecommon.TagScopeSubnetCRNamespacedName, util.CombineNamespaceName(name, ns))
	for i, nsxSubnet := range nsxSubnets {
		if *nsxSubnet.Id == id {
			continue
		}
		subnetsToDelete = append(subnetsToDelete, nsxSubnets[i])
	}
	return r.deleteSubnets(subnetsToDelete)
}
```

pkg/nsx/services/subnet/subnet_test.go
wenqiq: done.

pkg/controllers/subnet/subnet_controller.go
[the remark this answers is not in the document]
timdengyun: When user delete a subnet without deleting the subnetport below it, and then try to create another subnet with the same name.
This should be ok, in NSX side, there is a new subnet to be created with new UID, but we need to check clean stale subnet which has no CR but only nsx subnet left. For this case, you will not able to get subnet CR ID, you need someway to check which one is stale.

timdengyun, line 71: How to know subnetCR.UID is for deletion not for creation/update?

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -50,102 +48,144 @@ type SubnetReconciler struct {
...
-	portNums := len(r.SubnetPortService.GetPortsOfSubnet(*nsxSubnets[0].Id))
-	if portNums > 0 {
-		err := errors.New("subnet still attached by port")
-		log.Error(err, "", "ID", *nsxSubnets[0].Id)
-		return err
+	return nil
+}
+
+func (r *SubnetReconciler) deleteSubnetByName(name, ns string) error {
+	nsxSubnets := r.SubnetService.SubnetStore.GetByIndex(servicecommon.TagScopeSubnetCRNamespacedName, util.CombineNamespaceName(name, ns))
```
timdengyun, line 175: for the nsxSubnets gotten from store, we need to also get it's K8s uid to check if this CR is deleted or not, if CR uid is not in k8s, then we go to delete nsx subnet. if K8s uid CR exists, we skip to delete.

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -50,102 +48,144 @@ type SubnetReconciler struct {
 }
 
 func (r *SubnetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	obj := &v1alpha1.Subnet{}
-	log.Info("reconciling subnet CR", "subnet", req.NamespacedName)
+	startTime := time.Now()
+	defer func() {
+		log.Info("Finished reconciling Subnet", "Subnet", req.NamespacedName, "duration", time.Since(startTime))
```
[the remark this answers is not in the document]
wenqiq: Not adding. Just improved the original logic a bit.
```
log.Info("reconciling subnet CR", "subnet", req.NamespacedName)
```

pkg/controllers/subnetset/namespace_handler.go
```diff
@@ -39,7 +39,7 @@ func (e *EnqueueRequestForNamespace) Generic(_ context.Context, _ event.GenericE
 
 func (e *EnqueueRequestForNamespace) Update(_ context.Context, updateEvent event.UpdateEvent, l workqueue.RateLimitingInterface) {
 	obj := updateEvent.ObjectNew.(*v1.Namespace)
-	err := reconcileSubnetSet(e.Client, obj.Name, l)
+	err := requeueSubnetSet(e.Client, obj.Name, l)
```
timdengyun, line 42: reconcileSubnetSet name might be better?

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -50,102 +48,144 @@ type SubnetReconciler struct {
...
 				return ResultRequeue, err
 			}
-			log.V(1).Info("added finalizer on subnet CR", "subnet", req.NamespacedName)
+			return ResultNormal, nil
 		}
-		tags := r.SubnetService.GenerateSubnetNSTags(obj, obj.Namespace)
-		if tags == nil {
-			return ResultRequeue, errors.New("failed to generate subnet tags")
+		log.Error(err, "unable to fetch Subnet CR", "req", req.NamespacedName)
+		return ResultRequeue, err
```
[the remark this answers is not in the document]
timdengyun: I've discussed with wenqi in this case we should requeue it since  we will get client.IgnoreNotFound(err) == nil
`if client.IgnoreNotFound(err) == nil`

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -49,94 +49,100 @@ type SubnetSetReconciler struct {
 }
 
 func (r *SubnetSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	obj := &v1alpha1.SubnetSet{}
-	log.Info("reconciling subnetset CR", "subnetset", req.NamespacedName)
+	startTime := time.Now()
+	defer func() {
+		log.Info("Finished reconciling SubnetSet", "SubnetSet", req.NamespacedName, "time", time.Since(startTime))
```
timdengyun, line 54: Finished -> finished
we usually use the words starting with lower case.

pkg/controllers/subnet/subnet_controller.go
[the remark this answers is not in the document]
timdengyun: Finished -> finished
staring the words with lower case.

pkg/controllers/subnetset/subnetset_controller.go
wenqiq: IMO. Each log message should be treated as a complete sentence, so capitalizing the first letter of the first word is more appropriate.

pkg/controllers/subnet/subnet_controller.go
wenqiq: Updated.

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -50,102 +48,141 @@ type SubnetReconciler struct {
...
-		log.Info("no subnet found for subnet CR", "uid", string(obj.GetUID()))
-		return nil
-	}
-	portNums := len(r.SubnetPortService.GetPortsOfSubnet(*nsxSubnets[0].Id))
-	if portNums > 0 {
-		err := errors.New("subnet still attached by port")
-		log.Error(err, "", "ID", *nsxSubnets[0].Id)
+func (r *SubnetReconciler) deleteSubnetByID(subnetID string) error {
+	nsxSubnets := r.SubnetService.SubnetStore.GetByIndex(servicecommon.TagScopeSubnetCRUID, subnetID)
+	return r.deleteSubnets(nsxSubnets)
```
yanjunz97, line 150: Do we still need to list and check all the k8s crd ids when deleteSubnetByID?
Would you think it would be better to rename the current deleteSubnets to deleteStaleSubnets and have another function deleteSubnets only responsible for subnets deletion?

pkg/controllers/subnetset/subnetset_controller.go
yanjunz97, line 72: Shall we also remove this function as we do in Subnet controller?

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -288,33 +297,65 @@ func (r *SubnetSetReconciler) CollectGarbage(ctx context.Context) {
...
+		log.Error(err, "Failed to delete stale subnets", "SubnetSet", subnetSetName, "Namespace", ns)
+		return err
+	}
+	log.Info("Successfully cleaned stale subnets for SubnetSet", "SubnetSet", subnetSetName, "Namespace", ns, "num", len(subnetToDelete))
+	return nil
+}
+
+func (r *SubnetSetReconciler) deleteSubnetBySubnetSetName(subnetSetName, ns string) error {
+	nsxSubnets := r.SubnetService.SubnetStore.GetByIndex(servicecommon.TagScopeSubnetSetCRNamespacedName, util.CombineNamespaceName(subnetSetName, ns))
+	if err := r.deleteSubnets(nsxSubnets); err != nil {
```
yanjunz97, line 323: Shall we also check if those nsx subnets belongs to an existing subnetset with the same name/namespace by fetching all the existing subnetset cr uid, and the corresponding subnets uid to filter these subnets, which is similar to what we do in subnet controller?

pkg/controllers/subnet/namespace_handler.go
```diff
@@ -18,28 +18,28 @@ import (
...
+// Subnet controller should watch event of Namespace, when there are some updates of Namespace labels,
 // controller should build tags and update VpcSubnet according to new labels.
 
 type EnqueueRequestForNamespace struct {
 	Client client.Client
 }
 
 func (e *EnqueueRequestForNamespace) Create(_ context.Context, _ event.CreateEvent, _ workqueue.RateLimitingInterface) {
-	log.V(1).Info("namespace create event, do nothing")
+	log.V(1).Info("Namespace create event, do nothing")
```
dantingl, line 29: If I remember correctly, we use lower case for log message

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -50,102 +48,144 @@ type SubnetReconciler struct {
...
-				}
-				obj.Spec.IPv4SubnetSize = vpcNetworkConfig.DefaultSubnetSize
-			}
-			if err := r.Client.Update(ctx, obj); err != nil {
-				log.Error(err, "add finalizer", "subnet", req.NamespacedName)
-				updateFail(r, ctx, obj, err.Error())
+	subnetCR := &v1alpha1.Subnet{}
+	if err := r.Client.Get(ctx, req.NamespacedName, subnetCR); err != nil {
+		// IgnoreNotFound returns nil on NotFound errors.
+		if client.IgnoreNotFound(err) == nil {
```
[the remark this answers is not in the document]
dantingl: Not sure if you discussed this question offline, I have the same feeling we should also try to delete NSX Subnet for not found error.

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -50,102 +48,141 @@ type SubnetReconciler struct {
...
+func (r *SubnetReconciler) deleteSubnets(nsxSubnets []*model.VpcSubnet) error {
+	crdSubnetIDs, err := r.listSubnetCRs(context.Background())
+	if err != nil {
+		log.Error(err, "failed to list Subnet CRs")
 		return err
 	}
-	return r.SubnetService.DeleteSubnet(*nsxSubnets[0])
+	crdSubnetIDsSet := sets.NewString(crdSubnetIDs...)
+	for _, nsxSubnet := range nsxSubnets {
+		uid := nsxutil.FindTag(nsxSubnet.Tags, servicecommon.TagScopeSubnetCRUID)
```
dantingl, line 161: Why we need to compare UID again here?

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -266,87 +303,100 @@ func deleteSuccess(r *SubnetReconciler, _ context.Context, o *v1alpha1.Subnet) {
...
-		crdSubnetIDs.Insert(string(sr.UID))
-	}
-
-	for _, elem := range nsxSubnetList {
-		uid := util.FindTag(elem.Tags, servicecommon.TagScopeSubnetCRUID)
-		if crdSubnetIDs.Has(uid) {
+	crdSubnetIDsSet := sets.NewString(crdSubnetIDs...)
+	for _, nsxSubnet := range nsxSubnetList {
+		uid := nsxutil.FindTag(nsxSubnet.Tags, servicecommon.TagScopeSubnetCRUID)
+		if crdSubnetIDsSet.Has(uid) {
```
dantingl, line 388: The logic is changed here, you need to try to delete NSX Subnet event the CR is existing.
Subnet GC should two things:
1) scale down NSX Subnet if there is no port attached on it, this is current GC doing.
2) delete stale NSX Subnet if there is no Subnet CR existing

we can have a following patch or this patch for 2), but please also keep 1) logic in GC.

yanjunz97 (commented):
Add 2 more nits

pkg/controllers/subnet/subnet_controller.go
timdengyun: This comparing is needed to check if K8s CR of UID is still remaining  in K8s or not. So, we can know which CR is created or deleted in case that the CRs has the same ns/name.

pkg/controllers/subnetset/subnetset_controller.go
timdengyun: Yes, we don't need `cleanStaleSubnetsForSubnetSet` logic here, we should do it inside `client.IgnoreNotFound(err) == nil` condition.

pkg/controllers/subnetset/subnetset_controller.go
timdengyun: Right, that is true. It's better to move `deleteSubnets` to service for sharing function between subnet and subnetset controller. 
We also need get some K8s CR ids, these info can be passed from controller to service when calling `deleteSubnets` 

but this is a sort of refactoring work, we can refine it in the next patch.

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -50,102 +48,141 @@ type SubnetReconciler struct {
...
+	if err != nil {
+		log.Error(err, "failed to list Subnet CRs")
 		return err
 	}
-	return r.SubnetService.DeleteSubnet(*nsxSubnets[0])
+	crdSubnetIDsSet := sets.NewString(crdSubnetIDs...)
+	for _, nsxSubnet := range nsxSubnets {
+		uid := nsxutil.FindTag(nsxSubnet.Tags, servicecommon.TagScopeSubnetCRUID)
+		if crdSubnetIDsSet.Has(uid) {
+			log.Info("Skipping deletion, Subnet CRD still exists in K8s", "ID", *nsxSubnet.Id)
```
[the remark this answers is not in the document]
wenqiq: done.

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -266,87 +303,100 @@ func deleteSuccess(r *SubnetReconciler, _ context.Context, o *v1alpha1.Subnet) {
...
-		Scheme:            mgr.GetScheme(),
-		SubnetService:     subnetService,
-		SubnetPortService: subnetPortService,
-		VPCService:        vpcService,
-		Recorder:          mgr.GetEventRecorderFor("subnet-controller"),
-	}
-	if err := subnetReconciler.Start(mgr); err != nil {
-		log.Error(err, "failed to create controller", "controller", "Subnet")
-		return err
+func (r *SubnetReconciler) listSubnetCRs(ctx context.Context) ([]string, error) {
```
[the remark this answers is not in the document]
wenqiq: done.

pkg/controllers/subnet/namespace_handler.go
wenqiq: Why do we persist in following the wrong path until it's too late? It's better to make changes as early as possible. If we adjust a little bit with each patch, we will be in the right way sooner or later. It's not too late to mend the fold after the sheep are lost.

pkg/nsx/services/subnet/subnet.go
```diff
@@ -132,6 +132,7 @@ func (service *SubnetService) createOrUpdateSubnet(obj client.Object, nsxSubnet
 		Jitter:   0,
 		Steps:    6,
 	}
+	//
```
yanjunz97, line 135:
```suggestion
```

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -49,94 +51,96 @@ type SubnetSetReconciler struct {
...
-		return ResultNormal, client.IgnoreNotFound(err)
+	if err := r.Client.Get(ctx, req.NamespacedName, subnetsetCR); err != nil {
+		if apierrors.IsNotFound(err) {
+			if err := r.deleteSubnetBySubnetSetName(req.Name, req.Namespace); err != nil {
+				log.Error(err, "Failed to delete NSX Subnet", "SubnetSet", req.NamespacedName)
+				return ResultRequeue, err
+			}
+			return ResultNormal, nil
+		}
+		log.Error(err, "Unable to fetch Subnet CR", "req", req.NamespacedName)
```
yanjunz97, line 70:
```suggestion
		log.Error(err, "Unable to fetch SubnetSet CR", "req", req.NamespacedName)
```

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -49,94 +51,96 @@ type SubnetSetReconciler struct {
...
+	if !subnetsetCR.ObjectMeta.DeletionTimestamp.IsZero() {
+		metrics.CounterInc(r.SubnetService.NSXConfig, metrics.ControllerDeleteTotal, MetricResTypeSubnetSet)
+		err := r.deleteSubnetForSubnetSet(*subnetsetCR, false)
+		if err != nil {
+			log.Error(err, "Failed to delete NSX Subnet, retrying", "SubnetSet", req.NamespacedName)
+			deleteFail(r, ctx, subnetsetCR, err.Error())
+			return ResultRequeue, err
+		}
+		if err := r.Client.Delete(ctx, subnetsetCR); err != nil {
+			log.Error(err, "Failed to delete Subnet CR, retrying", "SubnetSet", req.NamespacedName)
```
yanjunz97, line 82:
```suggestion
			log.Error(err, "Failed to delete SubnetSet CR, retrying", "SubnetSet", req.NamespacedName)
```

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -49,94 +51,96 @@ type SubnetSetReconciler struct {
...
+			deleteFail(r, ctx, subnetsetCR, err.Error())
+			return ResultRequeue, err
+		}
+		deleteSuccess(r, ctx, subnetsetCR)
```
yanjunz97, line 86: Return after deleteSuccess?

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -288,33 +295,75 @@ func (r *SubnetSetReconciler) CollectGarbage(ctx context.Context) {
...
+}
+
+func (r *SubnetSetReconciler) deleteSubnetForSubnetSet(obj v1alpha1.SubnetSet, updateStatus bool) error {
 	nsxSubnets := r.SubnetService.SubnetStore.GetByIndex(servicecommon.TagScopeSubnetSetCRUID, string(obj.GetUID()))
-	hitError := false
-	hasStaleSubnetPorts := false
-	for _, subnet := range nsxSubnets {
-		r.SubnetService.LockSubnet(subnet.Path)
-		portNums := len(r.SubnetPortService.GetPortsOfSubnet(*subnet.Id))
+	if err := r.deleteStaleSubnets(nsxSubnets); err != nil {
```
yanjunz97, line 305: Use deleteSubnets as the nsxSubnets is fetch with subnet set UID instead of name so there is no new subnets are included?

pkg/nsx/services/subnet/subnet.go
```diff
@@ -194,6 +197,30 @@ func (service *SubnetService) ListSubnetSetID(ctx context.Context) sets.Set[stri
 	return subnetsetIDs
 }
 
+func (service *SubnetService) ListSubnetByName(ns, name string) []*model.VpcSubnet {
+	nsxSubnets := service.SubnetStore.GetByIndex(common.TagScopeNamespace, ns)
```
yanjunz97, line 201: Shall we add an index function for TagScopeNamespace?

yanjunz97 (commented):
Find one more comment

pkg/nsx/services/subnet/subnet.go
wenqiq: Thanks for reviewing. Updated.
Add two index functions, `common.TagScopeVMNamespace:    subnetIndexNamespaceFunc,` for Subnet and SubnetSet.
And `common.TagScopeNamespace:      subnetIndexScopeNamespaceFunc` for default Pod SubnetSet.

pkg/controllers/subnetset/subnetset_controller.go
wenqiq: updated.

pkg/controllers/subnetset/subnetset_controller.go
wenqiq: done.

pkg/controllers/subnetset/subnetset_controller.go
wenqiq: done.

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -50,102 +48,148 @@ type SubnetReconciler struct {
...
 }
 
-func (r *SubnetReconciler) DeleteSubnet(obj v1alpha1.Subnet) error {
-	nsxSubnets := r.SubnetService.SubnetStore.GetByIndex(servicecommon.TagScopeSubnetCRUID, string(obj.GetUID()))
-	if len(nsxSubnets) == 0 {
-		log.Info("no subnet found for subnet CR", "uid", string(obj.GetUID()))
-		return nil
+func (r *SubnetReconciler) deleteSubnetByID(subnetID string) error {
+	nsxSubnets := r.SubnetService.SubnetStore.GetByIndex(servicecommon.TagScopeSubnetCRUID, subnetID)
+	return r.deleteStaleSubnets(nsxSubnets)
```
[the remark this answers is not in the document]
wenqiq: Updated.

pkg/nsx/services/subnet/subnet.go
wenqiq: Updated. I added a comment here.

pkg/controllers/subnetset/subnetset_controller.go
wenqiq: done.

pkg/nsx/services/subnet/store.go
```diff
@@ -39,6 +39,24 @@ func subnetIndexFunc(obj interface{}) ([]string, error) {
 	}
 }
 
+func subnetIndexNamespaceFunc(obj interface{}) ([]string, error) {
```
yanjunz97, line 42: Would this be better as both the only difference between 2 tags is VM instead of Scope?
subnetIndexNamespaceFunc -> subnetIndexVMNamespaceFunc // match TagScopeVMNamespace
subnetIndexScopeNamespaceFunc -> subnetIndexNamespaceFunc // match TagScopeNamespace

yanjunz97 (commented):
Overall LGTM, only 1 nit

pkg/nsx/services/subnet/store.go
wenqiq: Good catch.Updated.

yanjunz97 (approved):

zhengxiexie (approved):
LGTM

PR #770, opened 2024-09-20
Add lock to subnet to avoid race between gc and controller

yanjunz97:
Subnetport/pod controller will acquire a subnet, create nsx subnetport on it and
save the subnetport to store after it is created. But Subnetset GC may delete the
subnet during the subnetport creation 

The PR adds locks for subnet by subnet path to avoid this race condition.

Testing done:
Decreased the subnetset garbage collector running interval to 10 seconds to increase
the potential of race. Created 8 pods, waited for all related subnetports created without
error and delete the pods. Repeated for 5 times and not found deadlock or race issue.

### DIFF
--- a/pkg/nsx/services/subnet/store.go
+++ b/pkg/nsx/services/subnet/store.go
@@ -2,6 +2,7 @@ package subnet
 
 import (
 	"errors"
+	"sync"
 
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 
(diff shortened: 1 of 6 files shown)


pkg/nsx/services/subnet/store.go
```diff
@@ -51,6 +52,34 @@ func subnetSetIndexFunc(obj interface{}) ([]string, error) {
 // SubnetStore is a store for subnet.
 type SubnetStore struct {
 	common.ResourceStore
+	pathLocks map[string]*sync.Mutex
```
zhengxiexie, line 55: The check-then-act pattern in the if statement is not atomic. Multiple goroutines could simultaneously determine that the lock doesn't exist and try to create a new one.

zhengxiexie: Consider using sync.Map?

zhengxiexie:
Besides, it's better to paste the testing process.

pkg/nsx/services/subnet/store.go
yanjunz97: Updated to use sync.Map, please take another look, thanks!

yanjunz97:
> Besides, it's better to paste the testing process.

Updated,thanks!

zhengxiexie (dismissed):
LGTM

zhengxiexie (approved):
LGTM

dantingl (approved):

PR #771, opened 2024-09-20
Add sleeping time to avoid leader election overlap

yanjunz97:


cmd/main.go
```diff
@@ -244,6 +244,12 @@ func electMaster(mgr manager.Manager, nsxClient *nsx.Client) {
 	log.Info("I'm trying to be elected as master")
 	<-mgr.Elected()
 	log.Info("I'm the master now")
+	// During the leader transitions, two operator may run in Kubernetes at the same time
```
zhengxiexie, line 247: > In HA mode, there can be a brief period where both the old and new leader operators are active simultaneously. This occurs when the new operator acquires the lease before the old operator recognizes it has lost the lease, leading to a potential race condition. To mitigate this, the new master operator is configured to wait for 15 seconds, which is the default Leader Election Lease Duration, ensuring a smooth transition.

How do you feel about expressing it this way？

cmd/main.go
```diff
@@ -244,6 +244,12 @@ func electMaster(mgr manager.Manager, nsxClient *nsx.Client) {
...
+	// when the new one acquire the lease while the old one does not realize it loses the lease.
+	// To avoid the race condition during such period, make the new master operator sleeping
+	// for 15 seconds, which is the default value of the Leader Election Lease Duration.
+	log.Info("Sleeping for 15 seconds...")
```
zhengxiexie, line 251: Waiting a 15-second delay to let the old instance know that it has lost its lease.

cmd/main.go
yanjunz97: It looks good to me. Updated, thanks!

cmd/main.go
yanjunz97: Updated, thanks!

cmd/main.go
```diff
@@ -244,6 +244,14 @@ func electMaster(mgr manager.Manager, nsxClient *nsx.Client) {
 	log.Info("I'm trying to be elected as master")
 	<-mgr.Elected()
 	log.Info("I'm the master now")
+	// In HA mode, there can be a brief period where both the old and new leader
+	// operators are active simultaneously. This occurs when the new operator
+	// acquires the lease before the old operator recognizes it has lost the lease,
+	// leading to a potential race condition. To mitigate this, the new master
+	// operator is configured to wait for 15 seconds, which is the default Leader
+	// Election Lease Duration, ensuring a smooth transition.
+	log.Info("waiting a 15-second delay to let the old instance know that it has lost its lease")
```
dantingl, line 253: Could we wait for 18 or 20 seconds to ensure old leader gone?

yanjunz97: Update comments as discussed offline, thanks

zhengxiexie (approved):
LGTM

PR #772, opened 2024-09-23
Add VPC network check in SecurityPolicy License error Annotation

timdengyun:
When securitypolicy controller tries to update SecurityPolicy obj by k8s client, it will pop up the error:
`2024-09-23 02:02:45.054	ERROR	securitypolicy/securitypolicy_controller.go:113	Failed to clean SecurityPolicy annotation	{"error": "no matches for kind \"SecurityPolicy\" in version \"nsx.vmware.com/v1alpha1\""}`

This is because there is no T1 CRD obj and VPC CRD obj API version conversion when updating SecurityPolicy, So, this patch is to add VPC network mode check to see if  SecurityPolicy conversion necessary.

Test Done:
1. Create SecurityPolicy in VPC mode to see there is no error any more.
2. Create SecurityPolicy in T1 mode  to see there is no error any more.


TaoZou1 (approved):

zhengxiexie (approved):
LGTM

PR #773, opened 2024-09-23
Remove private ip block store in VPC

seanpang-vmware:
The private ipblocks in VPC are now maintained by nsx, nsx operator will not do operations on these nsx resources, so vpc service do not need to build its store anymore.

Remove the private ip blocks store and its related logics.

### DIFF
--- a/pkg/controllers/networkinfo/networkinfo_controller.go
+++ b/pkg/controllers/networkinfo/networkinfo_controller.go
@@ -242,10 +242,6 @@ func (r *NetworkInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 						deleteFail(r, ctx, obj, &err, r.Client)
 						return common.ResultRequeueAfter10sec, err
 					}
-					if err := r.Service.DeleteIPBlockInVPC(*vpc); err != nil {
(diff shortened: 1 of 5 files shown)


zhengxiexie (dismissed):
LGTM

pkg/nsx/services/vpc/compare_test.go
```diff
@@ -0,0 +1,8 @@
+package vpc
+
+import (
+	"testing"
+)
+
+func TestIsVPCChanges(t *testing.T) {
```
dantingl, line 7: Do you want to add test in this patch?

seanpang-vmware: No, i just added an empty file since the file is already created, the test code is in another patch

pkg/nsx/services/vpc/compare_test.go
wenqiq, line 1: It would be better to delete this file. It's a bit weird to add a function without any content inside.

wenqiq (commented):
LGTM, just a nit comment.

wenqiq (approved):
It seems that the e2e test failure is not related to this PR, there is sth wrong with e2e testbed.

zhengxiexie (approved):
LGTM

PR #781, opened 2024-09-26
Fix bug which would create duplicated subnets when realization fails

zhengxiexie:
Subnets are allocated from subnet sets, it's id is suffixed with a random string, if the subnet failed to realize, and not being deleted, then a new subnet with different suffix will be created, thus leading a mount of duplicated subnets being left in the system.

Background
image
Found a large mount of stale subnets being created, the reason is 
```		index := util.GetRandomIndexString()
		nsxSubnet = &model.VpcSubnet{
			Id:             String(service.buildSubnetSetID(o, index)),
			AccessMode:     String(convertAccessMode(util.Capitalize(string(o.Spec.AccessMode)))),
			Ipv4SubnetSize: Int64(int64(o.Spec.IPv4SubnetSize)),
			DhcpConfig:     service.buildDHCPConfig(o.Spec.DHCPConfig.EnableDHCP, int64(o.Spec.IPv4SubnetSize-4)),
			DisplayName:    String(service.buildSubnetSetName(o, index)),
		}
```

The id is suffixed with a random string. When allocated a subnet from subnet sets, it will check the realization status, if it failed, we should delete it, or it will allocate a new subnet the next round.

Test done
Hack realization fail, observe it does delete the subnet, and no more stale subnets exist.

### DIFF
--- a/pkg/nsx/services/subnet/subnet.go
+++ b/pkg/nsx/services/subnet/subnet.go
@@ -128,6 +128,12 @@ func (service *SubnetService) createOrUpdateSubnet(obj client.Object, nsxSubnet 
 	}
 	if err = realizeService.CheckRealizeState(backoff, *nsxSubnet.Path, "RealizedLogicalSwitch"); err != nil {
 		log.Error(err, "failed to check subnet realization state", "ID", *nsxSubnet.Id)
+		// Delete the subnet if realization check fails, avoiding creating duplicate subnets continuously.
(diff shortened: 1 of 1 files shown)


seanpang-vmware (approved):

timdengyun (approved):
LGTM
@wenqiq please take a look if this is the issue we hit in the e2e sometimes.

PR #785, opened 2024-09-29
Remove rule index from rule/group ID/name for VPC and unify NSX resource name for VPC and T1

timdengyun:
This patch is to:
1. Remove SecurityPolicy rule index from rule ID and for VPC mode,
and keep T1 mode rule ID unchanged with rule index.
2. Remove SecurityPolicy rule index from group ID for VPC mode,
and keep T1 mode group ID unchanged with rule index.
3. Remove rule index from NSX group/rule name, and unify the NSX resource name for VPC and T1 network,
including SecurityPolicy, rule, and group.
4. Reduce length of rule hash string to 8 chars for VPC mode.

Test Done:
1. Create SecurityPolicy in VPC mode, do CRUD.
The SecurityPolicy CR name is: sp-app-access-policy-vpc
The generated NSX SecurityPolicy:
```
ID: sp-app-access-policy-vpc_36fcf0f9-170a-4346-bb14-9730e199bb49
Name: sp-app-access-policy-vpc
```

Policy appliedTo group:
```
ID: sp-app-access-policy-vpc_36fcf0f9-170a-4346-bb14-9730e199bb49_scope
Name: sp-app-access-policy-vpc_scope
```

For the rule with any ports:
```
    - direction: in
      action: drop
```
The build NSX Rule:
```
ID: sp-app-access-policy-vpc_36fcf0f9-170a-4346-bb14-9730e199bb49_d79c8ddf_all
Name: all_ingress_isolation
```

For rule:
```
    - direction: in
      action: allow
      sources:
        - vmSelector:
            matchLabels:
              app: coco
      ports:
        - protocol: TCP
          port: 8282
        - protocol: TCP
          port: 7000
          endPort: 7020
```
The build NSX rule:
```
ID: sp-app-access-policy-vpc_36fcf0f9-170a-4346-bb14-9730e199bb49_44b022be_8282_7000.7020
name: TCP.8282_TCP.7000.7020_ingress_allow
```

The rule appliedTo group:
```
ID: sp-app-access-policy-vpc_36fcf0f9-170a-4346-bb14-9730e199bb49_b7adc115_scope
Name:sp-app-access-policy-vpc_b7adc115_scope
```

Rule with user defined name:
```
  - direction: in
      name: rule-with-defined
      action: allow
      sources:
        - namespaceSelector:
            matchLabels:
              role: validate
          vmSelector:
            matchLabels:
              role: check
        - ipBlocks:
            - cidr: 100.64.232.1/32
      ports:
        - protocol: TCP
          port: 8081
```
          
```
ID: sp-app-access-policy-vpc_36fcf0f9-170a-4346-bb14-9730e199bb49_52658c89_8081
Name: rule-with-defined_ingress_allow
```

The source peer group for the this rule:
```
Path: /orgs/default/projects/dy-project1/infra/domains/default/groups/sp-app-access-policy-vpc_36fcf0f9-170a-4346-bb14-9730e199bb49_52658c89_src
ID: sp-app-access-policy-vpc_36fcf0f9-170a-4346-bb14-9730e199bb49_52658c89_src
Name: sp-app-access-policy-vpc_52658c89_src
```

The NSX Share for this source group:
```
ID: dy-project1_group_sp-app-access-policy-vpc_36fcf0f9-170a-4346-bb14-9730e199bb49_52658c89_src_share
Name: dy-project1_group_sp-app-access-policy-vpc_52658c89_src_share
```

2. Create NetworkPolicy in VPC mode, do CRUD.
The generated allow section SecurityPolicy:
```
ID: np-app-access_4a2afd8b-a5e8-4867-8515-fa4961b52f53_allow
Name: np-app-access
```
The generated isolation section SecurityPolicy:
```
ID: np-app-access_4a2afd8b-a5e8-4867-8515-fa4961b52f53_isolation
Name: np-app-access
```

3. Create SecurityPolicy in VPC mode, and do GC

4. Create NetworkPolicy in VPC mode, and do GC

5. Create SecurityPolicy with namedport in VPC mode
The rule with namedport:
```
      ports:
        - protocol: TCP
          port: db-port
        - protocol: UDP
          port: 5353
        - protocol: UDP
          port: 5354
```
The db-port will map to two ports number: 80 and 3366, so, the generated IPSet groups:
```
ID: sp-namedport-web_2cd759be-8d83-41e4-92d9-4eba6beab6cf_a347c628_3366_ipset
Name: TCP.db-port_UDP.5353_UDP.5354_TCP.3366_ingress_allow_ipset

ID: sp-namedport-web_2cd759be-8d83-41e4-92d9-4eba6beab6cf_a347c628_80_ipset
Name: TCP.db-port_UDP.5353_UDP.5354_TCP.80_ingress_allow_ipset
```
          
6. Create NetworkPolicy with namedport in VPC mode
For the same named port in the aforementioned, the generated IPSet groups:
```
ID: np-namedport-web_fd555fd3-04a6-4f6a-b035-79e587dd4767_allow_27cfee51_3366_ipset
Name: TCP.db-port_UDP.5353_UDP.5354_TCP.3366_ingress_allow_ipset

ID: np-namedport-web_fd555fd3-04a6-4f6a-b035-79e587dd4767_allow_27cfee51_80_ipset
Name: TCP.db-port_UDP.5353_UDP.5354_TCP.80_ingress_allow_ipset
```

7. Create SecurityPolicy in T1 mode, and do CRUD.

8. Create SecurityPolicy in T1 mode, and do GC.

9. T1 upgrade case:
Create a SecurityPolicy in v4.1.2 code, and start NSX operator with this patch to see if SecurityPolicy Name, group and rule name are changed as expected.
For the rule created in V4.1.2
```
    - direction: in
      action: allow
      sources:
        - namespaceSelector:
            matchLabels:
              role: test
      ports:
        - protocol: TCP
          port: 8082
          endPort: 8283
```
Before upgrade in V4.1.2         
```
ID:sp_e52e71f7-600a-4c67-b241-ce87c6dd191e_c2f3ed42f3e59c15731f8426d65e13676663e147_1_0_0
Name:sp-app-access-policy-1-0-0
```

After upgrade, ID is not changed.
```
ID:sp_e52e71f7-600a-4c67-b241-ce87c6dd191e_c2f3ed42f3e59c15731f8426d65e13676663e147_1_0_0
Name:TCP.8082.8283_ingress_allow
```

### DIFF
--- a/pkg/nsx/services/securitypolicy/builder.go
+++ b/pkg/nsx/services/securitypolicy/builder.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"sort"
 	"strconv"
 	"strings"
 
@@ -14,6 +13,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/vmware-tanzu/nsx-operator/pkg/apis/legacy/v1alpha1"
@@ -39,24 +39,15 @@ var (
 	Int64  = common.Int64
 )
 
-func (service *SecurityPolicyService) buildSecurityPolicyName(obj *v1alpha1.SecurityPolicy, createdFor string) string {
-	if IsVPCEnabled(service) {
-		// For VPC scenario, we use obj.Name as the NSX resource display name for both SecurityPolicy and NetworkPolicy.
-		return util.GenerateTruncName(common.MaxNameLength, obj.Name, "", "", "", "")
-	}
-	prefix := common.SecurityPolicyPrefix
-	if createdFor != common.ResourceTypeSecurityPolicy {
-		prefix = common.NetworkPolicyPrefix
-	}
-	// For T1 scenario, we use ns-name as the key resource name for SecurityPolicy, it is to be consistent with the
-	// previous solutions.
-	return util.GenerateTruncName(common.MaxNameLength, strings.Join([]string{obj.Namespace, obj.Name}, common.ConnectorUnderline), prefix, "", "", "")
+func (service *SecurityPolicyService) buildSecurityPolicyName(obj *v1alpha1.SecurityPolicy) string {
+	return util.GenerateTruncName(common.MaxNameLength, obj.Name, "", "", "", "")
 }
 
 func (service *SecurityPolicyService) buildSecurityPolicyID(obj *v1alpha1.SecurityPolicy, createdFor string) string {
 	if IsVPCEnabled(service) {
 		return util.GenerateIDByObject(obj)
 	}
+
 	prefix := common.SecurityPolicyPrefix
 	if createdFor != common.ResourceTypeSecurityPolicy {
 		prefix = common.NetworkPolicyPrefix
@@ -76,7 +67,7 @@ func (service *SecurityPolicyService) buildSecurityPolicy(obj *v1alpha1.Security
 	nsxSecurityPolicy := &model.SecurityPolicy{}
 
 	nsxSecurityPolicy.Id = String(service.buildSecurityPolicyID(obj, createdFor))
-	nsxSecurityPolicy.DisplayName = String(service.buildSecurityPolicyName(obj, createdFor))
+	nsxSecurityPolicy.DisplayName = String(service.buildSecurityPolicyName(obj))
 	// TODO: confirm the sequence number: offset
 	nsxSecurityPolicy.SequenceNumber = Int64(int64(obj.Spec.Priority))
 
@@ -93,7 +84,7 @@ func (service *SecurityPolicyService) buildSecurityPolicy(obj *v1alpha1.Security
 	currentSet := sets.Set[string]{}
 	for ruleIdx, r := range obj.Spec.Rules {
 		rule := r
-		// A rule containing named port may expand to multiple rules if the name maps to multiple port numbers.
+		// A rule containing named port may be expanded to multiple rules if the named ports map to multiple port numbers.
 		expandRules, buildGroups, buildGroupShares, err := service.buildRuleAndGroups(obj, &rule, ruleIdx, createdFor)
 		if err != nil {
 			log.Error(err, "failed to build rule and groups", "rule", rule, "ruleIndex", ruleIdx)
@@ -204,11 +195,6 @@ func (service *SecurityPolicyService) buildTargetTags(obj *v1alpha1.SecurityPoli
 	rule *v1alpha1.SecurityPolicyRule, ruleIdx int, createdFor string,
 ) []model.Tag {
 	basicTags := service.buildBasicTags(obj, createdFor)
-	sort.Slice(*targets, func(i, j int) bool {
-		k1, _ := json.Marshal((*targets)[i])
-		k2, _ := json.Marshal((*targets)[j])
-		return string(k1) < string(k2)
-	})
 	serializedBytes, _ := json.Marshal(*targets)
 	targetTags := []model.Tag{
 		{
@@ -242,7 +228,7 @@ func (service *SecurityPolicyService) buildTargetTags(obj *v1alpha1.SecurityPoli
 		targetTags = append(targetTags,
 			model.Tag{
 				Scope: String(common.TagScopeRuleID),
-				Tag:   String(service.buildRuleID(obj, rule, ruleIdx, createdFor)),
+				Tag:   String(service.buildRuleID(obj, ruleIdx, createdFor)),
 			},
 		)
 	}
@@ -379,7 +365,8 @@ func (service *SecurityPolicyService) buildAppliedGroupID(obj *v1alpha1.Security
 	if IsVPCEnabled(service) {
 		suffix := common.TargetGroupSuffix
 		if ruleIdx != -1 {
-			suffix = strings.Join([]string{strconv.Itoa(ruleIdx), suffix}, common.ConnectorUnderline)
+			ruleHash := service.buildLimitedRuleHashString(&(obj.Spec.Rules[ruleIdx]))
+			suffix = strings.Join([]string{ruleHash, suffix}, common.ConnectorUnderline)
 		}
 		return util.GenerateIDByObjectWithSuffix(obj, suffix)
 	}
@@ -415,17 +402,13 @@ func (service *SecurityPolicyService) buildAppliedGroupPath(obj *v1alpha1.Securi
 
 // build appliedTo group display name for both policy and rule levels.
 func (service *SecurityPolicyService) buildAppliedGroupName(obj *v1alpha1.SecurityPolicy, ruleIdx int) string {
-	var rule *v1alpha1.SecurityPolicyRule
 	if ruleIdx != -1 {
-		rule = &(obj.Spec.Rules[ruleIdx])
-		ruleName := strings.Join([]string{obj.Name, strconv.Itoa(ruleIdx)}, common.ConnectorUnderline)
-		if len(rule.Name) > 0 {
-			ruleName = rule.Name
-		}
-		return util.GenerateTruncName(common.MaxNameLength, ruleName, "", common.TargetGroupSuffix, "", "")
+		ruleHash := service.buildLimitedRuleHashString(&(obj.Spec.Rules[ruleIdx]))
+		suffix := strings.Join([]string{ruleHash, common.TargetGroupSuffix}, common.ConnectorUnderline)
+		return util.GenerateTruncName(common.MaxNameLength, obj.Name, "", suffix, "", "")
 	}
-	ruleName := strings.Join([]string{obj.Namespace, obj.Name}, common.ConnectorUnderline)
-	return util.GenerateTruncName(common.MaxNameLength, ruleName, "", common.TargetGroupSuffix, "", "")
+
+	return util.GenerateTruncName(common.MaxNameLength, obj.Name, "", common.TargetGroupSuffix, "", "")
 }
 
 func (service *SecurityPolicyService) buildRuleAndGroups(obj *v1alpha1.SecurityPolicy, rule *v1alpha1.SecurityPolicyRule,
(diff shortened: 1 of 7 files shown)


pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -204,11 +194,6 @@ func (service *SecurityPolicyService) buildTargetTags(obj *v1alpha1.SecurityPoli
 	rule *v1alpha1.SecurityPolicyRule, ruleIdx int, createdFor string,
 ) []model.Tag {
 	basicTags := service.buildBasicTags(obj, createdFor)
-	sort.Slice(*targets, func(i, j int) bool {
```
timdengyun, line 207: Remove this sort function for building target tags, the reason is that this function will change rule target content. the generated rule hash in buildRuleID is different from the one generated before sort. 
we will generate NSX rule ID during expanding rules, by this time, there is no rule sort. 
After sort, the rule hash will be changed when it builds rule hash for target group ID.

So, we need to keep rule hash consistency, we need to remove this rule target sort function.

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -957,13 +957,6 @@ func (service *SecurityPolicyService) buildPeerTags(obj *v1alpha1.SecurityPolicy
 		groupTypeTag = String(common.TagValueGroupSource)
 		peers = &rule.Sources
 	}
-
-	// TODO: abstract sort func for both peers and targets
```
timdengyun, line 961: It has the same reason why re removed this sort function for target tags.
This sort function will impact rule hash when generating rule hash for peer groups.

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -598,22 +580,43 @@ func (service *SecurityPolicyService) buildRuleOutGroup(obj *v1alpha1.SecurityPo
...
 	}
+	ruleIdxStr := fmt.Sprintf("%d", ruleIdx)
 	return util.GenerateID(fmt.Sprintf("%s", obj.UID), prefix, ruleHash, ruleIdxStr)
 }
 
-func (service *SecurityPolicyService) buildRuleDisplayName(rule *v1alpha1.SecurityPolicyRule, portIdx, portNumber int, hasNamedport bool, createdFor string) (string, error) {
+// A rule containing named port may be expanded to multiple NSX rules if the name ports map to multiple port numbers.
+// So, in VPC network, the rule port numbers, which either are defined in rule Port or resolved from named port, will be appended as CR rule baseID to distinguish them.
+// For T1, the portIdx and portAddressIdx are appended as suffix.
+func (service *SecurityPolicyService) buildExpandedRuleID(obj *v1alpha1.SecurityPolicy, rule *v1alpha1.SecurityPolicyRule, ruleIdx int,
```
wenyingd, line 601: Honestly, I don't like the flag `hasNamedPort` when building Rule's ID or display name, because it has no valid meaning in the final string. Can we make it more simple by directly telling the function what is the final port number or port range used in the output?

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -1888,6 +1882,51 @@ func (service *SecurityPolicyService) buildRulePortsString(ports *[]v1alpha1.Sec
...
+				portsNumString = strings.Join([]string{portsNumString, portNumString}, common.ConnectorUnderline)
+			}
+		}
+	}
+	return portsNumString
+}
+
+func (service *SecurityPolicyService) buildRuleHashString(rule *v1alpha1.SecurityPolicyRule) string {
+	serializedBytes, _ := json.Marshal(rule)
+	return fmt.Sprintf("%s", util.Sha1(string(serializedBytes)))
```
wenyingd, line 1927: `util.Sha1(string(serializedBytes))` returns a string, no need to use `fmt.Sprintf` ?

wenyingd, line 1927: Having checked function `util.Sha1`, the return string is with length of 40 chars, do we need to keep the complete string here as the unique part? Or can we truncate it with less chars?

pkg/nsx/services/securitypolicy/firewall_test.go
```diff
@@ -1073,11 +1073,11 @@ func TestCreateOrUpdateSecurityPolicy(t *testing.T) {
 	mockVPCService := common.MockVPCServiceProvider{}
 	fakeService.vpcService = &mockVPCService
 
-	podSelectorRule0IDPort000 := fakeService.buildExpandedRuleId(fakeService.buildRuleID(&spWithPodSelector, &spWithPodSelector.Spec.Rules[0], 0, common.ResourceTypeSecurityPolicy), 0, 0)
-	podSelectorRule1IDPort000 := fakeService.buildExpandedRuleId(fakeService.buildRuleID(&spWithPodSelector, &spWithPodSelector.Spec.Rules[1], 1, common.ResourceTypeSecurityPolicy), 0, 0)
+	podSelectorRule0IDPort000 := fakeService.buildExpandedRuleID(&spWithPodSelector, &spWithPodSelector.Spec.Rules[0], 0, 0, 0, false, -1, common.ResourceTypeSecurityPolicy)
+	podSelectorRule1IDPort000 := fakeService.buildExpandedRuleID(&spWithPodSelector, &spWithPodSelector.Spec.Rules[0], 1, 0, 0, false, -1, common.ResourceTypeSecurityPolicy)
```
wenyingd, line 1077: Why change here to use `spWithPodSelector.Spec.Rules[0]` for `Rule1ID`?

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -598,22 +580,44 @@ func (service *SecurityPolicyService) buildRuleOutGroup(obj *v1alpha1.SecurityPo
...
+		return util.GenerateIDByObjectWithSuffix(obj, ruleIndexHash)
 	}
+
 	prefix := common.SecurityPolicyPrefix
 	if createdFor == common.ResourceTypeNetworkPolicy {
 		prefix = common.NetworkPolicyPrefix
 	}
-	return util.GenerateID(fmt.Sprintf("%s", obj.UID), prefix, ruleHash, ruleIdxStr)
+	ruleIdxSuffix := fmt.Sprintf("%d", ruleIdx)
+	return util.GenerateID(fmt.Sprintf("%s", obj.UID), prefix, ruleIdxSuffix, ruleIndexHash)
```
wenyingd, line 596: This is using a different order from the previous solution in function `util.GenerateID` with T1 solution, old way: obj.UID, prefix, ruleHash, ruleId; new way: obj.UID, prefix, ruleID, ruleHash.

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -661,9 +665,9 @@ func (service *SecurityPolicyService) buildRuleDisplayName(rule *v1alpha1.Securi
 		// e.g. input: security policy's rule name: TCP.http_UDP.1234_ingress_allow,
 		// expand to NSX security policy rules with name TCP.http_UDP.1234_TCP.80_ingress_allow and TCP.http_UDP.1234_UDP.1234_ingress_allow.
 		// in case that user defined input security policy's rule name: sp_namedport_rule,
-		// expand to NSX security policy rules with name sp_namedport_rule.TCP.80_ingress_allow and sp_namedport_rule.UDP.1234_ingress_allow.
+		// expand to NSX security policy rules with name sp_namedport_rule_TCP.80_ingress_allow and sp_namedport_rule_UDP.1234_ingress_allow.
 		index := strings.Index(ruleName, common.ConnectorUnderline+suffix)
-		return util.GenerateTruncName(common.MaxNameLength, ruleName[:index]+"."+service.buildRulePortString(&rule.Ports[portIdx], true, portNumber), "", suffix, "", ""), nil
+		return util.GenerateTruncName(common.MaxNameLength, ruleName[:index]+common.ConnectorUnderline+service.buildRulePortString(&rule.Ports[portIdx], true, portNumber), "", suffix, "", ""), nil
```
wenyingd, line 670: I prefer `sp_namedport_rule.UDP.1234_ingress_allow`, in which "rule.UDP.1234" is a description for the rule itself. If we switch to `sp_namedport_rule_UDP.1234_ingress_allow`, it is difficult to parse the rule details from "rule_UDP.1234".

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -957,13 +958,6 @@ func (service *SecurityPolicyService) buildPeerTags(obj *v1alpha1.SecurityPolicy
 		groupTypeTag = String(common.TagValueGroupSource)
 		peers = &rule.Sources
 	}
-
-	// TODO: abstract sort func for both peers and targets
-	sort.Slice(*peers, func(i, j int) bool {
-		k1, _ := json.Marshal((*peers)[i])
-		k2, _ := json.Marshal((*peers)[j])
-		return string(k1) < string(k2)
-	})
```
wenyingd, line 966: I am confused about the use case of this tag, do we really have a consumer on the hash value of the selectors? If no, can we remove it? If yes, does the consumer use `sort` on the selectors to calculate the tag, then this change is possibly break it.

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -1888,6 +1883,51 @@ func (service *SecurityPolicyService) buildRulePortsString(ports *[]v1alpha1.Sec
...
+	// The built port number string is: 8282.8286
+	// - protocol: UDP
+	//   port: 3308
+	// The built port number string is: 3308
+	if !hasNamedport {
+		if port.EndPort != 0 {
+			return fmt.Sprintf("%s.%d", (port.Port).String(), port.EndPort)
+		}
+		return fmt.Sprintf("%s", (port.Port).String())
+	} else {
```
wenyingd, line 1901: Can we use this format?

```
func (service *SecurityPolicyService) buildRulePortNumberString(port v1alpha1.SecurityPolicyPort, portNumber int) string {
	if port.Port.Type == intstr.String {
		return fmt.Sprintf("%d", portNumber)
	}
	if port.EndPort != 0 {
		return fmt.Sprintf("%s.%d", (port.Port).String(), port.EndPort)
	}
	return (port.Port).String()
}
```

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -598,22 +580,44 @@ func (service *SecurityPolicyService) buildRuleOutGroup(obj *v1alpha1.SecurityPo
 	return nsxRuleDstGroup, nsxRuleSrcGroupPath, nsxRuleDstGroupPath, nsxGroupShare, nil
 }
 
-func (service *SecurityPolicyService) buildRuleID(obj *v1alpha1.SecurityPolicy, rule *v1alpha1.SecurityPolicyRule, ruleIdx int, createdFor string) string {
-	serializedBytes, _ := json.Marshal(rule)
-	ruleHash := fmt.Sprintf("%s", util.Sha1(string(serializedBytes)))
-	ruleIdxStr := fmt.Sprintf("%d", ruleIdx)
+func (service *SecurityPolicyService) buildRuleID(obj *v1alpha1.SecurityPolicy, ruleIdx int, createdFor string) string {
+	ruleIndexHash := service.buildRuleHashString(&(obj.Spec.Rules[ruleIdx]))
```
wenyingd, line 584: Move this line to line 590 so it is close to its consumer.

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -598,22 +580,44 @@ func (service *SecurityPolicyService) buildRuleOutGroup(obj *v1alpha1.SecurityPo
...
 	}
-	return util.GenerateID(fmt.Sprintf("%s", obj.UID), prefix, ruleHash, ruleIdxStr)
+	ruleIdxSuffix := fmt.Sprintf("%d", ruleIdx)
+	return util.GenerateID(fmt.Sprintf("%s", obj.UID), prefix, ruleIdxSuffix, ruleIndexHash)
+}
+
+// A rule containing named port may be expanded to multiple NSX rules if the name ports map to multiple port numbers.
+// So, in VPC network, the rule port numbers, which either are defined in rule Port or resolved from named port, will be appended as CR rule baseID to distinguish them.
+// For T1, the portIdx and portAddressIdx are appended as suffix.
+func (service *SecurityPolicyService) buildExpandedRuleID(obj *v1alpha1.SecurityPolicy, rule *v1alpha1.SecurityPolicyRule, ruleIdx int,
```
wenyingd, line 602: I saw you have remove parameter "rule" from function `buildRuleID` because we can use `obj *v1alpha1.SecurityPolicy` and `ruleIdx int` to calculate the rule inside the function, can we do it similarly in function `buildExpandedRuleID` to make it consistent?

pkg/nsx/services/securitypolicy/builder.go
timdengyun: In fact, current change is a bug fix to align with v4.1.2 ID format. 
https://github.com/vmware-tanzu/nsx-operator/blob/v4.1.2/pkg/nsx/services/securitypolicy/builder.go#L513
The bug was introduced in xie's refactor patch:
https://github.com/vmware-tanzu/nsx-operator/commit/0ed01de3c89f89e6c9c2f9c02f59065572aa98bf#diff-878f5907665c91bbee664334052881facd7f44a8617ab92547f46a730c8a97d9R571

pkg/nsx/services/securitypolicy/builder.go
timdengyun: At line#L588
`ruleIndexHash = service.buildLimitedRuleHashString(&(obj.Spec.Rules[ruleIdx]))`
We will reassign the value to this var, so, we put it in the beginning.

pkg/nsx/services/securitypolicy/builder.go
timdengyun: As we discussed, we don't have consumer for these hash value by `sort`. So, it won't break the logic.
In addition, the `sort` function will change rule selector order, it will change rule hash value after `sort`.
but we use this rule hash in our ID and name, we need to keep it consistent. That is why we removed it.

pkg/nsx/services/securitypolicy/firewall_test.go
timdengyun: That is a mistake, thanks for pointing it out. Corrected.

pkg/nsx/services/securitypolicy/builder.go
timdengyun: Good advice, revert the change back as before.

pkg/nsx/services/securitypolicy/builder.go
timdengyun: Thanks for suggestion, refactor the code as suggested.

pkg/nsx/services/securitypolicy/builder.go
timdengyun: It makes sense to me to do this change. It's good to keep code consistent as uniformed style. 
Remove `rule *v1alpha1.SecurityPolicyRule` from `buildExpandedRuleID` function as well.

wenyingd (dismissed):
Most of the code looks good to me. The only left comment is the current implementation related with named port is complicated. Because that logic is not introduced by this change, having offline sync with Yun, we would use another change to refine the functions.

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -1888,6 +1884,50 @@ func (service *SecurityPolicyService) buildRulePortsString(ports *[]v1alpha1.Sec
...
+	// Build the rule port number string name for non named port.
+	// This is a common case where the string is built from port definition. For instance,
+	// - protocol: TCP
+	//   port: 8282
+	//   endPort: 8286
+	// The built port number string is: 8282.8286
+	// - protocol: UDP
+	//   port: 3308
+	// The built port number string is: 3308
+	if port.Port.Type == intstr.String && portNumber != -1 {
```
timdengyun, line 1897: @wenyingd  Add more check for namedport case.
The UTs have been updated as well.

wenyingd: What about 
```
if port.Port.Type == intstr.String && portNumber > 0 {
...
```

timdengyun:
> Most of the code looks good to me. The only left comment is the current implementation related with named port is complicated. Because that logic is not introduced by this change, having offline sync with Yun, we would use another change to refine the functions.

The issue is open to track refactor work.
https://github.com/vmware-tanzu/nsx-operator/issues/795

pkg/nsx/services/securitypolicy/builder.go
timdengyun: It makes sense to me, Thanks for advice. Fixed.

pkg/nsx/services/securitypolicy/builder.go
```diff
@@ -598,22 +581,44 @@ func (service *SecurityPolicyService) buildRuleOutGroup(obj *v1alpha1.SecurityPo
...
+		return util.GenerateIDByObjectWithSuffix(obj, ruleIndexHash)
 	}
+
 	prefix := common.SecurityPolicyPrefix
 	if createdFor == common.ResourceTypeNetworkPolicy {
 		prefix = common.NetworkPolicyPrefix
 	}
-	return util.GenerateID(fmt.Sprintf("%s", obj.UID), prefix, ruleHash, ruleIdxStr)
+	ruleIdxStr := fmt.Sprintf("%d", ruleIdx)
+	return strings.Join([]string{prefix, string(obj.UID), ruleIndexHash, ruleIdxStr}, common.ConnectorUnderline)
```
timdengyun, line 597: @wenyingd   Using this way for connecting string to avoid sprintf.

zhengxiexie (approved):
LGTM

wenyingd (approved):

PR #786, opened 2024-09-29
Do not check cache when deleting unrealized vpc

seanpang-vmware:
In VPC creation procedure, if nsx vpc is created but failed to realize, it is not insert into vpc store, so in vpc deletion method if vpc is not found in store, it should also return OK.

### DIFF
--- a/pkg/nsx/services/vpc/vpc.go
+++ b/pkg/nsx/services/vpc/vpc.go
@@ -204,16 +204,13 @@ func (s *VPCService) ListVPC() []model.Vpc {
 	return vpcSet
 }
 
+// DeleteVPC will try to delete VPC resource from NSX.
 func (s *VPCService) DeleteVPC(path string) error {
 	pathInfo, err := common.ParseVPCResourcePath(path)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


pkg/nsx/services/vpc/vpc.go
```diff
@@ -204,16 +204,13 @@ func (s *VPCService) ListVPC() []model.Vpc {
 	return vpcSet
 }
 
+// DeleteVPC will try to delete VPC resource from NSX.
 func (s *VPCService) DeleteVPC(path string) error {
```
gran-vmv, line 208: Maybe we can `add skipCacheCheck bool` here to minimize change.

gran-vmv (approved):
LGTM if we accept additional DELETE API call.

pkg/nsx/services/vpc/vpc.go
seanpang-vmware: in the following logic, when trying to get vpc from cache retuning nil, it will skip the delete cache part, if we add parameter here that will involve changing other logic in service and test which will cause much bigger effort.

dantingl (approved):

PR #792, opened 2024-10-03
Remove SubnetPort/Pod Finalizer

yanjunz97:
This PR removes the Finalizer in SubnetPort and Pod controller to provide a smooth
deletion experience for the user while ensure the stale NSX SubnetPort will be deleted
as expected.

Testing done:
- Create a SubnetPort and delete it. Observed the deletion will not stuck and check
   in the log that corresponding NSX SubnetPort will be deleted by the controller
   when a delete event is detected.
- Create a Pod and delete it. The deletion will still need a short period as we still have
   lifecycle-controller/system.vmware.com finalizer on the pod. But the log shows the
   corresponding NSX SubnetPort will be deleted by the controller when a delete event
   is detected.

### DIFF
--- a/pkg/controllers/pod/pod_controller.go
+++ b/pkg/controllers/pod/pod_controller.go
@@ -7,18 +7,17 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimachineryruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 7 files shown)


wenqiq (dismissed):
lgtm, not related to this PR, just curious about the `podIsDeleted` function.
```
func podIsDeleted(pod *v1.Pod) bool {
	return !pod.ObjectMeta.DeletionTimestamp.IsZero() || pod.Status.Phase == "Succeeded" || pod.Status.Phase == "Failed"
}
```

yanjunz97:
> lgtm, not related to this PR, just curious about the `podIsDeleted` function.
> 
> ```
> func podIsDeleted(pod *v1.Pod) bool {
> 	return !pod.ObjectMeta.DeletionTimestamp.IsZero() || pod.Status.Phase == "Succeeded" || pod.Status.Phase == "Failed"
> }
> ```

I think it's because when the pod is in phase Succeeded/Failed, it means the pod will never run again and thus the corresponding subnetport is no longer required. Then we can delete the nsx subnetport.
Ref to https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase

timdengyun (approved):
LGTM

wenqiq (approved):

PR #636, opened 2024-07-12
Fix bug, could delete subnetset when there are still stale subnetport…

zhengxiexie:
When there are stale subnetports in one subnet of subnetset, we should not allow deleting it.
Tests done:

```
apiVersion: nsx.vmware.com/v1alpha1
kind: SubnetPort
metadata:
  name: pod-59
  namespace: zx-test-2
spec:
  subnetSet: pod-default
```

create subnetport and delete pod-default, it will fail reporting

```
2024-07-15 03:30:04.994	INFO	subnetset/subnetset_controller.go:57	reconciling subnetset CR	{"subnetset": {"name":"pod-default","namespace":"zx-test-2"}}
2024-07-15 03:30:04.994	ERROR	subnetset/subnetset_controller.go:122	deletion failed, delete all the subnetports first	{"subnetset": {"name":"pod-default","namespace":"zx-test-2"}, "error": "stale subnet ports found while deleting subnetset zx-test-2/pod-default"}
2024-07-15 03:30:05.034	INFO	subnetset/subnetset_controller.go:204	updated SubnetSet	{"Name": "pod-default", "Namespace": "zx-test-2", "New Conditions": [{"type":"Ready","status":"False","lastTransitionTime":"2024-07-15T03:30:04Z","reason":"SubnetNotReady","message":"stale subnet ports found while deleting subnetset zx-test-2/pod-default"}]}
2024-07-15 03:30:05.034	DEBUG	recorder/recorder.go:104	stale subnet ports found while deleting subnetset zx-test-2/pod-default	{"type": "Warning", "object": {"kind":"SubnetSet","namespace":"zx-test-2","name":"pod-default","uid":"779fb1b6-50c5-49cf-b870-04e6ea6f13c7","apiVersion":"nsx.vmware.com/v1alpha1","resourceVersion":"1897779"}, "reason": "FailUpdate"}
```

delete vm-default will succeed since it has no port attached in it.

PR #798, opened 2024-10-11
Improve Subnet/SubnetSet GC

wenqiq:
Improve Subnet/SubnetSet GC

Improve NSX Subnet CleanUp

For Subnet garbage collection (GC): 
Retrieve the Subnet CRUID values of all Subnets from the store, and delete the corresponding stale Subnets for those whose UID is not found in the existing Subnet CRs in Kubernetes.

For SubnetSet garbage collection (GC), two tasks need to be handled:

1. For each existing SubnetSet CR, delete any Subnet under it that does not have a SubnetPort attached.
2. Retrieve the SubnetSet CRUID values of all Subnets from the store, and delete the corresponding stale Subnets for those whose SubnetSet UID is not found in the existing SubnetSet CRs in Kubernetes.

### DIFF
--- a/pkg/controllers/subnet/subnet_controller.go
+++ b/pkg/controllers/subnet/subnet_controller.go
@@ -382,8 +382,8 @@ func (r *SubnetReconciler) collectGarbage(ctx context.Context) {
 	crdSubnetIDsSet := sets.New[string](crdSubnetIDs...)
 
 	subnetUIDs := r.SubnetService.ListSubnetIDsFromNSXSubnets()
-	subnetSetIDsToDelete := subnetUIDs.Difference(crdSubnetIDsSet)
-	for subnetSetID := range subnetSetIDsToDelete {
+	subnetIDsToDelete := subnetUIDs.Difference(crdSubnetIDsSet)
+	for subnetSetID := range subnetIDsToDelete {
 		nsxSubnets := r.SubnetService.ListSubnetCreatedBySubnet(string(subnetSetID))
 		metrics.CounterInc(r.SubnetService.NSXConfig, metrics.ControllerDeleteTotal, common.MetricResTypeSubnet)
 
--- a/pkg/controllers/subnet/subnet_controller.go
+++ b/pkg/controllers/subnet/subnet_controller.go
@@ -379,31 +379,20 @@ func (r *SubnetReconciler) collectGarbage(ctx context.Context) {
 		log.Error(err, "Failed to list Subnet CRs")
 		return
 	}
+	crdSubnetIDsSet := sets.New[string](crdSubnetIDs...)
 
-	var nsxSubnetList []*model.VpcSubnet
-	for _, crdSubnetID := range crdSubnetIDs {
-		nsxSubnetList = append(nsxSubnetList, r.SubnetService.ListSubnetCreatedBySubnet(crdSubnetID)...)
-	}
-	if len(nsxSubnetList) == 0 {
-		log.Info("No Subnets found in NSX, garbage collection complete")
-		return
-	}
-
-	crdSubnetIDsSet := sets.NewString(crdSubnetIDs...)
-	for _, nsxSubnet := range nsxSubnetList {
-		uid := nsxutil.FindTag(nsxSubnet.Tags, servicecommon.TagScopeSubnetCRUID)
-		if crdSubnetIDsSet.Has(uid) {
-			continue
-		}
-
-		log.Info("GC collected Subnet CR", "UID", uid)
+	subnetUIDs := r.SubnetService.ListSubnetIDsFromNSXSubnets()
+	subnetSetIDsToDelete := subnetUIDs.Difference(crdSubnetIDsSet)
+	for subnetSetID := range subnetSetIDsToDelete {
+		nsxSubnets := r.SubnetService.ListSubnetCreatedBySubnet(string(subnetSetID))
 		metrics.CounterInc(r.SubnetService.NSXConfig, metrics.ControllerDeleteTotal, common.MetricResTypeSubnet)
 
-		if err := r.SubnetService.DeleteSubnet(*nsxSubnet); err != nil {
-			log.Error(err, "Failed to delete NSX subnet", "NSX Subnet UID", uid)
+		log.Info("Subnet garbage collection, cleaning stale Subnets", "Count", len(nsxSubnets))
+		if err := r.deleteSubnets(nsxSubnets); err != nil {
+			log.Error(err, "Subnet garbage collection, failed to delete NSX subnet", "SubnetUID", subnetSetID)
 			metrics.CounterInc(r.SubnetService.NSXConfig, metrics.ControllerDeleteFailTotal, common.MetricResTypeSubnet)
 		} else {
-			log.Info("Successfully deleted NSX subnet", "NSX Subnet UID", uid)
+			log.Info("Subnet garbage collection, successfully deleted NSX subnet", "SubnetUID", subnetSetID)
 			metrics.CounterInc(r.SubnetService.NSXConfig, metrics.ControllerDeleteSuccessTotal, common.MetricResTypeSubnet)
 		}
 	}
--- a/pkg/controllers/subnet/subnet_controller.go
+++ b/pkg/controllers/subnet/subnet_controller.go
@@ -383,16 +383,16 @@ func (r *SubnetReconciler) collectGarbage(ctx context.Context) {
 
 	subnetUIDs := r.SubnetService.ListSubnetIDsFromNSXSubnets()
 	subnetIDsToDelete := subnetUIDs.Difference(crdSubnetIDsSet)
-	for subnetSetID := range subnetIDsToDelete {
-		nsxSubnets := r.SubnetService.ListSubnetCreatedBySubnet(string(subnetSetID))
+	for subnetID := range subnetIDsToDelete {
+		nsxSubnets := r.SubnetService.ListSubnetCreatedBySubnet(string(subnetID))
 		metrics.CounterInc(r.SubnetService.NSXConfig, metrics.ControllerDeleteTotal, common.MetricResTypeSubnet)
 
 		log.Info("Subnet garbage collection, cleaning stale Subnets", "Count", len(nsxSubnets))
 		if err := r.deleteSubnets(nsxSubnets); err != nil {
-			log.Error(err, "Subnet garbage collection, failed to delete NSX subnet", "SubnetUID", subnetSetID)
+			log.Error(err, "Subnet garbage collection, failed to delete NSX subnet", "SubnetUID", subnetID)
 			metrics.CounterInc(r.SubnetService.NSXConfig, metrics.ControllerDeleteFailTotal, common.MetricResTypeSubnet)
 		} else {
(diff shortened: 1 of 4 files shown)


pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -261,39 +261,30 @@ func (r *SubnetSetReconciler) CollectGarbage(ctx context.Context) {
...
 
-	subnetSetIDs := sets.New[string]()
-	for _, subnetSet := range subnetSetList.Items {
-		if err := r.deleteSubnetForSubnetSet(subnetSet, true); err != nil {
-			metrics.CounterInc(r.SubnetService.NSXConfig, metrics.ControllerDeleteFailTotal, MetricResTypeSubnetSet)
-		} else {
-			metrics.CounterInc(r.SubnetService.NSXConfig, metrics.ControllerDeleteSuccessTotal, MetricResTypeSubnetSet)
-		}
-		subnetSetIDs.Insert(string(subnetSet.UID))
-	}
```
yanjunz97, line 286: I notice 2 issues:

1. We need to keep this part. There are 2 steps in subnetset gc: 1) delete the subnets that belongs to a subnetset which does not have CR in k8s, which has already implemented in this PR, 2) for subnet under subnetset with CR in k8s, we also need to delete it if there is no subnetport under it, which is implemented in this deleted block.

2) When removing subnetset finalizer we slightly broke this function. If you take a look at the last version, where it defines `func (r *SubnetSetReconciler) DeleteSubnetForSubnetSet(obj v1alpha1.SubnetSet, updataStatus bool) (bool, error)` which does not return an error when `portNums > 0`. This is used for the case to skip deleting the subnet with subnetport.

wenqiq: For the first one:
Not understand why  `for subnet under subnetset with CR in k8s, we also need to delete it if there is no subnetport under it`?        
    
For the second one:
In the current code, it will return an error when `portNums > 0`. it is in the `deleteSubnets` function, which is called by `deleteSubnetForSubnetSet`.
```
func (r *SubnetSetReconciler) deleteSubnets(nsxSubnets []*model.VpcSubnet) error {
	for _, nsxSubnet := range nsxSubnets {
		portNums := len(r.SubnetPortService.GetPortsOfSubnet(*nsxSubnet.Id))
		if portNums > 0 {
			return fmt.Errorf("fail to delete Subnet/%s from SubnetSet CR, there is stale ports", *nsxSubnet.Id)
		}
```

timdengyun: Yes, @yanjunz97  is right, this delete the original fix from
https://github.com/vmware-tanzu/nsx-operator/pull/636
if there is portNums, you should continue instead of return, 
because there are multiple subnetsets under the same subnetset.

Also, you change lockSubnet order, which should be put before read portNum,
this is logic in @yanjunz97 PR:
https://github.com/vmware-tanzu/nsx-operator/pull/770/files#diff-ea8fdc4a9adcc30f1ee646f03f10d565526b3a2b6fec57ae9abaa316c14b8d89R296
Please change these logic back as before in DeleteSubnetForSubnetSet function.

timdengyun: For the first one: @yanjunz97 is right
this was mentioned in PR:
https://github.com/vmware-tanzu/nsx-operator/pull/769/files#r1784012526

For the question:
`for subnet under subnetset with CR in k8s, we also need to delete it if there is no subnetport under it?
`
subnetset has this specific GC process, 
The normal GC process is to delete stale NSX subnet if their corresponding K8s subnet/subnset CR has been deleted.

The other GC is to delete NSX subnet scale down in the subnetset CR,
when the nsx subnet allocated from subnetset CR has no more ports, we should GC this nsx subnet and reclaim it, but this GC shoud be only for 
subnetset GC @dantingl  please correct me if misunderstanding. 
Previous Janwei's GC logic is to do this logic, but has some issue,s
we should fix this logic and add the normal GC process as you are doing in this PR.

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -261,39 +261,30 @@ func (r *SubnetSetReconciler) CollectGarbage(ctx context.Context) {
...
-	for _, subnet := range nsxSubnetList {
-		if !r.SubnetService.IsOrphanSubnet(*subnet, subnetSetIDs) {
+	subnetSetUIDs := r.SubnetService.SubnetStore.ListIndexFuncValues(servicecommon.TagScopeSubnetSetCRUID)
```
timdengyun, line 272: subnetSetUIDs -> nsxSubnetSetUIDs is better, because this is gotten from NSX resource store.

timdengyun: there is a service function:
ListSubnetCreatedBySubnetSet to do same thing, we had better to call this service function rather than writing the same logic here.

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -261,39 +261,30 @@ func (r *SubnetSetReconciler) CollectGarbage(ctx context.Context) {
...
+	for uid := range subnetSetUIDs {
+		if crdSubnetSetIDsSet.Has(uid) {
 			continue
 		}
-		if err := r.SubnetService.DeleteSubnet(*subnet); err != nil {
+		nsxSubnets := r.SubnetService.SubnetStore.GetByIndex(servicecommon.TagScopeSubnetSetCRUID, string(uid))
```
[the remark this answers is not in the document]
wenqiq: Thanks for reviewing. Good catch. Updated.

pkg/controllers/subnetset/subnetset_controller.go
wenqiq: done.

pkg/controllers/subnetset/subnetset_controller.go
wenqiq: Updated.

pkg/controllers/subnetset/subnetset_controller.go
wenqiq: > subnetSetUIDs -> nsxSubnetSetUIDs is better, because this is gotten from NSX resource store.

After reviewing the source code, I think that seems to be the case that the `ListIndexFuncValues` returns the `SubnetSetCRUIDs`, instead of SubnetSet ID in NSX. 
For example, if there is `SubnetSet` cr-id1 in K8s, has two Subnet ns-cr-id1-xxx0, ns-cr-id-xxx1 in NSX, the `ListIndexFuncValues` will return `cr-id1`.

https://github.com/kubernetes/client-go/blob/c38ce11939a7cecae6e64542554e4de50d901919/tools/cache/thread_safe_store.go#L114
```
type Index map[string]sets.String
	func (i *storeIndex) getIndexValues(indexName string) []string {
		index := i.indices[indexName]
		names := make([]string, 0, len(index))
		for key := range index {
			names = append(names, key)
		}
		return names
	}
```

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -301,59 +299,60 @@ func (r *SubnetSetReconciler) deleteSubnetBySubnetSetName(ctx context.Context, s
...
-			return fmt.Errorf("fail to delete Subnet/%s from SubnetSet CR: %+v", *nsxSubnet.Id, err)
+			log.Error(fmt.Errorf("failed to delete NSX Subnet/%s: %+v", *nsxSubnet.Id, err), "Skipping to next subnet")
+			continue
 		}
 		r.SubnetService.UnlockSubnet(nsxSubnet.Path)
 	}
-	log.Info("Successfully deleted all Subnets", "subnetCount", len(nsxSubnets))
-	return nil
+	if hitError {
+		err = errors.New("one or more errors occurred while deleting subnets, ")
```
yanjunz97, line 343: Is this expected to end with ", "?

pkg/nsx/services/subnet/subnet.go
```diff
@@ -352,26 +342,56 @@ func (service *SubnetService) GetSubnetByPath(path string) (*model.VpcSubnet, er
...
+// type Index map[string]sets.String
+//
+//	func (i *storeIndex) getIndexValues(indexName string) []string {
+//		index := i.indices[indexName]
+//		names := make([]string, 0, len(index))
+//		for key := range index {
+//			names = append(names, key)
+//		}
+//		return names
+//	}
```
yanjunz97, line 366: stale code?

yanjunz97 (commented):
Overall LGTM, just some nits

pkg/nsx/services/subnet/subnet.go
wenqiq: I think it's not easy to understand here, so added some more annotations here.

pkg/controllers/subnetset/subnetset_controller.go
wenqiq: done.

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -379,31 +379,20 @@ func (r *SubnetReconciler) collectGarbage(ctx context.Context) {
...
-	crdSubnetIDsSet := sets.NewString(crdSubnetIDs...)
-	for _, nsxSubnet := range nsxSubnetList {
-		uid := nsxutil.FindTag(nsxSubnet.Tags, servicecommon.TagScopeSubnetCRUID)
-		if crdSubnetIDsSet.Has(uid) {
-			continue
-		}
-
-		log.Info("GC collected Subnet CR", "UID", uid)
+	subnetUIDs := r.SubnetService.ListSubnetIDsFromNSXSubnets()
+	subnetSetIDsToDelete := subnetUIDs.Difference(crdSubnetIDsSet)
```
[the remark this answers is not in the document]
wenqiq: done

yanjunz97 (dismissed):

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -379,31 +379,20 @@ func (r *SubnetReconciler) collectGarbage(ctx context.Context) {
...
-		uid := nsxutil.FindTag(nsxSubnet.Tags, servicecommon.TagScopeSubnetCRUID)
-		if crdSubnetIDsSet.Has(uid) {
-			continue
-		}
-
-		log.Info("GC collected Subnet CR", "UID", uid)
+	subnetUIDs := r.SubnetService.ListSubnetIDsFromNSXSubnets()
+	subnetIDsToDelete := subnetUIDs.Difference(crdSubnetIDsSet)
+	for subnetID := range subnetIDsToDelete {
+		nsxSubnets := r.SubnetService.ListSubnetCreatedBySubnet(string(subnetID))
```
zhengxiexie, line 387: Redundant type conversion for subnetID

wenqiq, line 387:
```suggestion
		nsxSubnets := r.SubnetService.ListSubnetCreatedBySubnet(subnetID)
```

zhengxiexie (dismissed):
LGTM

yanjunz97 (approved):

zhengxiexie (approved):
LGTM

PR #800, opened 2024-10-14
Fix the bug where the namespace cannot be deleted

zhengxiexie:
Fix the bug where the namespace cannot be deleted due to the username of the namespace deletion.
Use global log. 

Test Done:
On VC UI, namespace can be created and deleted succesfully.

### DIFF
--- a/pkg/controllers/namespace/namespace_controller.go
+++ b/pkg/controllers/namespace/namespace_controller.go
@@ -132,6 +132,33 @@ func (r *NamespaceReconciler) createDefaultSubnetSet(ns string, defaultSubnetSiz
 	return nil
 }
 
+func (r *NamespaceReconciler) deleteDefaultSubnetSet(ns string) error {
+	subnetSets := []string{
+		types.DefaultVMSubnetSet,
+		types.DefaultPodSubnetSet,
+	}
+	for _, name := range subnetSets {
+		if err := retry.OnError(retry.DefaultRetry, func(err error) bool {
+			return err != nil
+		}, func() error {
+			obj := &v1alpha1.SubnetSet{}
+			err := r.Client.Get(context.Background(), client.ObjectKey{
+				Namespace: ns,
+				Name:      name,
+			}, obj)
+			if err != nil {
+				return client.IgnoreNotFound(err)
+			}
+			log.Info("delete default SubnetSet", "Namespace", ns, "Name", name)
+			return r.Client.Delete(context.Background(), obj)
+		}); err != nil {
+			log.Error(err, "failed to delete SubnetSet", "Namespace", ns, "Name", name)
+			return err
+		}
+	}
+	return nil
+}
+
 func (r *NamespaceReconciler) namespaceError(ctx context.Context, k8sObj client.Object, msg string, err error) {
 	logErr := util.If(err == nil, errors.New(msg), err).(error)
 	log.Error(logErr, msg)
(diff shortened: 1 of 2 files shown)


pkg/controllers/namespace/namespace_controller.go
```diff
@@ -132,6 +132,33 @@ func (r *NamespaceReconciler) createDefaultSubnetSet(ns string, defaultSubnetSiz
...
+			err := r.Client.Get(context.Background(), client.ObjectKey{
+				Namespace: ns,
+				Name:      name,
+			}, obj)
+			if err != nil {
+				return client.IgnoreNotFound(err)
+			}
+			log.Info("delete default SubnetSet", "Namespace", ns, "Name", name)
+			return r.Client.Delete(context.Background(), obj)
+		}); err != nil {
```
TaoZou1, line 154: do you need to delete it from the store?

zhengxiexie: It initiates a delete event, but the actual deletion occurs during the subnet set reconciliation process.

pkg/controllers/subnetset/subnetset_webhook.go
```diff
@@ -65,17 +61,17 @@ func (v *SubnetSetValidator) Handle(ctx context.Context, req admission.Request)
...
 	} else {
 		err := v.decoder.Decode(req, subnetSet)
 		if err != nil {
-			subnetsetlog.Error(err, "error while decoding SubnetSet", "SubnetSet", req.Namespace+"/"+req.Name)
+			log.Error(err, "error while decoding SubnetSet", "SubnetSet", req.Namespace+"/"+req.Name)
 			return admission.Errored(http.StatusBadRequest, err)
 		}
 	}
-	subnetsetlog.Info("request user-info", "name", req.UserInfo.Username)
+	log.Info("request user-info", "name", req.UserInfo.Username)
```
lxiaopei, line 74: remove this debug log？

zhengxiexie: ok

pkg/controllers/namespace/namespace_controller.go
```diff
@@ -132,6 +132,33 @@ func (r *NamespaceReconciler) createDefaultSubnetSet(ns string, defaultSubnetSiz
...
+		if err := retry.OnError(retry.DefaultRetry, func(err error) bool {
+			return err != nil
+		}, func() error {
+			obj := &v1alpha1.SubnetSet{}
+			err := r.Client.Get(context.Background(), client.ObjectKey{
+				Namespace: ns,
+				Name:      name,
+			}, obj)
+			if err != nil {
+				return client.IgnoreNotFound(err)
```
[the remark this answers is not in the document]
zhengxiexie: if err when getting subnetset, it is not always "NotFound" error right?
yes, will retry
what happened for this IgnoreNotFound error?
if it is really NotFound, we ignore it and not delete, which is as expected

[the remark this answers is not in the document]
lxiaopei: if err != nil {
				return client.IgnoreNotFound(err)

we cannot know the err is NotFound error, if failed to connect to k8s api server, for Get, it also throws error but here return "client.IgnoreNotFound(err)"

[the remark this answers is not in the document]
zhengxiexie: The definition is 
```
func IgnoreNotFound(err error) error {
	if apierrors.IsNotFound(err) {
		return nil
	}
	return err
}
```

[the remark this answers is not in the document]
lxiaopei: OK

pkg/controllers/subnetset/subnetset_webhook.go
```diff
@@ -65,17 +61,17 @@ func (v *SubnetSetValidator) Handle(ctx context.Context, req admission.Request)
...
 	} else {
 		err := v.decoder.Decode(req, subnetSet)
 		if err != nil {
-			subnetsetlog.Error(err, "error while decoding SubnetSet", "SubnetSet", req.Namespace+"/"+req.Name)
+			log.Error(err, "error while decoding SubnetSet", "SubnetSet", req.Namespace+"/"+req.Name)
 			return admission.Errored(http.StatusBadRequest, err)
 		}
 	}
-	subnetsetlog.Info("request user-info", "name", req.UserInfo.Username)
+	log.V(1).Info("request user-info", "name", req.UserInfo.Username)
```
lxiaopei, line 74: remove？

TaoZou1 (approved):

lxiaopei (approved):

PR #809, opened 2024-10-21
Remove staticroute finalizer

TaoZou1:
The finalizer may block the deletion since the NSX resource may take long time to delete. Remove the finalizer so the deleting operation will return immediately


timdengyun:
Close this PR since there is a new PR https://github.com/vmware-tanzu/nsx-operator/pull/814/ to do the same sync with the latest changes.

TaoZou1:
Close it since Yun has done the job.

PR #815, opened 2024-10-21
Set is_recursive to true while deleting VPC

TaoZou1:
VPC deletion added a is_recursive flag to delete children resources. 
The default value is false. Set it to true while deleting VPC


seanpang-vmware (approved):
LGTM

wenyingd (approved):
LGTM

PR #820, opened 2024-10-22
Add webhook to prevent SubnetSet/Subnet deletion with stale SubnetPort

yanjunz97:
This PR adds webhook server for SubnetSet and Subnet to block the deletion of SubnetSet/Subnet if there are stale SubnetPorts under it.

Testing done:
- Create a Subnet and a SubnetPort using that Subnet, trying to delete that Subnet with be blocked with `Error from server (Forbidden): admission webhook "subnet.validating.crd.nsx.vmware.com" denied the request: Subnet ns-1/subnet-test with stale SubnetPorts cannot be deleted`
- Create a SubnetSet and a SubnetPort using that SubnetSet, trying to delete that SubnetSet with be blocked with `Error from server (Forbidden): admission webhook "subnetset.validating.crd.nsx.vmware.com" denied the request: SubnetSet ns-1/subnetset-test with stale SubnetPorts cannot be deleted`

### DIFF
--- a/pkg/controllers/subnet/subnet_webhook.go
+++ b/pkg/controllers/subnet/subnet_webhook.go
@@ -1,0 +1,69 @@
+package subnet
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	admissionv1 "k8s.io/api/admission/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/apis/vpc/v1alpha1"
+)
+
+var NSXOperatorSA = "system:serviceaccount:vmware-system-nsx:ncp-svc-account"
+
+// Create validator instead of using the existing one in controller-runtime because the existing one can't
+// inspect admission.Request in Handle function.
+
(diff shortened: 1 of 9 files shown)


zhengxiexie (dismissed):
LGTM

pkg/controllers/subnet/subnet_webhook.go
```diff
@@ -0,0 +1,69 @@
...
+				return admission.Denied(fmt.Sprintf("Subnet %s/%s with stale SubnetPorts cannot be deleted", subnet.Namespace, subnet.Name))
+			}
+		}
+	}
+	return admission.Allowed("")
+}
+
+func (v *SubnetValidator) checkSubnetPort(ctx context.Context, ns string, subnetName string) (bool, error) {
+	crdSubnetPorts := &v1alpha1.SubnetPortList{}
+	err := v.Client.List(ctx, crdSubnetPorts)
```
wenqiq, line 59: Here we could add an option to filter `Subnetports` results under a specific namespace?
```suggestion
	err := v.Client.List(ctx, crdSubnetPorts, client.InNamespace(ns))
```

yanjunz97: Updated, thanks!

wenqiq (approved):
lgtm

zhengxiexie (approved):

PR #825, opened 2024-10-25
Fix the Subnet creating with stale VPC issue

wenqiq:
When creating and deleting the same Namespace and creating a Subnet in that Namespace, the creating Subnet may fail.

### DIFF
--- a/pkg/controllers/networkinfo/networkinfo_controller.go
+++ b/pkg/controllers/networkinfo/networkinfo_controller.go
@@ -457,7 +457,7 @@ func (r *NetworkInfoReconciler) fetchStaleVPCsByNamespace(ctx context.Context, n
 		return nil, nil
 	}
 
-	return r.Service.GetVPCsByNamespace(ctx, ns), nil
+	return r.Service.GetVPCsByNamespace(ns), nil
 }
 
 func (r *NetworkInfoReconciler) deleteVPCsByName(ctx context.Context, ns string) error {
--- a/pkg/controllers/networkinfo/networkinfo_controller.go
+++ b/pkg/controllers/networkinfo/networkinfo_controller.go
@@ -447,7 +447,7 @@ func (r *NetworkInfoReconciler) fetchStaleVPCsByNamespace(ctx context.Context, n
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			// if the Namespace has been deleted, we never know whether it`s a shared Namespace, The GC will delete the stale VPCs
-			log.Info("Namespace does not exist, skipping deletion of NSX VPC", "Namespace", ns)
+			log.Info("Namespace does not exist while fetching stale VPCs", "Namespace", ns)
 			return nil, nil
 		}
 		return nil, fmt.Errorf("failed to check if Namespace is shared for NS %s: %w", ns, err)
--- a/pkg/controllers/networkinfo/networkinfo_controller.go
+++ b/pkg/controllers/networkinfo/networkinfo_controller.go
@@ -445,6 +445,11 @@ func (r *NetworkInfoReconciler) CollectGarbage(ctx context.Context) {
 func (r *NetworkInfoReconciler) fetchStaleVPCsByNamespace(ctx context.Context, ns string) ([]*model.Vpc, error) {
 	isShared, err := r.Service.IsSharedVPCNamespaceByNS(ctx, ns)
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// if the Namespace has been deleted, we never know whether it`s a shared Namespace, The GC will delete the stale VPCs
+			log.Info("Namespace does not exist, skipping deletion of NSX VPC", "Namespace", ns)
+			return nil, nil
+		}
 		return nil, fmt.Errorf("failed to check if Namespace is shared for NS %s: %w", ns, err)
 	}
 	if isShared {
--- a/pkg/nsx/services/vpc/store.go
+++ b/pkg/nsx/services/vpc/store.go
@@ -87,11 +87,11 @@ func (s *VPCStore) GetByIndex(key string, value string) []*model.Vpc {
 	return vpcs
 }
 
-func (s *VPCStore) GetVPCsByNamespace(ns string) []*model.Vpc {
+func (s *VPCStore) GetVPCsByNamespaceFromStore(ns string) []*model.Vpc {
 	return s.GetByIndex(common.TagScopeNamespace, ns)
 }
 
-func (s *VPCStore) GetVPCsByNamespaceID(namespaceID string) []*model.Vpc {
+func (s *VPCStore) GetVPCsByNamespaceIDFromStore(namespaceID string) []*model.Vpc {
 	return s.GetByIndex(common.TagScopeNamespaceUID, namespaceID)
 }
 
--- a/pkg/nsx/services/vpc/vpc.go
+++ b/pkg/nsx/services/vpc/vpc.go
@@ -183,7 +183,7 @@ func InitializeVPC(service common.Service) (*VPCService, error) {
 	return VPCService, nil
 }
 
-func (s *VPCService) GetVPCsByNamespaceUID(ctx context.Context, namespace string) []*model.Vpc {
+func (s *VPCService) GetCurrentVPCsByNamespace(ctx context.Context, namespace string) []*model.Vpc {
 	namespaceObj, sharedNamespace, err := s.resolveSharedVPCNamespace(ctx, namespace)
 	if err != nil {
 		log.Error(err, "Failed to get Namespace")
@@ -744,7 +744,7 @@ func (s *VPCService) CreateOrUpdateVPC(ctx context.Context, obj *v1alpha1.Networ
 		return nil, err
 	}
 
-	existingVPC := s.GetVPCsByNamespaceUID(ctx, ns)
+	existingVPC := s.GetCurrentVPCsByNamespace(ctx, ns)
 	updateVpc := len(existingVPC) != 0
 	if updateVpc && isShared { // We now consider only one VPC for one namespace
 		log.Info("The shared VPC already exist", "Namespace", ns)
@@ -1122,7 +1122,7 @@ func (s *VPCService) ListVPCInfo(ns string) []common.VPCResourceInfo {
 	}
 
 	// List VPCs from local store.
-	vpcs := s.GetVPCsByNamespaceUID(context.Background(), ns)
+	vpcs := s.GetCurrentVPCsByNamespace(context.Background(), ns)
 	log.Info("Got VPCs by Namespace from store", "Namespace", ns, "VPCCount", len(vpcs))
 	for _, v := range vpcs {
 		vpcResourceInfo, err := common.ParseVPCResourcePath(*v.Path)
--- a/pkg/nsx/services/vpc/vpc.go
+++ b/pkg/nsx/services/vpc/vpc.go
@@ -184,7 +184,7 @@ func InitializeVPC(service common.Service) (*VPCService, error) {
 	return VPCService, nil
 }
 
-func (s *VPCService) GetVPCsByFreshNamespace(ctx context.Context, namespace string) []*model.Vpc {
+func (s *VPCService) GetVPCsByNamespaceUID(ctx context.Context, namespace string) []*model.Vpc {
 	namespaceObj, sharedNamespace, err := s.resolveSharedVPCNamespace(ctx, namespace)
 	if err != nil {
 		log.Error(err, "Failed to get Namespace")
@@ -197,7 +197,6 @@ func (s *VPCService) GetVPCsByFreshNamespace(ctx context.Context, namespace stri
 }
 
 func (s *VPCService) GetVPCsByNamespace(ctx context.Context, namespace string) []*model.Vpc {
-	// only used in NetworkInfo controller, no need to check whether is a shared Namespace
 	namespaceObj, sharedNamespace, err := s.resolveSharedVPCNamespace(ctx, namespace)
 	if err != nil {
 		log.Error(err, "Failed to get Namespace")
@@ -757,7 +756,7 @@ func (s *VPCService) CreateOrUpdateVPC(ctx context.Context, obj *v1alpha1.Networ
 		return nil, err
 	}
 
-	existingVPC := s.GetVPCsByFreshNamespace(ctx, ns)
+	existingVPC := s.GetVPCsByNamespaceUID(ctx, ns)
 	updateVpc := len(existingVPC) != 0
 	if updateVpc && isShared { // We now consider only one VPC for one namespace
 		log.Info("The shared VPC already exist", "Namespace", ns)
@@ -1135,7 +1134,7 @@ func (s *VPCService) ListVPCInfo(ns string) []common.VPCResourceInfo {
 	}
 
 	// List VPCs from local store.
-	vpcs := s.GetVPCsByFreshNamespace(context.Background(), ns)
+	vpcs := s.GetVPCsByNamespaceUID(context.Background(), ns)
 	log.Info("Got VPCs by Namespace from store", "Namespace", ns, "VPCCount", len(vpcs))
 	for _, v := range vpcs {
 		vpcResourceInfo, err := common.ParseVPCResourcePath(*v.Path)
--- a/pkg/nsx/services/vpc/vpc.go
+++ b/pkg/nsx/services/vpc/vpc.go
@@ -10,7 +10,6 @@ import (
 	stderrors "github.com/vmware/vsphere-automation-sdk-go/lib/vapi/std/errors"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 	v1 "k8s.io/api/core/v1"
-	apirrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/retry"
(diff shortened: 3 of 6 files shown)


pkg/nsx/services/vpc/vpc.go
```diff
@@ -519,42 +522,57 @@ func (s *VPCService) DeleteLBPool(path string) error {
...
+	if err := s.Client.Get(ctx, types.NamespacedName{Name: name}, obj); err != nil {
+		log.Error(err, "Failed to fetch namespace", "Namespace", name)
+		return nil, err
+	}
+	return obj, nil
+}
+
+func (s *VPCService) getSharedVPCNamespaceFromNS(ns string) (*v1.Namespace, *v1.Namespace, error) {
+	obj, err := s.getNamespace(ns)
+	fmt.Println("00000", err, obj)
```
zhengxiexie, line 551: remove it

wenqiq: done

pkg/nsx/services/vpc/vpc.go
```diff
@@ -519,42 +522,56 @@ func (s *VPCService) DeleteLBPool(path string) error {
...
+		return obj, nil, nil
 	}
 
 	// If no annotation nsx.vmware.com/shared_vpc_namespace on ns, this is not a shared vpc
 	shared_ns, exist := annos[common.AnnotationSharedVPCNamespace]
 	if !exist {
-		return "", nil
+		return obj, nil, nil
+	}
+	if shared_ns == ns {
```
timdengyun, line 566: shared_ns -> sharedNs

pkg/nsx/services/vpc/vpc.go
```diff
@@ -519,42 +522,56 @@ func (s *VPCService) DeleteLBPool(path string) error {
...
-		return "", err
+	if err := s.Client.Get(ctx, types.NamespacedName{Name: name}, obj); err != nil {
+		log.Error(err, "Failed to fetch namespace", "Namespace", name)
+		return nil, err
+	}
+	return obj, nil
+}
+
+func (s *VPCService) getSharedVPCNamespaceFromNS(ns string) (*v1.Namespace, *v1.Namespace, error) {
+	obj, err := s.getNamespace(ns)
```
timdengyun, line 550: It's not good that mixing namespace obj of input `ns' and sharedNamespace in the one function.
Also, this func: `getSharedVPCNamespaceFromNS` as its name stated that it's for getting sharedNamespace from input ns. If you need to get input `ns` namespace obj, you should just call `getNamespace`.

wenqiq: Yes, I think renaming the function can improve clarity, as the name `getSharedVPCNamespaceFromNS` is a bit ambiguous. Since this function checks if a namespace is associated with a shared VPC namespace and retrieves both the original and shared VPC namespaces if applicable, a name that reflects these actions would be more descriptive.  Do you think renaming it `resolveSharedVPCNamespace` make sense?

pkg/nsx/services/vpc/vpc.go
wenqiq: done.

zhengxiexie (dismissed):
LGTM

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -445,6 +445,11 @@ func (r *NetworkInfoReconciler) CollectGarbage(ctx context.Context) {
 func (r *NetworkInfoReconciler) fetchStaleVPCsByNamespace(ctx context.Context, ns string) ([]*model.Vpc, error) {
 	isShared, err := r.Service.IsSharedVPCNamespaceByNS(ns)
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// if the Namespace has been deleted, we never know whether it`s a shared Namespace, The GC will delete the stale VPCs
+			log.Info("Namespace is not existing, skipping deletion of NSX VPC", "Namespace", ns)
```
wenyingd, line 450:
```suggestion
			log.Info("Namespace does not exist, skipping deletion of NSX VPC", "Namespace", ns)
```

pkg/nsx/services/vpc/store.go
```diff
@@ -85,10 +89,18 @@ func (vs *VPCStore) GetVPCsByNamespace(ns string) []*model.Vpc {
 	for _, vpc := range vpcs {
 		mvpc := vpc.(*model.Vpc)
 		tags := mvpc.Tags
+		nameMatch := false
+		idMatch := false
 		for _, tag := range tags {
-			if *tag.Scope == common.TagScopeNamespace && *tag.Tag == ns {
-				ret = append(ret, mvpc)
+			if *tag.Scope == common.TagScopeNamespace && *tag.Tag == ns.Name {
```
wenyingd, line 95: Why not update the `VPCStore` to configure with NS UID as index in `InitializeVPC`, then here we can directly use UUID as index to search the VPCs.

pkg/nsx/services/vpc/vpc.go
```diff
@@ -519,42 +522,58 @@ func (s *VPCService) DeleteLBPool(path string) error {
...
 	if !exist {
-		return "", nil
+		return obj, nil, nil
+	}
+	if sharedNs == ns {
+		return obj, obj, nil
+	}
+	sharedNamespace, err := s.getNamespace(sharedNs)
+	if err != nil {
+		// if sharedNamespace is not existing?
```
wenyingd, line 573:
```suggestion
		// if sharedNamespace does not exist
```

wenyingd: I don't think it is a valid case that a shared NS doesn't exist but is used as annotation on another NS. I suppose you add this check here is for security, maybe add comments to explain it is for security?

pkg/nsx/services/vpc/store.go
```diff
@@ -74,7 +75,10 @@ func (vs *VPCStore) Apply(i interface{}) error {
 	return nil
 }
 
-func (vs *VPCStore) GetVPCsByNamespace(ns string) []*model.Vpc {
+func (vs *VPCStore) GetVPCsByNamespace(ns *v1.Namespace) []*model.Vpc {
+	if ns == nil {
```
TaoZou1, line 79: when you use the GetVPCsByNamespace, it seems you didn't check if the result is nil, just iterate. How about also return empty []*model.Vpc if ns== nil?

pkg/nsx/services/vpc/store.go
```diff
@@ -85,10 +89,18 @@ func (vs *VPCStore) GetVPCsByNamespace(ns string) []*model.Vpc {
...
 		tags := mvpc.Tags
+		nameMatch := false
+		idMatch := false
 		for _, tag := range tags {
-			if *tag.Scope == common.TagScopeNamespace && *tag.Tag == ns {
-				ret = append(ret, mvpc)
+			if *tag.Scope == common.TagScopeNamespace && *tag.Tag == ns.Name {
+				nameMatch = true
 			}
+			if *tag.Scope == common.TagScopeNamespaceUID && *tag.Tag == string(ns.UID) {
```
TaoZou1, line 98: how about merge Line95/98? or if you think it's too long. you could check nameMatch first in line 98, if it's false, no need to compare the UID

pkg/nsx/services/vpc/store.go
wenqiq: The ret is nil in `var ret []*model.Vpc`, if no value is assigned to this value. It’s equal to nil.

pkg/nsx/services/vpc/store.go
wenqiq: A tag`s Scope can not be both `common.TagScopeNamespaceUID` and `common.TagScopeNamespace`, I don’t think it can be merged here. Do you have a change suggestion?

pkg/nsx/services/vpc/store.go
timdengyun: Yes, if that, we should add a new indexer function in VPC store, currently, there is VPC's ID as key in the VPC store.  We need to consider there are multiple NS UID for multiple VPCs, i.e, shared VPCs, also in the future, we might need to support one NS UID for multiple VPCs.

pkg/nsx/services/vpc/vpc.go
timdengyun: The new function name make sense if your prefer to return sharedns obj and normal ns obj one function

pkg/nsx/services/vpc/vpc.go
```diff
@@ -519,42 +522,59 @@ func (s *VPCService) DeleteLBPool(path string) error {
...
 
 	// If no annotation nsx.vmware.com/shared_vpc_namespace on ns, this is not a shared vpc
-	shared_ns, exist := annos[common.AnnotationSharedVPCNamespace]
+	sharedNs, exist := annos[common.AnnotationSharedVPCNamespace]
 	if !exist {
-		return "", nil
+		return obj, nil, nil
+	}
+	if sharedNs == ns {
+		return obj, obj, nil
```
timdengyun, line 569: For this function, the second output argument is to return sharedNS obj, and the first one is to return Ns obj for non-shared VPC case.
So, you need to when differentiate  the results to caller, it should be:
`return nil, obj, nil`

pkg/nsx/services/vpc/vpc.go
```diff
@@ -519,42 +522,59 @@ func (s *VPCService) DeleteLBPool(path string) error {
...
+	}
+	sharedNamespace, err := s.getNamespace(sharedNs)
+	if err != nil {
+		// if sharedNamespace does not exist, add this case for security,
+		// It shouldn't happen that a shared Namespace doesn't exist but is used as an annotation on another Namespace.
+		return nil, nil, err
 	}
-	return shared_ns, nil
+	return obj, sharedNamespace, nil
```
timdengyun, line 577: Likewise, it should be
`return nil, sharedNamespace, nil`
because this case is to return ns for shared VPCs not for non-shared VPCs

pkg/nsx/services/vpc/vpc.go
```diff
@@ -519,42 +522,59 @@ func (s *VPCService) DeleteLBPool(path string) error {
...
 	annos := obj.Annotations
 	// If no annotaion on ns, then this is not a shared VPC ns
 	if len(annos) == 0 {
-		return "", nil
+		return obj, nil, nil
 	}
 
 	// If no annotation nsx.vmware.com/shared_vpc_namespace on ns, this is not a shared vpc
-	shared_ns, exist := annos[common.AnnotationSharedVPCNamespace]
+	sharedNs, exist := annos[common.AnnotationSharedVPCNamespace]
```
timdengyun, line 564: `sharedNs` is confusing,  this var is used to represent the namespaces for shared VPC.
It's better to call it `nsForSharedVPCs`

pkg/nsx/services/vpc/store.go
wenqiq: Implement the VPC index in https://github.com/vmware-tanzu/nsx-operator/pull/825

pkg/nsx/services/vpc/vpc.go
wenqiq: done

zhengxiexie (dismissed):
LGTM

pkg/nsx/services/vpc/vpc.go
```diff
@@ -521,53 +536,64 @@ func (s *VPCService) DeleteLBPool(path string) error {
...
+	if err := s.Client.Get(ctx, types.NamespacedName{Name: name}, obj); err != nil {
+		log.Error(err, "Failed to fetch Namespace", "Namespace", name)
 		return nil, err
 	}
 	return obj, nil
 }
 
-func (s *VPCService) getSharedVPCNamespaceFromNS(ctx context.Context, ns string) (string, error) {
-	obj, err := getNamespace(s.Client, ctx, ns)
+// resolveSharedVPCNamespace will resolve the Namespace relationship based on VPC sharing,
```
wenyingd, line 564: I didn't get your point why this function must return both the original k8s NS object and the shared K8s NS object (if exists). If the purpose of this change is to return the K8s NS object (rather than the NS name) which is used to query the NSX VPC, I think you  can just add a 2nd K8s get operation to read the shared NS object based on the old implementation?

pkg/nsx/services/vpc/vpc.go
```diff
@@ -185,13 +184,29 @@ func InitializeVPC(service common.Service) (*VPCService, error) {
...
+	}
+	if sharedNamespace == nil {
+		return s.VpcStore.GetVPCsByNamespaceIDFromStore(string(namespaceObj.UID))
+	}
+	return s.VpcStore.GetVPCsByNamespaceIDFromStore(string(sharedNamespace.UID))
+}
+
 func (s *VPCService) GetVPCsByNamespace(ctx context.Context, namespace string) []*model.Vpc {
-	sns, err := s.getSharedVPCNamespaceFromNS(ctx, namespace)
+	// only used in NetworkInfo controller, no need to check whether is a shared Namespace
```
timdengyun, line 200: what this comment meaning? inside this func, we still need to check if is shared Namespace obj?

pkg/nsx/services/vpc/vpc.go
```diff
@@ -185,13 +184,29 @@ func InitializeVPC(service common.Service) (*VPCService, error) {
 	return VPCService, nil
 }
 
+func (s *VPCService) GetVPCsByFreshNamespace(ctx context.Context, namespace string) []*model.Vpc {
```
timdengyun, line 187: `GetVPCsByFreshNamespace` is not a good name, I understand you want to say this func will return existing VPCS for a given namesapce name, 
It's better to rename it as GetExistingVPCsByNamespace or GetCurrentVPCsByNamespace

`GetVPCsByNamespace` is to return all VPs in the store for a given namesapce name, including existing ones and deleted/staled VPCs

pkg/nsx/services/vpc/vpc.go
```diff
@@ -521,53 +536,64 @@ func (s *VPCService) DeleteLBPool(path string) error {
...
 	return false, err
 }
 
-func getNamespace(c client.Client, ctx context.Context, ns string) (*v1.Namespace, error) {
+func (s *VPCService) getNamespace(ctx context.Context, name string) (*v1.Namespace, error) {
 	obj := &v1.Namespace{}
-	if err := c.Get(ctx, types.NamespacedName{Name: ns}, obj); err != nil {
-		log.Error(err, "Failed to fetch Namespace", "Namespace", ns)
+	if err := s.Client.Get(ctx, types.NamespacedName{Name: name}, obj); err != nil {
+		log.Error(err, "Failed to fetch Namespace", "Namespace", name)
```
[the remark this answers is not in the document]
wenqiq: Capitalize the log sentence

pkg/nsx/services/vpc/vpc.go
wenqiq: If we use the old implementation method, we need to call the 'Get' K8s API twice to get the Na mespace(UID).

pkg/nsx/services/vpc/vpc.go
```diff
@@ -185,13 +183,20 @@ func InitializeVPC(service common.Service) (*VPCService, error) {
 	return VPCService, nil
 }
 
-func (s *VPCService) GetVPCsByNamespace(ctx context.Context, namespace string) []*model.Vpc {
-	sns, err := s.getSharedVPCNamespaceFromNS(ctx, namespace)
+func (s *VPCService) GetVPCsByNamespaceUID(ctx context.Context, namespace string) []*model.Vpc {
+	namespaceObj, sharedNamespace, err := s.resolveSharedVPCNamespace(ctx, namespace)
```
timdengyun, line 187: GetVPCsByNamespaceUID is a public interface function, it's not good to ask caller to pass `ctx`  argument since it's only used inside `resolveSharedVPCNamespace` not a part of logic necessary. We should provider the same interface contract with `GetVPCsByNamespace`

pkg/nsx/services/vpc/vpc.go
wenyingd: I mean the consumer actually only cares which K8s Namespace object can be used, but doesn't care it is a shared Namespace or its own Namespace, Why we must return two objects (plus error) in this function?

pkg/nsx/services/vpc/vpc.go
wenqiq: I think this is unreasonable. A public interface can't use context? How can we get the Namespace CR without using context? How do we obtain the UID of the Namespace? Use another function to get the Namespace CR? Or can you give a code suggestion?

wenqiq: Our goal is to get the VPC from the latest Namespace. The emphasis should be on the UID or the most recent Namespace, rather than focusing on the VPC. `GetVPCsByNamespaceUID ` is still not a good name. Renaming this function would be more appropriate.

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -445,14 +445,19 @@ func (r *NetworkInfoReconciler) CollectGarbage(ctx context.Context) {
 func (r *NetworkInfoReconciler) fetchStaleVPCsByNamespace(ctx context.Context, ns string) ([]*model.Vpc, error) {
 	isShared, err := r.Service.IsSharedVPCNamespaceByNS(ctx, ns)
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// if the Namespace has been deleted, we never know whether it`s a shared Namespace, The GC will delete the stale VPCs
+			log.Info("Namespace does not exist, skipping deletion of NSX VPC", "Namespace", ns)
```
wenyingd, line 450: The log is incorrect, this function is not about "deletion of NSX VPC".

wenyingd (commented):
One minor comment, others LGTM

pkg/nsx/services/vpc/vpc.go
wenqiq: If `resolveSharedVPCNamespace` is performing operations that may take time (e.g., making network calls or querying a database), the `ctx` can allow for a timeout or cancellation, preventing the function from hanging indefinitely. For example, if the function call takes too long, `ctx` could trigger a cancellation to release resources early.

timdengyun: I see, it makes sense. @wenyingd  How is you feel, it's good to pass ctx from caller for such case?  Thanks

wenyingd (approved):
LGTM

pkg/nsx/services/vpc/vpc.go
wenyingd: From antrea's experience, context is preferred to be passed as an argument of a public function if it is consumed inside the implementation, and let the caller decide how to construct the context.

timdengyun: I see, Thanks @wenyingd  for sharing your thoughts!

timdengyun (approved):
LGTM

PR #826, opened 2024-10-28
Remove ipaddressallocation finalizer

zhengxiexie:


pkg/controllers/ipaddressallocation/ipaddressallocation_controller_test.go
```diff
@@ -204,11 +201,10 @@ func TestIPAddressAllocationReconciler_Reconcile(t *testing.T) {
 	assert.NotEqual(t, ret, nil)
 	patch.Reset()
 
-	//  DeletionTimestamp.IsZero = true, Finalizers include util.FinalizerName, CreateorUpdateIPAddressAllocation succ
+	//  DeletionTimestamp.IsZero = true, Finalizers include util.FinalizerName, CreateOrUpdateIPAddressAllocation success
```
TaoZou1, line 204: do you still need the case?

zhengxiexie: ok

yanjunz97:
Do we need also delete the nsx IPAddressAllocation when the corresponding CR is not found, like what we do for other resources?
https://github.com/vmware-tanzu/nsx-operator/blob/09cf580c6b5683ee09d2331c493877e9762d380d/pkg/controllers/ipaddressallocation/ipaddressallocation_controller.go#L113

zhengxiexie:
> Do we need also delete the nsx IPAddressAllocation when the corresponding CR is not found, like what we do for other resources?
> 
> https://github.com/vmware-tanzu/nsx-operator/blob/09cf580c6b5683ee09d2331c493877e9762d380d/pkg/controllers/ipaddressallocation/ipaddressallocation_controller.go#L113

ok

TaoZou1 (approved):

PR #852, opened 2024-11-01
Move services_mock.go to pkg/mock

TaoZou1:
UT will exclude dir which name including mock. servcies_mock.go is a mock file and should'nt be included in UT.
Move servcies_mock.go to pkg/mock will increase the UT rate.

Reset the mockStaticRouteclient before use each time so the status will be reset to avoid random fail
else mockStaticRouteclient.EXPECT may return unexpected error

### DIFF
--- a/pkg/nsx/services/staticroute/staticroute_test.go
+++ b/pkg/nsx/services/staticroute/staticroute_test.go
@@ -338,10 +338,11 @@ func TestStaticRouteService_Cleanup(t *testing.T) {
 		Path: &staticRoutePath2,
 	}
 
-	service.StaticRouteStore.Add(staticRoute1)
-	service.StaticRouteStore.Add(staticRoute2)
-
 	t.Run("Successful cleanup", func(t *testing.T) {
+		service.StaticRouteStore.Add(staticRoute1)
+		service.StaticRouteStore.Add(staticRoute2)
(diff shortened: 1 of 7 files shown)


pkg/nsx/services/staticroute/staticroute_test.go
```diff
@@ -338,10 +338,11 @@ func TestStaticRouteService_Cleanup(t *testing.T) {
...
 	}
 
-	service.StaticRouteStore.Add(staticRoute1)
-	service.StaticRouteStore.Add(staticRoute2)
-
 	t.Run("Successful cleanup", func(t *testing.T) {
+		service.StaticRouteStore.Add(staticRoute1)
+		service.StaticRouteStore.Add(staticRoute2)
+		mockStaticRouteclient = mocks.NewMockStaticRoutesClient(mockController)
+		service.NSXClient.StaticRouteClient = mockStaticRouteclient
```
gran-vmv, line 345: Why this change will fix random failure? Could you explain this on PR description?

TaoZou1: reset the mockStaticRouteclient so the status will be reset

zhengxiexie (approved):
LGTM

timdengyun (approved):
LGTM

PR #879, opened 2024-11-11
Fix the potential race condition of InitializeSubnetService

zhengxiexie:
It is possible that an error occurs when trying to send to the already closed `fatalErrors` channel, resulting in a panic.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ bin/
 .DS_Store
 go.work
 go.work.sum
+.tool-versions
--- a/pkg/nsx/services/subnet/subnet.go
+++ b/pkg/nsx/services/subnet/subnet.go
@@ -65,20 +65,39 @@ func InitializeSubnetService(service common.Service) (*SubnetService, error) {
 		},
 	}
 
+	// Use sync.Once to ensure channel is closed only once
+	var closeOnceFatalError sync.Once
+	var closeOnceWgDone sync.Once
+	safeCloseFatalError := func(ce chan error) {
+		closeOnceFatalError.Do(func() {
+			close(ce)
+		})
+	}
+	safeCloseWgDone := func(ce chan bool) {
+		closeOnceWgDone.Do(func() {
+			close(ce)
+		})
+	}
+
 	wg.Add(1)
 	go subnetService.InitializeResourceStore(&wg, fatalErrors, ResourceTypeSubnet, nil, subnetService.SubnetStore)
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -2,4 +2,5 @@
 bin/
 .DS_Store
 go.work
-go.work.sum
\ No newline at end of file
+go.work.sum
+.tool-versions
```
wenyingd, line 6: Is this change needed?

pkg/nsx/services/subnet/subnet.go
```diff
@@ -65,20 +65,39 @@ func InitializeSubnetService(service common.Service) (*SubnetService, error) {
 		},
 	}
 
+	// Use sync.Once to ensure channel is closed only once
```
wenyingd, line 68: I would prefer to change like this,

```
func InitializeSubnetService(service common.Service) (*SubnetService, error) {
	wg := sync.WaitGroup{}
	fatalErrors := make(chan error, 1)
	defer close(fatalErrors)
	subnetService := &SubnetService{
		Service: service,
		SubnetStore: &SubnetStore{
			ResourceStore: common.ResourceStore{
				Indexer: cache.NewIndexer(keyFunc, cache.Indexers{
					common.TagScopeSubnetCRUID:    subnetIndexFunc,
					common.TagScopeSubnetSetCRUID: subnetSetIndexFunc,
					common.TagScopeVMNamespace:    subnetIndexVMNamespaceFunc,
					common.TagScopeNamespace:      subnetIndexNamespaceFunc,
				}),
				BindingType: model.VpcSubnetBindingType(),
			},
		},
	}

	wg.Add(1)
	go subnetService.InitializeResourceStore(&wg, fatalErrors, ResourceTypeSubnet, nil, subnetService.SubnetStore)
	wg.Wait()

	if len(fatalErrors) > 0 {
		err := <-fatalErrors
		return subnetService, err
	}

	return subnetService, nil
}
```

zhengxiexie: https://github.com/vmware-tanzu/nsx-operator/blob/main/pkg/nsx/services/securitypolicy/firewall.go#L116-L116
The initial reason is to return as early as possible if there are multiple resources.
You should consider closing channel `fatalErrors` in this way.

zhengxiexie: Yes, this version has no zombie routine.

PR #885, opened 2024-11-12
Fix realized state error check

yanjunz97:


pkg/nsx/services/realizestate/realize_state.go
```diff
@@ -55,6 +65,9 @@ func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, inte
 			if *result.State == model.GenericPolicyRealizedResource_STATE_REALIZED {
 				return nil
 			}
+			if *result.State == model.GenericPolicyRealizedResource_STATE_ERROR {
+				return NewRealizeStateError(fmt.Sprintf("%s realized with error", entityType))
```
dantingl, line 69: Can we also put any reason in result to the error massage?

yanjunz97: Added, thanks!

TaoZou1 (approved):

dantingl (approved):

PR #886, opened 2024-11-13
Check node label before handling it

TaoZou1:
For WCP supervisor cluster, the master node isn't a esxi host, just a VM.
Check the key of dict instead of value

### DIFF
--- a/pkg/controllers/common/utils.go
+++ b/pkg/controllers/common/utils.go
@@ -118,8 +118,8 @@ func getDefaultSubnetSetByNamespace(client k8sclient.Client, namespace string, r
 }
 
 func NodeIsMaster(node *v1.Node) bool {
-	for k := range node.Labels {
-		if k == LabelK8sMasterRole || k == LabelK8sControlRole {
+	for _, key := range [2]string{LabelK8sMasterRole, LabelK8sControlRole} {
+		if _, exists := node.Labels[key]; exists {
 			return true
 		}
 	}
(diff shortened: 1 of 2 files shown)


dantingl (dismissed):
Add UT？

pkg/controllers/common/utils.go
```diff
@@ -116,7 +116,7 @@ func getDefaultSubnetSetByNamespace(client k8sclient.Client, namespace string, r
 
 func NodeIsMaster(node *v1.Node) bool {
 	for k := range node.Labels {
-		if k == LabelK8sMasterRole || k == LabelK8sControlRole {
+		if strings.Contains(k, LabelK8sMasterRole) || strings.Contains(k, LabelK8sControlRole) {
```
wenqiq, line 119: Direct Key Access with `_, exists := map[key]`?

TaoZou1: Thanks. Done

dantingl (approved):

pkg/controllers/common/utils.go
```diff
@@ -118,8 +118,8 @@ func getDefaultSubnetSetByNamespace(client k8sclient.Client, namespace string, r
 }
 
 func NodeIsMaster(node *v1.Node) bool {
-	for k := range node.Labels {
-		if k == LabelK8sMasterRole || k == LabelK8sControlRole {
```
wenqiq, line 122: My thought is, can we directly extract this closure function here https://github.com/vmware-tanzu/nsx-operator/blob/07f48519c2622633711e1aa182d7d9e54e3ecdf0/test/e2e/framework.go#L174 and reuse it.

TaoZou1:
Drop. No need to change it. The bug has been closed.

PR #901, opened 2024-11-18
Implement the reconciler for SubnetConnectionBindingMap

wenyingd:
1. Implement the reconciler for SubnetConnectionBindingMap, it may update the SubnetConnectionBindingMap status with condition ready is false if its dependent Subnet or SubnetSet is not ready (or realized) or it hits errors when realizing NSX SubnetConnectionBindingMaps. It updates the status with ready condition as true if it is successfully realized on NSX. The reconciler also watches the Subnet/SubnetSet CR events to sync the connection binding maps.
2. The change also modifies the Subnet/SubnetSet reconciler to watch SubnetConnectionBindingMap CR events. If a Subnet/SubnetSet is used by a SubnetConnectionBindingMap, a finalizer is added on the corresponding Subnet/SubnetSet CR, and the finalizer is removed automatically if the CR is not used by any SubnetConnectionBindingMaps.

Test Done:
A child Subnet CR and a SubnetConnectionBindingMap CR are prepared,
```
# cat test.yaml 
kind: Subnet
metadata:
  name: subnet-child
  namespace: svc-tkg-domain-c10
spec:
  accessMode: Public
  subnetDHCPConfig:
    mode: DHCPDeactivated
  ipAddresses:
  - 192.168.200.0/24
---
metadata:
  name: binding1
  namespace: svc-tkg-domain-c10
spec:
  subnetName: subnet-child
  targetSubnetSetName: pod-default
  vlanTrafficTag: 201
---
```

1. When the dependent child Subnet is not ready 
```
# kubectl get subnetconnectionbindingmaps -n svc-tkg-domain-c10 -oyaml
apiVersion: v1
items:
- apiVersion: crd.nsx.vmware.com/v1alpha1
  metadata:
    annotations:
        {"apiVersion":"crd.nsx.vmware.com/v1alpha1","kind":"SubnetConnectionBindingMap","metadata":{"annotations":{},"name":"binding1","namespace":"svc-tkg-domain-c10"},"spec":{"subnetName":"subnet-child","targetSubnetSetName":"pod-default","vlanTrafficTag":201}}
    creationTimestamp: "2024-11-18T10:18:50Z"
    generation: 1
    name: binding1
    namespace: svc-tkg-domain-c10
    resourceVersion: "5527560"
    uid: 8abcf1c8-956b-45cc-bf18-5b4a33120a7a
  spec:
    subnetName: subnet-child
    targetSubnetSetName: pod-default
    vlanTrafficTag: 201
  status:
    conditions:
    - lastTransitionTime: "2024-11-18T10:19:35Z"
      message: Subnet CR subnet-child is not realized on NSX
      reason: DependencyNotReady
      status: "False"
      type: Ready
kind: List
metadata:
  resourceVersion: ""
```

After the SubnetConnectionBindingMap is successfully realized, its ready condition is updated as true
```
# kubectl get subnetbindings -n svc-tkg-domain-c10 -oyaml
apiVersion: v1
items:
- apiVersion: crd.nsx.vmware.com/v1alpha1
  metadata:
    annotations:
        {"apiVersion":"crd.nsx.vmware.com/v1alpha1","kind":"SubnetConnectionBindingMap","metadata":{"annotations":{},"name":"binding1","namespace":"svc-tkg-domain-c10"},"spec":{"subnetName":"subnet-child","targetSubnetSetName":"pod-default","vlanTrafficTag":201}}
    creationTimestamp: "2024-11-19T03:02:48Z"
    generation: 1
    name: binding1
    namespace: svc-tkg-domain-c10
    resourceVersion: "669501"
    uid: dcfbf90e-f20a-4b9e-b980-41f04eda1f4f
  spec:
    subnetName: subnet-child
    targetSubnetSetName: pod-default
    vlanTrafficTag: 201
  status:
    conditions:
    - lastTransitionTime: "2024-11-19T03:20:16Z"
      status: "True"
      type: Ready
kind: List
metadata:
  resourceVersion: ""
```

2. The finalizer is added on the dependent subnet and/or subnetset,
```
# kubectl get subnets -n svc-tkg-domain-c10 -oyaml subnet-child
kind: Subnet
metadata:
  annotations:
      {"apiVersion":"crd.nsx.vmware.com/v1alpha1","kind":"Subnet","metadata":{"annotations":{},"name":"subnet-child","namespace":"svc-tkg-domain-c10"},"spec":{"accessMode":"Public","ipAddresses":["192.168.200.0/24"],"subnetDHCPConfig":{"mode":"DHCPDeactivated"}}}
  creationTimestamp: "2024-11-19T04:23:28Z"
  finalizers:
  - subnet.nsx.vmware.com/finalizer
  generation: 2
  name: subnet-child
  namespace: svc-tkg-domain-c10
  resourceVersion: "759588"
  uid: 6cf18f78-5363-4e0c-8765-f039b20623e0
spec:
  accessMode: Public
  ipAddresses:
  - 192.168.200.0/24
  ipv4SubnetSize: 16
  subnetDHCPConfig:
    mode: DHCPDeactivated
status:
 ....


# kubectl get subnetsets -n svc-tkg-domain-c10 -oyaml pod-default
kind: SubnetSet
metadata:
  creationTimestamp: "2024-11-18T10:53:14Z"
  finalizers:
  - subnetset.nsx.vmware.com/finalizer
  generation: 2
  labels:
    nsxoperator.vmware.com/default-subnetset-for: Pod
  name: pod-default
  namespace: svc-tkg-domain-c10
  resourceVersion: "759589"
  uid: 8e223cfd-fd47-4011-9a14-11debe823d5d
spec:
  accessMode: PrivateTGW
  ipv4SubnetSize: 16
  subnetDHCPConfig: {}
status:
  ...
```
3. If deleting the dependent subnet (child subnet in the example), it is blocking and adding a condition deletionFailed with reason `SubnetInUse`
```
# kubectl delete subnets -n svc-tkg-domain-c10 subnet-child
subnet.crd.nsx.vmware.com "subnet-child" deleted
^C
# kubectl get subnets -n svc-tkg-domain-c10 -oyaml subnet-child
kind: Subnet
metadata:
  annotations:
      {"apiVersion":"crd.nsx.vmware.com/v1alpha1","kind":"Subnet","metadata":{"annotations":{},"name":"subnet-child","namespace":"svc-tkg-domain-c10"},"spec":{"accessMode":"Public","ipAddresses":["192.168.200.0/24"],"subnetDHCPConfig":{"mode":"DHCPDeactivated"}}}
  creationTimestamp: "2024-11-19T04:23:28Z"
  deletionGracePeriodSeconds: 0
  deletionTimestamp: "2024-11-19T05:35:13Z"
  finalizers:
  - subnet.nsx.vmware.com/finalizer
  generation: 3
  name: subnet-child
  namespace: svc-tkg-domain-c10
  resourceVersion: "761558"
  uid: 6cf18f78-5363-4e0c-8765-f039b20623e0
spec:
  accessMode: Public
  ipAddresses:
  - 192.168.200.0/24
  ipv4SubnetSize: 16
  subnetDHCPConfig:
    mode: DHCPDeactivated
status:
  conditions:
  - lastTransitionTime: "2024-11-19T04:23:29Z"
    message: NSX Subnet with DHCPDeactivated has been successfully created/updated
    reason: SubnetReady
    status: "True"
    type: Ready
  - lastTransitionTime: "2024-11-19T05:35:13Z"
    message: Subnet is used by SubnetConnectionBindingMap binding1 and not able to
      delete
    reason: SubnetInUse
    status: "True"
    type: DeletionFailed
  gatewayAddresses:
  - 192.168.200.1/24
  networkAddresses:
  - 192.168.200.0/24
```

4. After the SubnetConnectionBindingMap is deleted, the Subnet is deleted, and the finalizers are removed automatically.
```
# kubectl delete subnetbindings binding1 -n svc-tkg-domain-c10
subnetconnectionbindingmap.crd.nsx.vmware.com "binding1" deleted

# kubectl get subnets -n svc-tkg-domain-c10 -oyaml subnet-child
Error from server (NotFound): subnets.crd.nsx.vmware.com "subnet-child" not found

# kubectl get subnetsets -n svc-tkg-domain-c10 -oyaml pod-default
kind: SubnetSet
metadata:
  creationTimestamp: "2024-11-18T10:53:14Z"
  generation: 2
  labels:
    nsxoperator.vmware.com/default-subnetset-for: Pod
  name: pod-default
  namespace: svc-tkg-domain-c10
  resourceVersion: "763022"
  uid: 8e223cfd-fd47-4011-9a14-11debe823d5d
spec:
  accessMode: PrivateTGW
  ipv4SubnetSize: 16
  subnetDHCPConfig: {}
status:
  ...
```
5. SubnetSet scale-out and scale-in.
After scale-out the target SubnetSet, the number of generated NSX SubnetConnectionBindingMaps is increased from 1 to 2.
After scale-in the target SubnetSet, the number of generated NSX SubnetConnectionBindingMaps is decreased from 2 to 1,
No changes was updated to SubnetConnectionBindingMap CR unless errors occurred.

6. If the dependent SubnetSet has no NSX SubnetPort, and all NSX VpcSubnets are gc, the SubnetConnectionBindingMap CR status is changed to not-ready.
```
# kubectl describe subnetbindings -n svc-tkg-domain-c10 binding1 
Name:         binding1
Namespace:    svc-tkg-domain-c10
Labels:       <none>
Annotations:  <none>
API Version:  crd.nsx.vmware.com/v1alpha1
Kind:         SubnetConnectionBindingMap
Metadata:
  Creation Timestamp:  2024-11-19T06:57:29Z
  Generation:          1
  Resource Version:    866270
  UID:                 54dce36d-871e-4325-85fa-4ab523ad9f3b
Spec:
  Subnet Name:             subnet-child
  Target Subnet Set Name:  pod-default
  Vlan Traffic Tag:        201
Status:
  Conditions:
    Last Transition Time:  2024-11-19T08:11:28Z
    Message:               SubnetSet CR pod-default is not realized on NSX
    Reason:                DependencyNotReady
    Status:                False
    Type:                  Ready
Events:
  Type     Reason            Age                From                                   Message
  ----     ------            ----               ----                                   -------
  Normal   SuccessfulUpdate  38m                subnetconnectionbindingmap-controller  SubnetConnectionBindingMap CR has been successfully updated
  Warning  FailUpdate        2s (x13 over 22s)  subnetconnectionbindingmap-controller  no existing NSX VpcSubnet created by SubnetSet CR 'svc-tkg-domain-c10/pod-default'
```


pkg/controllers/subnetbinding/controller.go
```diff
@@ -0,0 +1,333 @@
...
+
+	r.StatusUpdater.IncreaseSyncTotal()
+
+	bindingMapCR := &v1alpha1.SubnetConnectionBindingMap{}
+	if err := r.Client.Get(ctx, req.NamespacedName, bindingMapCR); err != nil {
+		if apierrors.IsNotFound(err) {
+			// Try to delete stale NSX SubnetConnectionBindingMaps if exists
+			if err := r.SubnetBindingService.DeleteSubnetConnectionBindingMapsByCRName(req.Name, req.Namespace); err != nil {
+				log.Error(err, "Failed to delete NSX SubnetConnectionBindingMap", "SubnetConnectionBindingMap", req.NamespacedName)
+				return ResultRequeue, nil
```
wenyingd, line 97: Returns `nil` in Reconcile function even if an error happened because we want to use the ResultRequeue to control the delay to re-queue the request rather than the framework's default value. Otherwise, it may print such logs.
```
2024-11-19 08:16:55.803 INFO    controller/controller.go:314    Warning: Reconciler returned both a non-zero result and a non-nil error. The result will always be ignored if the error is non-nil and the non-nil error causes reqeueuing with exponential backoff. For more details, see: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler        {"controller": "subnetconnectionbindingmap", "controllerGroup": "crd.nsx.vmware.com", "controllerKind": "SubnetConnectionBindingMap", "SubnetConnectionBindingMap": {"name":"binding1","namespace":"svc-tkg-domain-c10"}, "namespace": "svc-tkg-domain-c10", "name": "binding1", "reconcileID": "b504bdcb-9a6b-4009-bc02-f00db6a9dfd1"}
2024-11-19 08:16:55.803 ERROR   controller/controller.go:316    Reconciler error        {"controller": "subnetconnectionbindingmap", "controllerGroup": "crd.nsx.vmware.com", "controllerKind": "SubnetConnectionBindingMap", "SubnetConnectionBindingMap": {"name":"binding1","namespace":"svc-tkg-domain-c10"}, "namespace": "svc-tkg-domain-c10", "name": "binding1", "reconcileID": "b504bdcb-9a6b-4009-bc02-f00db6a9dfd1", "error": "no existing NSX VpcSubnet created by SubnetSet CR 'svc-tkg-domain-c10/pod-default'"}
```

luolanzone (commented):
not finished.

pkg/controllers/subnetbinding/controller.go
```diff
@@ -0,0 +1,353 @@
...
+				retry:   true,
+			}
+		}
+	} else {
+		bms := r.SubnetBindingService.GetSubnetConnectionBindingMapsByChildSubnet(subnets[0])
+		if len(bms) > 0 {
+			dependency := r.SubnetBindingService.GetCRNameBySubnetConnectionBindingMap(bms[0])
+			return nil, &errorWithRetry{
+				message: fmt.Sprintf("Target Subnet CR %s is attached by %s", name, dependency),
+				error:   fmt.Errorf("target Subnet %s is already attached by SegmentConnectionBindingMap %s", name, dependency),
```
luolanzone, line 279: Why is this error not empty if Subnet is allowed to attached by multiple SubnetConnectionBindingMap?

pkg/controllers/subnetbinding/controller.go
```diff
@@ -0,0 +1,353 @@
...
+
+	// Check if the Subnet CR is nested.
+	if !isTarget {
+		bms := r.SubnetBindingService.GetSubnetConnectionBindingMapsByParentSubnet(subnets[0])
+		if len(bms) > 0 {
+			dependency := r.SubnetBindingService.GetCRNameBySubnetConnectionBindingMap(bms[0])
+			return nil, &errorWithRetry{
+				message: fmt.Sprintf("Subnet CR %s is working as target by %s", name, dependency),
+				error:   fmt.Errorf("Subnet %s already works as target in SegmentConnectionBindingMap %s", name, dependency),
+				retry:   true,
```
luolanzone, line 270: Is this a retryable error?

pkg/controllers/subnetbinding/controller.go
```diff
@@ -0,0 +1,353 @@
...
+			dependency := r.SubnetBindingService.GetCRNameBySubnetConnectionBindingMap(bms[0])
+			return nil, &errorWithRetry{
+				message: fmt.Sprintf("Target Subnet CR %s is attached by %s", name, dependency),
+				error:   fmt.Errorf("target Subnet %s is already attached by SegmentConnectionBindingMap %s", name, dependency),
+				retry:   true,
+			}
+		}
+	}
+
+	return subnets, nil
```
luolanzone, line 285: If we can ensure that the returned subnets array will always contains one valid subnet, we should return `subnets[0]`.

pkg/controllers/subnetbinding/controller.go
```diff
@@ -0,0 +1,353 @@
...
+		return nil, &errorWithRetry{
+			message: fmt.Sprintf("Unable to get SubnetSet CR %s", name),
+			error:   fmt.Errorf("failed to get SubnetSet %s in Namespace %s with error: %v", name, namespace, err),
+			retry:   false,
+		}
+	}
+
+	subnets := r.SubnetService.ListSubnetCreatedBySubnetSet(string(subnetSetCR.UID))
+	if len(subnets) == 0 {
+		log.Info("NSX VpcSubnets by SubnetSet CR do not exist", "key", subnetSetKey.String())
```
luolanzone, line 303: ditto, adding the UID info in the log as well

pkg/controllers/subnetbinding/controller.go
```diff
@@ -0,0 +1,353 @@
...
+				return
+			}
+			continue
+		}
+		newConditions = append(newConditions, cond)
+	}
+	bindingMap.Status.Conditions = newConditions
+	err := c.Status().Update(ctx, bindingMap)
+	if err != nil {
+		log.Error(err, "Failed to update SubnetConnectionBindingMap status", "Namespace", bindingMap.Namespace, "Name", bindingMap.Name)
```
luolanzone, line 350:
```suggestion
		log.Error(err, "Failed to update SubnetConnectionBindingMap status", "Namespace", bindingMap.Namespace, "Name", bindingMap.Name)
		return
```

pkg/controllers/subnetbinding/controller.go
```diff
@@ -0,0 +1,353 @@
...
+	condition := v1alpha1.Condition{
+		Type:    v1alpha1.Ready,
+		Status:  corev1.ConditionFalse,
+		Reason:  reason,
+		Message: msg,
+	}
+	updateBindingMapCondition(c, ctx, bindingMap, condition)
+}
+
+func updateBindingMapStatusWithReadyCondition(c client.Client, ctx context.Context, obj client.Object, _ metav1.Time, _ ...interface{}) {
```
luolanzone, line 326: Why not to define `func(client.Client, context.Context, client.Object, metav1.Time, ...interface{}){}` in the caller
and leave used parameters only in this function.

pkg/controllers/subnetbinding/subnets_handler.go
```diff
@@ -0,0 +1,102 @@
...
+func requeueSubnetConnectionBindingMapsBySubnet(ctx context.Context, c client.Client, namespace string, subnet string, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	bindingMapList := &v1alpha1.SubnetConnectionBindingMapList{}
+	err := c.List(ctx, bindingMapList, client.InNamespace(namespace))
+	if err != nil {
+		log.Error(err, "Failed to list SubnetConnectionBindingMaps with Subnet event", "Namespace", namespace, "Subnet", subnet)
+		return
+	}
+	for _, bm := range bindingMapList.Items {
+		if bm.Spec.SubnetName == subnet || bm.Spec.TargetSubnetName == subnet {
+			log.Info("Requeue SubnetConnectionBindingMap because because the dependent Subnet is ready", "Namespace", namespace, "Name", bm.Name)
```
luolanzone, line 68:
```suggestion
			log.Info("Requeue SubnetConnectionBindingMap because the dependent Subnet is ready", "Namespace", namespace, "Name", bm.Name)
```

edwardbadboy, line 68: In the above `PredicateFuncsSubnets`,  the condition `common.IsObjectUpdateToUnready(oldObj.Status.Conditions, newObj.Status.Conditions)` can also trigger an updated event. Should the log message change to " because the dependent Subnet realization state change"?

By the way, there are two repeated "because" in the log message.

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -242,12 +277,33 @@ func setSubnetReadyStatusFalse(client client.Client, ctx context.Context, obj cl
 			LastTransitionTime: transitionTime,
 		},
 	}
-	if err != nil {
+	if len(args) > 0 {
+		newConditions[0].Message = args[0].(string)
```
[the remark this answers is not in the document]
wenyingd: The current definition for `setSubnetReadyStatusFalse` must follow type `UpdateFailStatusFn` https://github.com/vmware-tanzu/nsx-operator/blob/main/pkg/controllers/common/utils.go#L168 . For subnet_controller, some caller needs to update message, some do not, so I have to use this format to cover the cases.

pkg/controllers/subnet/subnetbinding_handler.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+		if needFinalizer {
+			q.Add(req)
+		} else {
+			// Enqueue the Subnet request with 1s delay with the SubnetConnectionBindingMap deletion event.
+			// This is because when Subnet controller watches the deletion event at the same time as
+			// SubnetConnectionBindingMap Controller, and it may cause a wrong result of the Subnet dependency
+			// calculation. So we leave 1s to ensure the Subnet Controller reconciles Subnet status after
+			// the binding calculation is completed.
+			q.AddAfter(req, removeFinalizerDelay)
```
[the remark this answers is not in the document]
wenyingd: Removed this delay in the latest change as I update to create/delete finalizers on Subnet/SubnetSet according to the existence of SubnetConnectionBindingMap CR rather than the NSX resources.

pkg/controllers/subnetbinding/controller.go
```diff
@@ -0,0 +1,353 @@
...
+	SubnetBindingService *subnetbinding.BindingService
+	StatusUpdater        common.StatusUpdater
+}
+
+func StartSubnetBindingController(mgr ctrl.Manager, subnetService *subnet.SubnetService, subnetBindingService *subnetbinding.BindingService) {
+	reconciler := newReconciler(mgr, subnetService, subnetBindingService)
+	// Start the controller
+	if err := reconciler.setupWithManager(mgr); err != nil {
+		log.Error(err, "Failed to create controller", "controller", "SubnetConnectionBindingMap")
+		os.Exit(1)
```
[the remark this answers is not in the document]
wenyingd: I observed that some other modules also directly call `os.Exit` inside the controller modules, e.g., https://github.com/vmware-tanzu/nsx-operator/blob/main/pkg/controllers/pod/pod_controller.go#L155

pkg/controllers/subnetbinding/controller.go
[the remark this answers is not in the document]
wenyingd: The logs are like this, 
```
2024-11-27 10:08:51.934 INFO    controller/controller.go:183    Starting Controller     {"controller": "subnetconnectionbindingmap", "controllerGroup": "crd.nsx.vmware.com", "controllerKind": "SubnetConnectionBindingMap"}
```

Renamed the file.

pkg/controllers/subnetbinding/controller.go
wenyingd: updated

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -70,8 +73,44 @@ func (r *SubnetSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+				log.Error(err, "Failed to delete the finalizer", "subnet", req.NamespacedName)
+				msgFailDelFinalizer := "Failed to remove the finalizer on SubnetSet after no the dependency by SubnetConnectionBindingMaps"
+				r.StatusUpdater.UpdateFail(ctx, subnetsetCR, err, "Unable to remove the finalizer from SubnetSet",
+					setSubnetSetReadyStatusFalse, fmt.Sprint(msgFailDelFinalizer))
+				return ResultRequeue, err
+			}
+		}
+	}
+
 	if !subnetsetCR.ObjectMeta.DeletionTimestamp.IsZero() {
```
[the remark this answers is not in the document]
luolanzone: The finalizer removal logics should be inside of `if !subnetsetCR.ObjectMeta.DeletionTimestamp.IsZero(){}`. DeletionTimestamp will only be non-zero when a CR has a finalizer. So you should double check a few cases:
1. when the subnetsetCR.ObjectMeta.DeletionTimestamp is zero (create/update event) and getSubnetBindingCRsBySubnetSet returns non-empty bindingmaps, you should add the finalizer if there is no finalizer.

2. when subnetsetCR.ObjectMeta.DeletionTimestamp is not zero (it means Subnetset deletion event already happened), Check and clean up dependent resources created or owned by the SubnetsetCR, remove the finalizer if clean up is done or requeue when the clean up failed.

You should actually treat the CR deletion as an update event, you can still get the CR but DeletionTimestamp is not zero now. You can only get CR not found error when the finalizer is removed successfully.

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -70,8 +73,43 @@ func (r *SubnetSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		log.Error(err, "Unable to fetch SubnetSet CR", "SubnetSet", req.NamespacedName)
 		return ResultRequeue, err
 	}
+
+	hasBindings := r.subnetSetHasBindings(string(subnetsetCR.UID))
```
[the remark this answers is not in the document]
wenyingd: I prefer to call it as "SubnetBinding" rather than "BindingMap" like I what named the controller and service package. Rename the two functions as `getNSXSubnetBindingsBySubnetSet` and `getSubnetBindingCRsBySubnetSet`

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -70,8 +73,43 @@ func (r *SubnetSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	if len(hasBindings) > 0 {
+		if !controllerutil.ContainsFinalizer(subnetsetCR, servicecommon.SubnetSetFinalizerName) {
```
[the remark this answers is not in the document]
wenyingd: If the finalizer exists, the SubnetSet CR is not deleted cleanly from apiserver. So no need to add the check.

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -70,8 +73,43 @@ func (r *SubnetSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			if err := r.Client.Update(ctx, subnetsetCR); err != nil {
+				log.Error(err, "Failed to add finalizer", "subnetSet", req.NamespacedName)
+				msgFailAddFinalizer := fmt.Sprintf("Failed to add finalizer on SubnetSet for the dependency by SubnetConnectionBindingMap %s", hasBindings[0].Name)
+				r.StatusUpdater.UpdateFail(ctx, subnetsetCR, err, "Unable to add finalizer on SubnetSet used by SubnetConnectionBindingMap",
+					setSubnetSetReadyStatusFalse, msgFailAddFinalizer)
+				return ResultRequeue, err
+			}
+		}
+	} else {
+		if controllerutil.ContainsFinalizer(subnetsetCR, servicecommon.SubnetSetFinalizerName) {
```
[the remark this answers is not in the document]
wenyingd: No need, here we remove the finalizer only depends on if the SubnetBinding CRs exist or not. It is not about SubnetSet is to delete or not.

pkg/controllers/subnetset/subnetbinding_handler.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/vmware-tanzu/nsx-operator/pkg/controllers/common"
+	servicecommon "github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
+)
+
+var (
+	removeFinalizerDelay = 1 * time.Second
+)
+
+func requeueSubnetSetByBindingMapUpdate(ctx context.Context, c client.Client, objOld, objNew client.Object, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	newBM, _ := objNew.(*v1alpha1.SubnetConnectionBindingMap)
```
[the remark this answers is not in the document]
wenyingd: updated.

pkg/controllers/subnetset/subnetbinding_handler.go
```diff
@@ -0,0 +1,101 @@
...
+
+func enqueue(ctx context.Context, c client.Client, namespace, name string, needFinalizer bool, q workqueue.TypedRateLimitingInterface[reconcile.Request]) error {
+	subnetSetCR := &v1alpha1.SubnetSet{}
+	subnetSetKey := types.NamespacedName{Namespace: namespace, Name: name}
+	err := c.Get(ctx, subnetSetKey, subnetSetCR)
+	if err != nil {
+		log.Error(err, "Failed to get SubnetSet CR", "key", subnetSetKey.String())
+		return err
+	}
+	addedFinalizer := controllerutil.ContainsFinalizer(subnetSetCR, servicecommon.SubnetSetFinalizerName)
```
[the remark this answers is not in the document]
wenyingd: updated.

pkg/controllers/subnetset/subnetbinding_handler.go
```diff
@@ -0,0 +1,101 @@
...
+	if addedFinalizer != needFinalizer {
```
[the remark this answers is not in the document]
wenyingd: updated.

pkg/controllers/subnetset/subnetbinding_handler.go
```diff
@@ -0,0 +1,101 @@
...
+		req := reconcile.Request{
+			NamespacedName: types.NamespacedName{
+				Name:      name,
+				Namespace: namespace,
+			},
```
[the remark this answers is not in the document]
wenyingd: updated.

pkg/controllers/subnetset/subnetbinding_handler.go
```diff
@@ -0,0 +1,101 @@
...
+
+func requeueSubnetSetByBindingMapUpdate(ctx context.Context, c client.Client, objOld, objNew client.Object, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	newBM, _ := objNew.(*v1alpha1.SubnetConnectionBindingMap)
+	if newBM.Spec.TargetSubnetSetName == "" {
+		return
+	}
+	needFinalizer := common.IsObjectReady(newBM.Status.Conditions)
+	err := enqueue(ctx, c, newBM.Namespace, newBM.Spec.TargetSubnetSetName, needFinalizer, q)
+	if err != nil {
+		log.Error(err, "Failed to requeue SubnetSet", "Namespace", newBM.Namespace, "Name", newBM.Spec.TargetSubnetSetName)
```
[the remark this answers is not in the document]
wenyingd: Good catch, updated.

pkg/controllers/subnetset/subnetbinding_handler.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	oldBM := objOld.(*v1alpha1.SubnetConnectionBindingMap)
+	if common.IsObjectReady(oldBM.Status.Conditions) && oldBM.Spec.TargetSubnetSetName != "" &&
+		oldBM.Spec.TargetSubnetSetName != newBM.Spec.TargetSubnetSetName {
+		// Remove the finalizer from old target Subnet if it is not used.
+		err := enqueue(ctx, c, newBM.Namespace, oldBM.Spec.TargetSubnetSetName, false, q)
```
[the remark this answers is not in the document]
wenyingd: updated.

pkg/controllers/subnetset/subnetbinding_handler_test.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	req2 = reconcile.Request{
+		NamespacedName: types.NamespacedName{
+			Name:      "parentSet2",
+			Namespace: "default",
+		},
+	}
+)
+
+func TestRequeueSubnetByBindingMap(t *testing.T) {
```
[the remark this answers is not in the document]
wenyingd: updated.

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -70,8 +73,44 @@ func (r *SubnetSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+				r.StatusUpdater.UpdateFail(ctx, subnetsetCR, err, "Unable to remove the finalizer from SubnetSet",
+					setSubnetSetReadyStatusFalse, fmt.Sprint(msgFailDelFinalizer))
+				return ResultRequeue, err
+			}
+		}
+	}
+
 	if !subnetsetCR.ObjectMeta.DeletionTimestamp.IsZero() {
 		r.StatusUpdater.IncreaseDeleteTotal()
+		hasBindingsOnNSX := r.subnetSetHasBindings(string(subnetsetCR.UID))
```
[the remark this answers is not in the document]
wenyingd: > When SubnetSet is deleted, all its child Subnets will be deleted?

Yes.

> Will it be possible that one Subnet is in one SubnetSet, but it also has its own CR?

No.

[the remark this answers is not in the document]
wenyingd: updated.

pkg/controllers/subnetset/subnetbinding_handler.go
```diff
@@ -0,0 +1,101 @@
...
+	servicecommon "github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
+)
+
+var (
+	removeFinalizerDelay = 1 * time.Second
+)
+
+func requeueSubnetSetByBindingMapUpdate(ctx context.Context, c client.Client, objOld, objNew client.Object, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	newBM, _ := objNew.(*v1alpha1.SubnetConnectionBindingMap)
+	if newBM.Spec.TargetSubnetSetName == "" {
```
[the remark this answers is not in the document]
wenyingd: Do you mean user update SubnetBinding's target from TargetSubnetSetName to TargetSubnetName or from TargetSubnetName to TargetSubnetSetName?
It may happen, will update this logic.

pkg/controllers/subnetset/subnetset_controller.go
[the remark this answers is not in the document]
wenyingd: No, as I mentioned here https://github.com/vmware-tanzu/nsx-operator/pull/901#discussion_r1861463574, the finalizer should remove as long as no Binding CRs use this SubnetSet, no matter the SubnetSet CR is desired to delete or not.

wenyingd:
@luolanzone @edwardbadboy @lxiaopei In my latest change, I update the predicts to enqueue SubnetSet/Subnet events like this,

- SubnetConnectionBindingMap create event: It will enqueue Subnet/SubnetSet if it is used by the SubnetConnectionBindingMap
- SubnetConnectionBindingMap delete event: It will enqueue Subnet/SubnetSet if it is used by the SubnetConnectionBindingMap
- SubnetConnectionBindingMap update event: it will equeue Subnet/SubnetSet if the SubnetConnectionBindingMap's target is updated, these cases should include:  
    1. TargetSubnet1 -> TargetSubnet2
    2. TargetSubnet2 -> TargetSubnetSet1
    3. TargetSubnetSet1 -> TargetSubnetSet2
    4. TargetSubnetSet2 -> TargetSubnet1

Besides, I removed the filter on the compare bewteen  `hasSubnetBinding` and `wantSubnetBinding` from the requeueXXX functions, instead, the Subnet/SubnetSet request is enqueued directly, and let the reconciler to make the compare. This is to avoid the race conditions between the predicts and reconciler.

pkg/nsx/services/subnetbinding/tree.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+func (n *hNode) mergeChildNode(node *hNode) {
+	if node.resourceType == leafType {
+		n.childNodes = append(n.childNodes, node)
+		return
+	}
+
+	for _, cn := range n.childNodes {
+		if cn.resourceType == node.resourceType && cn.resourceID == node.resourceID {
```
edwardbadboy, line 27: This is linear search. In a single update/delete, is it possible that at some layer of the tree, the children nodes could be many? If each node at each layer would only have a few children, I think linear search should be OK.

For example, if we want to delete all subnetbindings, the orgs, projects, vpcs should be just a few (hundreds), but how many subnets and subnetbindings would b affected? Would linear search be an issue? Maybe we can generate some fake nodes and check how much CPU time it requires to merge all children into a big tree. I estimate that it won't be an issue because the hierarchical API at most only supports 5K children (as far as I know).

pkg/nsx/services/subnetbinding/tree.go
```diff
@@ -0,0 +1,221 @@
...
+		} else {
+			nsxBindingMap := nsxBindingMaps[*bm.Id]
+			s.BindingStore.Apply(&nsxBindingMap)
+		}
+	}
+
+	return nil
+}
+
+func (s *BindingService) hDeleteSubnetConnectionBindingMap(bindingMaps []*model.SubnetConnectionBindingMap) error {
```
edwardbadboy, line 199: When this is used in a total cleanup, is it possible for hDeleteSubnetConnectionBindingMap to exceed the maximum transaction size of a single hierarchical API call? Do we need to split the deletion into multiple calls?

edwardbadboy (commented):
Small comments. Overall looks good.

pkg/nsx/services/subnetbinding/tree.go
wenyingd: @edwardbadboy Thanks for catching this potential performance issue. I made some tests locally: I constructed 10W SubnetConnectionBindingMaps (1 org * 10 project * 20 VPC * 100 subnet * 5 binding), the total time cost by this function is about 5s.  I think this is acceptable for now. So I would prefer to improve it in another change separately.

pkg/nsx/services/subnetbinding/tree.go
wenyingd: Good catch, I would add paging logic to avoid the potential issue.

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -71,8 +74,44 @@ func (r *SubnetSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		log.Error(err, "Unable to fetch SubnetSet CR", "SubnetSet", req.NamespacedName)
 		return ResultRequeue, err
 	}
+
+	hasBindingCRs := r.getSubnetBindingCRsBySubnetSet(ctx, subnetsetCR)
+	if len(hasBindingCRs) > 0 {
+		if !controllerutil.ContainsFinalizer(subnetsetCR, servicecommon.SubnetSetFinalizerName) {
+			controllerutil.AddFinalizer(subnetsetCR, servicecommon.SubnetSetFinalizerName)
+			if err := r.Client.Update(ctx, subnetsetCR); err != nil {
+				log.Error(err, "Failed to add the finalizer", "subnetSet", req.NamespacedName)
```
yanjunz97, line 83:
```suggestion
				log.Error(err, "Failed to add the finalizer", "SubnetSet", req.NamespacedName)
```

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -279,7 +353,11 @@ func (r *SubnetSetReconciler) deleteSubnetBySubnetSetName(ctx context.Context, s
 
 func (r *SubnetSetReconciler) deleteSubnetForSubnetSet(subnetSet v1alpha1.SubnetSet, updateStatus, ignoreStaleSubnetPort bool) error {
 	nsxSubnets := r.SubnetService.SubnetStore.GetByIndex(servicecommon.TagScopeSubnetSetCRUID, string(subnetSet.GetUID()))
-	hasStaleSubnetPort, deleteErr := r.deleteSubnets(nsxSubnets)
+	// If ignoreStaleSubnetPort is true, we will actively delete the existing SubnetConnectionBindingMaps connected to the
+	// corresponding NSX Subnet. This happens in the GC case to scale-in the NSX Subnet if no subnet Port exists.
+	// For SubnetSet CR deletion event, we don't delete the existing SubnetConnectionBindingMaps but let the
+	// SubnetConnectionBindingMap controller do it after the binding CR is removed..
```
yanjunz97, line 359:
```suggestion
	// If ignoreStaleSubnetPort is true, we will actively delete the existing SubnetConnectionBindingMaps connected to the
	// corresponding NSX Subnet. This happens in the GC case to scale-in the NSX Subnet if no SubnetPort exists.
	// For SubnetSet CR deletion event, we don't delete the existing SubnetConnectionBindingMaps but let the
	// SubnetConnectionBindingMap controller do it after the binding CR is removed.
```

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -253,7 +327,7 @@ func (r *SubnetSetReconciler) CollectGarbage(ctx context.Context) {
 	for subnetSetID := range subnetSetIDsToDelete {
 		nsxSubnets := r.SubnetService.ListSubnetCreatedBySubnetSet(subnetSetID)
 		log.Info("SubnetSet garbage collection, cleaning stale Subnets for SubnetSet", "Count", len(nsxSubnets))
-		if _, err := r.deleteSubnets(nsxSubnets); err != nil {
+		if _, err := r.deleteSubnets(nsxSubnets, true); err != nil {
```
yanjunz97, line 330: Here we delete all the subnets under a subnetset if the subnetset CR is not exists. Just to confirm is this desired to also delete the subnetbindingmap here? It seems that you do not delete the subnetbindingmap in subnet GC when subnet CR does not exist.

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -0,0 +1,341 @@
...
+	r.StatusUpdater.IncreaseSyncTotal()
+
+	bindingMapCR := &v1alpha1.SubnetConnectionBindingMap{}
+	if err := r.Client.Get(ctx, req.NamespacedName, bindingMapCR); err != nil {
+		if apierrors.IsNotFound(err) {
+			r.StatusUpdater.IncreaseDeleteTotal()
+			// Try to delete NSX SubnetConnectionBindingMaps if exists
+			if err := r.SubnetBindingService.DeleteSubnetConnectionBindingMapsByCRName(req.Name, req.Namespace); err != nil {
+				log.Error(err, "Failed to delete NSX SubnetConnectionBindingMap", "SubnetConnectionBindingMap", req.NamespacedName)
+				return common.ResultRequeue, nil
```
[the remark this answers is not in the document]
wenyingd: I don't return error on demand. See this: https://github.com/vmware-tanzu/nsx-operator/pull/901#discussion_r1847900059

pkg/controllers/subnetset/subnetset_controller.go
wenyingd: Yes, here to delete the binding maps in SubnetSet Controller for the auto scale-in case, which is not triggered by by SubnetSet CR event. It is a nsx-operator internal behavior. But for Subnet CR, one CR is mapping one NSX VpcSubnet, and there is no internal behavior to auto scale-in/scale-out Subnets. So we don't introduce internal binding deletion behaviors.

pkg/nsx/services/subnetbinding/subnetbinding.go
```diff
@@ -0,0 +1,217 @@
...
+
+	return nil
+}
+
+// DeleteSubnetConnectionBindingMapsByCRName deletes all the SubnetConnectionBindingMaps generated by the given subnetBinding
+// CR's UID.
+func (s *BindingService) DeleteSubnetConnectionBindingMapsByCRName(bindingName string, bindingNamespace string) error {
+	bindingMaps := s.BindingStore.getBindingsByBindingMapCRName(bindingName, bindingNamespace)
+	return s.deleteSubnetConnectionBindingMaps(bindingMaps)
+}
```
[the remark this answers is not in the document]
wenyingd: Thinking more, I don't feel it is a valid case that a new SubnetConnectionBindingMap is created with the same namespaced name when`DeleteSubnetConnectionBindingMapsByCRName` is calling. This function is called by subnetbinding_controller when no SubnetConnectionBindingMap CR is found, and the Reconciler is running with a single goroutine, so we don't need to think the race condition here, right?

pkg/controllers/subnetbinding/subnets_handler.go
```diff
@@ -0,0 +1,108 @@
...
+func requeueBindingMapsBySubnetSetUpdate(ctx context.Context, c client.Client, objOld, objNew client.Object, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	subnetSet := objNew.(*v1alpha1.SubnetSet)
+	requeueSubnetConnectionBindingMapsBySubnetSet(ctx, c, subnetSet.Namespace, subnetSet.Name, q)
+}
+
+func requeueSubnetConnectionBindingMapsBySubnetSet(ctx context.Context, c client.Client, namespace string, subnetSet string, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	bindingMapList := &v1alpha1.SubnetConnectionBindingMapList{}
+	err := c.List(ctx, bindingMapList, client.InNamespace(namespace))
+	if err != nil {
+		log.Error(err, "Failed to list SubnetConnectionBindingMaps with SubnetSet event", "Namespace", namespace, "Subnet", subnetSet)
```
luolanzone, line 94:
```suggestion
		log.Error(err, "Failed to list SubnetConnectionBindingMaps with SubnetSet event", "Namespace", namespace, "SubnetSet", subnetSet)
```

pkg/controllers/subnetset/subnetbinding_handler.go
```diff
@@ -0,0 +1,98 @@
...
+func enqueue(ctx context.Context, c client.Client, namespace, name string, q workqueue.TypedRateLimitingInterface[reconcile.Request]) error {
+	subnetSetCR := &v1alpha1.SubnetSet{}
+	subnetSetKey := types.NamespacedName{Namespace: namespace, Name: name}
+	err := c.Get(ctx, subnetSetKey, subnetSetCR)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			log.Info("SubnetSet CR doesn't exist, ignore it", "key", subnetSetKey.String())
+			return nil
+		}
+		log.Error(err, "Failed to get SubnetSet CR", "key", subnetSetKey.String())
```
luolanzone, line 51: ditto 
```suggestion
		if errors.IsNotFound(err) {
			log.Info("SubnetSet CR doesn't exist, ignore it", "SubnetSet", subnetSetKey.String())
			return nil
		}
		log.Error(err, "Failed to get SubnetSet CR", "SubnetSet", subnetSetKey.String())
```

pkg/controllers/subnet/subnetbinding_handler.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+	}
+	if oldBM.Spec.TargetSubnetName != "" {
+		// Remove the finalizer from old target Subnet if it is not used.
+		enqueue(ctx, c, newBM.Namespace, oldBM.Spec.TargetSubnetName, q)
+	}
+}
+
+func enqueueSubnets(ctx context.Context, c client.Client, bindingMap *v1alpha1.SubnetConnectionBindingMap, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	if err := enqueue(ctx, c, bindingMap.Namespace, bindingMap.Spec.SubnetName, q); err != nil {
```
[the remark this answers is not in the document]
wenyingd: bindingMap.Spec.SubnetName can not be empty

pkg/controllers/common/dependency_watcher_test.go
```diff
@@ -0,0 +1,260 @@
...
+func TestPredicateFuncsBindingMap(t *testing.T) {
+	readyBM := &v1alpha1.SubnetConnectionBindingMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "bm1",
+			Namespace: "default",
+		},
+		Spec: v1alpha1.SubnetConnectionBindingMapSpec{
+			SubnetName:       "child",
+			TargetSubnetName: "parent1",
+			VLANTrafficTag:   202,
```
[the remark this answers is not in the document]
wenyingd: For subnet/subnetSet controller, no actions when vlan change. For subnet binding controller, it will enqueue, but the predict is not this one.

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -0,0 +1,356 @@
...
+	}()
+
+	r.StatusUpdater.IncreaseSyncTotal()
+
+	bindingMapCR := &v1alpha1.SubnetConnectionBindingMap{}
+	if err := r.Client.Get(ctx, req.NamespacedName, bindingMapCR); err != nil {
+		if apierrors.IsNotFound(err) {
+			r.StatusUpdater.IncreaseDeleteTotal()
+			// Try to delete NSX SubnetConnectionBindingMaps if exists
+			if err := r.SubnetBindingService.DeleteSubnetConnectionBindingMapsByCRName(req.Name, req.Namespace); err != nil {
```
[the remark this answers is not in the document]
wenyingd: It is already covered in the nsx service layer.

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -0,0 +1,343 @@
...
+		bms := r.SubnetBindingService.GetSubnetConnectionBindingMapsByChildSubnet(subnets[0])
+		if len(bms) > 0 {
+			dependency := r.SubnetBindingService.GetCRNameBySubnetConnectionBindingMap(bms[0])
+			return nil, &errorWithRetry{
+				message: fmt.Sprintf("Target Subnet CR %s is associated by %s", name, dependency),
+				error:   fmt.Errorf("target Subnet %s is already associated by SubnetConnectionBindingMap %s", name, dependency),
+				retry:   true,
+			}
+		}
+	}
```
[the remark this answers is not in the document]
wenyingd: No, the message is different between the two case since the child-parent relationship are different.

pkg/controllers/subnetset/subnetset_controller.go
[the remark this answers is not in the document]
wenyingd: It should not happen, the reconciler logic is following this workflow
1. SubnetSet CR is deleted in apiserver (not found the CR), line 65, delete NSX resources directly.
2. SubnetSet CR still exists in apiserver:
    2.1 No finalizer exists, but binding CR exists, add finalizer. This must be not a delete SubnetSet event (otherwise it shall go into case 1)
    2.2 Finalizer already exists, binding CR exists, no need to add finalizer. This could be a SubnetSet delete or update event, but this case we don't add finalizer.
    2.3 finalizer exists, binding CRs don't exist, need to delete finalizer. This could be a SubnetSet delete or update event, so this case we can't add pre-check to only support delete case

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -0,0 +1,343 @@
...
+			error:   fmt.Errorf("no existing NSX VpcSubnet created by SubnetSet CR '%s/%s'", namespace, name),
+			retry:   false,
+		}
+	}
+	return subnets, nil
+}
+
+func updateBindingMapStatusWithUnreadyCondition(c client.Client, ctx context.Context, obj client.Object, _ metav1.Time, _ error, args ...interface{}) {
+	bindingMap := obj.(*v1alpha1.SubnetConnectionBindingMap)
+	reason := args[0].(string)
```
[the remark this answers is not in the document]
wenyingd: Now this function has a single caller, and args[0] must set.

pkg/controllers/common/dependency_watcher.go
```diff
@@ -0,0 +1,91 @@
...
+
+var PredicateFuncsWithSubnetBindings = predicate.Funcs{
+	CreateFunc: func(e event.CreateEvent) bool {
+		return true
+	},
+	UpdateFunc: func(e event.UpdateEvent) bool {
+		oldBindingMap := e.ObjectOld.(*v1alpha1.SubnetConnectionBindingMap)
+		newBindingMap := e.ObjectNew.(*v1alpha1.SubnetConnectionBindingMap)
+		if oldBindingMap.Spec.TargetSubnetSetName != newBindingMap.Spec.TargetSubnetSetName ||
+			oldBindingMap.Spec.TargetSubnetName != newBindingMap.Spec.TargetSubnetName {
```
lxiaopei, line 82: for VLANTrafficTag change, do we need to handle?

pkg/nsx/services/subnetbinding/tree.go
```diff
@@ -0,0 +1,228 @@
...
+										bindingMap:   bindingMap,
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}, nil
```
luolanzone, line 96: It's quite long structure, consider to use variables:
```
	leafNode := &hNode{
		resourceID:   *bindingMap.Id,
		resourceType: leafType,
		bindingMap:   bindingMap,
	}

	vpcSubnetNode := &hNode{
		resourceID:   vpcInfo.ID,
		resourceType: "VpcSubnet",
		childNodes:   []*hNode{leafNode},
	}

	vpcNode := &hNode{
		resourceID:   vpcInfo.VPCID,
		resourceType: "Vpc",
		childNodes:   []*hNode{vpcSubnetNode},
	}

	projectNode := &hNode{
		resourceType: "Project",
		resourceID:   vpcInfo.ProjectID,
		childNodes:   []*hNode{vpcNode},
	}

	orgNode := &hNode{
		resourceType: "Org",
		resourceID:   vpcInfo.OrgID,
		childNodes:   []*hNode{projectNode},
	}
```

pkg/nsx/services/subnetbinding/subnetbinding.go
```diff
@@ -0,0 +1,236 @@
...
+
+// Apply sync bindingMaps on NSX and save into the store if succeeded to realize.
+func (s *BindingService) Apply(subnetPath string, bindingMaps []*model.SubnetConnectionBindingMap) error {
+	return s.hUpdateSubnetConnectionBindingMaps(subnetPath, bindingMaps)
+}
+
+// deleteSubnetConnectionBindingMaps uses HAPI call to delete multiple SubnetConnectionBindingMaps on NSX in one
+// transaction, and then delete these resources from local store. Since NSX can accept no more than 5K children Nodes
+// in on HAPI call, paging is used in this function to avoid scale issues.
```
luolanzone, line 172:
```suggestion
// in one HAPI call, paging is used in this function to avoid scale issues.
```

pkg/nsx/services/subnetbinding/subnetbinding.go
```diff
@@ -0,0 +1,236 @@
...
+func (s *BindingService) deleteSubnetConnectionBindingMaps(bindingMaps []*model.SubnetConnectionBindingMap) error {
+	// Add this check is for security purpose. The caller has similar pre-check and returned if no items exist in the bindingMaps.
```
luolanzone, line 174:
```suggestion
	// Add this check for security purpose. The caller has similar pre-check and returned if no items exist in the bindingMaps.
```

pkg/nsx/services/subnetbinding/subnetbinding.go
```diff
@@ -0,0 +1,236 @@
...
+	bindingMapsCount := len(bindingMaps)
+	if bindingMapsCount == 0 {
+		log.Info("No existing SubnetConnectionBindingMaps found in the store")
+		return nil
+	}
+	pages := (bindingMapsCount + hAPIPageSize - 1) / hAPIPageSize
+	for i := 1; i < pages+1; i++ {
```
luolanzone, line 181:
```suggestion
	for i := 1; i <= pages; i++ {
```

pkg/controllers/subnetbinding/subnets_handler.go
```diff
@@ -0,0 +1,108 @@
...
+		return
+	}
+	for _, bm := range bindingMapList.Items {
+		if bm.Spec.TargetSubnetSetName == subnetSet {
+			log.Info("Requeue SubnetConnectionBindingMap because because the dependent SubnetSet is ready or updated", "Namespace", namespace, "Name", bm.Name)
```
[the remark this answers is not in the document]
wenyingd: updated.

pkg/controllers/subnetset/subnetbinding_handler.go
wenyingd: updated.

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -0,0 +1,356 @@
...
+	bindingMapIdSetInStore := r.SubnetBindingService.ListSubnetConnectionBindingMapCRUIDsInStore()
+
+	if err = r.SubnetBindingService.DeleteMultiSubnetConnectionBindingMapsByCRs(bindingMapIdSetInStore.Difference(bindingMapIdSetByCRs)); err != nil {
+		log.Error(err, "Failed to delete stale SubnetConnectionBindingMaps")
+	}
+}
+
+var PredicateFuncsBindingMaps = predicate.Funcs{
+	UpdateFunc: func(e event.UpdateEvent) bool {
+		// Check the object type because multiple resource kinds are watched in this manager.
```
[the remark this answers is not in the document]
wenyingd: Renamed as `PredicateFuncsForBindingMaps`, `PredicateFuncsForSubnetSets`, `PredicateFuncsForSubnets`

pkg/controllers/common/dependency_watcher.go
wenyingd: This predict functions `PredicateFuncsWithSubnetBindings` is shared between Subnet Controller and SubnetSet Controller when watching SubnetConnectionBindingMaps. These two don't care about the VLAN changes, they only care the case that target subnet/subnetSets changed.

pkg/controllers/subnetbinding/subnets_handler.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func requeueSubnetConnectionBindingMapsBySubnetSet(ctx context.Context, c client.Client, namespace string, subnetSet string, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	bindingMapList := &v1alpha1.SubnetConnectionBindingMapList{}
+	err := c.List(ctx, bindingMapList, client.InNamespace(namespace))
+	if err != nil {
+		log.Error(err, "Failed to list SubnetConnectionBindingMaps with SubnetSet event", "Namespace", namespace, "Subnet", subnetSet)
+		return
+	}
+	for _, bm := range bindingMapList.Items {
```
lxiaopei, line 97: maybe we can use listOpts in List API to only return items we want?
listOpts := client.ListOptions{FieldSelector: fields.OneTermEqualSelector("spec.TargetSubnetSetName", subnetSet).String()}

pkg/controllers/subnetbinding/subnets_handler.go
```diff
@@ -0,0 +1,108 @@
...
+
+func requeueSubnetConnectionBindingMapsBySubnet(ctx context.Context, c client.Client, namespace string, subnet string, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	bindingMapList := &v1alpha1.SubnetConnectionBindingMapList{}
+	err := c.List(ctx, bindingMapList, client.InNamespace(namespace))
+	if err != nil {
+		log.Error(err, "Failed to list SubnetConnectionBindingMaps with Subnet event", "Namespace", namespace, "Subnet", subnet)
+		return
+	}
+	for _, bm := range bindingMapList.Items {
+		if bm.Spec.SubnetName == subnet || bm.Spec.TargetSubnetName == subnet {
```
lxiaopei, line 73: listOpts := client.ListOptions{FieldSelector: fields.OneTermEqualSelector("spec.TargetSubnetName", subnet.String()}

pkg/controllers/subnetset/subnetbinding_handler.go
```diff
@@ -0,0 +1,98 @@
...
+	newBM := objNew.(*v1alpha1.SubnetConnectionBindingMap)
+	oldBM := objOld.(*v1alpha1.SubnetConnectionBindingMap)
+
+	if newBM.Spec.TargetSubnetSetName == oldBM.Spec.TargetSubnetSetName {
+		return
+	}
+
+	if newBM.Spec.TargetSubnetSetName != "" {
+		if err := enqueue(ctx, c, newBM.Namespace, newBM.Spec.TargetSubnetSetName, q); err != nil {
+			return
```
lxiaopei, line 32: if there is error when enqueuing the new targetSubnetSet, the old targetSubnetSet will not be enqueued, will this cause issue?

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -71,8 +74,44 @@ func (r *SubnetSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	}
+
 	if !subnetsetCR.ObjectMeta.DeletionTimestamp.IsZero() {
 		r.StatusUpdater.IncreaseDeleteTotal()
+		bindingsOnNSX := r.getNSXSubnetBindingsBySubnetSet(string(subnetsetCR.UID))
+		if len(bindingsOnNSX) > 0 {
+			err := fmt.Errorf("SubnetSet %s is used by SubnetConnectionBindingMaps %s", req.String(), bindingsOnNSX[0].GetName())
+			log.Error(err, "Failed to delete SubnetSet CR because it is used by SubnetConnectionBindingMaps, retrying", "SubnetSet", req.NamespacedName)
+			r.StatusUpdater.DeleteFail(req.NamespacedName, nil, err)
+			msgDeleteInUse := fmt.Sprintf("SubnetSet is used by SubnetConnectionBindingMap %s and not able to delete", bindingsOnNSX[0].GetName())
```
lxiaopei, line 110: if multiple bindings found for the subnetSet, should we show all the bindings instead of just the first one?

pkg/controllers/subnet/subnetbinding_handler.go
```diff
@@ -0,0 +1,103 @@
...
+	newBM := objNew.(*v1alpha1.SubnetConnectionBindingMap)
+	oldBM := objOld.(*v1alpha1.SubnetConnectionBindingMap)
+
+	if newBM.Spec.TargetSubnetName == oldBM.Spec.TargetSubnetName {
+		return
+	}
+
+	if newBM.Spec.TargetSubnetName != "" {
+		if err := enqueue(ctx, c, newBM.Namespace, newBM.Spec.TargetSubnetName, q); err != nil {
+			return
```
lxiaopei, line 30: same here, if failed to enqueue new targetSubnet, seems the old targetSubnet does not have chance to enqueue?

pkg/controllers/subnetbinding/subnetbinding_controller_test.go
```diff
@@ -0,0 +1,899 @@
...
+
+func TestCollectGarbage(t *testing.T) {
+	for _, tc := range []struct {
+		name    string
+		patches func(t *testing.T, r *Reconciler) *gomonkey.Patches
+	}{
+		{
+			name: "Failed to list from CRs",
+			patches: func(t *testing.T, r *Reconciler) *gomonkey.Patches {
+				patches := gomonkey.ApplyPrivateMethod(reflect.TypeOf(r), "listBindingMapIDsFromCRs", func(_ *Reconciler, ctx context.Context) (sets.Set[string], error) {
```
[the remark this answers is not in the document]
wenyingd: I don't get your point, this is the mock function to ensure the main logic can be executed as expected. `patches()` is to execute the mock functions.

pkg/controllers/subnetbinding/subnets_handler.go
wenyingd: Let me try

pkg/controllers/subnetset/subnetset_controller.go
wenyingd: I don't think it is necessary to show all.

pkg/controllers/subnet/subnetbinding_handler.go
wenyingd: updated.

pkg/controllers/subnetbinding/subnets_handler.go
wenyingd: update: We can't use `spec.TargetSubnetSetName` as a filter since we didn't make indexes on the manager with these field(s). So I revert back to the previous version. 

If we want to do it, I would suggest to use another change for the improvement.

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -70,8 +73,41 @@ func (r *SubnetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+	if len(bindingCRs) > 0 {
+		if !controllerutil.ContainsFinalizer(subnetCR, servicecommon.SubnetFinalizerName) {
+			controllerutil.AddFinalizer(subnetCR, servicecommon.SubnetFinalizerName)
+			if err := r.Client.Update(ctx, subnetCR); err != nil {
+				log.Error(err, "Failed to add the finalizer", "Subnet", req.NamespacedName)
+				msgFailAddFinalizer := fmt.Sprintf("Failed to add the finalizer on a Subnet for the reference by SubnetConnectionBindingMap %s", bindingCRs[0].Name)
+				r.StatusUpdater.UpdateFail(ctx, subnetCR, err, "Failed to add the finalizer on Subnet used by SubnetConnectionBindingMaps", setSubnetReadyStatusFalse, msgFailAddFinalizer)
+				return ResultRequeue, err
+			}
+		}
```
luolanzone, line 87: I left a comment before, but it's hard to find the tread.
You should skip this part when `!subnetCR.DeletionTimestamp.IsZero()` is true. 
Per my understanding, we can't stop users to delete a subnet CR. As long as the user run kubectl delete subnet $cr, the subnetCR.DeletionTimestamp.IsZero() will be non-zero. And you may still have bindingMap CRs associated to it. Then this part will be executed to add a finalizer to a deleted CR which doesn't make sense.

pkg/controllers/subnetset/subnetset_controller.go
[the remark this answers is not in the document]
luolanzone: Same comment as the one in subnet_controller.

pkg/apis/vpc/v1alpha1/condition_types.go
```diff
@@ -12,6 +12,7 @@ const (
 	GatewayConnectionReady     ConditionType = "GatewayConnectionReady"
 	AutoSnatEnabled            ConditionType = "AutoSnatEnabled"
 	ExternalIPBlocksConfigured ConditionType = "ExternalIPBlocksConfigured"
+	DeleteFailure              ConditionType = "DeletionFailed"
```
yanjunz97, line 15: We added crd doc generation script recently, could you run `make generate-api-docs` to include the change here in doc?

luolanzone (commented):
LGTM overall, leave it to nsx operator owner to approve it.

yanjunz97 (approved):
LGTM

pkg/nsx/services/subnetbinding/subnetbinding.go
```diff
@@ -0,0 +1,235 @@
...
+// transaction, and then delete these resources from local store. Since NSX can accept no more than 5K children Nodes
+// in one HAPI call, paging is used in this function to avoid scale issues.
+func (s *BindingService) deleteSubnetConnectionBindingMaps(bindingMaps []*model.SubnetConnectionBindingMap) error {
+	// Add this check for security purpose. The caller has similar pre-check and returned if no items exist in the bindingMaps.
+	bindingMapsCount := len(bindingMaps)
+	if bindingMapsCount == 0 {
+		log.Info("No existing SubnetConnectionBindingMaps found in the store")
+		return nil
+	}
+	pages := (bindingMapsCount + hAPIPageSize - 1) / hAPIPageSize
```
edwardbadboy, line 179: Can you consider adding a unit test for the paging logic. Just to ensure the pages calculation and loop are correct.

edwardbadboy (commented):
Looks good. One minor comment.

pkg/nsx/services/subnetbinding/subnetbinding.go
edwardbadboy: We can do this in a follow-up patch.

edwardbadboy (approved):

pkg/controllers/common/dependency_watcher.go
lxiaopei: is it possible to change targetSubnet to targetSubnetSet? we can also cover that case right?

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -70,8 +73,41 @@ func (r *SubnetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+	}
+
 	if !subnetCR.DeletionTimestamp.IsZero() {
 		r.StatusUpdater.IncreaseDeleteTotal()
+		bindingsOnNSX := r.getNSXSubnetBindingsBySubnet(string(subnetCR.UID))
+		if len(bindingsOnNSX) > 0 {
+			err := fmt.Errorf("failed to delete Subnet CR %s", req.String())
```
lxiaopei, line 104: do we need to print the whole req? can just use req.NamespacedName?

pkg/controllers/subnet/subnetbinding_handler.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	if newBM.Spec.TargetSubnetName != "" {
+		if err := enqueue(ctx, c, newBM.Namespace, newBM.Spec.TargetSubnetName, q); err != nil {
+			log.Error(err, "Failed to enqueue the new target Subnet after SubnetConnectionBindingMap updates", "Namespace", newBM.Namespace, "Subnet", newBM.Spec.TargetSubnetName, "SubnetConnectionBindingMap", newBM.Name)
+		}
+	}
+	if oldBM.Spec.TargetSubnetName != "" {
+		// Enqueue to ensure the finalizer can be removed from the old target Subnet if it is not used.
+		enqueue(ctx, c, oldBM.Namespace, oldBM.Spec.TargetSubnetName, q)
```
lxiaopei, line 35: we better also log error here if enqueue failed?

pkg/controllers/subnet/subnetbinding_handler.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+}
+
+func enqueueSubnets(ctx context.Context, c client.Client, bindingMap *v1alpha1.SubnetConnectionBindingMap, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	if err := enqueue(ctx, c, bindingMap.Namespace, bindingMap.Spec.SubnetName, q); err != nil {
+		return
```
lxiaopei, line 41: add error log if some errors happen?

pkg/controllers/subnet/subnetbinding_handler.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	if bindingMap.Spec.TargetSubnetName != "" {
+		enqueue(ctx, c, bindingMap.Namespace, bindingMap.Spec.TargetSubnetName, q)
+	}
```
lxiaopei, line 46: ditto

lxiaopei (commented):
some minor comments, LGTM.

pkg/controllers/subnet/subnet_controller.go
wenyingd: The implementation of `req.String` would print "$namespac/$name", it is equal to req.NamespacedName. This is because the only property in Request is types.NamespacedName.

pkg/controllers/subnet/subnetbinding_handler.go
wenyingd: The error logs are provided in function `enqueue`, see https://github.com/vmware-tanzu/nsx-operator/pull/901/files/52bb2d785f0c1dbdb23a74efe0fc5b5d9ec67ed7#diff-51fe24c011631b55a1ac7cef9d82fd5764d219eb2952bd6308ea5defbb199e85R53

pkg/controllers/subnet/subnetbinding_handler.go
wenyingd: same as https://github.com/vmware-tanzu/nsx-operator/pull/901#discussion_r1877069840

lxiaopei (approved):

pkg/nsx/services/subnetbinding/subnetbinding.go
wenyingd: Sure, I will add the UT in the follow-up change.

pkg/controllers/common/dependency_watcher.go
wenyingd: yes, it is supported.

PR #906, opened 2024-11-19
Fix checking realized state error

TaoZou1:
For default project, the realized-state path should be "/policy/api/infra/realized-state/realized-entities?". There is no orgs and projects info in it

Using realizedEntitiesClient under infra to replace that under projects

Test Done:
1. replace the manager in QE env
2. run the test and check if query resources under /orgs/default/projects/default realized successfully

### DIFF
--- a/pkg/nsx/services/realizestate/realize_state.go
+++ b/pkg/nsx/services/realizestate/realize_state.go
@@ -45,15 +45,11 @@ func IsRealizeStateError(err error) bool {
 // backoff defines the maximum retries and the wait interval between two retries.
 func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, intentPath, entityType string) error {
 	// TODO， ask NSX if there were multiple realize states could we check only the latest one?
-	vpcInfo, err := common.ParseVPCResourcePath(intentPath)
-	if err != nil {
-		return err
-	}
 	return retry.OnError(backoff, func(err error) bool {
 		// Won't retry when realized state is `ERROR`.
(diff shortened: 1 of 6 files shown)


pkg/nsx/services/realizestate/realize_state.go
```diff
@@ -45,15 +45,11 @@ func IsRealizeStateError(err error) bool {
...
-	vpcInfo, err := common.ParseVPCResourcePath(intentPath)
-	if err != nil {
-		return err
-	}
 	return retry.OnError(backoff, func(err error) bool {
 		// Won't retry when realized state is `ERROR`.
 		return !IsRealizeStateError(err)
 	}, func() error {
-		results, err := service.NSXClient.RealizedEntitiesClient.List(vpcInfo.OrgID, vpcInfo.ProjectID, intentPath, nil)
+		results, err := service.NSXClient.RealizedEntitiesClient.List(intentPath, nil)
```
timdengyun, line 52: Do you test this RealizedEntitiesClient is working for non-default case.
Previously, we use RealizedEntitiesClient under project/infra to check VPC related resources. 
I think we should handle both default project and non-default case using different client be checking intenPath is 
`/orgs/default/projects/default  or /orgs/default/projects/<project-id>`

TaoZou1: yes, it worked.

timdengyun (approved):
LGTM

yanjunz97 (approved):

PR #913, opened 2024-11-20
Fix checking realized state error

TaoZou1:
For default project, the realized-state path should be "/policy/api/infra/realized-state/realized-entities?". There is no orgs and projects info in it

Test Done:
1. create resources under /orgs/default/projects/default
2. check if resources created


yanjunz97 (approved):

timdengyun (approved):
LGTM

PR #923, opened 2024-11-22
Use read lock on subnet for subnetport/pod creation

yanjunz97:
Previously we add lock on subnet to prevent the race between SubnetPort/Pod controller and SubnetSet GC.
In this PR we replace the mutex lock by read-write lock to improve the parallelism of the SubnetPort creation.

Testing done:
- Created 32 SubnetPort at a time, each SubnetPort reconcile finished in 2-3 seconds

### DIFF
--- a/pkg/nsx/services/common/services.go
+++ b/pkg/nsx/services/common/services.go
@@ -32,6 +32,8 @@ type SubnetServiceProvider interface {
 	GenerateSubnetNSTags(obj client.Object) []model.Tag
 	LockSubnet(path *string)
 	UnlockSubnet(path *string)
+	RLockSubnet(path *string)
+	RUnlockSubnet(path *string)
 }
 
 type SubnetPortServiceProvider interface {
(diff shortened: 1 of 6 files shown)


pkg/nsx/services/common/services.go
```diff
@@ -30,8 +30,8 @@ type SubnetServiceProvider interface {
 	GetSubnetsByIndex(key, value string) []*model.VpcSubnet
 	CreateOrUpdateSubnet(obj client.Object, vpcInfo VPCResourceInfo, tags []model.Tag) (string, error)
 	GenerateSubnetNSTags(obj client.Object) []model.Tag
-	LockSubnet(path *string)
-	UnlockSubnet(path *string)
+	LockSubnet(path *string, read bool)
```
zhengxiexie, line 33: Just a personal opinion, I prefer creating new methods, RWLockSubnet instead of passing in arguments, it is more straightforward.

pkg/nsx/services/common/services.go
```diff
@@ -30,8 +30,8 @@ type SubnetServiceProvider interface {
...
+	UnlockSubnet(path *string, read bool)
```
zhengxiexie, line 34: ditto

pkg/nsx/services/common/services.go
yanjunz97: Updated, thanks!

zhengxiexie (approved):
LGTM

dantingl (approved):

PR #932, opened 2024-11-25
Add AllowOverwrite for some clients

TaoZou1:
Some clients may operate resources created by other PI so 'X-Allow-Overwrite' should be True. Add the header for those clients.
Remove headerconfig.go since it's only used for deprecated rest connector

Test Done:
1. use user=admin to run cleanup
2. check if subnetport has been deleted which created by WCP user


pkg/nsx/cluster.go
```diff
@@ -163,12 +163,23 @@ func (cluster *Cluster) CreateServerUrl(host string, scheme string) string {
 
 // NewRestConnector creates a RestConnector used for SDK client.
 // HeaderConfig is used to use http header for request, it could be ignored if no extra header needed.
```
timdengyun, line 165: Since we don't use Headconfig, please remove this comment.

pkg/nsx/client.go
```diff
@@ -176,13 +179,13 @@ func GetClient(cf *config.NSXOperatorConfig) *Client {
 	staticRouteClient := vpcs.NewStaticRoutesClient(restConnector(cluster))
 	natRulesClient := nat.NewNatRulesClient(restConnector(cluster))
 	vpcGroupClient := vpcs.NewGroupsClient(restConnector(cluster))
-	portClient := subnets.NewPortsClient(restConnector(cluster))
+	portClient := subnets.NewPortsClient(restConnectorAllowOverwrite(cluster))
```
timdengyun, line 182: Do we confirm that NSX only exposes PI checking for subnetPort and ipAddressAllocation resource?

TaoZou1: Due to security reason, we only add the parameter when it's necessary.
we need to overwrite when handling the pre-created vpc.
so only subnetport/ipAddressAllocation needed.

timdengyun (approved):
LGTM

PR #933, opened 2024-11-26
[e2e] Add test cases for pre-created VPC

wenyingd:
This change is to add a new end-to-end test for the pre-created VPC, it includes these steps,
1. Create a VPC/LBS/Attachment on NSX
2. Create a Namespace on vCenter or K8s, which depends on if the setup is a wcp testbed or not. Note, it planned to run the test on wcp testbed, please provide a valid pair of vc-user and vc-password which has the permission to create instance on vCenter
3. Create LoadBalancer typed Service inside the Namespace
4. Create server Pod for the LoadBalancer typed Service
5. Create client Pod in the same Namespace
6. Verified the traffic from the client Pod to the LoadBalancer's external IP
7. Delete LoadBalancer typed Service and verified that the realted NSX resources are removed as expected
8. Delete the Namespace from vCenter or K8s
10. Delete the VPC from NSX

### DIFF
--- a/test/e2e/framework.go
+++ b/test/e2e/framework.go
@@ -15,14 +15,19 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
-
-	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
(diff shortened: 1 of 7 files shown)


zhengxiexie (approved):
lgtm

zhengxiexie (approved):

test/e2e/vclient.go
wenqiq, line 1: The current implementation is OK. It seems that there is a Go library for vSphere API. [vSphere API](https://github.com/vmware/govmomi)
Maybe it can be used.

wenqiq (commented):
LGTM overall. Just a nit comment.

test/e2e/vclient.go
wenyingd: The REST APIs for VPC namespaces in VC is just released, I am not sure it is supported in the library. So I would prefer to use the REST API directly rather than waiting for the support in go-library since the API call is simple.

test/e2e/framework.go
```diff
@@ -431,6 +455,7 @@ func (data *TestData) deploymentWaitForIPsOrNames(timeout time.Duration, namespa
 	}
 	return podIPStrings.List(), podNames, nil
 }
+*/
```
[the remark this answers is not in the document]
zhengxiexie: Leave it alone, let me reorganize it in developing security policy traffic.

wenqiq (approved):

PR #940, opened 2024-12-04
Check all resources realized status when creating subnet

TaoZou1:
Subnet includes child resources which also need checking realized status. Check subnet for all child resources

### DIFF
--- a/pkg/nsx/services/realizestate/realize_state.go
+++ b/pkg/nsx/services/realizestate/realize_state.go
@@ -41,9 +41,10 @@ func IsRealizeStateError(err error) bool {
 	return ok
 }
 
-// CheckRealizeState allows the caller to check realize status of entityType with retries.
-// backoff defines the maximum retries and the wait interval between two retries.
-func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, intentPath, entityType string) error {
+// CheckRealizeState allows the caller to check realize status of intentPath with retries.
+// Backoff defines the maximum retries and the wait interval between two retries.
+// Check all the entities, all entities should be in the REALIZED state to be treated as REALIZED
+func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, intentPath string) error {
 	// TODO， ask NSX if there were multiple realize states could we check only the latest one?
 	return retry.OnError(backoff, func(err error) bool {
 		// Won't retry when realized state is `ERROR`.
@@ -54,12 +55,11 @@ func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, inte
 		if err != nil {
 			return err
 		}
+		entitiesRealized := 0
 		for _, result := range results.Results {
-			if entityType != "" && result.EntityType != nil && *result.EntityType != entityType {
-				continue
-			}
 			if *result.State == model.GenericPolicyRealizedResource_STATE_REALIZED {
-				return nil
+				entitiesRealized++
+				continue
 			}
 			if *result.State == model.GenericPolicyRealizedResource_STATE_ERROR {
 				var errMsg []string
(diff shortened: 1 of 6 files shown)


pkg/nsx/services/subnetport/subnetport.go
```diff
@@ -176,7 +176,7 @@ func (service *SubnetPortService) CheckSubnetPortState(obj interface{}, nsxSubne
 		Jitter:   0,
 		Steps:    6,
 	}
-	if err := realizeService.CheckRealizeState(backoff, *nsxSubnetPort.Path, "RealizedLogicalPort"); err != nil {
+	if err := realizeService.CheckRealizeState(backoff, *nsxSubnetPort.Path, realizestate.RealizedLogicalPort); err != nil {
```
dantingl, line 179: I think we should not specify the entity type for any resource

pkg/nsx/services/vpc/vpc.go
```diff
@@ -854,7 +854,7 @@ func (s *VPCService) createNSXVPC(createdVpc *model.Vpc, nc *common.VPCNetworkCo
 func (s *VPCService) checkVPCRealizationState(createdVpc *model.Vpc, newVpcPath string) error {
 	log.V(2).Info("Check VPC realization state", "VPC", *createdVpc.Id)
 	realizeService := realizestate.InitializeRealizeState(s.Service)
-	if err := realizeService.CheckRealizeState(util.NSXTDefaultRetry, newVpcPath, "RealizedLogicalRouter"); err != nil {
+	if err := realizeService.CheckRealizeState(util.NSXTDefaultRetry, newVpcPath, realizestate.RealizedLogicalRouter); err != nil {
```
dantingl, line 857: same here

TaoZou1: Done

pkg/nsx/services/subnetport/subnetport.go
TaoZou1: Done

pkg/nsx/services/realizestate/realize_state.go
```diff
@@ -54,12 +55,14 @@ func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, inte
 		if err != nil {
 			return err
 		}
+		entitiesRealized := 0
 		for _, result := range results.Results {
 			if entityType != "" && result.EntityType != nil && *result.EntityType != entityType {
 				continue
 			}
```
yanjunz97, line 62: Would you think we can remove this part or even the entityType parameter? As if the entityType is passed, this part will skip the count of realized resource which makes the final entitiesRealized!=len(results.Results). And if there is no use case for entityType, maybe we can just remove this parameter in this function?

TaoZou1: Done

pkg/nsx/services/realizestate/realize_state.go
```diff
@@ -41,9 +41,10 @@ func IsRealizeStateError(err error) bool {
 	return ok
 }
 
-// CheckRealizeState allows the caller to check realize status of entityType with retries.
-// backoff defines the maximum retries and the wait interval between two retries.
-func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, intentPath, entityType string) error {
+// CheckRealizeState allows the caller to check realize status of intentPath with retries.
```
wenyingd, line 44: Is this code more reasonable?
```
// CheckRealizeState allows the caller to check realize status of entityType with retries.
// backoff defines the maximum retries and the wait interval between two retries.
// if entityType == "", check all the entities, all entities should be in the REALIZED state to be tereated as REALIZED
func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, intentPath, entityType string) error {
   // TODO， ask NSX if there were multiple realize states could we check only the latest one?
   return retry.OnError(backoff, func(err error) bool {
      // Won't retry when realized state is `ERROR`.
      return !IsRealizeStateError(err)
   }, func() error {
      results, err := service.NSXClient.RealizedEntitiesClient.List(intentPath, nil)
      err = nsxutil.TransNSXApiError(err)
      if err != nil {
         return err
      }
      for _, result := range results.Results {
         if entityType != "" && result.EntityType != nil && *result.EntityType != entityType {
            continue
         }
            var errMsg []string
            for _, alarm := range result.Alarms {
               if alarm.Message != nil {
                  errMsg = append(errMsg, *alarm.Message)
               }
            }
            return NewRealizeStateError(fmt.Sprintf("%s realized with errors: %s", *result.EntityType, errMsg))
         }
         if *result.State != model.GenericPolicyRealizedResource_STATE_REALIZED {
            return fmt.Errorf("%s not realized", entityType)
         }
      }
      return nil
   })
}
```

yanjunz97 (approved):

PR #948, opened 2024-12-06
Fix the issue that NSX VPC not deleted when the Namespace is terminating

wenyingd:
This change is to fix an issue with the NetworkInfo deletion logic that NSX VPC is not removed if the CR's Namesapce still exists.

The fix is to filter out the Namespace if it is already mark for deletion when a NetworkInfo CR is deleted.

Test Done:
With this fix, after the e2e is completed, the new created VPC by the test cases are deleted in time.

### DIFF
--- a/pkg/controllers/networkinfo/networkinfo_controller.go
+++ b/pkg/controllers/networkinfo/networkinfo_controller.go
@@ -357,6 +357,10 @@ func (r *NetworkInfoReconciler) setupWithManager(mgr ctrl.Manager) error {
 				ipBlocksInfoService: r.IPBlocksInfoService,
 			},
 			builder.WithPredicates(VPCNetworkConfigurationPredicate)).
+		Watches(
+			&corev1.Namespace{},
+			&NamespaceHandler{},
+			builder.WithPredicates(NamespacePredicate)).
 		Complete(r)
 }
 
@@ -478,15 +482,15 @@ func (r *NetworkInfoReconciler) fetchStaleVPCsByNamespace(ctx context.Context, n
 }
 
 func (r *NetworkInfoReconciler) deleteVPCsByName(ctx context.Context, ns string) error {
+	staleVPCs := r.Service.GetVPCsByNamespace(ns)
+	if len(staleVPCs) == 0 {
+		return nil
+	}
+
 	_, idSet, err := r.listNamespaceCRsNameIDSet(ctx)
 	if err != nil {
 		log.Error(err, "Failed to list Kubernetes Namespaces")
 		return fmt.Errorf("failed to list Kubernetes Namespaces while deleting VPCs: %v", err)
-	}
-
-	staleVPCs, err := r.fetchStaleVPCsByNamespace(ctx, ns)
-	if err != nil {
-		return err
 	}
 
 	var vpcToDelete []*model.Vpc
@@ -538,11 +542,10 @@ func (r *NetworkInfoReconciler) deleteVPCs(ctx context.Context, staleVPCs []*mod
 	}
 
 	// Update the VPCNetworkConfiguration Status
-	ncName, err := r.Service.GetNetworkconfigNameFromNS(ctx, ns)
-	if err != nil {
-		return fmt.Errorf("failed to get VPCNetworkConfiguration for Namespace when deleting stale VPCs %s: %w", ns, err)
+	vpcNetConfig := r.Service.GetVPCNetworkConfigByNamespace(ns)
+	if vpcNetConfig != nil {
+		deleteVPCNetworkConfigurationStatus(ctx, r.Client, vpcNetConfig.Name, staleVPCs, r.Service.ListVPC())
 	}
-	deleteVPCNetworkConfigurationStatus(ctx, r.Client, ncName, staleVPCs, r.Service.ListVPC())
 	return nil
 }
 
--- a/pkg/controllers/networkinfo/networkinfo_controller.go
+++ b/pkg/controllers/networkinfo/networkinfo_controller.go
@@ -136,7 +136,7 @@ func (r *NetworkInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	networkInfoCR := &v1alpha1.NetworkInfo{}
 	if err := r.Client.Get(ctx, req.NamespacedName, networkInfoCR); err != nil {
 		if apierrors.IsNotFound(err) {
-			if err := r.deleteVPCsByName(ctx, req.Namespace); err != nil {
+			if err := r.deleteVPCsByNamespace(ctx, req.Namespace); err != nil {
 				r.StatusUpdater.DeleteFail(req.NamespacedName, nil, err)
 				return common.ResultRequeue, err
 			}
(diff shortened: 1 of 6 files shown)


wenqiq:
This change is intended to consider Namespaces in the terminating state when deleting stale VPCs, rather than `ignore` them, correct?  That is, VPCs under a Namespace in the terminating state should be deleted.

zhengxiexie (dismissed):
lgtm

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -408,8 +408,11 @@ func (r *NetworkInfoReconciler) listNamespaceCRsNameIDSet(ctx context.Context) (
 	nsSet := sets.Set[string]{}
 	idSet := sets.Set[string]{}
 	for _, ns := range namespaces.Items {
-		nsSet.Insert(ns.Name)
-		idSet.Insert(string(ns.UID))
+		// Ignore the terminating Namespaces in the list results.
+		if ns.DeletionTimestamp.IsZero() {
```
lxiaopei, line 412: ns.ObjectMeta.DeletionTimestamp.IsZero()?

wenyingd: They should take the same effect. Since only field ObjectMeta has property "DeletionTimeStamp", when we access Namespace.DeletionTimestamp, it actually accesses Namespace.ObjectMeta.DeletionTimestamp.

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -501,20 +486,12 @@ func (r *NetworkInfoReconciler) deleteVPCsByName(ctx context.Context, ns string)
 	return r.deleteVPCs(ctx, vpcToDelete, ns)
 }
 
+// deleteVPCsByID is called when the NetworkInfo CR still exists in the K8s api-server, and the CR's
+// deletionTimestamp is not 0. In this case, it's Namespace must exist, and no other Namespaces configured
+// with the same name. So we can directly use the Namespace's name to find the stale VPCs to delete.
```
lxiaopei, line 491: seems id param is not used, could we remove it?

lxiaopei: Do we need consider that NetworkInfo CR is deleted by accidently but the namespace still there?
In this case we should not delete VPCs?

wenyingd: What do you think if I remove function `deleteVPCsByID`, and call `deleteVPCsByName` instead? Does it resolve your concern?

pkg/controllers/networkinfo/networkinfo_controller.go
timdengyun: Yes. that is the same.

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -459,36 +462,18 @@ func (r *NetworkInfoReconciler) CollectGarbage(ctx context.Context) {
...
-			log.Info("Namespace does not exist while fetching stale VPCs", "Namespace", ns)
-			return nil, nil
-		}
-		return nil, fmt.Errorf("failed to check if Namespace is shared for NS %s: %w", ns, err)
-	}
-	if isShared {
-		log.Info("Shared Namespace, skipping deletion of NSX VPC", "Namespace", ns)
-		return nil, nil
+func (r *NetworkInfoReconciler) deleteVPCsByName(ctx context.Context, ns string) error {
+	staleVPCs := r.Service.GetVPCsByNamespace(ns)
```
timdengyun, line 466: I see the major change compared with previous is that:
We don't check if this namespace is sharedVPC namespace or not.
So, we always assume the requested namespace in deleteVPCsByName which has a networkinfo CR and this CR has been deleted, so we need to delete the VPC only for this VPC in this namespace, is it?

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -538,14 +515,22 @@ func (r *NetworkInfoReconciler) deleteVPCs(ctx context.Context, staleVPCs []*mod
...
 	}
-	deleteVPCNetworkConfigurationStatus(ctx, r.Client, ncName, staleVPCs, r.Service.ListVPC())
 	return nil
 }
 
+func (r *NetworkInfoReconciler) listVPCsByNetworkConfigName(ncName string) []*model.Vpc {
+	namespacesUsingNC := r.Service.GetNamespacesByNetworkconfigName(ncName)
+	aliveVPCs := make([]*model.Vpc, 0)
+	for _, namespace := range namespacesUsingNC {
+		aliveVPCs = append(aliveVPCs, r.Service.GetVPCsByNamespace(namespace)...)
```
lxiaopei, line 529: if ns is deleted but its VPC is not deleted from VPC store yet, GetVPCsByNamespace will also return the "stale" vpc in aliveVPCs list?

pkg/controllers/networkinfo/networkinfo_controller.go
lxiaopei: ok, got it.

pkg/controllers/networkinfo/networkinfo_controller.go
lxiaopei: yes, deleteVPCsByName should be fine.

pkg/controllers/networkinfo/networkinfo_controller.go
wenyingd: This function is called after the "stale VPC" is removed from memory store (including deleted on NSX)

pkg/controllers/networkinfo/networkinfo_utils.go
```diff
@@ -191,35 +191,33 @@ func getGatewayConnectionStatus(ctx context.Context, nc *v1alpha1.VPCNetworkConf
...
-			newVPCInfos = append(newVPCInfos, vpc)
+
+	if getAliveVPCsFn != nil {
+		aliveVPCs := sets.New[string]()
+		for _, vpc := range getAliveVPCsFn(ncName) {
+			aliveVPCs.Insert(*vpc.DisplayName)
 		}
+		var newVPCInfos []v1alpha1.VPCInfo
+		for _, vpcInfo := range nc.Status.VPCs {
+			if aliveVPCs.Has(vpcInfo.Name) {
```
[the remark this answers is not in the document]
wenyingd: Here I expect to be consistent with the previous output of the status. The purpose of this function is to remove the stale logic from the VPCNetworkConfiguration.status.vpcs, but it doesn't desire to reload the latest status of the live VPC.

Shall we enhance it in another change if the original logic is incorrect?

pkg/controllers/networkinfo/networkinfo_controller.go
wenyingd: No, the change is to resolve the bug that, when a NetworkInfo CR is deleted, its Namespace may not remove from apiserver. However, as long as the Namespace is desired to delete, we should delete the corresponding NSX VPC. The original code doesn't correctly check the Namespace is desired to delete or not, instead, it thinks the NS as valid as long as it exists in the apiserver. So we actually do not delete the NSX VPC even if the user expect to delete NS, which caused many stale resources left on NSX before next round of GC.

timdengyun (approved):
LGTM, thanks

pkg/controllers/networkinfo/networkinfo_utils.go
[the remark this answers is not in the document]
lxiaopei: OK，I think the original logic is not correct. We need a following patch.

lxiaopei (approved):
LGTM

pkg/controllers/networkinfo/networkinfo_utils.go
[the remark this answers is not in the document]
wenyingd: I created an issue https://github.com/vmware-tanzu/nsx-operator/issues/968 to track it.

PR #950, opened 2024-12-10
Remove entityType from CheckRealizeState

TaoZou1:
Remove entityType from CheckRealizeState. Since we're not sure if the realized 
items for intentPath will be changed or not

### DIFF
--- a/pkg/nsx/services/realizestate/realize_state.go
+++ b/pkg/nsx/services/realizestate/realize_state.go
@@ -46,10 +46,10 @@ func IsRealizeStateError(err error) bool {
 	return ok
 }
 
-// CheckRealizeState allows the caller to check realize status of entityType with retries.
-// backoff defines the maximum retries and the wait interval between two retries.
-// if entityType == "", check all the entities, all entities should be in the REALIZED state to be tereated as REALIZED
-func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, intentPath, entityType string) error {
+// CheckRealizeState allows the caller to check realize status of intentPath with retries.
+// Backoff defines the maximum retries and the wait interval between two retries.
+// Check all the entities, all entities should be in the REALIZED state to be treated as REALIZED
+func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, intentPath string) error {
 	// TODO， ask NSX if there were multiple realize states could we check only the latest one?
 	return retry.OnError(backoff, func(err error) bool {
 		// Won't retry when realized state is `ERROR`.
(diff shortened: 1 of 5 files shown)


pkg/nsx/services/realizestate/realize_state.go
```diff
@@ -46,10 +46,10 @@ func IsRealizeStateError(err error) bool {
 	return ok
 }
 
-// CheckRealizeState allows the caller to check realize status of entityType with retries.
-// backoff defines the maximum retries and the wait interval between two retries.
-// if entityType == "", check all the entities, all entities should be in the REALIZED state to be tereated as REALIZED
-func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, intentPath, entityType string) error {
+// CheckRealizeState allows the caller to check realize status of intentPath with retries.
+// Backoff defines the maximum retries and the wait interval between two retries.
+// Check all the entities, all entities should be in the REALIZED state to be tereated as REALIZED
```
yanjunz97, line 51:
```suggestion
// Check all the entities, all entities should be in the REALIZED state to be treated as REALIZED
```

yanjunz97 (commented):
Overall LGTM, just found one typo

pkg/nsx/services/realizestate/realize_state.go
```diff
@@ -62,9 +62,6 @@ func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, inte
 		}
 		entitiesRealized := 0
 		for _, result := range results.Results {
-			if entityType != "" && result.EntityType != nil && *result.EntityType != entityType {
-				continue
-			}
 			if *result.State == model.GenericPolicyRealizedResource_STATE_REALIZED {
```
wenyingd, line 65: I prefer to do it like this with function. No need to explicitly compare the count of model.GenericPolicyRealizedResource_STATE_REALIZED resources. Instead, I assume the intent of this function is all NSX returned resources should be with `model.GenericPolicyRealizedResource_STATE_REALIZED` , then we can directly return error if any is not in this state. 

```
func() error {
		results, err := service.NSXClient.RealizedEntitiesClient.List(intentPath, nil)
		err = nsxutil.TransNSXApiError(err)
		if err != nil {
			return err
		}
		for _, result := range results.Results {
				var errMsg []string
				for _, alarm := range result.Alarms {
					if alarm.Message != nil {
						errMsg = append(errMsg, *alarm.Message)
					}
				}
				return NewRealizeStateError(fmt.Sprintf("%s realized with errors: %s", *result.EntityType, errMsg))
			}
			if *result.State != model.GenericPolicyRealizedResource_STATE_REALIZED {
				return fmt.Errorf("%s not realized", entityType)
			}
		}
		return nil
	}
```

TaoZou1: It seems there is a problem.
If you have 3 items. One is unrealized, one is error, one is realized.
Your logic depends on the return order.

pkg/nsx/services/realizestate/realize_state.go
TaoZou1: done

wenyingd (approved):

yanjunz97 (approved):

PR #951, opened 2024-12-10
Add to store after all entities realized

TaoZou1:
Creating VPC need to check realized entities including vpc/lbs/ attachment.
Only add those entities to store after the realization successful

### DIFF
--- a/pkg/nsx/services/vpc/vpc.go
+++ b/pkg/nsx/services/vpc/vpc.go
@@ -814,17 +814,24 @@ func (s *VPCService) CreateOrUpdateVPC(ctx context.Context, obj *v1alpha1.Networ
 		return nil, err
 	}
 
-	if err := s.VpcStore.Add(&newVpc); err != nil {
-		return nil, err
-	}
-
 	// Check LBS realization
-	if err := s.checkLBSRealization(createdLBS, createdVpc, nc, *newVpc.Path); err != nil {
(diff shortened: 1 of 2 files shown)


dantingl (dismissed):

pkg/nsx/services/vpc/vpc.go
```diff
@@ -871,16 +878,15 @@ func (s *VPCService) checkVPCRealizationState(createdVpc *model.Vpc, newVpcPath
...
+func (s *VPCService) checkLBSRealization(createdLBS *model.LBService, createdVpc *model.Vpc, nc *common.VPCNetworkConfigInfo, newVpcPath string) (*model.LBService, error) {
 	if createdLBS == nil {
-		return nil
+		return nil, nil
 	}
 	newLBS, err := s.NSXClient.VPCLBSClient.Get(nc.Org, nc.NSXProject, *createdVpc.Id, *createdLBS.Id)
 	if err != nil || newLBS.ConnectivityPath == nil {
 		log.Error(err, "Failed to read LBS object after creating or updating", "LBS", createdLBS.Id)
-		return err
+		return &newLBS, err
```
yanjunz97, line 888: &newLBS -> nil? Otherwise there will be error when err is not nil and newLBS is nil?

yanjunz97 (commented):
Please improve the UT if possible as the patch coverage is only 10%

PR #954, opened 2024-12-12
Register all networkconfigurations before controller starting

TaoZou1:
Previous networkconfiguration register depends on networkinfo events. But other controller will depends on the map. Register all networkconfigurations before controller starting

Test Done:
1. run in the env to check if all the networkconfigurations have been registered

### DIFF
--- a/pkg/controllers/networkinfo/networkinfo_controller.go
+++ b/pkg/controllers/networkinfo/networkinfo_controller.go
@@ -362,7 +362,11 @@ func (r *NetworkInfoReconciler) setupWithManager(mgr ctrl.Manager) error {
 
 // Start setup manager and launch GC
 func (r *NetworkInfoReconciler) Start(mgr ctrl.Manager) error {
-	err := r.setupWithManager(mgr)
+	err := r.RegisterAllNetworkInfo(context.Background())
+	if err != nil {
+		return err
+	}
+	err = r.setupWithManager(mgr)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -373,6 +377,27 @@ func (r *NetworkInfoReconciler) Start(mgr ctrl.Manager) error {
 	return nil
 }
 
+func (r *NetworkInfoReconciler) RegisterAllNetworkInfo(ctx context.Context) error {
+	log.Info("Register all NetworkConfigurations")
+	var networkConfigList v1alpha1.VPCNetworkConfigurationList
+	if err := r.Client.List(ctx, &networkConfigList); err != nil {
+		log.Error(err, "Failed to list NetworkInfo")
```
wenyingd, line 384:
```suggestion
		log.Error(err, "Failed to list VPCNetworkConfiguration")
```

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -373,6 +377,27 @@ func (r *NetworkInfoReconciler) Start(mgr ctrl.Manager) error {
...
+		return err
+	}
+	for _, nconfig := range networkConfigList.Items {
+		vname := nconfig.GetName()
+		ninfo, err := buildNetworkConfigInfo(nconfig)
+		if err != nil {
+			log.Error(err, "build network config failed")
```
wenyingd, line 391: Would you print the failed VPCNetworkConfiguration name in the error log?

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -373,6 +377,27 @@ func (r *NetworkInfoReconciler) Start(mgr ctrl.Manager) error {
...
+			return err
+		}
+		log.Info("Create network config and update to map", "NetworkConfigInfo", ninfo)
```
wenyingd, line 394: Better to add this log after the job is done (line 395)

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -362,7 +362,11 @@ func (r *NetworkInfoReconciler) setupWithManager(mgr ctrl.Manager) error {
 
 // Start setup manager and launch GC
 func (r *NetworkInfoReconciler) Start(mgr ctrl.Manager) error {
-	err := r.setupWithManager(mgr)
+	err := r.RegisterAllNetworkInfo(context.Background())
```
wenyingd, line 365: This change may not work for the standby Pod with the HA mode. It calls `r.RegisterAllNetworkInfo` before r.setupWithManager(mgr), it means loading all VPCNetworkConfigurations before manager.Start. It may have issues in HA mode on the standby Pod. For example, at time1, the pod loads all VPC Network configurations from kube apiserver, then it stands by and waits for itself to be the master. During this time, more VPC Network Configurations are created, but they are not existing in the cache. After the Pod is elected as the master, the Pod actually does not have the new resources, it still needs the Reconcile to sync the resources?

seanpang-vmware: the k8s client in manager struct is not available before setupWithManager, will cause nil memory error

pkg/controllers/networkinfo/networkinfo_controller.go
TaoZou1: Done

pkg/controllers/networkinfo/networkinfo_controller.go
TaoZou1: slave will be blocked before start controller.

wenyingd (approved):
LGTM

seanpang-vmware (approved):

PR #957, opened 2024-12-12
[e2e] Use Subnet UID to get NSX resources in the test

wenyingd:
The flaky issue may happen if the stale VpcSubnet resources exist on NSX in anohter VPC configured with the same Namespace name.
The fix is using the tested Subnet CR's UID to get NSX resources. This is because the K8s can ensure the CR's UID is unique.

Fix: #956

PR #959, opened 2024-12-12
[Chery-Pick][e2e] Use Subnet UID to get NSX resources in the test (#957)

wenyingd:
The flaky issue may happen if stale VpcSubnet resources exist on NSX in anohter VPC configured with the same Namespace name.
The fix is using the tested Subnet CR's UID to get NSX resources. This is because the K8s can ensure the CR's UID is unique.


timdengyun (approved):
LGTM

PR #960, opened 2024-12-13
Register all networkconfigurations before controller starting

TaoZou1:
Previous networkconfiguration register depends on networkinfo events. But other controller will depends on the map. Register all networkconfigurations before controller starting

Test Done:
1. run in the env to check if all the networkconfigurations have been registered


wenyingd (approved):
LGTM

seanpang-vmware (approved):

PR #969, opened 2024-12-13
[Cherry-Pick] Fix the issue that NSX VPC not deleted when the Namespace is terminating (#948)

wenyingd:
This change is to fix an issue with the NetworkInfo deletion logic that NSX VPC is not removed if the CR's Namesapce still exists.

The fix is to add a wather on Namespace deletion event and ensure the NSX VPC is deleted when the K8s Namespace is deleted.


timdengyun (approved):
LGTM

PR #970, opened 2024-12-17
[Cherry-Pick][e2e] Add test cases for pre-created VPC (#933)

wenyingd:
This change is to add a new end-to-end test for the pre-created VPC, it includes these steps,
1. Create a VPC/LBS/Attachment on NSX
2. Create a Namespace on vCenter or K8s, which depends on if the setup is a wcp testbed or not. Note, it planned to run the test on wcp testbed, please provide a valid pair of vc-user and vc-password which has the permission to create instance on vCenter
3. Create LoadBalancer typed Service inside the Namespace
4. Create server Pod for the LoadBalancer typed Service
5. Create client Pod in the same Namespace
6. Verified the traffic from the client Pod to the LoadBalancer's external IP
7. Delete LoadBalancer typed Service and verified that the realted NSX resources are removed as expected
8. Delete the Namespace from vCenter or K8s
9. Delete the VPC from NSX

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -65,13 +65,16 @@ require (
 	github.com/google/gnostic-models v0.6.8 // indirect
 	github.com/google/go-cmp v0.6.0 // indirect
 	github.com/google/gofuzz v1.2.0 // indirect
+	github.com/gorilla/websocket v1.5.0 // indirect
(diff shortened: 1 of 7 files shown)


wenqiq (approved):
lgtm

zhengxiexie (approved):
lgtm

PR #971, opened 2024-12-18
Reopen e2e security policy traffic

zhengxiexie:


test/e2e/nsx_security_policy_test.go
```diff
@@ -84,12 +82,10 @@ func testSecurityPolicyBasicTraffic(t *testing.T) {
...
 	assert.NoError(t, testData.waitForResourceExistOrNot(ns, common.ResourceTypeRule, ruleName1, true))
 
-	// Temporarily disable traffic check
-	/*
-		// Ping from pod
-		err = testData.runPingCommandFromPod(ns, busybox, iPs, 4)
-		assertNotNil(t, err, "Error when running ping command from test Pod %s", busybox)
-	*/
+	// Test traffic from client Pod to server Pod
+	trafficErr = assureTraffic(ns, clientPodName, containerName, iPs.ipv4.String(), podPort, 1*time.Second, 5*time.Second)
```
wenyingd, line 86: Maybe you can define constants or global variables for traffic verification interval and timeout?

test/e2e/nsx_security_policy_test.go
```diff
@@ -322,3 +313,24 @@ func assureSecurityPolicyReady(t *testing.T, ns, spName string) {
 	})
 	require.NoError(t, err)
 }
+
+func assureTraffic(ns, podname, containername, ip string, port int32, interval, timeout time.Duration) error {
```
wenyingd, line 317: It looks most code is copied from precreated_vpc_test.go? If you expect to share them in multiple files, better to place such logic into a function in framework.go, and call the function in each test file?

test/e2e/nsx_security_policy_test.go
```diff
@@ -186,36 +180,33 @@ func testSecurityPolicyMatchExpression(t *testing.T) {
...
-		t.Logf("Pods are %v", iPs)
-		assert.NoError(t, err, "Error when waiting for IP for Pod %s", podA)
-
-		// Ping from pod
-		err = testData.runPingCommandFromPod(ns, clientA, iPs, 4)
-		assert.NoError(t, err, "Error when running ping command from Pod %s", clientA)
-		err = testData.runPingCommandFromPod(ns, clientB, iPs, 4)
-		assert.NoError(t, err, "Error when running ping command from Pod %s", clientB)
-	*/
+	clientA := "client-a"
```
wenyingd, line 191: Personally I don't prefer to mixed use manifests and kube-client in one case, which is not friendly for both test design and issue troubleshooting. For example, this style requires that the Pod name we used in the test.go must be consistent with manifests, if the dev doesn't know this precondition, a change on the Pod name in test.go may block the cases, and it is not easier to trouble shoot.

test/e2e/framework.go
```diff
@@ -293,6 +295,78 @@ func (data *TestData) createNamespace(namespace string, mutators ...func(ns *cor
...
+				DefaultSubnetSize: 16,
+			},
+		},
+	}
+	dataJson, err := json.Marshal(vcNamespace)
+	if err != nil {
+		log.Error(err, "Unable convert vcNamespace object to json bytes", "namespace", namespace)
+		return fmt.Errorf("unable convert vcNamespace object to json bytes: %v", err)
+	}
+	urlPath := "/api/vcenter/namespaces/instances/v2"
```
wenyingd, line 321: We can define a constants for this url path?

test/e2e/nsx_security_policy_test.go
```diff
@@ -48,31 +50,27 @@ func testSecurityPolicyBasicTraffic(t *testing.T) {
 	ruleName0 := "all_ingress_isolation"
 	ruleName1 := "all_egress_isolation"
 	var err error
-	setupTest(t, ns)
-	defer teardownTest(t, ns, defaultTimeout)
+	testData.createVCNamespace(ns)
```
wenyingd, line 53: You should check the returned error of this function, and do not continue the following up functions if an error returns.

test/e2e/framework.go
```diff
@@ -293,6 +295,78 @@ func (data *TestData) createNamespace(namespace string, mutators ...func(ns *cor
...
+	defer func() {
+		testData.vcClient.closeSession()
+	}()
+
+	_ = testData.vcClient.deleteNamespace(namespace)
+	// wait for the namespace on k8s terminating
+	err := wait.PollUntilContextTimeout(context.TODO(), 1*time.Second, defaultTimeout, false, func(ctx context.Context) (done bool, err error) {
+		ns, err := data.clientset.CoreV1().Namespaces().Get(context.TODO(), namespace, metav1.GetOptions{})
+		if err != nil {
+			log.Error(err, "Check namespace existence", "namespace", namespace)
```
wenyingd, line 358: Here we should check NotFoundError and return true in the case.

test/e2e/framework.go
```diff
@@ -293,6 +295,78 @@ func (data *TestData) createNamespace(namespace string, mutators ...func(ns *cor
...
+}
+
+// deleteVCNamespace deletes the provided VC namespace and waits for deletion to actually complete.
+func (data *TestData) deleteVCNamespace(namespace string) error {
+	_ = testData.vcClient.startSession()
+	defer func() {
+		testData.vcClient.closeSession()
+	}()
+
+	_ = testData.vcClient.deleteNamespace(namespace)
```
wenyingd, line 353: This function is supposed to be called only if the VC namespace is successfully created?

test/e2e/manifest/testSecurityPolicy/rule-in-rule-applied-to-with-src.yaml
```diff
@@ -35,7 +35,7 @@ spec:
       hostname: web-deployment
       containers:
         - name: web
-          image: "harbor-repo.vmware.com/dockerhub-proxy-cache/humanux/http_https_echo:latest"
+          image: "netfvt-docker-local.packages.vcfd.broadcom.net/humanux/http_https_echo:latest"
```
wenyingd, line 38: This is another reason I don't like to use manifest in e2e test, we need to modify many files any time the image is updated

test/e2e/nsx_security_policy_test.go
```diff
@@ -186,36 +180,33 @@ func testSecurityPolicyMatchExpression(t *testing.T) {
 	ruleName := "expression-policy-1-rule"
 
 	var err error
-	setupTest(t, ns)
-	defer teardownTest(t, ns, defaultTimeout)
+	testData.createVCNamespace(ns)
```
wenyingd, line 183: Question: is it possible we deploy a CI testbed not using WCP (VC is not used)? If yes, we may not create namespace on VC, but directly on K8s?

test/e2e/framework.go
```diff
@@ -293,6 +295,78 @@ func (data *TestData) createNamespace(namespace string, mutators ...func(ns *cor
 	return nil
 }
 
+// createVCNamespace creates a VC namespace with the provided namespace.
+func (data *TestData) createVCNamespace(namespace string) error {
+	_ = testData.vcClient.startSession()
```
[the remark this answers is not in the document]
zhengxiexie: ok

test/e2e/framework.go
```diff
@@ -293,6 +295,78 @@ func (data *TestData) createNamespace(namespace string, mutators ...func(ns *cor
...
+			return true, nil
+		}
+		return false, nil
+	})
+	return err
+}
+
+// deleteVCNamespace deletes the provided VC namespace and waits for deletion to actually complete.
+func (data *TestData) deleteVCNamespace(namespace string) error {
+	_ = testData.vcClient.startSession()
```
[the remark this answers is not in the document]
zhengxiexie: ok

test/e2e/nsx_security_policy_test.go
zhengxiexie: ok

test/e2e/nsx_security_policy_test.go
zhengxiexie: ok

test/e2e/framework.go
zhengxiexie: ok

test/e2e/nsx_security_policy_test.go
zhengxiexie: ok

test/e2e/nsx_security_policy_test.go
zhengxiexie: Yes, the previous k8s testbed is what you said, running for a while.

test/e2e/nsx_security_policy_test.go
zhengxiexie: ok, we will have a discussion before long.

test/e2e/manifest/testSecurityPolicy/rule-in-rule-applied-to-with-src.yaml
zhengxiexie: ok

test/e2e/framework.go
```diff
@@ -293,6 +296,84 @@ func (data *TestData) createNamespace(namespace string, mutators ...func(ns *cor
...
+	if err != nil {
+		log.Error(err, "Failed to prepare http request with vcNamespace data", "namespace", namespace)
+		return fmt.Errorf("failed to parepare http request with vcNamespace data: %v", err)
+	}
+	if _, err = data.vcClient.handleRequest(request, nil); err != nil {
+		log.Error(err, "Failed to create VC namespace", "namespace", namespace)
+		return err
+	}
+	// wait for the namespace on k8s running
+	err = wait.PollUntilContextTimeout(context.TODO(), 1*time.Second, defaultTimeout, false, func(ctx context.Context) (done bool, err error) {
```
wenyingd, line 335: Would you delete the VC Namespace if the K8s namespace is not ready in time.

test/e2e/framework.go
```diff
@@ -293,6 +296,84 @@ func (data *TestData) createNamespace(namespace string, mutators ...func(ns *cor
...
+		ns, err := data.clientset.CoreV1().Namespaces().Get(context.TODO(), namespace, metav1.GetOptions{})
+		if err != nil {
+			log.Error(err, "Check namespace existence", "namespace", namespace)
+			return false, err
+		}
+		if ns.Status.Phase == corev1.NamespaceActive {
```
wenyingd, line 341: A K8s Namespace's phase is already Active after it is created and before it is deleted. I think you check the Namespace's condition type "NamespaceNetworkReady" with status "True" which is introduced in nsx-operator to identify if the VPC is realized?

test/e2e/framework.go
```diff
@@ -815,3 +803,24 @@ func (data *TestData) deleteService(nsName string, svcName string) error {
 func (data *TestData) useWCPSetup() bool {
 	return data.vcClient != nil
 }
+
+func checkTrafficByCurl(ns, podname, containername, ip string, port int32, interval, timeout time.Duration) error {
```
wenyingd, line 807: Does it work if we call this function precreated_vpc_test.go when validating the LB traffic?

test/e2e/framework.go
zhengxiexie: yes, in each case, it applies to pattern

	err := testData.createVCNamespace(ns)
	if err != nil {
		t.Fatalf("Failed to create VC namespace: %v", err)
	}
	defer func() {
		err := testData.deleteVCNamespace(ns)
		if err != nil {
			t.Fatalf("Failed to delete VC namespace: %v", err)
		}
	}(

test/e2e/framework.go
zhengxiexie: ok

test/e2e/framework.go
zhengxiexie: Yes, you can refactor it after it checked in.

wenqiq:
The e2e test failed, please check.

test/e2e/nsx_security_policy_test.go
```diff
@@ -43,36 +50,41 @@ func TestSecurityPolicy(t *testing.T) {
...
-		assert.NoError(t, err, "Error when running ping command from test Pod %s", busybox)
-	*/
+	err := testData.createVCNamespace(ns)
+	if err != nil {
+		t.Fatalf("Failed to create VC namespace: %v", err)
+	}
+	defer func() {
+		err := testData.deleteVCNamespace(ns)
+		if err != nil {
+			t.Fatalf("Failed to delete VC namespace: %v", err)
```
wenqiq, line 66: I think if cleanup fails during defer, it should logs the error instead of halting the test (t.Fatalf replaced with t.Logf).

test/e2e/nsx_security_policy_test.go
```diff
@@ -185,37 +193,42 @@ func testSecurityPolicyMatchExpression(t *testing.T) {
...
-	defer teardownTest(t, ns, defaultTimeout)
+	err := testData.createVCNamespace(ns)
+	if err != nil {
+		testData.deleteVCNamespace(ns)
+		t.Fatalf("Failed to create VC namespace: %v", err)
+	}
+	defer func() {
+		err := testData.deleteVCNamespace(ns)
+		if err != nil {
+			t.Fatalf("Failed to delete VC namespace: %v", err)
```
wenqiq, line 204: ditto

test/e2e/nsx_security_policy_test.go
```diff
@@ -185,37 +193,42 @@ func testSecurityPolicyMatchExpression(t *testing.T) {
 	securityPolicyName := "expression-policy-1"
 	ruleName := "expression-policy-1-rule"
 
-	var err error
-	setupTest(t, ns)
-	defer teardownTest(t, ns, defaultTimeout)
+	err := testData.createVCNamespace(ns)
+	if err != nil {
+		testData.deleteVCNamespace(ns)
```
wenqiq, line 198: The testData.deleteVCNamespace(ns) is called both in the error-handling block and the defer statement?

test/e2e/nsx_security_policy_test.go
zhengxiexie: The test should only pass if the namespace is deleted successfully; otherwise, it could impact subsequent runs.
What do you think?

test/e2e/nsx_security_policy_test.go
zhengxiexie: slip of the hand，fixed

wenyingd (approved):
LGTM

wenqiq (approved):
LGTM overall. Regarding the use of YAML or code for creating resources in e2e testing, IMO, using YAML makes debugging more convenient, while using code allows errors to be detected promptly when CRD fields change. Of course, this is not a strict requirement.

PR #975, opened 2024-12-23
Add StoragePolicy and ContentLibrary when creating VC Namespace

zhengxiexie:
This PR is prepared for VM creating in e2e test.
If we want create VM in a namespace, it should be created by VC API, the namespace should have Storage, ContentLibrary, and VmClass attached.

### DIFF
--- a/test/e2e/framework.go
+++ b/test/e2e/framework.go
@@ -307,17 +307,40 @@ func (data *TestData) createVCNamespace(namespace string) error {
 	}()
 
 	svID, _ := data.vcClient.getSupervisorID()
+	storagePolicyID, _ := data.vcClient.getStoragePolicyID()
+	log.V(1).Info("Get storage policy", "storagePolicyID", storagePolicyID)
+	contentLibraryID, _ := data.vcClient.getContentLibraryID()
+	log.V(1).Info("Get content library", "contentLibraryID", contentLibraryID)
 	vcNamespace := &VCNamespaceCreateSpec{
 		Supervisor: svID,
 		Namespace:  namespace,
+		StorageSpecs: []InstancesStorageSpec{
+			{
+				Policy: storagePolicyID,
+			},
+		},
+		ContentLibraries: []InstancesContentLibrarySpec{
+			{
+				ContentLibrary: contentLibraryID,
+			},
+		},
 		NetworkSpec: InstancesNetworkConfigInfo{
 			NetworkProvider: "NSX_VPC",
 			VpcNetwork: InstancesVpcNetworkInfo{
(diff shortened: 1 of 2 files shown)


test/e2e/vclient.go
```diff
@@ -143,13 +163,55 @@ func (c *vcClient) getSupervisorID() (string, error) {
...
+	for _, po := range response {
+		log.Info("Checking storage policy", "policy", po.Name, "description", po.Description)
+		if strings.Contains(po.Name, "global") || strings.Contains(po.Name, "local") {
+			return po.Name, nil
+		}
+	}
+	return "", fmt.Errorf("no valid storage policy found on vCenter")
+}
+
+func (c *vcClient) getContentLibraryID() (string, error) {
```
wenyingd, line 195: It looks you are trying to get the first content library? How do you ensure this can work for the tests? Could we filter with the test preferred resource to avoid unexpected failure?

BTW, could you add comments in the function?

wenyingd (commented):
Could you add more information in the commit message about why we need these VC resources in e2e test?

And is VM class needed in the VM creation? If yes, do we need to get this configuration?

test/e2e/vclient.go
zhengxiexie: Getting the first content library should be OK.

test/e2e/framework.go
```diff
@@ -307,17 +307,40 @@ func (data *TestData) createVCNamespace(namespace string) error {
 	}()
 
 	svID, _ := data.vcClient.getSupervisorID()
+	storagePolicyID, _ := data.vcClient.getStoragePolicyID()
+	log.Info("", "storagePolicyID", storagePolicyID)
```
wenyingd, line 311: It is strange to use an empty string in the log sentence?

zhengxiexie: ok

wenyingd (approved):
The change looks good to me.

It is better to add more details in the commit message to help record what changes this PR has introduced and what problem it is to resolve.

seanpang-vmware (approved):

test/e2e/framework.go
```diff
@@ -307,17 +307,40 @@ func (data *TestData) createVCNamespace(namespace string) error {
 	}()
 
 	svID, _ := data.vcClient.getSupervisorID()
+	storagePolicyID, _ := data.vcClient.getStoragePolicyID()
```
dantingl, line 310: Will content library be heavy load or have performance for Namespace, can we add a parameter in `createVCNamespace` to add storage policy and content library only if they are needed in specific test cases?

PR #976, opened 2024-12-24
Improve SubnetDHCPConfig validation rule

yanjunz97:
Subnet controller will update the SubnetSpec as the following when
either accessMode or ipv4SubnetSize is not provided.
```
spec:
  accessMode: Public
  ipv4SubnetSize: 64
  subnetDHCPConfig:{}
```
Thus we need another validation rule on subnetDHCPConfig to prevent
subnetDHCPConfig being updated from no mode specified to
DHCPServer/DHCPRelay mode.


dantingl (approved):

lxiaopei (approved):
LGTM

PR #980, opened 2024-12-30
Fix bug which should delete node from store when node is not found

zhengxiexie:
```
8153:2024-12-24 13:20:56.232	INFO	node/node_controller.go:40	reconciling node	{"node": {"name":"192-163-127-8"}}
8154:2024-12-24 13:20:56.234	INFO	node/node_controller.go:46	node not found	{"req": {"name":"192-163-127-8"}}
8596:2024-12-24 13:48:43.606	INFO	node/node_controller.go:40	reconciling node	{"node": {"name":"192-163-127-8"}}
8597:2024-12-24 13:48:43.607	INFO	node/node.go:78	node alreay cached	{"node.Fqdn": "192-163-127-8", "node.UniqueId": "78dac851-95a4-4b78-84f0-acd69c7244e1"}
9179:2024-12-24 13:58:41.968	ERROR	pod/pod_controller.go:86	failed to get node ID for pod	{"pod.Name": {"name":"test-0","namespace":"ns-2"}, "pod.UID": "a84b008a-230c-4d5a-a088-0e1c5d230962", "node": "192-163-127-8", "error": "multiple node IDs found for node 192-163-127-8: [78dac851-95a4-4b78-84f0-acd69c7244e1 d90046e7-4992-4739-822e-663c95a9dfaf]"}
```
From the log, we can see when node is deleted, there is node not found error, and there are still two node ids in store.
The phase node.ObjectMeta.DeletionTimestamp being not zero is intermittent, and it is more possible the node is not found, we should delete the node from store.

### DIFF
--- a/pkg/controllers/node/node_controller.go
+++ b/pkg/controllers/node/node_controller.go
@@ -45,6 +45,10 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		if errors.IsNotFound(err) {
 			log.Info("node not found", "req", req.NamespacedName)
 			deleted = true
+			if err := r.Service.SyncNodeStore(req.NamespacedName.Name, deleted); err != nil {
+				log.Error(err, "failed to sync node store", "req", req.NamespacedName)
+				return common.ResultNormal, err
+			}
 		} else {
 			log.Error(err, "unable to fetch node", "req", req.NamespacedName)
 		}
@@ -53,7 +57,7 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	if common.NodeIsMaster(node) {
 		// For WCP supervisor cluster, the master node isn't a transport node.
 		log.Info("skipping handling master node", "node", req.NamespacedName)
-		return ctrl.Result{}, nil
+		return common.ResultNormal, nil
 	}
 	if !node.ObjectMeta.DeletionTimestamp.IsZero() {
 		log.Info("node is being deleted", "node", req.NamespacedName)
@@ -62,9 +66,9 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 
 	if err := r.Service.SyncNodeStore(node.Name, deleted); err != nil {
 		log.Error(err, "failed to sync node store", "req", req.NamespacedName)
-		return ctrl.Result{}, err
+		return common.ResultNormal, err
 	}
-	return ctrl.Result{}, nil
+	return common.ResultNormal, nil
 }
 
 // SetupWithManager sets up the controller with the Manager.
--- a/pkg/controllers/node/node_controller_test.go
+++ b/pkg/controllers/node/node_controller_test.go
@@ -2,6 +2,7 @@ package node
 
 import (
 	"context"
+	"errors"
 	"testing"
 
 	"github.com/agiledragon/gomonkey/v2"
@@ -16,6 +17,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/event"
 
 	"github.com/vmware-tanzu/nsx-operator/pkg/config"
+	"github.com/vmware-tanzu/nsx-operator/pkg/controllers/common"
 	"github.com/vmware-tanzu/nsx-operator/pkg/metrics"
 	servicecommon "github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/node"
@@ -59,7 +61,7 @@ func TestNodeReconciler_Reconcile(t *testing.T) {
 		{
 			name:           "Node not found",
 			existingNode:   nil,
-			expectedResult: ctrl.Result{},
+			expectedResult: common.ResultNormal,
 			expectedErr:    false,
 		},
 		{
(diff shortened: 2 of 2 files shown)


pkg/controllers/node/node_controller_test.go
```diff
@@ -102,6 +150,16 @@ func TestNodeReconciler_Reconcile(t *testing.T) {
 				},
 			}
 
+			if tt.name == "Node not found" {
```
lxiaopei, line 153: Can we use below to mock SyncNodeStore return nil or error? then I think we do not need to change files like node.go and store.go under nsx/services/node.
patches.ApplyMethod(reflect.TypeOf(r.NodeService), "SyncNodeStore", func(_ *node.Service, nodeName string, deleted bool) error {
					return nil				})

zhengxiexie: ok, fixed

pkg/controllers/node/node_controller.go
```diff
@@ -45,10 +45,15 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		if errors.IsNotFound(err) {
 			log.Info("node not found", "req", req.NamespacedName)
 			deleted = true
+			if err := r.Service.SyncNodeStore(req.NamespacedName.Name, deleted); err != nil {
+				log.Error(err, "failed to sync node store", "req", req.NamespacedName)
+				return ctrl.Result{}, err
```
yanjunz97, line 50: Maybe replace all `ctrl.Result{}` in this file by `common.ResultNormal` to make it consistent?

pkg/controllers/node/node_controller.go
```diff
@@ -45,10 +45,15 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+			}
+			return ctrl.Result{}, nil
 		} else {
 			log.Error(err, "unable to fetch node", "req", req.NamespacedName)
+			return common.ResultNormal, client.IgnoreNotFound(err)
```
yanjunz97, line 55: Maybe just return err as in this else case client.IgnoreNotFound(err) will always be err?
```suggestion
			return common.ResultNormal, err
```

Or you can put it after the conditions.

```
if err := r.Client.Get(ctx, req.NamespacedName, node); err != nil {
		if errors.IsNotFound(err) {
			log.Info("node not found", "req", req.NamespacedName)
			deleted = true
			if err := r.Service.SyncNodeStore(req.NamespacedName.Name, deleted); err != nil {
				log.Error(err, "failed to sync node store", "req", req.NamespacedName)
				return common.ResultNormal, err
			}
		} else {
			log.Error(err, "unable to fetch node", "req", req.NamespacedName)
		}
                 return common.ResultNormal, client.IgnoreNotFound(err)
	}
```

pkg/controllers/node/node_controller.go
zhengxiexie: ok

pkg/controllers/node/node_controller.go
zhengxiexie: ok

pkg/controllers/node/node_controller.go
```diff
@@ -45,6 +45,11 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		if errors.IsNotFound(err) {
 			log.Info("node not found", "req", req.NamespacedName)
 			deleted = true
+			if err := r.Service.SyncNodeStore(req.NamespacedName.Name, deleted); err != nil {
+				log.Error(err, "failed to sync node store", "req", req.NamespacedName)
+				return common.ResultNormal, err
```
lxiaopei, line 50: this line is not needed since we have it in #52?

pkg/controllers/node/node_controller.go
```diff
@@ -45,6 +45,11 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+			}
+			return common.ResultNormal, nil
```
lxiaopei, line 52: ditto

pkg/controllers/node/node_controller_test.go
```diff
@@ -90,18 +90,28 @@ func TestNodeReconciler_Reconcile(t *testing.T) {
...
 				NamespacedName: types.NamespacedName{
-					Name: "test-node",
+					Name: nodeName,
 				},
 			}
 
+			if tt.name != "Master Node" {
+				patchesSyncNodeStore := gomonkey.ApplyFunc((*node.NodeService).SyncNodeStore,
+					func(n *node.NodeService, nodeName string, deleted bool) error {
+						return nil
```
lxiaopei, line 110: could you also add a testcase for SyncNodeStore return error？

lxiaopei: for expectedResult，since you change to use common.ResultNormal, could you also use  common.ResultNormal instead of ctrl.Result{}？

pkg/controllers/node/node_controller.go
zhengxiexie: ok

pkg/controllers/node/node_controller_test.go
zhengxiexie: fixed

yanjunz97 (approved):
LGTM

lxiaopei (approved):
LGTM

PR #981, opened 2024-12-31
Remove API code for deprecated DHCP config

yanjunz97:
The PR remove the old API code for deprecated DHCP config and update the Subnet e2e test with DHCP switch check.


dantingl (approved):
Do we need to update SDK repo or NSX API schema is not changed?

yanjunz97:
> Do we need to update SDK repo or NSX API schema is not changed?

NSX API schema is not changed, and they still keep the dhcp_config field in nsx subnet.

TaoZou1 (approved):

PR #982, opened 2025-01-02
[CP]Fix bug which should delete node from store when there is node not found(980)

zhengxiexie:
When node is deleted, the phase node.ObjectMeta.DeletionTimestamp being not zero is intermittent, and it is more possible the node is not found, we should delete the node from store.


yanjunz97 (approved):

lxiaopei (approved):
LGTM

PR #986, opened 2025-01-03
Fix bug which overrides tags.tag when query dlb

zhengxiexie:
Search query supports multiple tags.scope and tags.tag, no need to construct a map.


wenyingd (approved):
LGTM

yanjunz97 (approved):
LGTM

PR #991, opened 2025-01-06
Fix Subnet update error detection

yanjunz97:
Operator modifies the existing Subnet in store before updating the NSX Subnet,
which results in the next reconcile will find the updated Subnet already in store,
skip the update, and overwrite the Status as Ready.
This PR uses a copy when updating the Subnet to avoid this issue.

### DIFF
--- a/pkg/nsx/services/subnet/builder.go
+++ b/pkg/nsx/services/subnet/builder.go
@@ -70,7 +70,7 @@ func (service *SubnetService) buildSubnet(obj client.Object, tags []model.Tag) (
 		if dhcpMode == "" {
 			dhcpMode = v1alpha1.DHCPConfigModeDeactivated
 		}
-		nsxSubnet.SubnetDhcpConfig = service.buildSubnetDHCPConfig(dhcpMode)
+		nsxSubnet.SubnetDhcpConfig = service.buildSubnetDHCPConfig(dhcpMode, nil)
 		nsxSubnet.IpAddresses = o.Spec.IPAddresses
 	case *v1alpha1.SubnetSet:
 		// The index is a random string with the length of 8 chars. It is the first 8 chars of the hash
@@ -87,7 +87,7 @@ func (service *SubnetService) buildSubnet(obj client.Object, tags []model.Tag) (
 		if dhcpMode == "" {
 			dhcpMode = v1alpha1.DHCPConfigModeDeactivated
 		}
-		nsxSubnet.SubnetDhcpConfig = service.buildSubnetDHCPConfig(dhcpMode)
+		nsxSubnet.SubnetDhcpConfig = service.buildSubnetDHCPConfig(dhcpMode, nil)
 	default:
 		return nil, SubnetTypeError
 	}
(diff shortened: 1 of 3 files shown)


pkg/nsx/services/subnet/subnet.go
```diff
@@ -428,36 +438,39 @@ func (service *SubnetService) UpdateSubnetSet(ns string, vpcSubnets []*model.Vpc
...
+		changed := common.CompareResource(SubnetToComparable(vpcSubnets[i]), SubnetToComparable(comparableSubnet))
 		if !changed {
 			log.Info("NSX Subnet unchanged, skipping update", "Subnet", *vpcSubnet.Id)
 			continue
 		}
 
-		vpcSubnets[i].Tags = newTags
+		// Avoid updating vpcSubnets[i] to ensure Subnet store
+		// is only updated after the updating succeeds.
+		updatedSubnet := *vpcSubnets[i]
```
dantingl, line 460: Why we need two new variables `comparableSubnet` and `updatedSubnet` here

pkg/nsx/services/subnet/subnet_test.go
```diff
@@ -398,7 +398,7 @@ func TestSubnetService_UpdateSubnetSet(t *testing.T) {
 	})
 
 	patchesCreateOrUpdateSubnet := gomonkey.ApplyFunc((*SubnetService).createOrUpdateSubnet,
-		func(r *SubnetService, obj client.Object, nsxSubnet *model.VpcSubnet, vpcInfo *common.VPCResourceInfo) (*model.VpcSubnet, error) {
+		func(r *SubnetService, obj client.Object, nsxSubnet *model.VpcSubnet, vpcInfo *common.VPCResourceInfo, isUpdate bool) (*model.VpcSubnet, error) {
```
dantingl, line 401: Do you think we need a new UT for `createOrUpdateSubnet`?

pkg/nsx/services/subnet/subnet.go
```diff
@@ -138,6 +143,11 @@ func (service *SubnetService) createOrUpdateSubnet(obj client.Object, nsxSubnet
 		Jitter:   0,
 		Steps:    6,
 	}
+
+	// For update case, wait for some time to make sure the realized state is updated.
+	if isUpdate {
+		time.Sleep(1 * time.Second)
+	}
```
TaoZou1, line 150: it's better that NSX fix that issue. After NSX patch API called, the realized status should be updated in sync mode instead

pkg/nsx/services/subnet/subnet.go
yanjunz97: It makes sense to merge them, use only updatedSubnet instead, thanks!

pkg/nsx/services/subnet/subnet_test.go
yanjunz97: Yes, some part of the code in createOrUpdateSubnet is not covered yet. Added the unit test, thanks

pkg/nsx/services/subnet/subnet.go
yanjunz97: Opened a bug for NSX team, let's see how they rely on this observations, thanks

dantingl (dismissed):

pkg/nsx/services/subnet/builder.go
```diff
@@ -113,6 +113,15 @@ func (service *SubnetService) buildSubnetDHCPConfig(mode string) *model.SubnetDh
 	return subnetDhcpConfig
 }
 
+func (service *SubnetService) updateSubnetDHCPConfig(subnetDHCPConfig *model.SubnetDhcpConfig, mode string) *model.SubnetDhcpConfig {
+	nsxMode := nsxutil.ParseDHCPMode(mode)
+	subnetDhcpConfig := &model.SubnetDhcpConfig{
+		DhcpServerAdditionalConfig: subnetDHCPConfig.DhcpServerAdditionalConfig,
+		Mode:                       &nsxMode,
+	}
+	return subnetDhcpConfig
```
TaoZou1, line 122: since the subnetDHCPConfig is pointer, could we just update the subnetDHCPConfig.Mode instead of create a new one and return? it seems no need to return it

yanjunz97: updatedSubnet.SubnetDhcpConfig is a pointer to vpcSubnets[i].SubnetDhcpConfig before the modification. If we change the Mode on it, the vpcSubnets[i].SubnetDhcpConfig.Mode will also be modified and result in the content in store updated before updating success. So I need to create a new subnetDHCPConfig instead of updating it.

TaoZou1 (approved):

dantingl (approved):

PR #994, opened 2025-01-07
Remove id comparison when deleting nsx resource by NamespacedName

yanjunz97:


pkg/controllers/staticroute/staticroute_controller.go
```diff
@@ -47,41 +47,16 @@ type StaticRouteReconciler struct {
...
 	for _, item := range nsxStaticRoutes {
-		uid := util.FindTag(item.Tags, commonservice.TagScopeStaticRouteCRUID)
-		if CRPolicySet.Has(uid) {
-			log.Info("skipping deletion, StaticRoute CR still exists in K8s", "staticrouteUID", uid, "nsxStatciRouteId", *item.Id)
-			continue
-		}
-
-		log.Info("deleting StaticRoute", "StaticRouteUID", uid, "nsxStaticRouteId", *item.Id)
+		log.Info("Deleting StaticRoute", "Namespace", ns, "Name", name, "nsxStaticRouteId", *item.Id)
 		path := strings.Split(*item.Path, "/")
```
wenyingd, line 54: Not introduced by this change, I prefer to use `ParseVPCResourcePath` to parse the IDs from the path to make the code consistent.

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -342,7 +341,12 @@ func (r *SubnetSetReconciler) CollectGarbage(ctx context.Context) {
 
 func (r *SubnetSetReconciler) deleteSubnetBySubnetSetName(ctx context.Context, subnetSetName, ns string) error {
 	nsxSubnets := r.SubnetService.ListSubnetBySubnetSetName(ns, subnetSetName)
-	return r.deleteStaleSubnets(ctx, nsxSubnets)
+	// We also actively delete the existing SubnetConnectionBindingMaps connected to the stale NSX Subnets.
```
wenyingd, line 344:
```suggestion
	// We also actively delete the SubnetConnectionBindingMaps associated with the empty NSX Subnet that has no SubnetPorts .
```

wenyingd (commented):
Overall LGTM, two nits

pkg/controllers/staticroute/staticroute_controller.go
yanjunz97: Updated, thanks!

pkg/controllers/staticroute/staticroute_controller.go
```diff
@@ -47,41 +47,19 @@ type StaticRouteReconciler struct {
...
 	nsxStaticRoutes := r.Service.ListStaticRouteByName(ns, name)
 	for _, item := range nsxStaticRoutes {
-		uid := util.FindTag(item.Tags, commonservice.TagScopeStaticRouteCRUID)
-		if CRPolicySet.Has(uid) {
-			log.Info("skipping deletion, StaticRoute CR still exists in K8s", "staticrouteUID", uid, "nsxStatciRouteId", *item.Id)
-			continue
+		log.Info("Deleting StaticRoute", "Namespace", ns, "Name", name, "nsxStaticRouteId", *item.Id)
+		vpcInfo, err := commonservice.ParseVPCResourcePath(*item.Path)
+		if err != nil {
+			return err
```
wenyingd, line 56: Better to print the error log if a failure happens.

wenyingd (dismissed):
One nit.

pkg/controllers/staticroute/staticroute_controller.go
yanjunz97: Updated, thanks!

wenyingd (dismissed):
LGTM

pkg/controllers/staticroute/staticroute_controller.go
```diff
@@ -47,41 +47,20 @@ type StaticRouteReconciler struct {
...
+		if err != nil {
+			log.Error(err, "Failed to parse NSX VPC path for StaticRoute", "path", *item.Path)
+			return err
 		}
-
-		log.Info("deleting StaticRoute", "StaticRouteUID", uid, "nsxStaticRouteId", *item.Id)
-		path := strings.Split(*item.Path, "/")
-		if err := r.Service.DeleteStaticRouteByPath(path[2], path[4], path[6], *item.Id); err != nil {
-			log.Error(err, "failed to delete StaticRoute", "StaticRouteUID", uid, "nsxStaticRouteId", *item.Id)
+		if err := r.Service.DeleteStaticRouteByPath(vpcInfo.OrgID, vpcInfo.ProjectID, vpcInfo.VPCID, *item.Id); err != nil {
```
timdengyun, line 59: I feel that we had better change https://github.com/vmware-tanzu/nsx-operator/blob/main/pkg/nsx/services/staticroute/staticroute.go#L107 DeleteStaticRouteByPath to pass a *model.StaticRoutes obj to do deletion like other components do. So, we don't need to handle VPC path logic here.

yanjunz97: Refactor the deletion function, please take another look, thanks!

timdengyun (approved):
LGTM, Thanks

pkg/controllers/staticroute/staticroute_controller.go
timdengyun: Thanks, Yanjun!

pkg/nsx/services/staticroute/staticroute.go
```diff
@@ -104,24 +103,25 @@ func (service *StaticRouteService) patch(orgId string, projectId string, vpcId s
 	return nil
 }
 
-func (service *StaticRouteService) DeleteStaticRouteByPath(orgId string, projectId string, vpcId string, id string) error {
+func (service *StaticRouteService) DeleteStaticRoute(nsxStaticRoute *model.StaticRoutes) error {
 	staticRouteClient := service.NSXClient.StaticRouteClient
-	staticroute := service.StaticRouteStore.GetByKey(id)
-	if staticroute == nil {
-		return nil
+	vpcInfo, err := common.ParseVPCResourcePath(*nsxStaticRoute.Path)
```
wenyingd, line 108: Better to check nil pointer before consuming a resource (`nsxStaticRoute`) if we dedicate on the current function `DeleteStaticRoute`. Not a strong option if we could the safeguard on it from the caller side.

yanjunz97: Currently we assume fields like Id and Path always exists in the NSX resource and not checking before dereferencing. Let's keep this as consistency. We can consider check them all in a future PR if necessary.

wenyingd (approved):
LGTM

PR #996, opened 2025-01-09
Update dhcp validation to prevent switch from DHCPServer/DHCPRelay to DHCPDeactivated

yanjunz97:
According to NSX change that we would not support static IP Allocation modification.
We add rules to only allow the DHCP mode switch between DHCPServer and DHCPRelay.

Testing done:
- subnetDHCPConfig.mode can switch between DHCPServer and DHCPRelay, but not between DHCPDeactivated and DHCPServer/DHCPRelay
- those 3 conditions can switch between each other: subnetDHCPConfig.mode=DHCPDeactivated, subnetDHCPConfig.mode=null, subnetDHCPConfig=null
- In case of subnetDHCPConfig.mode=null or subnetDHCPConfig=null, we cannot set subnetDHCPConfig.mode to  DHCPServer or DHCPRelay


pkg/apis/vpc/v1alpha1/subnet_types.go
```diff
@@ -92,12 +95,12 @@ type SubnetList struct {
 }
 
 // SubnetDHCPConfig is DHCP configuration for Subnet.
-// +kubebuilder:validation:XValidation:rule="has(oldSelf.mode) || !has(self.mode) || self.mode=='DHCPDeactivated'", message="subnetDHCPConfig cannot switch from DHCPDeactivated to other modes"
+// +kubebuilder:validation:XValidation:rule="has(oldSelf.mode)==has(self.mode) || (has(oldSelf.mode) && !has(self.mode)  && oldSelf.mode=='DHCPDeactivated') || (!has(oldSelf.mode) && has(self.mode) && self.mode=='DHCPDeactivated')", message="subnetDHCPConfig mode can only switch between DHCPServer and DHCPRelay"
```
lxiaopei, line 98: do you test if it is allowed if !has(oldSelf.mode) && !has(self.mode)?

pkg/apis/vpc/v1alpha1/subnet_types.go
```diff
@@ -20,7 +20,7 @@ const (
 )
 
 // SubnetSpec defines the desired state of Subnet.
-// +kubebuilder:validation:XValidation:rule="has(oldSelf.subnetDHCPConfig) || !has(self.subnetDHCPConfig) || !has(self.subnetDHCPConfig.mode) || self.subnetDHCPConfig.mode=='DHCPDeactivated'", message="subnetDHCPConfig cannot switch from DHCPDeactivated to other modes"
+// +kubebuilder:validation:XValidation:rule="has(oldSelf.subnetDHCPConfig)==has(self.subnetDHCPConfig) || (has(oldSelf.subnetDHCPConfig) && !has(self.subnetDHCPConfig) && (!has(oldSelf.subnetDHCPConfig.mode) || oldSelf.subnetDHCPConfig.mode=='DHCPDeactivated')) || (!has(oldSelf.subnetDHCPConfig) && has(self.subnetDHCPConfig) && (!has(self.subnetDHCPConfig.mode) || self.subnetDHCPConfig.mode=='DHCPDeactivated'))", message="subnetDHCPConfig mode can only switch between DHCPServer and DHCPRelay"
```
lxiaopei, line 23: we should also allow if !has(oldSelf.subnetDHCPConfig) && !has(self.subnetDHCPConfig)?

pkg/apis/vpc/v1alpha1/subnet_types.go
yanjunz97: Yes, it is allowed, has(oldSelf.mode)==has(self.mode) allows the case for both `!has(oldSelf.mode) && !has(self.mode)` and `has(oldSelf.mode) && has(self.mode)`

pkg/apis/vpc/v1alpha1/subnet_types.go
yanjunz97: Similarly, `has(oldSelf.subnetDHCPConfig)==has(self.subnetDHCPConfig)` allows the case for both `!has(oldSelf.subnetDHCPConfig) && !has(self.subnetDHCPConfig)` and `has(oldSelf.subnetDHCPConfig) && has(self.subnetDHCPConfig)`

lxiaopei (dismissed):
LGTM

yanjunz97:
Removed the e2e test for dhcp switch for now as the current e2e testbed does not apply crd for each run and the current e2e test for switch is not well-designed for the new case.
Will add back the e2e test with better cases after the crd files are applied the e2e testbed.

timdengyun (approved):
LGTM

PR #995, opened 2025-01-08
New string generator for the hash string in NSX resources

wenyingd:
1. Use chars 0-9,a-z,A-Z to generate the hash string for ID/DisplayName in NSX resources with VPC scenario. It also works on the Security Policy related resources' display name in T1.
2. The length is 6 chars when using the new hash string function.

Test Done:
1. VPC pipeline has passed, the supervisor is running, and the cluster status is ready.
2. T1 pipeline has passed.

Manual Verification
1. Created a subnetset with a long name which is longer than 80 char:
```
# k get subnetset -n antrea-test
NAME                                                                           ACCESSMODE   IPV4SUBNETSIZE   NETWORKADDRESSES
subnetset-sample-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa   Private      64               172.26.0.0/26
```
2. Created a VM using the subnetset to enforce a NSX subnet is created,
3. Check the NSX subnet configurations, and observed its display name is shorter than 80 chars and all chars are allowed.
```
{
    "ipv4_subnet_size": 64,
    "access_mode": "Private",
    "ip_addresses": [
        "172.26.0.0/26"
    ],
    "advanced_config": {
        "static_ip_allocation": {
            "enabled": true
        },
        "connectivity_state": "CONNECTED",
        "gateway_addresses": [
            "172.26.0.1/26"
        ]
    },
    "ip_blocks": [
        "/orgs/default/projects/project-quality/infra/ip-blocks/antrea-test_5110d687-625f-4d36-8a69-3642a46eaa15-172_26_0_0_16"
    ],
    "subnet_dhcp_config": {
        "mode": "DHCP_DEACTIVATED"
    },
    "resource_type": "VpcSubnet",
    "id": "subnetset-sample-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa_ccfb9059-fc0f-4eb9-945e-10c525630f25_7316e37e",
    "display_name": "subnetset-sample-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa_qepgkF_7316e37e",
    "tags": [
        {
            "scope": "nsx-op/cluster",
            "tag": "1ea68964-9a4a-465d-98ec-e0cf5b49088e"
        },
        {
            "scope": "nsx-op/version",
            "tag": "1.0.0"
        },
[...]

PR #997, opened 2025-01-10
[Cherry-Pick] New string generator for the hash string in NSX resources (#995)

wenyingd:
1. Use chars 0-9,a-z,A-Z to generate the hash string for ID/DisplayName in NSX resources with VPC scenario. It also works on the Security Policy related resources' display name in T1.
2. The length is 6 chars when using the new hash string function.


timdengyun (approved):
LGTM

lxiaopei (approved):
LGTM

PR #1005, opened 2025-01-15
Refine NetworkInfo CR VPCState.LoadBalancerIPAddresses description

timdengyun:
In NSX LB, we don't set NSX LN SNAT IPs in NetworkInfo CR VPCState.LoadBalancerIPAddresses and leave it as empty.
So, removing the description of adding NSX LB IPs not to make users confuse.


timdengyun:
not needed any more, it's handled by https://github.com/vmware-tanzu/nsx-operator/pull/1018

PR #1009, opened 2025-01-16
Cleanup enhancement

wenyingd:
The original logic is to one-by-one clean up the NSX resources according to the type. This may introduce many un-needed API request introduced when deleting the VPC sub-resources after the corresponding VPC is already removed.

This change is to refine the workflow as below,
1. Deleting the resources on NSX before removing the VPC. This step is to remove the resources whose existence may block the VPC deletion, e.g., SubnetPort.
2. Deleting the VPCs with option "recursive=true".
3. Deleting the other resources created in the infra.

HAPI is used to delete NSX resources by which we can reduce the number of API requests, and it is trying to delete the resources in parallel.

Test Done:
1. With a benchmark for 50 VPCs, 5 subnets/ 100 subnet ports / 5 subnet connection binding maps / 1 static routes / 1 vpc IP address allocation / 1 security policy configured per VPC, the new logic takes about 3m to complete the clean up job, while the original logic may take almost 30m to do the same thing.


pkg/nsx/services/common/policy_tree.go
```diff
@@ -0,0 +1,527 @@
...
+		leafType:      modelFormat[len(modelFormat)-1],
+
+		leafWrapper: leafWrapper[T],
+		pathGetter:  getNSXResourcePath[T],
+		idGetter:    getNSXResourceId[T],
+	}, nil
+}
+
+func PagingDeleteResources[T any](ctx context.Context, builder *PolicyTreeBuilder[T], objs []T, pageSize int, nsxClient *nsx.Client, delFn func(deletedObjs []T)) error {
+	if len(objs) == 0 {
```
TaoZou1, line 514: name delFun  to deleteObjectsFromStore?

pkg/nsx/services/securitypolicy/cleanup.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	return builder.PagingDeleteResources(ctx, cachedShares, common.DefaultHAPIChildrenCount, nsxClient, func(deletedObjs []*model.Share) {
+		store.DeleteMultipleObjects(deletedObjs)
+	})
+}
+
+func cleanGroups(ctx context.Context, store *GroupStore, builder *common.PolicyTreeBuilder[*model.Group], nsxClient *nsx.Client) error {
+	cachedObjs := store.List()
+	log.Info("Cleaning up Groups", "Count", len(cachedObjs))
```
yanjunz97, line 163: Also check cachedObjs length before moving on like this?
https://github.com/vmware-tanzu/nsx-operator/blob/0f6c0d5779200a4e5eb9fdf8dc932d48bf3f3371/pkg/nsx/services/securitypolicy/cleanup.go#L145-L147

pkg/nsx/services/common/policy_tree.go
```diff
@@ -0,0 +1,535 @@
...
+		segmentsCount += 1
+	}
+	if segmentsCount != len(b.pathFormat)*2 {
+		return nil, fmt.Errorf("invalid input path: %s", inputPath)
+	}
+
+	return pathSegments, nil
+}
+
+func (b *PolicyTreeBuilder[T]) DeleteMultipleResourcesOnNSX(objects []T, nsxClient *nsx.Client) error {
```
yanjunz97, line 444: The function name is a little bit confusing as it looks like the function is for update instead of deletion. It deletes the objects only if the caller sets markForDelete to true. Would you think it's better to rename to UpdateMultipleResourcesOnNSX or add some comments to clarify this?
If you choose to rename the function, maybe also update PagingDeleteResources?

yanjunz97 (dismissed):
LGTM

TaoZou1 (approved):

yanjunz97 (approved):

PR #1020, opened 2025-02-10
nsxserviceaccount: update PI cert when there is PI while there is no cluster-control-plane

liu4480:
Each time when nsx-operator reconciles NSXServiceAccount, it creates a new cert, and uses the new cert to create PI, then create the cluster-control-plane resource.

If the PI creation was successful, but cluster-control-plane creation fails, it doesn't delete the PI, but it continue with next retry. The next retry will generate a new cert, but the new cert is not updated to the existing PI, and the new cert is used for creating cluster-control-plane. In this bug, at first the license is missing, so PI is created with an older cert, and then license is configured, so the cluster-control-plane is created with a newer cert. That's why the clusters created before the license was applied run into such symptom.

the key problem is this code snippet:
```
else if !hasPI != (piObj == nil) {
                return "", fmt.Errorf("old PI exists")
}
```
This code is originally for checking whether a cluster with the same name has already been registered but it's with a different uuid. This means probably there is some leftover data in NSX.

Leftover data case:
hasPI: false
!hasPI: true
(piObj == nil): false
!hasPI != (piObj == nil): true
So it reports old PI exists error.

PI creation successful but cluster-control-plane creation failure case:
hasPI: true
!hasPI: false
(piObj == nil): false
!hasPI != (piObj == nil): false
So it doesn't report error, in this case, nsxserviceaccounts server shall update the PI with new cert

### DIFF
--- a/pkg/nsx/services/nsxserviceaccount/cluster.go
+++ b/pkg/nsx/services/nsxserviceaccount/cluster.go
@@ -186,12 +186,36 @@ func (s *NSXServiceAccountService) RestoreRealizedNSXServiceAccount(ctx context.
 	hasCCP := len(s.ClusterControlPlaneStore.GetByIndex(common.TagScopeNSXServiceAccountCRUID, string(obj.UID))) > 0
 	piObj := s.PrincipalIdentityStore.GetByKey(normalizedClusterName)
 	ccpObj := s.ClusterControlPlaneStore.GetByKey(normalizedClusterName)
+	var pi *mpmodel.PrincipalIdentity
+	var certificate mpmodel.Certificate
+	detail := true
+	if piObj != nil {
+		pi = piObj.(*mpmodel.PrincipalIdentity)
+		certificate, _ = s.NSXClient.CertificatesClient.Get(*(pi.CertificateId), &detail)
+	}
+	// read Secret
+	secretName := obj.Status.Secrets[0].Name
+	secretNamespace := obj.Status.Secrets[0].Namespace
+	secret := &v1.Secret{}
+	if err := s.Client.Get(ctx, types.NamespacedName{Name: secretName, Namespace: secretNamespace}, secret); err != nil {
+		return err
+	}
+	cert := secret.Data[SecretCertName]
+
+	log.V(1).Info("RestoreRealizedNSXServiceAccount: ",
+		"certificate.PemEncoded==nil", certificate.PemEncoded == nil,
+		"certificate.PemEncoded", certificate.PemEncoded,
+		"cert", string(cert))
 	if hasPI && hasCCP && piObj != nil && ccpObj != nil {
+		if string(cert) != "" && (certificate.PemEncoded == nil || *(certificate.PemEncoded) != string(cert)) {
+			return s.updatePICert(pi, normalizedClusterName, string(cert))
+		}
 		return nil
 	} else if hasPI || hasCCP || (piObj != nil) || (ccpObj != nil) {
 		return fmt.Errorf("PI/CCP doesn't match")
 	}
 	_, err := s.NSXClient.ClusterControlPlanesClient.Get(siteId, enforcementpointId, normalizedClusterName)
+	log.V(1).Info("RestoreRealizedNSXServiceAccount s.NSXClient.ClusterControlPlanesClient.Get", "err", err)
 	err = nsxutil.TransNSXApiError(err)
 	if err == nil {
 		return fmt.Errorf("CCP store is not synchronized")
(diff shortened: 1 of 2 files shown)


pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -249,8 +249,14 @@ func (s *NSXServiceAccountService) createPIAndCCP(normalizedClusterName string,
 			return "", err
 		}
 		s.PrincipalIdentityStore.Add(&pi)
-	} else if !hasPI != (piObj == nil) {
+	} else if !hasPI && (piObj == nil) {
```
gran-vmv, line 252: No need to change this, this branch is to handle same NSXSA name with different UID.

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -482,6 +469,31 @@ func (s *NSXServiceAccountService) updatePIAndCCPCert(normalizedClusterName, uid
 	return nil
 }
 
+func (s *NSXServiceAccountService) updatePIAndCCPCert(normalizedClusterName, uid, cert string) error {
```
gran-vmv, line 472: Move this func prior to `updatePICert` to avoid many lines changed.

pkg/nsx/services/nsxserviceaccount/cluster.go
liu4480: done

pkg/nsx/services/nsxserviceaccount/cluster.go
liu4480: done

liu4480:
@edwardbadboy @dantingl @lxiaopei can you have a look? thanks

gran-vmv (commented):
LGTM

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -250,7 +269,15 @@ func (s *NSXServiceAccountService) createPIAndCCP(normalizedClusterName string,
 		}
 		s.PrincipalIdentityStore.Add(&pi)
 	} else if !hasPI != (piObj == nil) {
+		log.Error(fmt.Errorf("old PI exists %v %v", !hasPI, piObj == nil), "")
```
edwardbadboy, line 272: Better to add the variable name in the log message, too. After some time, no one knows "old PI exists true false" means, especially the first values is negation of hasPI, it will easily confuses people. I'd like to suggest `"old PI exists: '!hasPI' = 'true/false', 'piObj==nil' = 'true/false'"`.

By the way, I think it's better to leverage the existing structured log formatter of log.Error. Like:
```
log.Error(fmt.Errorf("conflicting old PI exists"), "Failed to create PI and CCPN", "!hasPI", !hasPI, "piObj == nil", piObj == nil)
```

pkg/nsx/services/nsxserviceaccount/cluster.go
```diff
@@ -186,12 +186,36 @@ func (s *NSXServiceAccountService) RestoreRealizedNSXServiceAccount(ctx context.
...
+		return err
+	}
+	cert := secret.Data[SecretCertName]
+
+	log.V(1).Info("RestoreRealizedNSXServiceAccount: ",
+		"certificate.PemEncoded==nil", certificate.PemEncoded == nil,
+		"certificate.PemEncoded", certificate.PemEncoded,
+		"cert", string(cert))
 	if hasPI && hasCCP && piObj != nil && ccpObj != nil {
+		if string(cert) != "" && (certificate.PemEncoded == nil || certificate.PemEncoded != nil && *(certificate.PemEncoded) != string(cert)) {
```
edwardbadboy, line 210: I think `(certificate.PemEncoded == nil || certificate.PemEncoded != nil && *(certificate.PemEncoded) != string(cert))` can be simplified a bit to:

```
(certificate.PemEncoded == nil || *(certificate.PemEncoded) != string(cert))
```

pkg/nsx/services/nsxserviceaccount/cluster.go
liu4480: done

pkg/nsx/services/nsxserviceaccount/cluster.go
liu4480: done

edwardbadboy (approved):

TaoZou1 (approved):

gran-vmv (approved):

PR #1022, opened 2025-02-17
Fix main.go startup failure

seanpang-vmware:
In init method, it could not use cf, err:= ..., it will overwirte the global var cf, thus when main() try to use the cf var, its value will be nil.


zhengxiexie (approved):
LGTM

TaoZou1 (approved):

PR #1025, opened 2025-02-19
Remove cache stores for VPCNetworkConfig

yanjunz97:
Current we maintain 2 caches, VPCNSNetworkConfigStore and VPCNetworkConfigStore,
for mapping between Namespace, VPCNetworkConfig name and VPCNetworkConfig Info.
But those information are parsed based Namespace and VPCNetworkConfig CRs, which
are cached in controller runtime client. Thus this PR
- Removes the cache in NSX Operator and relies on controller runtime client cache to get
those information.
- Removes defaultNetworkConfigCR and get it by listing VPCNetworkConfiguration

Testing done:
Create a Namespace and create VM on the Namespace. The VM can power on with IP.

### DIFF
--- a/pkg/controllers/namespace/namespace_controller.go
+++ b/pkg/controllers/namespace/namespace_controller.go
@@ -43,9 +43,9 @@ type NamespaceReconciler struct {
 }
 
 func (r *NamespaceReconciler) getDefaultNetworkConfigName() (string, error) {
-	exist, nc := r.VPCService.GetDefaultNetworkConfig()
-	if !exist {
-		return "", errors.New("default network config not found")
+	nc, err := r.VPCService.GetDefaultNetworkConfig()
+	if err != nil {
+		return "", fmt.Errorf("default NetworkConfig not found: %w", err)
 	}
 	return nc.Name, nil
 }
@@ -163,33 +163,6 @@ func (r *NamespaceReconciler) namespaceError(ctx context.Context, k8sObj client.
 	util.UpdateK8sResourceAnnotation(r.Client, ctx, k8sObj, changes)
 }
 
-func (r *NamespaceReconciler) insertNamespaceNetworkconfigBinding(ns string, anno map[string]string) error {
-	ncName, useDefault := "", false
-	var err error
-	if anno == nil {
-		log.V(2).Info("Empty annotation for Namespace, using default network config", "Namespace", ns)
-		useDefault = true
-	} else {
-		annoNC, ncExist := anno[types.AnnotationVPCNetworkConfig]
-		if !ncExist {
-			useDefault = true
-		} else {
-			ncName = annoNC
-		}
-	}
-
-	if useDefault {
-		ncName, err = r.getDefaultNetworkConfigName()
-		if err != nil {
-			return err
-		}
-	}
-
-	log.Info("Record Namespace and network config mapping relation", "Namespace", ns, "NetworkConfig", ncName)
-	r.VPCService.RegisterNamespaceNetworkconfigBinding(ns, ncName)
-	return nil
-}
-
 /*
 	VPC creation strategy:
 
@@ -224,33 +197,25 @@ func (r *NamespaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		log.Info("Start processing Namespace create/update event", "Namespace", ns)
 
 		ctx := context.Background()
-		annotations := obj.GetAnnotations()
-		err := r.insertNamespaceNetworkconfigBinding(ns, annotations)
+		ncName, err := r.VPCService.GetNetworkconfigNameFromNS(ctx, ns)
 		if err != nil {
-			log.Error(err, "Failed to build Namespace and NetworkConfig bindings", "Namespace", ns)
-			return common.ResultRequeueAfter10sec, err
-		}
-		ncName, ncExist := annotations[types.AnnotationVPCNetworkConfig]
-
-		// If ns do not have network config name tag, then use default vpc network config name
-		if !ncExist {
-			log.Info("NetworkConfig name not found on Namespace, using default NetworkConfig", "Namespace", ns)
-			ncName, err = r.getDefaultNetworkConfigName()
-			if err != nil {
-				log.Error(err, "Failed to get default network config name", "Namespace", ns)
-				return common.ResultRequeueAfter10sec, err
-			}
+			log.Error(err, "Failed to get NetworkConfig name", "Namespace", ns)
+			return common.ResultRequeueAfter10sec, nil
 		}
 
-		nc, ncExist := r.VPCService.GetVPCNetworkConfig(ncName)
+		nc, ncExist, err := r.VPCService.GetVPCNetworkConfig(ncName)
+		if err != nil {
+			log.Error(err, "Failed to get NetworkConfig", "Namespace", ncName)
+			return common.ResultRequeue, nil
+		}
 		if !ncExist {
-			message := fmt.Sprintf("missing network config %s for Namespace %s", ncName, ns)
+			message := fmt.Sprintf("missing NetworkConfig %s for Namespace %s", ncName, ns)
 			r.namespaceError(ctx, obj, message, nil)
 			return common.ResultRequeueAfter10sec, errors.New(message)
 		}
-		if !r.VPCService.ValidateNetworkConfig(nc) {
+		if !r.VPCService.ValidateNetworkConfig(*nc) {
 			// if network config is not valid, no need to retry, skip processing
-			message := fmt.Sprintf("invalid network config %s for Namespace %s, missing private cidr", ncName, ns)
+			message := fmt.Sprintf("invalid NetworkConfig %s for Namespace %s, missing private cidr", ncName, ns)
 			r.namespaceError(ctx, obj, message, nil)
 			return common.ResultRequeueAfter10sec, errors.New(message)
 		}
@@ -264,7 +229,6 @@ func (r *NamespaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return common.ResultNormal, nil
 	} else {
 		metrics.CounterInc(r.NSXConfig, metrics.ControllerDeleteTotal, common.MetricResTypeNamespace)
-		r.VPCService.UnRegisterNamespaceNetworkconfigBinding(obj.GetNamespace())
 		// actively delete default SubnetSet, so that SubnetSet webhook can admit the delete request
 		if err := r.deleteDefaultSubnetSet(ns); err != nil {
 			return common.ResultRequeueAfter10sec, err
--- a/pkg/controllers/networkinfo/networkinfo_utils.go
+++ b/pkg/controllers/networkinfo/networkinfo_utils.go
@@ -191,7 +191,7 @@ func getGatewayConnectionStatus(ctx context.Context, nc *v1alpha1.VPCNetworkConf
 	return gatewayConnectionReady, reason
 }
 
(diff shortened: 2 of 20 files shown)


pkg/nsx/services/vpc/vpc.go
```diff
@@ -1318,18 +1316,73 @@ func (s *VPCService) GetAllVPCsFromNSX() map[string]model.Vpc {
...
-	return nsVpcMap
+	return nsVpcMap, nil
 }
 
 func IsPreCreatedVPC(nc common.VPCNetworkConfigInfo) bool {
 	return nc.VPCPath != ""
 }
+
+func buildNetworkConfigInfo(vpcConfigCR *v1alpha1.VPCNetworkConfiguration) (*common.VPCNetworkConfigInfo, error) {
+	org, project, err := nsxProjectPathToId(vpcConfigCR.Spec.NSXProject)
```
TaoZou1, line 1347: move it to build.go?

pkg/nsx/services/vpc/vpc.go
```diff
@@ -1318,18 +1316,73 @@ func (s *VPCService) GetAllVPCsFromNSX() map[string]model.Vpc {
...
+	parts := strings.Split(path, "/")
+	if len(parts) < 4 {
+		return "", "", errors.New("invalid NSX project path")
+	}
+	return parts[2], parts[len(parts)-1], nil
+}
+
+func isDefaultNetworkConfigCR(vpcConfigCR *v1alpha1.VPCNetworkConfiguration) bool {
+	annos := vpcConfigCR.GetAnnotations()
+	val, exist := annos[common.AnnotationDefaultNetworkConfig]
```
TaoZou1, line 1378: the same function isDefaultNetworkConfigCR/nsxProjectPathToId in the vpcnetworkconfig_handler.go, could you  reuse them?

pkg/nsx/services/vpc/vpc.go
```diff
@@ -48,83 +50,79 @@ type VPCNetworkInfoStore struct {
...
+	nc, ncExist, err := s.GetVPCNetworkConfig(ncName)
+	if err != nil {
+		log.Error(err, "Failed to get network config info using network config name", "Name", ncName)
+		return nil, err
+	}
 	if !ncExist {
-		log.Info("Failed to get network config info using network config name", "Name", ncName)
-		return nil
+		log.Info("Network config info does not exist", "Name", ncName)
+		return nil, nil
```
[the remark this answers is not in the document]
yanjunz97: Sounds good. There are many occurrence of network config, I replaced them all by NetworkConfig, thanks

pkg/nsx/services/vpc/vpc.go
yanjunz97: Updated, thanks!

pkg/controllers/namespace/namespace_controller.go
```diff
@@ -225,32 +198,26 @@ func (r *NamespaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 		ctx := context.Background()
 		annotations := obj.GetAnnotations()
-		err := r.insertNamespaceNetworkconfigBinding(ns, annotations)
+
+		ncName, err := r.VPCService.GetNetworkconfigNameFromAnnotation(ns, annotations)
```
seanpang-vmware, line 202: could we shelve the annotation related logics into the service method:
eg:

r.VPCService.GetNetworkConfigNameByNamespace(ns)

hide the get k8s ns obj, get ns obj annoation logic into service method.
in this way, maybe in controller logic, we only need to care reconcile part, and don't need to retrieve k8s objs.
I see in following lines for VPC service interface, you already impl it this way, hence we do not need to leave this code still in reconciler anymore.

pkg/nsx/services/common/services.go
```diff
@@ -13,15 +13,13 @@ import (
...
-	GetVPCNetworkConfig(ncCRName string) (VPCNetworkConfigInfo, bool)
+	GetNamespacesByNetworkconfigName(nc string) ([]string, error)
+	GetVPCNetworkConfig(ncCRName string) (*VPCNetworkConfigInfo, bool, error)
 	ValidateNetworkConfig(nc VPCNetworkConfigInfo) bool
-	GetVPCNetworkConfigByNamespace(ns string) *VPCNetworkConfigInfo
-	GetDefaultNetworkConfig() (bool, *VPCNetworkConfigInfo)
+	GetVPCNetworkConfigByNamespace(ns string) (*VPCNetworkConfigInfo, error)
+	GetDefaultNetworkConfig() (*VPCNetworkConfigInfo, error)
 	ListVPCInfo(ns string) []VPCResourceInfo
+	GetNetworkconfigNameFromAnnotation(ns string, annos map[string]string) (string, error)
```
seanpang-vmware, line 22: according to comments in namespace controller, suggest to change it to GetNetworkConfigNameByNamespace, for public method, it seems unnecessary for invoker to retrieve k8s ns annoations to call this method.

pkg/controllers/networkinfo/networkinfo_utils.go
```diff
@@ -191,7 +191,7 @@ func getGatewayConnectionStatus(ctx context.Context, nc *v1alpha1.VPCNetworkConf
 	return gatewayConnectionReady, reason
 }
 
-func updateVPCNetworkConfigurationStatusWithAliveVPCs(ctx context.Context, client client.Client, ncName string, getAliveVPCsFn func(ncName string) []*model.Vpc) {
+func updateVPCNetworkConfigurationStatusWithAliveVPCs(ctx context.Context, client client.Client, ncName string, getAliveVPCsFn func(ncName string) ([]*model.Vpc, error)) {
```
seanpang-vmware, line 194: this method contains a new return value error, but it is never used?

pkg/controllers/networkinfo/networkinfo_utils.go
```diff
@@ -202,7 +202,12 @@ func updateVPCNetworkConfigurationStatusWithAliveVPCs(ctx context.Context, clien
 
 	if getAliveVPCsFn != nil {
 		aliveVPCs := sets.New[string]()
-		for _, vpc := range getAliveVPCsFn(ncName) {
+		vpcList, err := getAliveVPCsFn(ncName)
+		if err != nil {
+			log.Error(err, "Failed to get alive VPCs", "Name", ncName)
+			return
```
seanpang-vmware, line 208: I guess you plan to return nil, err here?

pkg/controllers/networkinfo/vpcnetworkconfig_handler.go
```diff
@@ -71,24 +61,23 @@ func (h *VPCNetworkConfigurationHandler) Generic(_ context.Context, _ event.Gene
...
 	}
 
-	// update network config info in store
-	info, err := buildNetworkConfigInfo(*newNc)
+	nss, err := h.vpcService.GetNamespacesByNetworkconfigName(newNc.Name)
 	if err != nil {
-		log.Error(err, "Failed to process network config update event")
-		return
+		log.Error(err, "Failed to get Namespaces with NetworkConfig", "VPCNetworkConfig", newNc.Name)
+		req := reconcile.Request{
```
seanpang-vmware, line 76: maybe a dumb question, since we do not need to update cache anymore, why do we need to trigger a new networkinfo event here? i see in line #91, it is already triggered, won't this event be a redundant one?

pkg/controllers/namespace/namespace_controller.go
yanjunz97: Sounds good to me to expose GetNetworkconfigNameFromNS instead of GetNetworkconfigNameFromAnnotation. Updated, thanks!
But we may still need to retrieve k8s objs in controller to determine whether this is a deleted ns.

pkg/nsx/services/common/services.go
yanjunz97: Updated as above, thanks!

pkg/controllers/networkinfo/networkinfo_utils.go
yanjunz97: It might look a little bit confusing to read but that is the return value for getAliveVPCsFn instead of updateVPCNetworkConfigurationStatusWithAliveVPCs. The value is used for log as the following
```
                vpcList, err := getAliveVPCsFn(ncName)
		if err != nil {
			log.Error(err, "Failed to get alive VPCs", "Name", ncName)
			return
		}
```

pkg/controllers/networkinfo/vpcnetworkconfig_handler.go
yanjunz97: I'm thinking about if there is an error from GetNamespacesByNetworkconfigName which means there might be a network issue when listing the namespace gets a miss, we may not want to ignore this update and need to retry later. So here I trigger a new VPCNetworkConfiguration event.
Added a return statement here to avoid confusion.

pkg/controllers/networkinfo/networkinfo_utils.go
seanpang-vmware: maybe i did not explain it clearly, I mean the method gives a return value type, but the value is never used in any return statement in the method logic.

pkg/controllers/networkinfo/networkinfo_utils.go
seanpang-vmware: same as above

pkg/controllers/networkinfo/vpcnetworkconfig_handler.go
yanjunz97: Good caught, removed the namespace info, thanks!

pkg/controllers/networkinfo/vpcnetworkconfig_handler.go
```diff
@@ -71,24 +61,23 @@ func (h *VPCNetworkConfigurationHandler) Generic(_ context.Context, _ event.Gene
...
+			NamespacedName: client.ObjectKey{Name: newNc.GetName(), Namespace: newNc.GetNamespace()},
```
seanpang-vmware: could you use some POC code to confirm if this function works.
the reconciler queue is for networkInfo, which means it only handles NetworkInfo CR event type, if you push a network config CR into the queue, it will not be handled.

the queue is not for entire k8s, but only for this network info reconciler.  so pushing other events into queue, eg: namespace, pod, subnet won't trigger its corresponding reconciler or handler, so does networkconfig CR.

yanjunz97: I see. I had a misunderstanding on the queue and thought it would be used by both reconciler and handler. I checked other handlers found that the errors are usually ignored under this condition, so I remove the requeue. Please let me know if you have better solution, thanks!

pkg/nsx/services/ipaddressallocation/ipaddressallocation.go
```diff
@@ -106,6 +106,7 @@ func (service *IPAddressAllocationService) Apply(nsxIPAddressAllocation *model.V
 	ns := service.GetIPAddressAllocationNamespace(nsxIPAddressAllocation)
 	VPCInfo := service.VPCService.ListVPCInfo(ns)
 	if len(VPCInfo) == 0 {
+		log.Error(nil, "Failed to list VPCInfo for IPAddressAllocation")
```
seanpang-vmware, line 109: how about move 109 after 110 to use the err in log.Error(err, ... ), using log.Error with error value nil seems weired.

yanjunz97: Updated, thanks!

TaoZou1 (approved):

PR #1033, opened 2025-03-06
Block SubnetPort allocation on an IP exhausted Subnet

yanjunz97:
For Pod case, we will create and delete SubnetPort for image fetcher.
As the IP release needs 2 minutes, it may happen that the SubnetPort
number on the Subnet is less than capacity but larger than the number
of available IPs. For this condition, we will get IP exhausted error when
creating the SubnetPort, and we would like to avoid allocating more
SubnetPorts on this IP-exhausted Subnet in the following 2 minutes.

Testing done:
Create 16 pods, delete them and re-create them. Check the log that
once the Subnet is found as IP-exhausted, the following SubnetPort
will be allocated on other Subnet.

### DIFF
--- a/pkg/nsx/services/subnetport/store.go
+++ b/pkg/nsx/services/subnetport/store.go
@@ -3,6 +3,7 @@ package subnetport
 import (
 	"errors"
 	"sync"
+	"time"
 
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 	"k8s.io/apimachinery/pkg/types"
@@ -98,8 +99,9 @@ type CountInfo struct {
 	// dirtyCount defines the number of SubnetPorts under creation in the Subnet
 	dirtyCount int
 	lock       sync.Mutex
-	// totalIp defines the number of available IP in the Subnet
-	totalIp int
+	// totalIP defines the number of available IP in the Subnet
+	totalIP            int
+	exhaustedCheckTime time.Time
 }
 
 func (vs *SubnetPortStore) Apply(i interface{}) error {
--- a/pkg/nsx/services/subnetport/subnetport.go
+++ b/pkg/nsx/services/subnetport/subnetport.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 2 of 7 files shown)


pkg/nsx/services/subnetport/subnetport.go
```diff
@@ -387,6 +393,9 @@ func (service *SubnetPortService) AllocatePortFromSubnet(subnet *model.VpcSubnet
 		info.totalIp = totalIP - 4
 	}
 
+	if time.Since(info.exhausted) < IPReallocationTime {
```
zhengxiexie, line 396: what does it mean?

yanjunz97: It checks if 2 minutes have passed since last time we received the IP exhausted error. We would like to avoid allocating more SubnetPorts on this IP-exhausted Subnet in 2 minutes.

zhengxiexie: rename to exhaustedCheckTime?

yanjunz97: Updated, thanks!

pkg/nsx/services/subnetport/store.go
```diff
@@ -99,7 +100,8 @@ type CountInfo struct {
 	dirtyCount int
 	lock       sync.Mutex
 	// totalIp defines the number of available IP in the Subnet
-	totalIp int
+	totalIp   int
```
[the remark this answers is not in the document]
yanjunz97: Updated, thanks!

pkg/nsx/util/errors.go
```diff
@@ -606,15 +607,16 @@ var (
 )
 
 type RealizeStateError struct {
-	message string
+	Message string
```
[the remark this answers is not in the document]
yanjunz97: I see, it's not used directly by external source. Changed it back, thanks.

zhengxiexie (dismissed):
lgtm

pkg/nsx/util/errors.go
```diff
@@ -607,14 +608,15 @@ var (
 
 type RealizeStateError struct {
 	message string
+	Code    int
```
timdengyun, line 611: `Code` is also not for external access, as the same as message, it it right?

timdengyun (commented):
a nit comment, Thanks.

pkg/nsx/util/errors.go
timdengyun: I missed that you have used it in subnetport to check error code, so, it should be ok.

timdengyun (approved):
LGTM

zhengxiexie (approved):

PR #1042, opened 2025-03-19
Fix inventory pod has no name

TaoZou1:
compareContainerApplicationInstance will check if pre is nil. 
The nil value was returned by GetByKey if store has no record. 
Pass nil to func compareResources if no record found

Test Done:
1. run the manager, add one pod
2. check if the pod name existed in the inventory

### DIFF
--- a/pkg/nsx/services/inventory/build.go
+++ b/pkg/nsx/services/inventory/build.go
@@ -20,14 +20,14 @@ func (s *InventoryService) BuildPod(pod *corev1.Pod) (retry bool) {
 	// Calculate the services related to this Pod from pendingAdd or inventory store.
 	containerApplicationIds := []string{}
 	if s.pendingAdd[string(pod.UID)] != nil {
-		containerApplicationInstance := s.pendingAdd[string(pod.UID)].(containerinventory.ContainerApplicationInstance)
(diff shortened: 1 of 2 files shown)


zhengxiexie (approved):
lgtm

pkg/nsx/services/inventory/build.go
```diff
@@ -89,7 +89,7 @@ func (s *InventoryService) BuildPod(pod *corev1.Pod) (retry bool) {
 		OriginProperties:        originProperties,
 		Status:                  status,
 	}
-
+	log.V(1).Info("Build pod", "current instance", containerApplicationInstance, "pre instance", preContainerApplicationInstance)
 	operation, _ := s.compareAndMergeUpdate(preContainerApplicationInstance, containerApplicationInstance)
```
yanjunz97, line 93: It looks like preContainerApplicationInstance is a value while containerApplicationInstance is a pointer?

yanjunz97: By the way, may I learn about why you use value for preContainerApplicationInstance? If there is record in the store, the pre will also not but nil even if it is a pointer?

yanjunz97 (approved):

PR #1046, opened 2025-03-24
[CI]Add namedport e2e test case

timdengyun:
Add one namedport e2e test cases to verify namedport SecurirtyPolicy traffic.

### DIFF
--- a/test/e2e/framework.go
+++ b/test/e2e/framework.go
@@ -556,6 +556,40 @@ func (data *TestData) podWaitForIPs(timeout time.Duration, name, namespace strin
 	return ips, nil
 }
 
+// deploymentWaitForIPsOrNames polls the K8s apiServer until the specified Pod in deployment has an IP address
+func (data *TestData) deploymentWaitForIPsOrNames(timeout time.Duration, namespace, deployment string) ([]string, []string, error) {
+	podIPStrings := sets.NewString()
+	var podNames []string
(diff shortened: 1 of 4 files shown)


test/e2e/framework.go
```diff
@@ -556,6 +556,40 @@ func (data *TestData) podWaitForIPs(timeout time.Duration, name, namespace strin
 	return ips, nil
 }
 
+// deploymentWaitForIPsOrNames polls the K8s apiServer until the specified Pod in deployment has an IP address
+func (data *TestData) deploymentWaitForIPsOrNames(timeout time.Duration, namespace, deployment string) ([]string, []string, error) {
```
zhengxiexie, line 560: I think we can merge deploymentWaitForIPsOrNames and deploymentWaitForNames to deploymentWaitFor like podWaitFor/vmWaitFor?

timdengyun: I checked the code, `deploymentWaitForNames` is for polling pecific pods are created, no matter they are running or not.
but here, `deploymentWaitForIPsOrNames`  is specific for running state pods, so pods have IPs and names. 
Two functions are close but used in different use cases. 
So, I think we can keep both functions instead of merging into one func.

zhengxiexie (approved):
lgtm

TaoZou1 (approved):

PR #1051, opened 2025-03-26
Add GC for pod

TaoZou1:
Add GC for pod

1. stop the nsx-operator
2. delete the pod
3. restart the nsx-operator, check if the inventory of pod has been deleted


pkg/nsx/services/inventory/pod.go
```diff
@@ -82,3 +91,15 @@ func (s *InventoryService) DeleteContainerApplicationInstance(externalId string,
 	}
 	return nil
 }
+
+func (s *InventoryService) DeleteStalePods() error {
```
zhengxiexie, line 95: I think we don't need this, isn't CleanStaleInventoryApplicationInstance enough?

TaoZou1: DeleteStalePods is specific for stale pod which doesn't have project_id (antrea have this scenario). 
		if applicationInstance.ContainerProjectId == "" {
				// ContainerProjectId is the index for applicationInstanceStore and cannot be empty.
				// The case should not happen and we need to put this one into stale cache and clean it during resync if it did happen.
CleanStaleInventoryApplicationInstance only for normal GC case.

pkg/controllers/inventory/inventory_controller.go
```diff
@@ -169,5 +170,39 @@ func (c *InventoryController) syncInventoryKeys() {
...
 func (c *InventoryController) CleanStaleInventoryObjects() error {
 	log.Info("Clean stale inventory objects")
+	err := c.CleanStaleInventoryApplicationInstance()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (c *InventoryController) CleanStaleInventoryApplicationInstance() error {
```
zhengxiexie, line 180: Move it to services/inventory/pod.go, keep this file not specific CR related.

zhengxiexie: Could you refactor it to 
CleanupInventoryApplicationInstance(onlyStale bool)
then we can reuse it in the following Cleanup feature.

if onlyStale, keep this logic as it is, if all, then not to check c.service.IsPodDeleted.

TaoZou1: Good idea. move it to pod.go
The cleanup/gc has different logic. it seems no need to merge them.

yanjunz97 (approved):
LGTM

zhengxiexie (approved):
lgtm

PR #1059, opened 2025-04-02
[WIP] Modify the log format to comply with ABNF standards

heypnus:


pkg/logger/logger.go
```diff
@@ -36,6 +37,17 @@ func InitLog(log *logr.Logger) {
 	Log = *log
 }
 
+func getGoroutineID() int {
```
zhengxiexie, line 40: It seems Go team discourages this way for poor performance.
Go standard library’s internal uses (like in the http/2 package) only employ this technique for debugging purposes.

pkg/logger/logger.go
```diff
@@ -24,7 +25,7 @@ var (
 		enc.AppendString(t.Format(logTmFmtWithMS))
 	}
 	customCallerEncoder = func(caller zapcore.EntryCaller, enc zapcore.PrimitiveArrayEncoder) {
-		enc.AppendString(colorYellow + caller.TrimmedPath() + colorReset)
+		enc.AppendString(fmt.Sprintf("nsx %d [nsx-op@%d caller=\"%s\"]", os.Getpid(), getGoroutineID(), caller.TrimmedPath()))
```
zhengxiexie, line 28: Could we add a switch like 
	flag.IntVar(&UseABNFLog)  default true
to control it?

TaoZou1, line 28: Could we cache the Getpid()?
getGoroutineID() uses lots of CPU time.

TaoZou1, line 28: and use operator '+' here has better performance than fmt.Sprintf

PR #1067, opened 2025-04-10
Implement GC for networkpolicy

zhengxiexie:


pkg/nsx/services/inventory/networkpolicy.go
```diff
@@ -6,11 +6,26 @@ import (
 
 	"github.com/antihax/optional"
 	nsxt "github.com/vmware/go-vmware-nsxt"
+	"github.com/vmware/go-vmware-nsxt/containerinventory"
 	networkingv1 "k8s.io/api/networking/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 )
 
+func (s *InventoryService) IsNetworkPolicyDeleted(namespace, name, externalId string, networkPolicy *networkingv1.NetworkPolicy) bool {
```
yanjunz97, line 15: Why we need this `networkPolicy *networkingv1.NetworkPolicy`? It looks like there is no use case to call this function with non-nil networkPolicy?

zhengxiexie: @TaoZou1 copy from ingress, seems support outside obj passing?

yanjunz97: Ingress uses this function here to reuse the ingress after checking if it is deleted
https://github.com/vmware-tanzu/nsx-operator/blob/ae4f611b19d3223bad28d5bf44c77db002045470/pkg/nsx/services/inventory/ingresspolicy.go#L43-L52
Though I'm not sure why this is different from other services.

zhengxiexie: ok, fixed

yanjunz97 (approved):
LGTM

TaoZou1 (approved):

PR #1068, opened 2025-04-16
Add SubnetAdvancedConfig in the subnet/subnetset CRD

heypnus:
This patch will allow the customer to specify the property
SubnetAdvancedConfig.StaticIpAllocation.enabled for VPC subnet in the subnet/subnetset CR.

If staticIPAllocation is not set explicitly in the subnet/subnetset CR, the static
IP pool allocation will be automatically enabled when the DHCP mode is DHCPDeactivated
or not set and disabled when the DHCP mode is DHCPServer or DHCPRelay.
The static IP pool allocation and DHCP mode cannot both be enabled simultaneously
in subnet.

In the NSXAPI doc, the default value of SubnetAdvancedConfig.StaticIpAllocation.enabled
is false.

Testing done:

1. create a new subnet with spec.advancedConfig.staticIPAllocation.enabled=true and
   spec.subnetDHCPConfig.mode=DHCPServer, the CR cannot be created and the validation
   report error as expected:
   > kubectl apply -f subnet-05-dhcp-and-static-ip-alloc.yaml
   > The Subnet "subnet-debug-05" is invalid: spec: Invalid value: "object": Static IP pool allocation and subnet DHCP configuration cannot both be enabled simultaneously in subnet
2. create a new subnet with spec.advancedConfig.staticIPAllocation.enabled=false and
   spec.subnetDHCPConfig=null, the subnet can be created; then create a new port on
   the subnet, the port also can be created and its status doesn't have the IP/MAC.
3. create a new subnet only with spec.subnetDHCPConfig.mode=DHCPDeactivated and a subnet
   port on this subnet, both the NSX subnet and port can be realized as expected.
4. create a new subnet only with spec.subnetDHCPConfig.mode=DHCPServer and a subnet
   port on this subnet, both the NSX subnet and port can be realized as expected.


build/yaml/crd/vpc/crd.nsx.vmware.com_subnets.yaml
```diff
@@ -63,6 +63,19 @@ spec:
...
                 - message: Value is immutable
                   rule: self == oldSelf
+              advancedConfig:
+                description: VPC Subnet advanced configuration.
+                properties:
+                  staticIPAllocation:
+                    description: StaticIPAllocation configuration for subnet ports
+                      with VIF attachment.
+                    properties:
+                      enable:
```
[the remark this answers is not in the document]
heypnus: updated

pkg/apis/vpc/v1alpha1/subnet_types.go
```diff
@@ -104,6 +108,23 @@ type SubnetDHCPConfig struct {
...
+	// mode is DHCPServer or DHCPRelay.
+	// The static IP pool allocation and DHCP mode cannot both be enabled simultaneously
+	// in subnet.
+	StaticIPAllocation StaticIPAllocation `json:"staticIPAllocation,omitempty"`
+}
+
+// Static IP allocation for VPC Subnet ports with VIF attachement.
+type StaticIPAllocation struct {
+	// Activate or Deactivate static ip allocation for VPC Subnet ports with VIF attachement.
+	Enable *bool `json:"enable,omitempty"`
```
tnqn, line 125: The field is named "enabled" in NSX API, not "enable". Should it be consistent with NSX API?

pkg/apis/vpc/v1alpha1/subnet_types.go
```diff
@@ -104,6 +108,23 @@ type SubnetDHCPConfig struct {
...
 }
 
+// AdvancedConfig is Subnet advanced configuration.
+type AdvancedConfig struct {
+	// StaticIPAllocation configuration for subnet ports with VIF attachment.
+	// If this field is not set, the static IP pool allocation will be automatically
+	// enabled when the DHCP mode is DHCPDeactivated or not set and disabled when the DHCP
+	// mode is DHCPServer or DHCPRelay.
+	// The static IP pool allocation and DHCP mode cannot both be enabled simultaneously
+	// in subnet.
```
tnqn, line 118: This defaulting means that the value of `advanced_config.subnet_dhcp_config.enabled` could be true/false when the value in the CR is unset. When CCI API shows the field, it doesn't know whether the field should be explicitly set or not. For example, `enabled=false` in NSX API could be got from `enabled=nil` and `enabled=false` when DHCP is activivated; `enabled=true` in NSX API could be got from `enabled=nil` and `enabled=true` when DHCP is deactivated. Users may see different configurations in CCI and Supervisor API.

@jianjuns, does it violate the consistency you defined?

And to keep consistent defaulting behavior, in CCI we also need to make this field optional and determine its value based on DHCP configuration, instead of following NSX's default value.

pkg/apis/vpc/v1alpha1/subnet_types.go
```diff
@@ -104,6 +108,23 @@ type SubnetDHCPConfig struct {
...
+	StaticIPAllocation StaticIPAllocation `json:"staticIPAllocation,omitempty"`
+}
+
+// Static IP allocation for VPC Subnet ports with VIF attachement.
+type StaticIPAllocation struct {
+	// Activate or Deactivate static ip allocation for VPC Subnet ports with VIF attachement.
```
tnqn, line 124:
```suggestion
	// Activate or Deactivate static ip allocation for VPC Subnet ports with VIF attachment.
```

yanjunz97 (commented):
You may also run `make generate-api-docs` to update our api docs

pkg/apis/vpc/v1alpha1/subnet_types.go
heypnus: My bad. Thanks a lot for pointing it out!

pkg/apis/vpc/v1alpha1/subnet_types.go
heypnus: Fixed, thanks!

pkg/apis/vpc/v1alpha1/subnet_types.go
```diff
@@ -104,6 +108,23 @@ type SubnetDHCPConfig struct {
...
+	// enabled when the DHCP mode is DHCPDeactivated or not set and disabled when the DHCP
+	// mode is DHCPServer or DHCPRelay.
+	// The static IP pool allocation and DHCP mode cannot both be enabled simultaneously
+	// in subnet.
+	StaticIPAllocation StaticIPAllocation `json:"staticIPAllocation,omitempty"`
+}
+
+// Static IP allocation for VPC Subnet ports with VIF attachment.
+type StaticIPAllocation struct {
+	// Activate or Deactivate static ip allocation for VPC Subnet ports with VIF attachment.
```
jianjuns, line 124: ip -> IP

pkg/apis/vpc/v1alpha1/subnet_types.go
```diff
@@ -104,6 +108,23 @@ type SubnetDHCPConfig struct {
 	Mode DHCPConfigMode `json:"mode,omitempty"`
 }
 
+// AdvancedConfig is Subnet advanced configuration.
+type AdvancedConfig struct {
+	// StaticIPAllocation configuration for subnet ports with VIF attachment.
```
jianjuns, line 113: Subnet Ports?

Maybe change VIF attachment to "Network Interface attachment" as we do not expose VIF in net-op APIs.

pkg/apis/vpc/v1alpha1/subnet_types.go
jianjuns: @tnqn : could you provide your suggested definition?

heypnus:
Abandoning this PR because the VLAN extension subnet doesn't support static_ip_allocation.enabled=False. Will reopen it after we determine the new solution.

jianjuns:
> Abandoning this PR because the VLAN extension subnet doesn't support static_ip_allocation.enabled=False.

I think static_ip_allocation is supported, but DHCP is not. Please watch for the discussion with IPAM team.

PR #1074, opened 2025-04-17
Create network errors from pod status message

zhengxiexie:
Implemented logic to capture pod status messages when the pod is not running. These errors are now stored in the `ContainerApplicationInstance` for better troubleshooting. Added a new test case to validate this behavior and ensure accurate error reporting.

### DIFF
--- a/pkg/nsx/services/inventory/build.go
+++ b/pkg/nsx/services/inventory/build.go
@@ -56,6 +56,27 @@ func (s *InventoryService) BuildPod(pod *corev1.Pod) (retry bool) {
 		status = InventoryStatusUnknown
 	}
 
+	// Create network errors from pod conditions and status message
+	var networkErrors []common.NetworkError
+	networkStatus := NetworkStatusHealthy
+	if pod.Status.Phase != corev1.PodRunning {
+		networkStatus = NetworkStatusUnhealthy
+		// Check for a message in pod status
+		if pod.Status.Message != "" {
+			networkErrors = append(networkErrors, common.NetworkError{
+				ErrorMessage: pod.Status.Message,
+			})
+		}
+		// Check conditions
+		for _, condition := range pod.Status.Conditions {
+			if condition.Message != "" {
+				networkErrors = append(networkErrors, common.NetworkError{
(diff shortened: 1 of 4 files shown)


pkg/nsx/services/inventory/build.go
```diff
@@ -85,7 +105,7 @@ func (s *InventoryService) BuildPod(pod *corev1.Pod) (retry bool) {
 		ContainerClusterId:      s.NSXConfig.Cluster,
 		ContainerProjectId:      string(namespace.UID),
 		ExternalId:              string(pod.UID),
-		NetworkErrors:           nil,
+		NetworkErrors:           networkErrors,
 		NetworkStatus:           "",
```
TaoZou1, line 109: Should we also update NetworkStatus?
and also need to update compare function

pkg/nsx/services/inventory/build.go
```diff
@@ -56,6 +56,26 @@ func (s *InventoryService) BuildPod(pod *corev1.Pod) (retry bool) {
...
+		// Check for a message in pod status
+		if pod.Status.Message != "" {
+			networkErrors = append(networkErrors, common.NetworkError{
+				ErrorMessage: pod.Status.Message,
+			})
+		}
+
+		// Check conditions
+		for _, condition := range pod.Status.Conditions {
+			if condition.Status == corev1.ConditionFalse {
```
yanjunz97, line 71: Why we collect all the condition message and sorted them for node but just false condition message for Pod?
https://github.com/vmware-tanzu/nsx-operator/blob/6095f84246ed681b215ea32852a806432bdd2cb0/pkg/nsx/services/inventory/build.go#L427-L435

zhengxiexie: image
Pod is different from VM, the status message carries more significance in this context.

pkg/nsx/services/inventory/build.go
zhengxiexie: fixed

yanjunz97 (approved):

TaoZou1 (approved):

PR #1077, opened 2025-04-22
Support localhost connection to k8s api server

yanjunz97:


cmd/main.go
```diff
@@ -279,7 +279,7 @@ func electMaster(mgr manager.Manager, nsxClient *nsx.Client) {
 
 func main() {
 	log.Info("Starting NSX Operator")
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	mgr, err := ctrl.NewManager(pkgutil.GetConfig(), ctrl.Options{
```
heypnus, line 282: So the API server address switch only can occur in the startup stage, right? Then if the eth1 down during the NSX operator runtime, what will happen?

yanjunz97: I've checked this in HA mode, and found operator will restart after eht1 down automatically because the lease renewal failed.

pkg/util/kubernetes.go
```diff
@@ -0,0 +1,55 @@
...
+	healthUrl := cfg.Host + "/healthz"
+	resp, err := client.Get(healthUrl)
+	if err != nil {
+		log.Error(err, "Failed to connect to Kubernetes API Server", "url", healthUrl)
+		return false
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		log.Error(nil, "Kubernetes healthz check failed", "status", resp.Status)
```
dantingl: Is there a way to distinguish TCP connection error, TLS handshake validation error?
I mean we should only switch to localhost for TCP connection error, right?

pkg/util/kubernetes.go
```diff
@@ -0,0 +1,55 @@
...
+	"strings"
+
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+const (
+	localhostIP   = "127.0.0.1"
+	localhostIPv6 = "::1"
+	localhostPort = "6443"
```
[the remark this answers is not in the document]
yanjunz97: Sure, I found KUBERNETES_SERVICE_PORT shall be set by default in pod. Use it first.

pkg/util/kubernetes.go
yanjunz97: Sounds great, I updated to not switch for client creation error and status error, thanks!

heypnus (approved):
LGTM

PR #1081, opened 2025-04-25
Support Pod restore

yanjunz97:
Testing done:
Create 2 Pods, delete the NSX SubnetPort for one of the Pod, force the restore mode and restart NSX Operator.
Observed the NSX SubnetPort is created with the same IP and MAC, and Pod is annotated with `nsx/restore: eth0`
Then labeled the restored Pod with test=true, observed the annotation was updated on SubnetPort and other fields
like attachment/mac/ip are not changed.

### DIFF
--- a/pkg/controllers/pod/pod_controller.go
+++ b/pkg/controllers/pod/pod_controller.go
@@ -5,8 +5,11 @@ package pod
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net"
 	"reflect"
+	"strings"
 	"time"
 
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
@@ -14,8 +17,10 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apimachineryruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/retry"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -27,6 +32,7 @@ import (
 	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
 	servicecommon "github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/subnetport"
+	"github.com/vmware-tanzu/nsx-operator/pkg/util"
 )
 
 var (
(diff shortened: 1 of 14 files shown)


pkg/controllers/pod/pod_controller.go
```diff
@@ -293,23 +354,40 @@ func (r *PodReconciler) CollectGarbage(ctx context.Context) error {
 	return nil
 }
 
+func (r *PodReconciler) getSubnetByIP(pod *v1.Pod, subnetSetUID string) (string, error) {
```
heypnus, line 357: Maybe `getSubnetByIP` -> `getSubnetByPod`? Because the input is pod, not IP..

yanjunz97: Good caught. Updated this for both Pod and SubnetPort, thanks!

heypnus (commented):
For the testing done, did you check the update case after the restore? e.g. label the pod after its subnetport is restored... You know I hit some issue in this case for ipaddressallocation restore.. The restored NSX resource may be different from the normal created resource...

pkg/controllers/pod/pod_controller.go
```diff
@@ -174,12 +181,32 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
...
+			}
+			err = util.UpdateK8sResourceAnnotation(r.Client, ctx, pod, podAnnotationChanges)
+			if err != nil {
+				log.Error(err, "Failed to update Pod annotation", "pod.Name", req.NamespacedName, "pod.UID", pod.UID, "podAnnotationChanges", podAnnotationChanges)
+				return common.ResultNormal, err
+			}
+		}
 		r.StatusUpdater.UpdateSuccess(ctx, pod, setPodReadyStatusTrue)
+		if r.restoreMode {
+			// Add restore annotation on Pod to notity Spherelet
```
dantingl, line 203: notify

pkg/controllers/pod/pod_controller.go
```diff
@@ -174,12 +181,32 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
...
+			retry.OnError(util.K8sClientRetry, func(err error) bool {
+				return err != nil
+			}, func() error {
+				return common.UpdateRestoreAnnotation(r.Client, ctx, pod, "eth0")
```
dantingl, line 207: I'm thinking maybe we first add annotation key before PodVM supports multi-nic

pkg/controllers/pod/pod_controller.go
yanjunz97: updated, thanks

pkg/controllers/pod/pod_controller.go
yanjunz97: I replaced eth0 by true for better readable, please let me know if you prefer other value. Thanks!

yanjunz97:
> > For the testing done, did you check the update case after the restore? e.g. label the pod after its subnetport is restored... You know I hit some issue in this case for ipaddressallocation restore.. The restored NSX resource may be different from the normal created resource...
> 
> Yeah, does nsx-operator update mac annotation on Pod after upgrade?

I updated the test done part with the tag update on restored object case.

I also checked that upgrading nsx-operator from previous version to this version will be able to add mac annotation on the Pods.

heypnus (approved):

heypnus (approved):

dantingl (approved):

PR #1084, opened 2025-04-30
Start controller in goroutine

yanjunz97:
Controller runtime manager needs to start before calling List/Get on manager client.
Otherwise we will get error `the cache is not started, can not read objects`


heypnus (approved):
Thanks for the fix.

PR #754, opened 2024-09-06
Support Subnet/Subnetset IPv4SubnetSize and Subnet IPAddresses immutable

timdengyun:
For NSX subnet, once it is created, it's not allowed to change IPv4SubnetSize and IPAddresses.
So, this patch is to
1. Change Subnet CR definition to make Subnet IPv4SubnetSize and IPAddresses are immutable once the CR is created.
2. Change Subnetset CR definition to make Subnetset IPv4SubnetSize is immutable once the CR is created.

Aslo, this patch is to
1. Add check for IPv4SubnetSize for Subnetset CR to see if is power of 2.

Test done:
subnet Test yaml:
cat vpctest-subnet-dhcp.yaml
```
apiVersion: crd.nsx.vmware.com/v1alpha1
kind: Subnet
metadata:
  name: test-subnet
  namespace: test
spec:
  DHCPConfig:
    enableDHCP: true
  ipv4SubnetSize: 33
  accessMode: Private
  #ipAddresses:
  #- 172.58.40.0/24
```
1. Create subnet CR with invalid size
kubectl apply -f vpctest-subnet-dhcp.yaml
output:
status:
  conditions:
  - lastTransitionTime: "2024-09-06T10:28:58Z"
    message: NSX Subnet could not be created/updated
    reason: SubnetNotReady

and change ipv4SubnetSize: 32 as valid size, which is not allowed to change ipv4SubnetSize field. User needs to delete the failed CR and recreate it. 
```
kubectl apply -f ./vpctest-subnet-dhcp.yaml
The Subnet "test-subnet" is invalid: spec.ipv4SubnetSize: Invalid value: "integer": Value is immutable    
```

2. Create subnet CR with valid size:32
```
 kubectl get subnet test-subnet -n test
NAME          ACCESSMODE   IPV4SUBNETSIZE   NETWORKADDRESSES
test-subnet   Private      32               172.48.0.0/27
```
and change ipv4SubnetSize to 64, 
```
kubectl apply -f ./vpctest-subnet-dhcp.yaml
The Subnet "test-subnet" is invalid: spec.ipv4SubnetSize: Invalid value: "integer": Value is immutable
```

3. Create subnet CR without ipv4SubnetSize, it will use the default ipv4SubnetSize:32 from vpcNetworkconfiguratinon CR.
```
apiVersion: crd.nsx.vmware.com/v1alpha1
kind: Subnet
metadata:
  name: test-subnet
  namespace: test
spec:
  DHCPConfig:
[...]

PR #1085, opened 2025-05-06
Support IPAddressAllocation restore

heypnus:
Testing done:
```
Case A
  Step 1. create an IPAddressAllocation CR with spec.allocationSize=32.
  Step 2. after the NSX VPC IP address allocation is created, stop the NSX operator.
  Step 3. delete the NSX VPC IP address allocation by manual and hack the ncpconfig
          CR to the restore mode.
  Step 4. start the NSX operator and check the NSX VPC IP address allocation is
          re-created and the allocation_ips matches the CR status.
  Step 5. rename the NSX VPC IP address allocation, hack the ncpconfig CR to quit
          the restore mode, then start the NSX operator again.
  Step 6. check the NSX VPC IP address allocation's name can be updated back.

Case B
  Step 1. create an IPAddressAllocation CR with spec.allocationSize=1.
  Step 2-4. similar to step 2-4 in case A.
  Step 5. hack the ncpconfig CR to quit the restore mode, start the NSX operator again.
  Step 6. check that the log can print "IPAddressAllocation is not changed".
```

### DIFF
--- a/pkg/controllers/ipaddressallocation/ipaddressallocation_controller.go
+++ b/pkg/controllers/ipaddressallocation/ipaddressallocation_controller.go
@@ -5,6 +5,7 @@ package ipaddressallocation
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	v1 "k8s.io/api/core/v1"
@@ -41,6 +42,7 @@ type IPAddressAllocationReconciler struct {
 	VPCService    servicecommon.VPCServiceProvider
 	Recorder      record.EventRecorder
 	StatusUpdater common.StatusUpdater
+	restoreMode   bool
 }
 
 func setReadyStatusFalse(client client.Client, ctx context.Context, obj client.Object, transitionTime metav1.Time, err error, _ ...interface{}) {
@@ -60,7 +62,7 @@ func setReadyStatusFalse(client client.Client, ctx context.Context, obj client.O
 	ipaddressallocation.Status.Conditions = conditions
 	e := client.Status().Update(ctx, ipaddressallocation)
 	if e != nil {
-		log.Error(e, "unable to update IPAddressAllocation status", "IPAddressAllocation", ipaddressallocation)
+		log.Error(e, "Unable to update IPAddressAllocation status", "IPAddressAllocation", ipaddressallocation)
 	}
 }
 
@@ -78,13 +80,13 @@ func setReadyStatusTrue(client client.Client, ctx context.Context, obj client.Ob
 	ipaddressallocation.Status.Conditions = conditions
 	e := client.Status().Update(ctx, ipaddressallocation)
 	if e != nil {
-		log.Error(e, "unable to update IPAddressAllocation status", "IPAddressAllocation", ipaddressallocation)
+		log.Error(e, "Unable to update IPAddressAllocation status", "IPAddressAllocation", ipaddressallocation)
 	}
 }
 
 func (r *IPAddressAllocationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	obj := &v1alpha1.IPAddressAllocation{}
-	log.Info("reconciling IPAddressAllocation CR", "IPAddressAllocation", req.NamespacedName)
+	log.Info("Reconciling IPAddressAllocation CR", "IPAddressAllocation", req.NamespacedName)
 	r.StatusUpdater.IncreaseSyncTotal()
 	if err := r.Client.Get(ctx, req.NamespacedName, obj); err != nil {
 		if apierrors.IsNotFound(err) {
@@ -106,7 +108,7 @@ func (r *IPAddressAllocationReconciler) Reconcile(ctx context.Context, req ctrl.
 
 func (r *IPAddressAllocationReconciler) handleUpdate(ctx context.Context, obj *v1alpha1.IPAddressAllocation) (ctrl.Result, error) {
 	r.StatusUpdater.IncreaseUpdateTotal()
-	updated, err := r.Service.CreateOrUpdateIPAddressAllocation(obj)
+	updated, err := r.Service.CreateOrUpdateIPAddressAllocation(obj, r.restoreMode)
 	if err != nil {
 		r.StatusUpdater.UpdateFail(ctx, obj, err, "", setReadyStatusFalse)
 		return resultRequeue, err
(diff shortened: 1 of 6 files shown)


pkg/nsx/services/ipaddressallocation/ipaddressallocation.go
```diff
@@ -82,11 +84,28 @@ func (service *IPAddressAllocationService) CreateOrUpdateIPAddressAllocation(obj
 		return false, err
 	}
 	log.V(1).Info("Existing ipaddressallocation", "ipaddressallocation", existingIPAddressAllocation)
+
+	if existingIPAddressAllocation != nil && existingIPAddressAllocation.AllocationIps != nil && existingIPAddressAllocation.AllocationSize == nil {
+		// For the restored NSX VPC IPAddressAllocation, its allocation_size is null.
+		// We cannot update it in the normal reconcile after the restore, because the error:
+		// "nsx error code: 612866, message: Properties IP block, allocation IPs, allocation IP, IP block visibility, allocation size and IPAddressType of existing Vpc IP address allocation: <VPC IP address allocation path> can not be modified."
```
yanjunz97, line 91: When new tag is added to the restored IPAddressAllocation, will the update be triggered and resulted in this error?
Shall we propagate the AllocationIps from existingIPAddressAllocation to nsxIPAddressAllocation and remove AllocationSize? Or just updating the tags field of the existingIPAddressAllocation?

heypnus: I updated the logic, now we remove the `AllocationSize` from desired `nsxIPAddressAllocation` instead of filling `AllocationSize` to `existingIPAddressAllocation`, I think this can cover the case that the properties like `tags`, `display_name` need to be updated. I think we don't need to propagate the `AllocationIps` from `existingIPAddressAllocation` to `nsxIPAddressAllocation` because the error 612866 indicates that the `AllocationIps` is immutable and we don't compare this field in the comparison function.

pkg/nsx/services/ipaddressallocation/ipaddressallocation.go
```diff
@@ -82,11 +82,21 @@ func (service *IPAddressAllocationService) CreateOrUpdateIPAddressAllocation(obj
...
+	if existingIPAddressAllocation != nil && existingIPAddressAllocation.AllocationIps != nil && existingIPAddressAllocation.AllocationSize == nil {
+		// For the restored NSX VPC IPAddressAllocation, its allocation_size is null.
+		// After the restore, if the built variable nsxIPAddressAllocation still has the nonempty
+		// allocation_size, we need to remove it from the parameters, otherwise the NSX operator
+		// will keep reporting the following error and retrying:
+		// "nsx error code: 612866, message: Properties IP block, allocation IPs, allocation IP, IP block visibility, allocation size and IPAddressType of existing Vpc IP address allocation: <VPC IP address allocation path> can not be modified."
+		// For this kind of cases, we manually populate the allocation_size before the comparison.
+		nsxIPAddressAllocation.AllocationSize = nil
+	}
 	ipAddressAllocationUpdated := common.CompareResource(IpAddressAllocationToComparable(existingIPAddressAllocation),
```
TaoZou1, line 95: if restoreMode, no need to CompareResource, right?
ipaddressallocation doesn't support UPDATE. 

There is a issue here, if the user update the AllocationSize later, there is no error shown.

heypnus: > if restoreMode, no need to CompareResource, right? ipaddressallocation doesn't support UPDATE.

I agree with you that the ipaddressallocation doesn't support update. But here the update ability is reserved, because I think in the future maybe we have the requirement to change the tags or naming rule, etc.

> There is a issue here, if the user update the AllocationSize later, there is no error shown.

Yes I found this issue in the test. I think this should be fixed from the starting point of the customer operation, not here. i.e. we need to deny the update via a **CRD change**. I think we can limit the modifications for the CRD in a new patch, the implementation looks like https://github.com/vmware-tanzu/nsx-operator/pull/754/files#diff-5dfd130d45825356caf06b49a0496416931c47ea61ded657a5899a39ec545f1fR23 . cc @dantingl

pkg/controllers/ipaddressallocation/ipaddressallocation_controller.go
```diff
@@ -172,9 +174,43 @@ func (r *IPAddressAllocationReconciler) CollectGarbage(ctx context.Context) erro
...
+	}
+	r.restoreMode = false
+	if len(errorList) > 0 {
+		return errors.Join(errorList...)
+	}
 	return nil
 }
 
+func (r *IPAddressAllocationReconciler) getRestoreList() ([]types.NamespacedName, error) {
+	nsxIPAddressAllocationCRIDs := r.Service.ListIPAddressAllocationID()
```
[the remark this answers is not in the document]
heypnus: > nsxIPAddressAllocationCRIDs -> nsxIPAddressAllocationIDs ?
I changed `nsxIPAddressAllocationCRIDs` to `ipAddressAllocationCRIDs` because the NSX ID and CR ID is different, and here it's CR ID.

> ipAddressAllocationList -> ipAddressAllocationCRList ?
done

TaoZou1 (dismissed):

pkg/nsx/services/ipaddressallocation/ipaddressallocation.go
TaoZou1: ipallocation also not check if size=0, will it be also verified by webhook?

pkg/nsx/services/ipaddressallocation/ipaddressallocation.go
yanjunz97: I refer to this case: When we update tags, we will call the patch API with `nsxIPAddressAllocation`. And this `nsxIPAddressAllocation` generated by `service.BuildIPAddressAllocation(obj)` in the normal reconcile does not contains the field AllocationIps. But I think if the previous IPAddressAllocation is restored with AllocationIps, and we update it without AllocationIps will lead to this error 612866?
I'm not sure if I have missed something. Have you tested the case update tags for a restored IPAddressAllocation?

heypnus: Many thanks for your good caught! It's indeed an issue, I updated the code to propagate the `AllocationIps` from `existingIPAddressAllocation` to `nsxIPAddressAllocation` as a fix and adding the test steps in case A, step 5-6, please check.

pkg/nsx/services/ipaddressallocation/ipaddressallocation.go
heypnus: I think we can verify it like https://github.com/vmware-tanzu/nsx-operator/blob/ecd2d6a3daa322f94bd655ab4a50ae8661931163/pkg/apis/vpc/v1alpha1/staticroute_types.go#L21 , webhook is not needed.

dantingl: Does NSX API support size update?
If size is 1, will NSX update `allocation_ip` or `allocation_ips`

yanjunz97 (approved):
LGTM

TaoZou1 (approved):

pkg/nsx/services/ipaddressallocation/ipaddressallocation.go
heypnus: > Does NSX API support size update? 

no

> If size is 1, will NSX update `allocation_ip` or `allocation_ips`

allocation_ips

PR #1086, opened 2025-05-07
Support Subnet/SubnetSet restore

yanjunz97:
Testing done:
- Subnet case:
  - Created a Subnet, deleted the NSX Subnet, forced restore mode and restarted NSX Operator. Observed the NSX Subnet is created with the same CIDR. Added label on the Namespace, observed the NSX Subnet tag is updated.
  - Created a Pod, deleted the NSX SubnetPort and NSX Subnet for the Pod, forced restore mode and restarted NSX Operator. Observed the NSX Subnet is created with same CIDR and NSX SubnetPort is created with the same IP/MAC. Added label on the Namespace, observed the NSX Subnet tag is updated.
  - Created 16 Pods in a Namespace with default Subnet size as 16, deleted the 16 NSX SubnetPorts and 2 NSX Subnets, forced restore mode and restarted NSX Operator. Observed the 2 NSX Subnets are created with the same CIDR and NSX SubnetPorts are created with the same IP/MAC.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -137,6 +137,7 @@ func startServiceController(mgr manager.Manager, nsxClient *nsx.Client) {
 
 	var vpcService *vpc.VPCService
 	var hookServer webhook.Server
+	var subnetSetReconcile *subnetset.SubnetSetReconciler
 
 	if cf.CoeConfig.EnableVPCNetwork {
 		// Check NSX version for VPC networking mode
@@ -206,12 +207,13 @@ func startServiceController(mgr manager.Manager, nsxClient *nsx.Client) {
 		}
 
 		// Create controllers which only supports VPC
+		subnetSetReconcile = subnetset.NewSubnetSetReconciler(mgr, subnetService, subnetPortService, vpcService, subnetBindingService)
 		reconcilerList = append(
 			reconcilerList,
 			networkinfocontroller.NewNetworkInfoReconciler(mgr, vpcService, ipblocksInfoService),
 			namespacecontroller.NewNamespaceReconciler(mgr, cf, vpcService),
 			subnet.NewSubnetReconciler(mgr, subnetService, subnetPortService, vpcService, subnetBindingService),
-			subnetset.NewSubnetSetReconciler(mgr, subnetService, subnetPortService, vpcService, subnetBindingService),
+			subnetSetReconcile,
 			node.NewNodeReconciler(mgr, nodeService),
 			staticroutecontroller.NewStaticRouteReconciler(mgr, staticRouteService),
 			subnetport.NewSubnetPortReconciler(mgr, subnetPortService, subnetService, vpcService),
@@ -233,6 +235,7 @@ func startServiceController(mgr manager.Manager, nsxClient *nsx.Client) {
 	}
 
 	if restoreMode {
+		subnetSetReconcile.EnableRestoreMode()
 		err := pkgutil.ProcessRestore(reconcilerList, mgr.GetClient())
 		if err != nil {
 			log.Error(err, "Failed to process restore")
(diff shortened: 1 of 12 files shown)


cmd/main.go
```diff
@@ -233,6 +235,7 @@ func startServiceController(mgr manager.Manager, nsxClient *nsx.Client) {
 	}
 
 	if restoreMode {
+		subnetSetReconcile.EnableRestoreMode()
```
heypnus, line 238: In the other reconciler, we set `restoreMode = true` in the `RestoreReconcile` itself. Could you help explain why the subnetset is different?

yanjunz97: Sure, it's because in SubnetSet GC, there is a step will update status of SubnetSet for scaling down case. It will delete the SubnetInfo in subnetset.status.subnets if the corresponding NSX Subnet is not in store, while in restore mode we would like to keep this SubnetInfo as source of truth to restore the NSX Subnet.

https://github.com/vmware-tanzu/nsx-operator/blob/1816f9fc78583356aa38c54fb94b56f1a80d3fec/pkg/controllers/subnetset/subnetset_controller.go#L390-L393

Thus we need a flag in GC to skip the status update when a SubnetInfo without SubnetPort exists. As GC is called before RestoreReconcile, I set the restoreMode to true in main unlike other reconciler.

dantingl:
For test case, do you think we should test multi-subnet case for subnetset

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -332,9 +333,40 @@ func getExistingConditionOfType(conditionType v1alpha1.ConditionType, existingCo
...
 
+func (r *SubnetReconciler) getRestoreList() ([]types.NamespacedName, error) {
+	restoreList := []types.NamespacedName{}
+	subnetList := &v1alpha1.SubnetList{}
+	if err := r.Client.List(context.TODO(), subnetList); err != nil {
+		return restoreList, err
+	}
+	for _, subnet := range subnetList.Items {
+		// Restore a Subnet if it has NetworkAddresses
+		// Not filter the Subnet in cache as it might be updated
```
dantingl, line 362: Can you help explain more why we do not compare with NSX subnet in store?

yanjunz97: I'm thinking about the case as the following:
Create a Subnet -> Backup -> Update the Subnet -> Create a SubnetPort on the Subnet -> Restore
In such a case, if we do not reconcile the Subnet in store, we will create the SubnetPort before updating the Subnet.

Currently Subnet can switch between dhcpServer and dhcpRelay, which may have impact on how the SubnetPort being created. Though it's not a common usage, I'm thinking about if in the future some other updates are supported for Subnet, we may want to ensure the update happens before SubnetPort creation, i.e. reconcile Subnet to check if it is updated in restore process.

yanjunz97:
> For test case, do you think we should test multi-subnet case for subnetset

Sure, tested the case and updated the testing done part, thanks

dantingl (dismissed):

pkg/nsx/services/subnet/subnet.go
```diff
@@ -87,9 +87,47 @@ func InitializeSubnetService(service common.Service) (*SubnetService, error) {
 	return subnetService, nil
 }
 
+func (service *SubnetService) RestoreSubnetSet(obj *v1alpha1.SubnetSet, vpcInfo common.VPCResourceInfo, tags []model.Tag) error {
+	nsxSubnets := service.SubnetStore.GetByIndex(common.TagScopeSubnetSetCRUID, string(obj.UID))
+	var errList []error
+	for _, subnetInfo := range obj.Status.Subnets {
+		nsxSubnet, err := service.buildSubnet(obj, tags, subnetInfo.NetworkAddresses)
+		if err != nil {
+			log.Error(err, "Failed to build Subnet")
```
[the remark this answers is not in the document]
yanjunz97: Updated, thanks!

cmd/main.go
yanjunz97: Considering NSX Operator will restart after the restore process ends, I think it might be better to remove the `restoreMode=false` in other reconcilers as there is no use case. Thanks for the reminder!

heypnus (approved):

dantingl (approved):

PR #1087, opened 2025-05-08
Add basic validations for IPAddressAllocation CRD

heypnus:
Testing done:
```
Case 1. create CR with allocationSize=0, the K8s reports error : ... spec.allocationSize in body should be greater than or equal to 1
Case 2. create CR with allocationSize=1 then try to modify it to 2, the K8s reports error : ... Value is immutable
Case 3. create CR without ipAddressBlockVisibility, then apply a new configuration with ipAddressBlockVisibility=Private, the apply operation can be done.
Case 4. create CR with ipAddressBlockVisibility=Private, then try to modify it to External, the K8s reports error: spec.ipAddressBlockVisibility: Invalid value: "string": Value is immutable
```


zhengxiexie (approved):
LGTM

dantingl (approved):

PR #1095, opened 2025-05-22
Fix the issue that image fetcher subnet port can't be deleted

zhengxiexie:
The previous `ListSubnetPortByName` function only includes subnet ports for VMs, but it does not include subnet ports with the `TagScopeNamespace` tag, image fetcher subnet port uses this tag; this issue can prevent the successful deletion of the VPC.


yanjunz97 (dismissed):
LGTM

pkg/nsx/services/subnetport/subnetport.go
```diff
@@ -332,10 +332,13 @@ func (service *SubnetPortService) ListSubnetPortIDsFromCRs(ctx context.Context)
 
 func (service *SubnetPortService) ListSubnetPortByName(ns string, name string) []*model.VpcSubnetPort {
```
wenyingd, line 333: Could you add unit test for the change?

yanjunz97 (approved):

wenyingd (approved):

PR #1096, opened 2025-05-23
Add validation on bindingmap for pre-created Subnet

yanjunz97:
After introducing the pre-created Subnet, we need 2 more validations on the SubnetBindingMap
1. Parent Subnet cannot be pre-created Subnet
2. Parent Subnets and child subnet should be in the same VPC

Testing done:
- Case 1: Create a BindingMap with pre-created Subnet as child subnet and set a SubnetSet as parent. The BindingMap is created as expected.
```
metadata:
  name: binding1
  namespace: ns-1
spec:
  subnetName: subnet-precreated-1
  targetSubnetSetName: vm-default
  vlanTrafficTag: 201
```
- Case 2: Create a BindingMap with pre-created Subnet as child subnet and set a Subnet as parent. The BindingMap is created as expected.
```
metadata:
  name: binding2
  namespace: ns-1
spec:
  subnetName: subnet-precreated-1
  targetSubnetName: subnet-test-1
  vlanTrafficTag: 201
```
- Case 3: Create a BindingMap with pre-created Subnet as parent. The CR reports error `Target Subnet ns-1/subnet-precreated-1 is a pre-created Subnet`
```
metadata:
  name: binding3
  namespace: ns-1
spec:
  subnetName: subnet-test-1
  targetSubnetName: subnet-precreated-1
  vlanTrafficTag: 201
```
- Case 4: Create a BindingMap, the child pre-created Subnet has vpcName as `:kube-system_857f4b84-894d-42ac-a9ec-34d1c8f6a928` and parent Subnet has vpcName as `:ns-1_9c32dcad-dd4c-4481-a351-e462e9eb40e6`. The CR reports error 
```
Subnet /orgs/default/projects/project-quality/vpcs/kube-system_a9e5eaf4-6daa-4bb6-84a4-a6d852888dd3/subnets/subnet-precreated-2
and target Subnet /orgs/default/projects/project-quality/vpcs/ns-1_ad49141b-d3ce-48cd-9e03-c14ce2b28128/subnets/subnet-test-1_b2d0077e-c5d3-402e-8eb4-44d87280348d
are in different VPCs
```
```
metadata:
  name: binding4
  namespace: ns-1
spec:
  subnetName: subnet-precreated-2
  targetSubnetName: subnet-test-1
  vlanTrafficTag: 201
```

### DIFF
--- a/pkg/controllers/subnetbinding/subnetbinding_controller.go
+++ b/pkg/controllers/subnetbinding/subnetbinding_controller.go
@@ -6,7 +6,6 @@ import (
 	"reflect"
 	"time"
 
-	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -102,7 +101,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	// Create or update SubnetConnectionBindingMap
 	r.StatusUpdater.IncreaseUpdateTotal()
-	childSubnet, parentSubnets, err := r.validateDependency(ctx, bindingMapCR)
+	childSubnetPath, parentSubnetPaths, err := r.validateDependency(ctx, bindingMapCR)
 	if err != nil {
 		// Update SubnetConnectionBindingMap with not-ready condition
 		r.StatusUpdater.UpdateFail(ctx, bindingMapCR, err, "dependent Subnets are not ready", updateBindingMapStatusWithUnreadyCondition, "DependencyNotReady", err.message)
@@ -113,7 +112,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return common.ResultRequeueAfter60sec, nil
 	}
 
-	if err := r.SubnetBindingService.CreateOrUpdateSubnetConnectionBindingMap(bindingMapCR, childSubnet, parentSubnets); err != nil {
+	if err := r.SubnetBindingService.CreateOrUpdateSubnetConnectionBindingMap(bindingMapCR, childSubnetPath, parentSubnetPaths); err != nil {
 		// Update SubnetConnectionBindingMap with not-ready condition
 		r.StatusUpdater.UpdateFail(ctx, bindingMapCR, err, "failure to configure SubnetConnectionBindingMaps on NSX", updateBindingMapStatusWithUnreadyCondition, "ConfigureFailed", fmt.Sprintf("Failed to realize SubnetConnectionBindingMap %s on NSX", req.Name))
 		return common.ResultRequeue, nil
@@ -203,41 +202,84 @@ func (r *Reconciler) listBindingMapIDsFromCRs(ctx context.Context) (sets.Set[str
 	return bmIDs, nil
 }
 
+func getVpcPath(subnetPath string) (string, *errorWithRetry) {
+	info, err := servicecommon.ParseVPCResourcePath(subnetPath)
+	if err != nil {
+		return "", &errorWithRetry{
+			message: fmt.Sprintf("Invalid Subnet path %s", subnetPath),
+			retry:   false,
+			error:   fmt.Errorf("failed to parse Subnet path %s", subnetPath),
+		}
+	}
+	return info.GetVPCPath(), nil
+}
+
 // validateDependency validates the following conditions:
 //  1. the dependent Subnet/SubnetSet is not realized. In this case, a not-retry error is returned, and the
 //     Subnet/SubnetSet readiness update will actively trigger a requeue event
 //  2. the associated Subnet is already used as a target Subnet in another SubnetConnectionBindingMap CR, or the target
 //     Subnet already has associated SubnetConnectionBindingMap CR. In this case, a retry error is returned.
-func (r *Reconciler) validateDependency(ctx context.Context, bindingMap *v1alpha1.SubnetConnectionBindingMap) (*model.VpcSubnet, []*model.VpcSubnet, *errorWithRetry) {
-	childSubnets, err := r.validateVpcSubnetsBySubnetCR(ctx, bindingMap.Namespace, bindingMap.Spec.SubnetName, false)
+//  3. the target Subnet is a pre-created Subnet. In this case, a not-retry error is returned.
+//  4. the associated Subnet is a pre-created Subnet in a VPC different from the target Subnet Namespace VPC
+//     In this case, not-retry error is returned.
+func (r *Reconciler) validateDependency(ctx context.Context, bindingMap *v1alpha1.SubnetConnectionBindingMap) (string, []string, *errorWithRetry) {
+	childSubnetPaths, childSubnetCR, err := r.validateVpcSubnetsBySubnetCR(ctx, bindingMap.Namespace, bindingMap.Spec.SubnetName, false)
 	if err != nil {
-		return nil, nil, err
+		return "", nil, err
 	}
-	childSubnet := childSubnets[0]
+	childSubnetPath := childSubnetPaths[0]
 
+	var parentSubnetPaths []string
 	if bindingMap.Spec.TargetSubnetName != "" {
-		parentSubnets, err := r.validateVpcSubnetsBySubnetCR(ctx, bindingMap.Namespace, bindingMap.Spec.TargetSubnetName, true)
+		var parentSubnetCR *v1alpha1.Subnet
+		parentSubnetPaths, parentSubnetCR, err = r.validateVpcSubnetsBySubnetCR(ctx, bindingMap.Namespace, bindingMap.Spec.TargetSubnetName, true)
 		if err != nil {
-			return nil, nil, err
+			return "", nil, err
 		}
-		return childSubnet, parentSubnets, nil
+		// Check if the target Subnet is pre-created Subnet
+		if _, ok := parentSubnetCR.GetAnnotations()[servicecommon.AnnotationAssociatedResource]; ok {
+			return "", nil, &errorWithRetry{
+				message: fmt.Sprintf("Target Subnet %s/%s is a pre-created Subnet", bindingMap.Namespace, bindingMap.Spec.TargetSubnetName),
+				error:   fmt.Errorf("pre-created Subnet %s/%s cannot be a target Subnet", bindingMap.Namespace, bindingMap.Spec.TargetSubnetName),
+				retry:   false,
+			}
+		}
+	} else {
+		parentSubnetPaths, err = r.validateVpcSubnetsBySubnetSetCR(ctx, bindingMap.Namespace, bindingMap.Spec.TargetSubnetSetName)
+		if err != nil {
+			return "", nil, err
+		}
 	}
 
-	parentSubnets, err := r.validateVpcSubnetsBySubnetSetCR(ctx, bindingMap.Namespace, bindingMap.Spec.TargetSubnetSetName)
-	if err != nil {
-		return nil, nil, err
+	// If child Subnet is a pre-created Subnet, check if it is in the same vpc as parent Subnet
+	if _, ok := childSubnetCR.GetAnnotations()[servicecommon.AnnotationAssociatedResource]; ok {
+		childVpcPath, err := getVpcPath(childSubnetPath)
+		if err != nil {
+			return "", nil, err
+		}
+		parentVpcPath, err := getVpcPath(parentSubnetPaths[0])
+		if err != nil {
+			return "", nil, err
+		}
+		if childVpcPath != parentVpcPath {
+			return "", nil, &errorWithRetry{
+				message: fmt.Sprintf("Subnet %s and target Subnet %s are in different VPCs", childSubnetPath, parentSubnetPaths[0]),
+				retry:   false,
+				error:   fmt.Errorf("Subnet and target Subnet are in different VPCs"),
+			}
+		}
 	}
-	return childSubnet, parentSubnets, nil
+	return childSubnetPath, parentSubnetPaths, nil
 }
 
-func (r *Reconciler) validateVpcSubnetsBySubnetCR(ctx context.Context, namespace, name string, isTarget bool) ([]*model.VpcSubnet, *errorWithRetry) {
+func (r *Reconciler) validateVpcSubnetsBySubnetCR(ctx context.Context, namespace, name string, isTarget bool) ([]string, *v1alpha1.Subnet, *errorWithRetry) {
 	subnetCR := &v1alpha1.Subnet{}
 	subnetKey := types.NamespacedName{Namespace: namespace, Name: name}
 	// Check the Subnet CR existence.
 	err := r.Client.Get(ctx, subnetKey, subnetCR)
 	if err != nil {
 		log.Error(err, "Failed to get Subnet CR", "Subnet", subnetKey.String())
-		return nil, &errorWithRetry{
+		return nil, subnetCR, &errorWithRetry{
 			message: fmt.Sprintf("Unable to get Subnet CR %s", name),
 			retry:   false,
 			error:   fmt.Errorf("failed to get Subnet %s in Namespace %s with error: %v", name, namespace, err),
@@ -245,10 +287,44 @@ func (r *Reconciler) validateVpcSubnetsBySubnetCR(ctx context.Context, namespace
 	}
 
 	// Check the Subnet CR realization.
-	subnets := r.SubnetService.ListSubnetCreatedBySubnet(string(subnetCR.UID))
-	if len(subnets) == 0 {
-		log.Info("NSX VpcSubnets by subnet CR do not exist", "Subnet", subnetKey.String())
-		return nil, &errorWithRetry{
+	var subnetPaths []string
+	if anno, ok := subnetCR.GetAnnotations()[servicecommon.AnnotationAssociatedResource]; ok {
+		realized := false
+		for _, con := range subnetCR.Status.Conditions {
+			if con.Type == v1alpha1.Ready && con.Status == corev1.ConditionTrue {
+				realized = true
+				break
+			}
+		}
+		if !realized {
+			return nil, subnetCR, &errorWithRetry{
+				message: fmt.Sprintf("Subnet CR %s is not realized on NSX", name),
+				retry:   false,
+				error:   err,
+			}
+		}
+		path, err := common.GetSubnetPathFromAssociatedResource(anno)
+		if err != nil {
+			// No need to retry as not support associated resource annotation
+			// changing after Subnet creation.
+			log.Error(err, "Failed to get NSX Subnet path for shared Subnet", "Subnet", subnetKey.String())
+			return nil, subnetCR, &errorWithRetry{
+				message: fmt.Sprintf("Failed to get NSX Subnet path for shared Subnet %s", name),
+				retry:   false,
+				error:   err,
+			}
+		}
+		subnetPaths = append(subnetPaths, path)
+	} else {
+		subnets := r.SubnetService.ListSubnetCreatedBySubnet(string(subnetCR.UID))
+		for _, subnet := range subnets {
+			subnetPaths = append(subnetPaths, *subnet.Path)
+		}
+	}
+
+	if len(subnetPaths) == 0 {
+		log.Info("NSX VpcSubnets by Subnet CR do not exist", "Subnet", subnetKey.String())
+		return nil, subnetCR, &errorWithRetry{
 			message: fmt.Sprintf("Subnet CR %s is not realized on NSX", name),
 			retry:   false,
 			error:   fmt.Errorf("not found NSX VpcSubnets created by Subnet CR '%s/%s'", namespace, name),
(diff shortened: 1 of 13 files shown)


pkg/nsx/services/vpc/vpc.go
```diff
@@ -50,9 +51,9 @@ type VPCService struct {
 	LbsStore *LBSStore
 }
 
-func (s *VPCService) GetDefaultNetworkConfig() (*v1alpha1.VPCNetworkConfiguration, error) {
+func GetDefaultNetworkConfig(client client.Client) (*v1alpha1.VPCNetworkConfiguration, error) {
 	vpcNetworkConfigList := &v1alpha1.VPCNetworkConfigurationList{}
-	err := s.Client.List(context.Background(), vpcNetworkConfigList)
+	err := client.List(context.Background(), vpcNetworkConfigList)
```
wenyingd, line 56: Is this changed for test?

pkg/nsx/services/common/builder.go
```diff
@@ -54,3 +55,13 @@ func ParseVPCResourcePath(nsxResourcePath string) (VPCResourceInfo, error) {
 	info.ParentID = layers[size-3]
 	return info, nil
 }
+
+// parse org id and project id from nsxProject path
+// example /orgs/default/projects/nsx_operator_e2e_test
+func NSXProjectPathToId(path string) (string, string, error) {
+	parts := strings.Split(path, "/")
+	if len(parts) < 4 {
```
wenyingd, line 63: I guess "/orgs/default/projects" can pass this validation, but it is supposed not valid?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -208,36 +209,94 @@ func (r *Reconciler) listBindingMapIDsFromCRs(ctx context.Context) (sets.Set[str
 //     Subnet/SubnetSet readiness update will actively trigger a requeue event
 //  2. the associated Subnet is already used as a target Subnet in another SubnetConnectionBindingMap CR, or the target
 //     Subnet already has associated SubnetConnectionBindingMap CR. In this case, a retry error is returned.
+//  3. the target Subnet is a pre-created Subnet. In this case, a not-retry error is returned.
+//  4. the target Subnet and associated Subnet are not in the same vpc. In this case, not-retry error is returned.
```
wenyingd, line 213: Do you mean the target Subnet CR and associated Subnet CR are located in the same K8s Namespace, but the corresponding NSX VpcSubnets arr in the different VPCs?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -208,36 +209,94 @@ func (r *Reconciler) listBindingMapIDsFromCRs(ctx context.Context) (sets.Set[str
...
 	}
 	childSubnet := childSubnets[0]
 
 	if bindingMap.Spec.TargetSubnetName != "" {
-		parentSubnets, err := r.validateVpcSubnetsBySubnetCR(ctx, bindingMap.Namespace, bindingMap.Spec.TargetSubnetName, true)
+		parentSubnets, parentSubnetCR, err := r.validateVpcSubnetsBySubnetCR(ctx, bindingMap.Namespace, bindingMap.Spec.TargetSubnetName, true)
 		if err != nil {
 			return nil, nil, err
 		}
+		// check vpcName of child/parent Subnet
```
wenyingd, line 226: Is it possible the vpcName are the same, but vpc paths are different, e.g., NSX orgs are different?
Why not directly compare the two VPC path since we already find the child/parent NSX VpcSubnet?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -270,15 +329,26 @@ func (r *Reconciler) validateVpcSubnetsBySubnetCR(ctx context.Context, namespace
...
 				error:   fmt.Errorf("target Subnet %s is already associated by SubnetConnectionBindingMap %s", name, dependency),
 				retry:   true,
 			}
 		}
 	}
 
-	return subnets, nil
+	// Check if the Subnet CR is for pre-created Subnet
+	if isTarget {
+		if subnetCR.Status.Shared {
```
wenyingd, line 342: Can we move this check to function `func (r *Reconciler) validateDependency` within the condition `if bindingMap.Spec.TargetSubnetName != ""`

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -208,36 +209,94 @@ func (r *Reconciler) listBindingMapIDsFromCRs(ctx context.Context) (sets.Set[str
...
 	if bindingMap.Spec.TargetSubnetName != "" {
-		parentSubnets, err := r.validateVpcSubnetsBySubnetCR(ctx, bindingMap.Namespace, bindingMap.Spec.TargetSubnetName, true)
+		parentSubnets, parentSubnetCR, err := r.validateVpcSubnetsBySubnetCR(ctx, bindingMap.Namespace, bindingMap.Spec.TargetSubnetName, true)
 		if err != nil {
 			return nil, nil, err
 		}
+		// check vpcName of child/parent Subnet
+		if childSubnetCR.Spec.VPCName != parentSubnetCR.Spec.VPCName {
+			return nil, nil, &errorWithRetry{
+				message: fmt.Sprintf("Parent Subnet %s vpcName %s and child Subnet %s vpcName %s are different", bindingMap.Spec.TargetSubnetName, parentSubnetCR.Spec.VPCName, bindingMap.Spec.SubnetName, childSubnetCR.Spec.VPCName),
```
wenyingd, line 229: I would prefer to remove "Parent" "Child" in the message since we don't have these words int the CRs.

You can use `fmt.Sprintf("Subnet %s vpcName %s and target Subnet %s vpcName %s are different", bindingMap.Spec.SubnetName, childSubnetCR.Spec.VPCName, bindingMap.Spec.TargetSubnetName, parentSubnetCR.Spec.VPCName)`?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -208,36 +209,94 @@ func (r *Reconciler) listBindingMapIDsFromCRs(ctx context.Context) (sets.Set[str
...
+	var parentVPCName string
+	if vpcInfo.ProjectID == defaultProject {
+		parentVPCName = fmt.Sprintf(":%s", vpcInfo.VPCID)
+	} else {
+		parentVPCName = fmt.Sprintf("%s:%s", vpcInfo.ProjectID, vpcInfo.VPCID)
+	}
+
+	if childSubnetVPCName != parentVPCName {
+		return &errorWithRetry{
+			message: fmt.Sprintf("Parent Subnet vpcName %s and child Subnet vpcName %s are different", parentVPCName, childSubnetVPCName),
```
wenyingd, line 284: ditto

pkg/nsx/services/vpc/vpc.go
yanjunz97: I was thinking about calling this function in bindingmap controller without introducing the dependency of vpc. But I feel it might be better to introduce the dependency. Removed those changes, thanks

pkg/nsx/services/common/builder.go
yanjunz97: Updated 4 to 5, thanks!

pkg/controllers/subnetbinding/subnetbinding_controller.go
yanjunz97: Good point, updated the code to use Path directly. Thanks!

pkg/controllers/subnetbinding/subnetbinding_controller.go
yanjunz97: Updated, thanks!

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -203,52 +206,105 @@ func (r *Reconciler) listBindingMapIDsFromCRs(ctx context.Context) (sets.Set[str
...
+func getVpcPath(subnetPath string) (string, *errorWithRetry) {
+	info, err := servicecommon.ParseVPCResourcePath(subnetPath)
+	if err != nil {
+		return "", &errorWithRetry{
+			message: fmt.Sprintf("Invalid Subnet path %s", subnetPath),
+			retry:   false,
+			error:   fmt.Errorf("failed to parse Subnet path %s", subnetPath),
+		}
+	}
+	return fmt.Sprintf("/orgs/%s/projects/%s/vpcs/%s", info.OrgID, info.ProjectID, info.VPCID), nil
```
wenyingd, line 218: VPCResourceInfo has the public function `GetVPCPath`, we can call it directly?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -270,15 +326,82 @@ func (r *Reconciler) validateVpcSubnetsBySubnetCR(ctx context.Context, namespace
...
 				error:   fmt.Errorf("target Subnet %s is already associated by SubnetConnectionBindingMap %s", name, dependency),
 				retry:   true,
 			}
 		}
 	}
+	return subnets, subnetCR, nil
+}
 
-	return subnets, nil
+func (r *Reconciler) getSharedSubnetFromNSX(subnet *v1alpha1.Subnet) (*model.VpcSubnet, *errorWithRetry) {
```
wenyingd, line 339: Is it better if we move this function to SubnetService? Even more, we can let the SubnetService to check if the Subnet CR is shared or not, then decide how to list the NSX VpcSubnets, e.g., from local cache or from NSX API request. I guess this logic is also callable by SubnetController when reconciling (including deleting) the CR ?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -270,15 +326,82 @@ func (r *Reconciler) validateVpcSubnetsBySubnetCR(ctx context.Context, namespace
...
+		if parts[0] == vpcInfo.ProjectID && parts[1] == vpcInfo.VPCID && parts[2] == vpcInfo.ID {
+			nsxSubnet, err := r.SubnetService.GetSubnetFromNSX(vpcInfo)
+			if err != nil {
+				return nil, &errorWithRetry{
+					message: fmt.Sprintf("Failed to get Subnet from NSX, error: %v", err),
+					error:   err,
+					retry:   true,
+				}
+			}
+			return nsxSubnet, nil
```
wenyingd, line 397: Do you mean only one Subnet in a single VPC (`nc.Spec.Subnets`) can be shared?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -270,15 +326,82 @@ func (r *Reconciler) validateVpcSubnetsBySubnetCR(ctx context.Context, namespace
...
+	}
+	// associatedResource has the format projectID:vpcID:subnetID or :vpcID:subnetID for default project
+	parts := strings.Split(associatedResource, ":")
+	if parts[0] == "" {
+		nc, err := r.VPCService.GetDefaultNetworkConfig()
+		if err != nil {
+			return nil, &errorWithRetry{
+				message: fmt.Sprintf("Failed to get default NetworkConfig, error: %v", err),
+				error:   err,
+				retry:   true,
```
wenyingd, line 357: Could you add comments to explain what error is retrieable and what is not?

pkg/nsx/services/subnet/subnet.go
```diff
@@ -459,3 +459,12 @@ func (service *SubnetService) UpdateSubnetSet(ns string, vpcSubnets []*model.Vpc
 	}
 	return nil
 }
+
+func (service *SubnetService) GetSubnetFromNSX(vpcInfo common.VPCResourceInfo) (*model.VpcSubnet, error) {
```
wenyingd, line 463:
```suggestion
func (service *SubnetService) GetSubnetFromNSX(vpcSubnetInfo common.VPCResourceInfo) (*model.VpcSubnet, error) {
```

pkg/nsx/services/common/services.go
```diff
@@ -31,6 +31,7 @@ type SubnetServiceProvider interface {
 	GenerateSubnetNSTags(obj client.Object) []model.Tag
 	ListSubnetByName(ns, name string) []*model.VpcSubnet
 	ListSubnetBySubnetSetName(ns, subnetSetName string) []*model.VpcSubnet
+	GetSubnetFromNSX(vpcInfo VPCResourceInfo) (*model.VpcSubnet, error)
```
wenyingd, line 34: ditto

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -270,15 +326,82 @@ func (r *Reconciler) validateVpcSubnetsBySubnetCR(ctx context.Context, namespace
...
 
-	return subnets, nil
+func (r *Reconciler) getSharedSubnetFromNSX(subnet *v1alpha1.Subnet) (*model.VpcSubnet, *errorWithRetry) {
+	anno := subnet.GetAnnotations()
+	associatedResource, ok := anno[servicecommon.AnnotationAssociatedResource]
+	if !ok {
+		return nil, &errorWithRetry{
+			message: fmt.Sprintf("No associated resource annotation in Shared Subnet %s/%s", subnet.Namespace, subnet.Name),
+			error:   fmt.Errorf("no associated resource annotation in Shared Subnet %s/%s", subnet.Namespace, subnet.Name),
+			retry:   false,
```
wenyingd, line 346: Is it possible to add the annotation later?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -270,15 +326,82 @@ func (r *Reconciler) validateVpcSubnetsBySubnetCR(ctx context.Context, namespace
...
+	// Get Subnet path from NetworkConfig
+	nc, err := r.VPCService.GetVPCNetworkConfigByNamespace(subnet.Namespace)
+	if err != nil {
+		return nil, &errorWithRetry{
+			message: fmt.Sprintf("Failed to get NetworkConfig for Namespace %s", subnet.Namespace),
+			error:   err,
+			retry:   true,
+		}
+	}
+	for _, subnetPath := range nc.Spec.Subnets {
```
wenyingd, line 379: I don't get why we need this for-loop. If we have get `projectID:vpcID:subnetID` from the associatedResource, can't we directly construct the NSX path of the target VpcSubnet?

pkg/controllers/subnetbinding/subnetbinding_controller.go
yanjunz97: Updated, thanks!

pkg/controllers/subnetbinding/subnetbinding_controller.go
yanjunz97: This part of code is removed. But I think the answer for this question is as we are looking for the child subnet referred by bindingmap, it is fine to return the nsx Subnet once we find one match the path.

pkg/controllers/subnetbinding/subnetbinding_controller.go
yanjunz97: Added to subnet uts as the function is moved there, thanks!

pkg/nsx/services/subnet/subnet.go
yanjunz97: This function is removed, thanks

pkg/controllers/subnetbinding/subnetbinding_controller.go
yanjunz97: No, we assume the annotation is added when the shared Subnet CR is created; and the shared Subnet CR is only created by NSX Operator.

pkg/nsx/services/subnet/subnet.go
```diff
@@ -459,3 +461,39 @@ func (service *SubnetService) UpdateSubnetSet(ns string, vpcSubnets []*model.Vpc
...
+	if !ok {
+		return nil, fmt.Errorf("no associated resource annotation in Shared Subnet %s/%s", subnet.Namespace, subnet.Name), false
+	}
+	// associatedResource has the format projectID:vpcID:subnetID or :vpcID:subnetID for default project
+	parts := strings.Split(associatedResource, ":")
+	if len(parts) != 3 {
+		return nil, fmt.Errorf("failed to parse associated resource annotation %s", associatedResource), false
+	}
+	if parts[0] == "" {
+		nc, err := service.vpcService.GetDefaultNetworkConfig()
```
lxiaopei, line 479: for ":vpcID:subnetID", it means the default NSX Project, not the namespace's Project.

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -203,52 +203,110 @@ func (r *Reconciler) listBindingMapIDsFromCRs(ctx context.Context) (sets.Set[str
...
 			return nil, nil, err
 		}
-		return childSubnet, parentSubnets, nil
 	}
 
-	parentSubnets, err := r.validateVpcSubnetsBySubnetSetCR(ctx, bindingMap.Namespace, bindingMap.Spec.TargetSubnetSetName)
-	if err != nil {
-		return nil, nil, err
+	// If child Subnet is a pre-created Subnet, check if it is in the same vpc as parent Subnet
+	if childSubnetCR.Status.Shared {
```
lxiaopei, line 256: could we just check the "nsx.vmware.com/associated_resource" annotation to check if the parent and child Subnet are in the same VPC? we do not need to get the real NSX Subnet.

pkg/nsx/services/subnet/subnet.go
yanjunz97: Updated to use NSX API call to determine the default project, thanks

pkg/nsx/services/subnet/subnet.go
```diff
@@ -459,3 +461,39 @@ func (service *SubnetService) UpdateSubnetSet(ns string, vpcSubnets []*model.Vpc
...
+	vpcInfoList := service.vpcService.ListVPCInfo(subnet.Namespace)
+	if len(vpcInfoList) == 0 {
+		return nil, fmt.Errorf("failed to get VPC Info for Namespace %s", subnet.Namespace), true
+	}
+	nsxSubnet, err := service.NSXClient.SubnetsClient.Get(vpcInfoList[0].OrgID, parts[0], parts[1], parts[2])
+	if err != nil {
+		err = nsxutil.TransNSXApiError(err)
+		return nil, err, true
+	}
+	return &nsxSubnet, nil, false
```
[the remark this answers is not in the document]
yanjunz97: Checked the Subnet CR condition for the realization state, thanks!

pkg/controllers/subnetbinding/subnetbinding_controller.go
yanjunz97: We always need to get child subnet path for bindingmap creation. But as only the path is required, we can save the API call to get the real NSX Subnet by changing the return value of those bindingmap functions. I updated the code, please take another look, thanks!

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -203,52 +202,127 @@ func (r *Reconciler) listBindingMapIDsFromCRs(ctx context.Context) (sets.Set[str
...
+		for _, con := range subnetCR.Status.Conditions {
+			if con.Type == v1alpha1.Ready && con.Status == corev1.ConditionTrue {
+				realized = true
+				break
+			}
+		}
+		if !realized {
+			return nil, subnetCR, &errorWithRetry{
+				message: fmt.Sprintf("Subnet CR %s is not realized on NSX", name),
+				retry:   false,
```
lxiaopei, line 302: shall we retry if the pre-created subnet is not realized?

yanjunz97: Subnetbindingmap has Subnet handler to requeue the Subnet if its readiness condition changes, so we do not need to retry here.

pkg/controllers/subnetbinding/subnetbinding_controller.go
lxiaopei: make sense.

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -203,52 +202,127 @@ func (r *Reconciler) listBindingMapIDsFromCRs(ctx context.Context) (sets.Set[str
...
+			return "", nil, err
 		}
-		return childSubnet, parentSubnets, nil
 	}
 
-	parentSubnets, err := r.validateVpcSubnetsBySubnetSetCR(ctx, bindingMap.Namespace, bindingMap.Spec.TargetSubnetSetName)
-	if err != nil {
-		return nil, nil, err
+	// If child Subnet is a pre-created Subnet, check if it is in the same vpc as parent Subnet
+	if childSubnetCR.Status.Shared {
```
lxiaopei, line 255: same here, use associated_resource annotation.

pkg/controllers/subnetbinding/subnetbinding_controller.go
lxiaopei: good to know.

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -203,52 +202,127 @@ func (r *Reconciler) listBindingMapIDsFromCRs(ctx context.Context) (sets.Set[str
...
+				retry:   false,
+				error:   err,
+			}
+		}
+		path, err := r.SubnetService.GetSharedSubnetPath(subnetCR)
+		if err != nil {
+			log.Info("Failed to get NSX Subnet path for shared Subnet", "Subnet", subnetKey.String(), "error", err)
+			return nil, subnetCR, &errorWithRetry{
+				message: fmt.Sprintf("Failed to get NSX Subnet path for shared Subnet %s", name),
+				retry:   true,
```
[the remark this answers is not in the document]
lxiaopei: I think if shared subnet does not exist(actually this will be check in WCP so it is unexpected case), we do not need requeue.

pkg/nsx/services/subnet/subnet.go
```diff
@@ -459,3 +461,22 @@ func (service *SubnetService) UpdateSubnetSet(ns string, vpcSubnets []*model.Vpc
...
+	associatedResource, ok := anno[common.AnnotationAssociatedResource]
+	if !ok {
+		return "", fmt.Errorf("no associated resource annotation in Shared Subnet %s/%s", subnet.Namespace, subnet.Name)
+	}
+	// associatedResource has the format projectID:vpcID:subnetID
+	parts := strings.Split(associatedResource, ":")
+	if len(parts) != 3 {
+		return "", fmt.Errorf("failed to parse associated resource annotation %s", associatedResource)
+	}
+	vpcInfoList := service.vpcService.ListVPCInfo(subnet.Namespace)
```
lxiaopei, line 477: here we use this just to get the orgID, which we only supports "default" here, I feel we can just use "default"

pkg/nsx/services/subnet/subnet.go
```diff
@@ -459,3 +461,22 @@ func (service *SubnetService) UpdateSubnetSet(ns string, vpcSubnets []*model.Vpc
 	}
 	return nil
 }
+
+// GetSharedSubnetPath generates Subnet Path for shared Subnet
+func (service *SubnetService) GetSharedSubnetPath(subnet *v1alpha1.Subnet) (string, error) {
+	anno := subnet.GetAnnotations()
+	associatedResource, ok := anno[common.AnnotationAssociatedResource]
```
lxiaopei, line 468: the annotation was checked before "GetSharedSubnetPath" func is invoked. I feel we only need to pass the associatedResource annotation in this func.

pkg/controllers/subnetbinding/subnetbinding_controller.go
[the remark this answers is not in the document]
yanjunz97: As discussed offline, this part of code returns error for getting subnet path. Changed the retry to false as we removed the getvpcinfo part. Thanks

pkg/nsx/services/subnet/subnet.go
yanjunz97: Updated, thanks!

pkg/nsx/services/subnet/subnet.go
yanjunz97: Updated, thanks!

lxiaopei (approved):
LGTM

wenyingd (approved):
LGTM

PR #1109, opened 2025-06-10
Enhancing Log Level Clarity for Improved Operational Visibility

zhengxiexie:
# Enhancing Log Level Clarity for Improved Operational Visibility

Error, because it is an unexpected failure.
image
WARNING, since it doesn't prevent normal operation.
image

Our current logging implementation using "github.com/go-logr/logr" and "github.com/go-logr/zapr" lacks a critical WARNING level, resulting in misleading error categorization. Many expected operational scenarios, particularly retryable reconciliation events, are incorrectly flagged as ERROR logs, creating unnecessary alarm and confusion.

This PR introduces a strategic solution by repurposing log.V(1).Info as a WARNING level with distinctive orange-red highlighting that maintains visual prominence while properly distinguishing it from true ERROR conditions (bright red). This approach aligns with industry best practices where:

- ERROR logs should exclusively indicate unexpected system failures requiring immediate attention
- WARNING logs should highlight potential issues that don't prevent normal operation but may require monitoring

This implementation will significantly improve operational clarity and reduce alert fatigue by properly categorizing retryable events as warnings rather than errors.

Next steps involve systematically reclassifying appropriate ERROR level messages to log.V(1).INFO to complete this logging enhancement.


yanjunz97:
I feel usually the log level is like Error < Warning < Info < Debug. It might be strange that if I set the log level to 0, I only get the Error and Info log but not get the Warning level log. From user perspective, I feel it's better use log.Info as warning and log.V(1).Info as Info, though from the developer perspective log.Info does not mean Info level is also a little bit strange. Would you think it's necessary for us to have our own log type implementing log.Error,  log.Warning,  log.Info, log.Debug, log.V(int).Info?

pkg/nsx/services/inventory/build.go
```diff
@@ -247,7 +247,7 @@ func normalize(name string, maxLength int) string {
 	hashId := sha1.Sum([]byte(name))
 	nameLength := maxLength - 9
 	newname := fmt.Sprintf("%s-%s", name[:nameLength], hashId[:8])
-	log.Info("Name exceeds max length of supported by NSX. Truncate name to newname",
+	log.V(2).Info("Name exceeds max length of supported by NSX. Truncate name to newname",
 		"maxLength", maxLength, "name", name, "newname", newname)
```
TaoZou1, line 251: how about keep using log.Info?

pkg/logger/logger.go
```diff
@@ -26,6 +33,23 @@ var (
 	customCallerEncoder = func(caller zapcore.EntryCaller, enc zapcore.PrimitiveArrayEncoder) {
 		enc.AppendString(colorYellow + caller.TrimmedPath() + colorReset)
 	}
+	customLevelEncoder = func(level zapcore.Level, enc zapcore.PrimitiveArrayEncoder) {
+		if level == zapcore.InfoLevel-1 {
+			enc.AppendString(colorOrange + "WARNING" + colorReset)
+		} else if level == zapcore.InfoLevel-2 {
+			enc.AppendString(colorPurple + "DEBUG" + colorReset)
+		} else if level < zapcore.InfoLevel-2 {
```
TaoZou1, line 41: do we need to support loglevel < zapcore.InfoLevel-2

pkg/logger/logger.go
```diff
@@ -60,19 +117,29 @@ func ZapLogger(cfDebug bool, cfLogLevel int) logr.Logger {
...
+		Core: baseCore,
+		// Filter recorder/recorder.go log
+		filter: filterRecorderLogs,
+	}
+
+	zapLogger := zap.New(filteredCore, zap.AddCaller(), zap.AddCallerSkip(0))
+
+	defer func(zapLogger *zap.Logger) {
+		_ = zapLogger.Sync()
+	}(zapLogger)
```
TaoZou1, line 143: the same as "defer zapLogger.Sync()"?

pkg/logger/logger.go
```diff
@@ -36,8 +60,41 @@ func InitLog(log *logr.Logger) {
...
+	if strings.Contains(entry.Caller.File, "recorder/recorder.go") {
+		return true
+	}
+	return false
+}
+
+// If debug set in configmap, set the log level to 1.
+// If loglevel set in the command line and greater than the debug log level, set it to command line level.
 func getLogLevel(cfDebug bool, cfLogLevel int) int {
 	logLevel := 0
```
TaoZou1, line 99: when log level was info, you need to show Warning level log.

pkg/controllers/securitypolicy/namespace_handler.go
```diff
@@ -27,15 +27,15 @@ type EnqueueRequestForNamespace struct {
 }
 
 func (e *EnqueueRequestForNamespace) Create(_ context.Context, _ event.CreateEvent, _ workqueue.TypedRateLimitingInterface[reconcile.Request]) {
-	log.V(1).Info("NameSpace create event, do nothing")
+	log.Info("NameSpace create event, do nothing")
```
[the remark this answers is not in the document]
zhengxiexie: fixed

pkg/nsx/services/inventory/build.go
```diff
@@ -86,7 +86,7 @@ func (s *InventoryService) BuildPod(pod *corev1.Pod) (retry bool) {
 	} else if len(pod.Status.PodIPs) == 2 {
 		ips = pod.Status.PodIPs[0].IP + "," + pod.Status.PodIPs[1].IP
 	} else {
-		log.Info("Unexpected Pod IPs found", "Pod ips", pod.Status.PodIPs)
+		log.V(2).Info("Unexpected Pod IPs found", "Pod ips", pod.Status.PodIPs)
 	}
```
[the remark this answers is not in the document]
zhengxiexie: fixed

pkg/logger/logger.go
zhengxiexie: refer latest version

pkg/logger/logger.go
zhengxiexie: It is to avoid IDE warning non error handling

pkg/logger/logger.go
zhengxiexie: refer latest version

pkg/logger/logger.go
```diff
@@ -15,43 +17,156 @@ const (
...
+		entry.Message = strings.TrimPrefix(entry.Message, "+debug")
+		// Change the log level to Debug
+		entry.Level = zapcore.DebugLevel
 	}
-	return realLogLevel
+	return c.Core.Check(entry, ce)
+}
+
+func (c *FilteredCore) Write(entry zapcore.Entry, fields []zapcore.Field) error {
+	if c.filter(entry) {
```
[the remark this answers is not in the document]
zhengxiexie: fixed

pkg/logger/logger.go
```diff
@@ -15,43 +17,156 @@ const (
...
+	if strings.Contains(entry.Caller.File, "recorder/recorder.go") {
+		return true
+	}
+	return false
 }
 
 func ZapLogger(cfDebug bool, cfLogLevel int) logr.Logger {
-	logLevel := getLogLevel(cfDebug, cfLogLevel)
+	logLevel := cfLogLevel
+	if cfDebug {
```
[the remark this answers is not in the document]
zhengxiexie: Let's be simple, if set to Debug, use 0 by priority; otherwise, use the command argument.

pkg/logger/logger.go
```diff
@@ -15,43 +17,156 @@ const (
...
+}
+
+func (c *FilteredCore) Write(entry zapcore.Entry, fields []zapcore.Field) error {
+	if c.filter(entry) {
+		return nil
+	}
+
+	// Check if the message has warned prefix
+	if strings.HasPrefix(entry.Message, "+warn") {
+		// Remove the prefix from the message
```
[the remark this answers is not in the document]
zhengxiexie: fixed

pkg/logger/logger.go
```diff
@@ -13,45 +15,132 @@ import (
...
+// Info logs an info message with the given key-value pairs
+func (l *CustomLogger) Info(msg string, keysAndValues ...interface{}) {
+	l.logger.Info(msg, keysAndValues...)
+}
+
+// Debug logs a debug message with the given key-value pairs
+func (l *CustomLogger) Debug(msg string, keysAndValues ...interface{}) {
+	// Don't know why l.logger.V(-1).Info is not working as Debug, hack it
+	l.logger.V(-1).Info(debugPrefix+msg, keysAndValues...)
+}
```
yanjunz97, line 57: I feel a little bit confused about the verbose level here. Based on the godoc, it seems Warn/Info/Debug actually calls the same log level and the only difference is whether we add warnPrefix/debugPrefix. Is this intended? If so why we still add V(x) for Warn/Debug?

```
func (l logr.Logger) V(level int) logr.Logger
V returns a new Logger instance for a specific verbosity level, relative to this Logger. In other words, V-levels are additive. A higher verbosity level means a log message is less important. Negative V-levels are treated as 0.
```

I was thinking about using
log.Info for warning
log.Info.V(1) for Info
log.Info.V(2) for debug

yanjunz97:
Hi @zhengxiexie you can refer to my draft at https://github.com/yanjunz97/nsx-operator/tree/relog-draft

With which provides the following mapping
From user perspective, it will be
-1: warning + error
0: info + warning + error
1: debug + info + warning + err
2: debug + info + warning + err
3+: V(level) + debug + info + warning + err

Screenshot 2025-07-18 at 1 21 54 PM

But there are 2 issues in this implementation:
- I use log.V(2) to implement log.Debug(). If we use level 0 for info level then we have to make both level 1 and 2 for debug level, because only for level 3 and above we can distinguish it from debug log.
- The third party, like controller runtime, may use logf as log. Based on our zapcore color format, info log from controller runtime will be marked as WARN level.

The missing of -1 level for warning in logf makes the mapping always not perfect. Please feel free to investigate any other implementation.

yanjunz97:
I just notice your hack with prefix should be able to mock the level -1...
Please take another look at https://github.com/yanjunz97/nsx-operator/tree/relog-draft-2

This should be better with the following mapping from user perspective:

-1: warning + error
0: info + warning + error
1: debug + info + warning + error
2+: V(level) + debug + info + warning + error

Screenshot 2025-07-18 at 2 12 05 PM

PR #1117, opened 2025-06-24
Fix invalid port issue while creating ServiceEntry

heypnus:
The NSX doesn't support the port range format likes "6000-6000" in the NSX security policy, it reports the following error:

> Invalid port provided while creating/updating Service / ServiceEntry

This patch will change "6000-6000" to "6000" to avoid this issue.

Testing done:
create the following network policy and the NSX security policy can be created:
```
kind: NetworkPolicy
metadata:
  annotations:
    ncp/log_traffic: ALL
  name: np-ingress-allow-multiselector-namedport-endport
spec:
  ingress:
  - from:
    - podSelector:
        matchLabels:
          app: tcp
    ports:
    - endPort: 6000
      port: 6000
      protocol: TCP
    - endPort: 6010
      port: 6001
      protocol: UDP
  podSelector:
    matchLabels:
      app: coffee
  policyTypes:
  - Ingress
```

### DIFF
--- a/pkg/nsx/services/securitypolicy/builder.go
+++ b/pkg/nsx/services/securitypolicy/builder.go
@@ -489,7 +489,7 @@ func buildRuleServiceEntries(port v1alpha1.SecurityPolicyPort) *data.StructValue
 	// Note: the caller ensures the given port.Port type is Int. For named port case, the caller should
 	// convert to a new SecurityPolicyPort using the correct port number.
 	// In case that the destination_port in NSX-T is 0.
-	if port.EndPort == 0 {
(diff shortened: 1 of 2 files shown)


timdengyun (approved):
LGTM

timdengyun:
Thanks Qian, do we need a UT for this case?

heypnus:
> Thanks Qian, do we need a UT for this case?

Done. But the UT won't improve the coverage... 😉

dantingl (approved):

timdengyun:
> > Thanks Qian, do we need a UT for this case?
> 
> Done. But the UT won't improve the coverage... 😉

Accumulated small changes make big improvements:)

PR #1128, opened 2025-06-27
Fix null Subnet CR in SubnetPort controller

yanjunz97:


heypnus (approved):

lxiaopei (approved):
LGTM，it might return nil if SubnetSet , right?

PR #1129, opened 2025-06-27
Fix 'unknown NSX resource type' error

TaoZou1:
While building resource tree, there was an error "unknown NSX resource type" in leafWrapper function.
Fix the issue by adding model.SharedResource in leafWrapper 
Change the log position to print the dump error


TaoZou1:
use other patch to resolve the issue

PR #1134, opened 2025-07-02
Add webhook for ipaddressallocation

TaoZou1:
Search all the service under the same namespace. If the ip has been used in the service, operator will reject the deletion of ipaddressallocation

Test Done:
1. create an ipaddressallocation with ip 192.168.0.50
2. update svc with loadBalancerIP: 192.168.0.50
3. wait for svc ready
4. delete the ipaddressallocation
5. check if return error like: Error from server (Forbidden): admission webhook
"ipaddressallocation.validating.crd.nsx.vmware.com" denied the request: cannot delete IPAddressAllocation ip2: IP 192.168.0.50 is still in use by Service milk-svc

### DIFF
--- a/build/yaml/webhook/manifests.yaml
+++ b/build/yaml/webhook/manifests.yaml
@@ -68,3 +68,24 @@ webhooks:
     resources:
     - subnets
   sideEffects: None
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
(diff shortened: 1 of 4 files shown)


yanjunz97 (approved):
LGTM

yanjunz97 (approved):

zhengxiexie (approved):

PR #1135, opened 2025-07-03
Remove node checking

TaoZou1:
While reporting pod inventory, operator will check if node's ready. But NCP didn't check it. Follow the NCP style and remove the node checking.
Rename build.go to builder.go

Test Done:
1. create a pod with status pending
2. check from UI if pod inventory reported
3. check if networkworking status with error '"0/6 nodes are available: 3 node(s) didn't match Pod's node affinity/selector, 3 node(s) had untolerated taint {node-role.kubernetes.io/control-plane:"
4. check if pod status is 'Down'


timdengyun (approved):
LGTM

PR #1138, opened 2025-07-03
[cherry-pick]Block SubnetPort allocation on an IP exhausted Subnet (#1033)

yanjunz97:
For Pod case, we will create and delete SubnetPort for image fetcher. As the IP release needs 2 minutes, it may happen that the SubnetPort number on the Subnet is less than capacity but larger than the number of available IPs. For this condition, we will get IP exhausted error when creating the SubnetPort, and we would like to avoid allocating more SubnetPorts on this IP-exhausted Subnet in the following 2 minutes.

Testing done:
Create 16 pods, delete them and re-create them. Check the log that once the Subnet is found as IP-exhausted, the following SubnetPort will be allocated on other Subnets.


dantingl (approved):

yanjunz97:
e2e test passed on local testbed, merged the PR

PR #1142, opened 2025-07-04
Fix AddressBinding webhook validation for default interface

gran-vmv:
Testing done:
```
root@422df4dbea4c38c7dd45af0238d16b69 [ ~ ]# cat ab1
kind: AddressBinding
metadata:
  name: gran1
  namespace: default
spec:
  vmName: vm1
---
kind: AddressBinding
metadata:
  name: gran2
  namespace: default
spec:
  vmName: vm1
  interfaceName: inf1

root@422df4dbea4c38c7dd45af0238d16b69 [ ~ ]# kubectl create -f ab1
addressbinding.crd.nsx.vmware.com/gran1 created
Error from server (Forbidden): error when creating "ab1": admission webhook "addressbinding.validating.crd.nsx.vmware.com" denied the request: interface already has AddressBinding
root@422df4dbea4c38c7dd45af0238d16b69 [ ~ ]# kubectl delete -f ab1
addressbinding.crd.nsx.vmware.com "gran1" deleted
Error from server (NotFound): error when deleting "ab1": addressbindings.crd.nsx.vmware.com "gran2" not found


root@422df4dbea4c38c7dd45af0238d16b69 [ ~ ]# vi ab1
root@422df4dbea4c38c7dd45af0238d16b69 [ ~ ]# cat ab1
kind: AddressBinding
metadata:
  name: gran1
  namespace: default
spec:
  vmName: vm1
  interfaceName: inf1
---
kind: AddressBinding
metadata:
  name: gran2
  namespace: default
spec:
  vmName: vm1

root@422df4dbea4c38c7dd45af0238d16b69 [ ~ ]# kubectl create -f ab1
addressbinding.crd.nsx.vmware.com/gran1 created
Error from server (Forbidden): error when creating "ab1": admission webhook "addressbinding.validating.crd.nsx.vmware.com" denied the request: interface already has AddressBinding

```

### DIFF
--- a/pkg/controllers/subnetport/addressbinding_webhook.go
+++ b/pkg/controllers/subnetport/addressbinding_webhook.go
@@ -46,8 +46,17 @@ func (v *AddressBindingValidator) Handle(ctx context.Context, req admission.Requ
 			log.Error(err, "failed to list AddressBinding from cache", "indexValue", abIndexValue)
 			return admission.Errored(http.StatusInternalServerError, err)
 		}
+		hasDefault := ab.Spec.InterfaceName == ""
+		if !hasDefault {
+			for _, existingAddressBinding := range existingAddressBindingList.Items {
+				if existingAddressBinding.Spec.InterfaceName == "" {
+					hasDefault = true
+					break
+				}
+			}
+		}
 		for _, existingAddressBinding := range existingAddressBindingList.Items {
-			if ab.Name != existingAddressBinding.Name && ab.Spec.InterfaceName == existingAddressBinding.Spec.InterfaceName {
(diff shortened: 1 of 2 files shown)


yanjunz97 (approved):
LGTM.
Just to check, as I know currently we do not support using default interface when the VM has multiple interfaces, will we support that in the future? If that is supported then we need to update the validation here as an ab with default interface name might be created with an ab with the second interface name?

gran-vmv:
> LGTM. Just to check, as I know currently we do not support using default interface when the VM has multiple interfaces, will we support that in the future? If that is supported then we need to update the validation here as an ab with default interface name might be created with an ab with the second interface name?

I think the answer is "No", since we cannot determine which interface is "default" when VM has multiple interfaces.

heypnus (approved):

PR #1144, opened 2025-07-07
Use id for Subnet vpcName

yanjunz97:
Change vpcName in Subnet from project-name:vpc-name to project-id:vpc:id to be consistent with VCFA.

Testing done:
- Create a shared Subnet in workload-ns-1, observed vpcName is set as `project-quality:workload-ns-1_d843cda5-e1c8-4785-b18f-feb3ae495565`
- Create an auto-created Subnet in workload-ns-1, observed vpcName is set as ` project-quality:workload-ns-1_d843cda5-e1c8-4785-b18f-feb3ae495565`

### DIFF
--- a/pkg/controllers/namespace/subnet_share_controller.go
+++ b/pkg/controllers/namespace/subnet_share_controller.go
@@ -47,7 +47,7 @@ func (r *NamespaceReconciler) createSharedSubnetCR(ctx context.Context, ns strin
 		return err
 	}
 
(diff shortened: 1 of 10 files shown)


zhengxiexie (approved):
LGTM

dantingl (commented):
Do we need to change `vpcName` in CRD definition?

yanjunz97:
> Do we need to change `vpcName` in CRD definition?

As discussed in chat, it's fine to use vpcName as vcfa also uses id as vpc name

PR #1145, opened 2025-07-07
Update svc status if has error annotation

TaoZou1:
Fix the svc network status/status not uploading issue if svc has ep

Test Done:
1. create pod tea
2. create svc with labels: app: tea which VIP has been used
3. check inventory if networkstatus is DOWN, status is DOWN, error is reported


zhengxiexie (approved):
LGTM

PR #1164, opened 2025-07-18
Use CidrList for IPBlock

yanjunz97:


yanjunz97:
Close as we need to refactor this part by supporting both cidr_list and range_list

PR #1169, opened 2025-07-22
Allow Operator to update shared subnet ip

yanjunz97:
We should only block the ipAddresses change by non-operator user,
otherwise the shared Subnet spec update will fail


lxiaopei (approved):
LGTM

PR #1175, opened 2025-07-28
Add E2E tests for NSX inventory synchronization

zhengxiexie:


test/e2e/nsx_inventory_test.go
```diff
@@ -0,0 +1,324 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestInventorySync(t *testing.T) {
```
yanjunz97, line 14: Do we need to test the ContainerCluster?

test/e2e/nsx_inventory_test.go
```diff
@@ -0,0 +1,324 @@
...
+	err := testData.createVCNamespace(ns)
+	if err != nil {
+		t.Fatalf("Failed to create VC namespace: %v", err)
+	}
+	defer func() {
+		err := testData.deleteVCNamespace(ns)
+		if err != nil {
+			t.Fatalf("Failed to delete VC namespace: %v", err)
+		}
+	}()
```
yanjunz97, line 40: Do we need this defer as we delete the namespace at line 47-50?

test/e2e/nsx_inventory_test.go
```diff
@@ -0,0 +1,324 @@
...
+	assert.NoError(t, err, "Namespace was not removed from NSX inventory")
+}
+
+// testPodSync tests that a Kubernetes pod is synced to the NSX inventory as a ContainerApplicationInstance
+func testPodSync(t *testing.T) {
+	_, deadlineCancel := context.WithTimeout(context.Background(), defaultTimeout)
+	defer deadlineCancel()
+
+	// Create a namespace with a unique name
+	ns := fmt.Sprintf("test-pod-sync-%s", getRandomString())
```
yanjunz97, line 63: Do you think we can share the same namespace for all inventory tests, at least for resources other than namespace sync?
I'm just worry about some unexpected failures may lead to nsx vpc deletion failure. And creating too many namespace in test may make it easier to reach the edge capacity

test/e2e/nsx_inventory_test.go
```diff
@@ -0,0 +1,324 @@
...
+	t.Run("testNodeSync", func(t *testing.T) { testNodeSync(t) })
+	t.Run("testNetworkPolicySync", func(t *testing.T) { testNetworkPolicySync(t) })
+	t.Run("testIngressSync", func(t *testing.T) { testIngressSync(t) })
+}
+
+// testNamespaceSync tests that a Kubernetes namespace is synced to the NSX inventory as a ContainerProject
+func testNamespaceSync(t *testing.T) {
+	_, deadlineCancel := context.WithTimeout(context.Background(), defaultTimeout)
+	defer deadlineCancel()
+
```
yanjunz97, line 27: Why we have these in each function? It seems the deadline context is never used?

test/e2e/nsx_inventory_test.go
zhengxiexie: added

test/e2e/nsx_inventory_test.go
yanjunz97: Not quite understand the relationship between parallel test and separate namespace. Basically I feel the more namespaces we create the higher possibility we may hit an error in vpc deletion failure. But overall, we can put an eye on it to see if the edge capacity will be an issue.

TaoZou1 (approved):

PR #1177, opened 2025-07-29
Deduplicate network error messages in inventory using a map

zhengxiexie:


TaoZou1:
could you add ut for duplicate network error?
could you also your test done in the description?

TaoZou1 (approved):

PR #1181, opened 2025-07-31
Set Ipv4SubnetSize default value only when IPAddresses has no value

TaoZou1:
Only set the default value of Ipv4SubnetSize when IPAddresses has no value, and when the Ipv4SubnetSize value is also missing. 
When building a subnet, only set the Ipv4SubnetSize value when the IPAddresses has no value

Test Done:
case 1:
1. create a subnet 
kind: Subnet
metadata:
  name: sub2
spec:
  accessMode: Private
  ipAddresses: ["172.26.2.32/28"]
2.  check there was no ipv4SubnetSize in the spec
case 2:
1. create a subnet with both ipAddresses and ipv4SubnetSize
kind: Subnet
metadata:
  name: sub2
spec:
  accessMode: Private
  ipAddresses: ["172.26.2.32/28"]
  ipv4SubnetSize: 28
2. check there was no ipv4SubnetSize in the spec
case 3:
1. create a subnet
kind: Subnet
metadata:
  name: sub2
spec:
  accessMode: Private
 2. check the ipv4SubnetSize = 32


yanjunz97 (approved):
LGTM

zhengxiexie (approved):
LGTM

pkg/controllers/subnet/subnet_controller.go
```diff
@@ -155,7 +155,7 @@ func (r *SubnetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		specChanged = true
 	}
 
-	if subnetCR.Spec.IPv4SubnetSize == 0 {
+	if subnetCR.Spec.IPv4SubnetSize == 0 && len(subnetCR.Spec.IPAddresses) == 0 {
```
dantingl, line 158: Why we need to update IPv4SubnetSize even if IPAddresses is not set, can we leave it empty

TaoZou1: to be compatible with previous behavior.
if only ipAddressBlockVisibility provided in crd, the allocationSize value will be set default size

PR #1186, opened 2025-07-31
Remove redundant code

TaoZou1:
Remove redundant code.
Use switch to replace some if clause.
Cast type in switch instead of in case clause


zhengxiexie:

PR #1191, opened 2025-08-04
Check enable_inventory before init inventory service

TaoZou1:
Check enable_inventory before init inventory service

Test Done:
Case 1:
1. set enable_inventory to false
2. reset nsx-ncp pod
3. check if no inventory service running

Case 2:
1. set enable_inventory to true
2. reset nsx-ncp pod
3. check if inventory service running


yanjunz97 (approved):
LGTM

PR #1200, opened 2025-08-13
Change SecurityPolicyPort port/endPort from integer to integer pointer

timdengyun:
In NSX SecurityPolicy service entries, for TCP/UDP protocol, the destination_ports can be 0 and 65535 or ANY if user doesn't specify any ports. Current integer data type for port/endPort will have default zero value given by kubeduilder when the user doesn't set it. This is will be same value if user explicity set zero value in port field. So, there is no way to make a difference between port 0 and the default empty value.

This case only happens for SecurityPolicy, not for NetworkPolicy. For NetworkPolicy port, the allowed port value range is: between 1 and 65535, or ANY.

Hence, to distinguish port 0 and ANY, we need to know if user has specify ports number in spec or not. It's needed to make port/endPort as integer pointer, so that, we can know this field is set or not by checking pointer is nil.

Testing Done:


zhengxiexie:

PR #1201, opened 2025-08-13
Add protection logic when updating k8s ns

seanpang-vmware:
Add protection and error handling when updating k8s namespace resource. When NetworkInfo CR is created, the errors on namespace annotations should be removed if there is error before.


zhengxiexie:
There are other places still UpdateK8sResourceAnnotation without checking error.

seanpang-vmware:
> There are other places still UpdateK8sResourceAnnotation without checking error.

that are error logics, when error happens, it could not handle error in error handling., it will return error and retry anyway, so i only add protection code in success logic to make sure if there are error in anntoation, it will be cleared

TaoZou1 (approved):

PR #1210, opened 2025-08-18
Add condition for updating inventory service status when hasAddr

zhengxiexie:
### Service Inventory Fix
- **Fix service status logic** for services without selectors in inventory service
- **Prevent incorrectly marking** valid external/manual services as `InventoryStatusDown`

## 🎯 Motivation

**Fixes:** [Bug #3562809](https://bugzilla-vcf.lvn.broadcom.net/show_bug.cgi?id=3562809) - LoadBalancer services without selectors show "Service endpoint status unknown"
Compatible with [Bug #3541307](https://bugzilla-vcf.lvn.broadcom.net/show_bug.cgi?id=3541307) - The networking status should be down for service with errors
**Issue**: LoadBalancer services without selectors incorrectly show "Service endpoint status unknown" in NSX application inventory data, even when endpoints pods are up and functioning correctly.

**Root Cause Analysis:**
The original logic didn't distinguish between services that **should** have endpoints (those with selectors) vs services that **may not** have endpoints by design:

- **External services** point to external endpoints and don't have pod selectors
- **Manual endpoint services** have manually created endpoints without selectors  
- **Headless services** may not have ready endpoints but are still valid

#### ✅ Test Case 1: LoadBalancer Service Without Selector (Fixed)
**Purpose**: Validate that LoadBalancer services without selectors show correct status when manual endpoints exist.

**Configuration**:
```yaml
cat ipa.yml lb.yml dep.yml ep.yml 

kind: IPAddressAllocation
metadata:
  name: ipallocation-external-1
  namespace: ns1
spec:
  allocationSize: 1
  ipAddressBlockVisibility: External

apiVersion: v1
kind: Service
metadata:
  name: tcp-svc-lb
  namespace: ns1
spec:
  type: LoadBalancer
  ports:
    - name: tcp-80
      port: 80
      protocol: TCP
      targetPort: 80
  loadBalancerIP: 192.168.0.12

apiVersion: apps/v1
kind: Deployment
metadata:
  name: tcp-deployment
  namespace: ns1
spec:
  replicas: 2
  selector:
    matchLabels:
      role: web
  template:
    metadata:
      labels:
        deployment: tcp-deployment
        role: web
    spec:
      hostname: web-deployment
      containers:
        - name: web
          image: "netfvt-docker-local.packages.vcfd.broadcom.net/humanux/http_https_echo:latest"
          imagePullPolicy: IfNotPresent
          ports:
            - containerPort: 80
              name: web-port

apiVersion: v1
kind: Endpoints
metadata:
  name: tcp-svc-lb
  namespace: ns1
subsets:
  - addresses:
      - ip: 10.246.0.2
      - ip: 10.246.0.4
    ports:
      - port: 80
        protocol: TCP
```

**Result**: ✅ **SUCCESS** - NSX inventory now shows correct status
```json
 "status" : "UNKNOWN",
    "network_status" : "UNHEALTHY",
    "network_errors" : [ {
      "error_message" : "Service endpoint status is unknown",
      "error_code" : "",
      "spec" : ""
    } ],
    "resource_type" : "ContainerApplication",
    "display_name" : "tcp-svc-lb",
```

### DIFF
--- a/pkg/nsx/services/inventory/builder.go
+++ b/pkg/nsx/services/inventory/builder.go
@@ -354,8 +354,6 @@ func (s *InventoryService) BuildService(service *corev1.Service) (retry bool) {
 	// Initialize as an empty slice to ensure NSX receives [] instead of null when clearing errors
 	networkErrors := make([]common.NetworkError, 0)
 	// Get pods from the endpoint
-	netStatus := NetworkStatusHealthy
-	status := InventoryStatusUp
 	podIDs, hasAddr := GetPodIDsFromEndpoint(context.TODO(), s.Client, service.Name, service.Namespace)
 
 	// Check if a service has a selector - services without selectors are valid
@@ -364,46 +362,8 @@ func (s *InventoryService) BuildService(service *corev1.Service) (retry bool) {
 	// Use a map to track unique error messages
 	uniqueErrors := make(map[string]bool)
 
-	if len(podIDs) > 0 {
-		status = InventoryStatusUp
-	} else if hasAddr {
-		status = InventoryStatusUnknown
-		errorMessage := "Service endpoint status is unknown"
-		uniqueErrors[errorMessage] = true
-		networkErrors = append(networkErrors, common.NetworkError{
-			ErrorMessage: errorMessage,
-		})
-	} else if hasSelector {
-		// Only mark as down if service has selector but no endpoints
-		status = InventoryStatusDown
-		netStatus = NetworkStatusUnhealthy
-		errorMessage := "Failed to get endpoints for Service"
-		uniqueErrors[errorMessage] = true
-		networkErrors = append(networkErrors, common.NetworkError{
-			ErrorMessage: errorMessage,
-		})
-	} else {
-		// Service without a selector is valid even without endpoints
-		status = InventoryStatusUp
-	}
-
-	// Check for NCP errors in service annotations
-	for key, value := range service.Annotations {
-		if util.Contains(ServiceNCPErrors, key) {
-			errorMessage := key + ":" + value
-			if !uniqueErrors[errorMessage] {
-				uniqueErrors[errorMessage] = true
-				networkErrors = append(networkErrors, common.NetworkError{
-					ErrorMessage: errorMessage,
-				})
-			}
-		}
-	}
-
-	if len(networkErrors) > 0 {
-		status = InventoryStatusDown
-		netStatus = NetworkStatusUnhealthy
-	}
+	// Determine complete service status with all available information
+	status, netStatus := s.determineServiceStatus(podIDs, hasAddr, hasSelector, service.Annotations, &networkErrors, uniqueErrors)
 
 	// Update the Pods' service IDs, which are related to this service
 	retry = s.synchronizeServiceIDsWithApplicationInstances(podIDs, service)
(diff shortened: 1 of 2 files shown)


pkg/nsx/services/inventory/builder.go
```diff
@@ -400,7 +402,10 @@ func (s *InventoryService) BuildService(service *corev1.Service) (retry bool) {
 		}
 	}
 
-	if len(networkErrors) > 0 {
+	if !isAnnotationError && hasAddr {
```
TaoZou1, line 405: && len(podIDs)==0 ？

zhengxiexie: fixed

pkg/nsx/services/inventory/builder.go
```diff
@@ -400,7 +402,10 @@ func (s *InventoryService) BuildService(service *corev1.Service) (retry bool) {
...
 	}
 
-	if len(networkErrors) > 0 {
+	if len(podIDs) == 0 && !isAnnotationError && hasAddr {
+		status = InventoryStatusUnknown
+		netStatus = NetworkStatusUnhealthy
+	} else if len(networkErrors) > 0 {
 		status = InventoryStatusDown
 		netStatus = NetworkStatusUnhealthy
 	}
```
yanjunz97, line 411: Can we merge the conditions here and the ones in line 367-388?
 https://github.com/vmware-tanzu/nsx-operator/blob/cb819f4d69f1062f1a385127d4aa502890182a70/pkg/nsx/services/inventory/builder.go#L367-L388
I feel the current logic is complicated, under some condition the status is set back and forth. Could we have a clear summary on how the service status is determined?

zhengxiexie: fixed

pkg/nsx/services/inventory/builder.go
```diff
@@ -448,6 +416,67 @@ func (s *InventoryService) BuildService(service *corev1.Service) (retry bool) {
...
+		// Only mark as down if service has selector but no endpoints
+		status = InventoryStatusDown
+		netStatus = NetworkStatusUnhealthy
+		errorMessage := "Failed to get endpoints for Service"
+		uniqueErrors[errorMessage] = true
+		*networkErrors = append(*networkErrors, common.NetworkError{
+			ErrorMessage: errorMessage,
+		})
+	}
+	return status, netStatus
```
TaoZou1, line 443: what's if has ep, but no pidIDs, no selector?
it seesm return "", ""
the default value of status, netStatus  maybe be InventoryStatusDown ,  NetworkStatusUnhealthy?

pkg/nsx/services/inventory/builder.go
```diff
@@ -448,6 +410,53 @@ func (s *InventoryService) BuildService(service *corev1.Service) (retry bool) {
...
+			if !uniqueErrors[errorMessage] {
+				uniqueErrors[errorMessage] = true
+				*networkErrors = append(*networkErrors, common.NetworkError{
+					ErrorMessage: errorMessage,
+				})
+			}
+		}
+	}
+
+	if len(podIDs) > 0 {
```
yanjunz97, line 432: Just to double confirm if those cases are expected:
- When a service without Selector is created, as it does not get podIDs/address and also not set Selector, it will be shown as UP. And after the address is configured, the service status will change from UP to Unknown?
- If we get the podIDs, the service will always be shown as UP even if there is error annotation on the service. I'm not sure if there will be a case like this, if not, it should be fine.

pkg/nsx/services/inventory/builder.go
zhengxiexie: fixed

yanjunz97 (approved):
LGTM

pkg/nsx/services/inventory/builder.go
TaoZou1: def _convert_service(self, key, service):
    deleted = []
    added = []
    project_id = None
    retry = False
    network_errors, network_status = self.update_network_errors(
        key, self._get_external_id(key, service))

status depends on ep status
network_status depends on anno error

yanjunz97 (approved):

pkg/nsx/services/inventory/builder.go
```diff
@@ -448,6 +408,56 @@ func (s *InventoryService) BuildService(service *corev1.Service) (retry bool) {
...
+		status = InventoryStatusUp
+		netStatus = NetworkStatusHealthy
+		if hasAnnotationError {
+			netStatus = NetworkStatusUnhealthy
+		}
+	} else if hasAddr {
+		status = InventoryStatusUnknown
+		if hasAnnotationError {
+			status = InventoryStatusDown
+		}
```
dantingl, line 440: This is different with NCP, NCP determine status based on podIDs and hasAddr

PR #1211, opened 2025-08-19
Remove redundant condition for inventory service status when hasAddr

zhengxiexie:


dantingl (dismissed):

pkg/nsx/services/inventory/builder.go
```diff
@@ -435,9 +435,6 @@ func (s *InventoryService) determineServiceStatus(podIDs []string, hasAddr, hasS
 		}
 	} else if hasAddr {
 		status = InventoryStatusUnknown
-		if hasAnnotationError {
-			status = InventoryStatusDown
-		}
 		netStatus = NetworkStatusUnhealthy
```
dantingl, line 438: set netStatus to NetworkStatusUnhealthy only when hasAnnotationError

zhengxiexie: ok

dantingl: Also line 447, if status is down, network status will always be unhealthy?

dantingl (approved):

PR #1217, opened 2025-08-27
Refactor SecurityPolicy deletion logic for non-default project

timdengyun:
Previously, SecurityPolicy, rules and groups, and shares will be deleted together with one HAPI call. However, if some resources, like a group or a share, are being used by other security policies, the whole deletion will fail, and NSX SecurityPolicy rules remain. But this SecurityPolicy is already deleted in K8s, which leads security leak due to the inconsistency between the K8s SecurityPolicy CR and NSX SecurityPolicy resource.

This patch is to refactor the SecurityPolicy deletion logic for non-default projects, and separate SecurityPolicy/rules deletion from groups/shares. So, NSX SecurityPolicy will be deleted as long as the SecurityPolicy CR is deleted. The groups and shares could be deleted in the GC loop if they failed to be deleted with the NSX SecurityPolicy at the same time.

This refactor logic also applies to the NetworkPolicy internal SecurityPolicy CR deletion logic.

Testing Done:
1. Testing Normal SecurityPolicy CRUD: 
- Create SecurityPolicy and do update/deletion
2. Test SecurityPolicy is deleted and groups is GC:
 - Create SecurityPolicy
 - Create a DFW in NSX to reference the one group created by SecurityPolicy
 - Delete SecurityPolicy CR and verify NSX SecurityPolicy/rules are deleted
 - Remove group references in NSX DFW rule and verify the stale groups/shares will GC
 3. Test GC SecurityPolicy with groups/shares in one time
 - Create SecurityPolicy 
 - Delete SecurityPolicy CR and make nsx-operator restart at the same time, so the NSX SecurityPolicy resources remain
 - Check NSX SecurityPolicy/rules, groups, shares will be GC after restart
 4.  Testing Normal NetworkPolicy CRUD: 
 - Create NetworkPolicy and do update/deletion
 5. Test NetworkPolicy is deleted and groups is GC:
 - Create NetworkPolicy
 - Create a DFW in NSX to reference the one group created by NetworkPolicy
 - Delete NetworkPolicy CR and verify the corresponding NSX SecurityPolicy/rules are deleted
 - Remove group references in NSX DFW rule and verify the stale groups/shares will GC
 6. Test GC NetworkPolicy with groups/shares in one time
 - Create NetworkPolicy 
 - Delete NetworkPolicy CR and make nsx-operator restart at the same time, so the NSX SecurityPolicy resources remain
 - Check NSX SecurityPolicy/rules, groups, shares will be GC after restart

### DIFF
--- a/pkg/nsx/services/securitypolicy/firewall.go
+++ b/pkg/nsx/services/securitypolicy/firewall.go
@@ -245,7 +245,7 @@ func (service *SecurityPolicyService) CreateOrUpdateSecurityPolicy(obj interface
 			err = service.createOrUpdateVPCSecurityPolicy(obj.(*v1alpha1.SecurityPolicy), common.ResourceTypeSecurityPolicy)
 		} else {
 			// For T1 network SecurityPolicy create/update
-			err = service.createOrUpdateSecurityPolicy(obj.(*v1alpha1.SecurityPolicy), common.ResourceTypeSecurityPolicy)
+			err = service.createOrUpdateT1SecurityPolicy(obj.(*v1alpha1.SecurityPolicy), common.ResourceTypeSecurityPolicy)
 		}
 	}
 	return err
@@ -523,7 +523,7 @@ func (service *SecurityPolicyService) getFinalSecurityPolicyResource(obj *v1alph
 	}
 }
 
-func (service *SecurityPolicyService) createOrUpdateSecurityPolicy(obj *v1alpha1.SecurityPolicy, createdFor string) error {
+func (service *SecurityPolicyService) createOrUpdateT1SecurityPolicy(obj *v1alpha1.SecurityPolicy, createdFor string) error {
 	finalSecurityPolicy, finalGroups, _, _, isChanged, err := service.getFinalSecurityPolicyResource(obj, createdFor, false)
 	if err != nil {
 		log.Error(err, "Failed to get SecurityPolicy resources from CR", "securityPolicyUID", obj.UID)
@@ -584,7 +584,8 @@ func (service *SecurityPolicyService) createOrUpdateSecurityPolicy(obj *v1alpha1
 
 func (service *SecurityPolicyService) createOrUpdateVPCSecurityPolicy(obj *v1alpha1.SecurityPolicy, createdFor string) error {
 	var err error
-	var finalGetNSXSecurityPolicy model.SecurityPolicy
+	var finalGetNSXSecurityPolicy *model.SecurityPolicy
+
 	isDefaultProject := false
 	vpcInfo, err := service.getVPCInfo(obj.ObjectMeta.Namespace)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


pkg/nsx/services/securitypolicy/firewall.go
```diff
@@ -730,89 +732,127 @@ func (service *SecurityPolicyService) deleteVPCSecurityPolicy(sp types.UID, isGC
...
-	existingNsxInfraShares := infraShareStore.GetByIndex(indexScope, string(sp))
-	existingNsxInfraShareGroups := infraGroupStore.GetByIndex(indexScope, string(sp))
+	existingSecurityPolices := securityPolicyStore.GetByIndex(indexScope, string(spUID))
+	existingNsxInfraShares := infraShareStore.GetByIndex(indexScope, string(spUID))
+	existingNsxInfraShareGroups := infraGroupStore.GetByIndex(indexScope, string(spUID))
+	// There is no NSX groups/rules in the security policy retrieved from securityPolicy store.
+	// The groups/rules associated the deleting security policy can only be gotten from group/rule store.
+	existingGroups := groupStore.GetByIndex(indexScope, string(spUID))
+	service.markDeleteGroups(existingGroups, nsxGroups, spUID)
+
```
heypnus, line 750: This function `deleteVPCSecurityPolicy` is too long. In the next refactor patch, please consider to move L743-L749, L761, L771-L775, L777-L786, L801-L809 out and assemble them to a new function `markSecurityPolicyResourcesDelete` and its returned values can be existingXXXs and vpcInfo.

pkg/nsx/services/securitypolicy/wrap.go
```diff
@@ -227,18 +227,25 @@ func (service *SecurityPolicyService) wrapProject(sp *model.SecurityPolicy, gs [
...
-		return nil, err
+	var securityPolicyChildren []*data.StructValue
+	if sp != nil {
+		var err error
+		securityPolicyChildren, err = service.wrapSecurityPolicy(sp)
+		if err != nil {
+			return nil, err
+		}
 	}
+
```
heypnus, line 248: Please change L230-L250 to:
```
	var resourceReferenceChildren []*data.StructValue
	if sp != nil {
		securityPolicyChildren, err := service.wrapSecurityPolicy(sp)
		if err != nil {
			return nil, err
		}
		rulesChildren, err := service.wrapRules(sp.Rules)
		if err != nil {
			return nil, err
		}
		sp.Rules = nil
		sp.Children = rulesChildren
		sp.ResourceType = &common.ResourceTypeSecurityPolicy
		resourceReferenceChildren = append(resourceReferenceChildren, securityPolicyChildren...)
	}
```

pkg/nsx/services/securitypolicy/firewall.go
```diff
@@ -730,89 +732,127 @@ func (service *SecurityPolicyService) deleteVPCSecurityPolicy(sp types.UID, isGC
...
+				return err
+			}
+			log.Info("Successfully deleted NSX SecurityPolicy and rules only", "nsxSecurityPolicyUID", spUID)
+		}
+
+		err = service.deleteNSXSecurityPolicyGroupShare(*nsxGroups, *nsxShares, *nsxShareGroups, &vpcInfo)
+		// Ignore error here to make groups/shares to deleted in GC.
+		// Because NSX SecurityPolicy is deleted, it's unable to get SecurityPolicy from SecurityPolicyStore by requeuing the error.
+		if err != nil {
+			log.Info("NSX SecurityPolicy is deleted, and groups, shares in VPC will be GC", "nsxSecurityPolicyUID", spUID)
```
heypnus, line 829: Please polish the sentences including "`, and groups, shares`", they looks not very fluent.

pkg/nsx/services/securitypolicy/firewall.go
```diff
@@ -926,49 +966,93 @@ func (service *SecurityPolicyService) markDeleteShares(existingShares []*model.S
...
+	err = nsxutil.TransNSXApiError(err)
+	if err != nil {
+		log.Error(err, "Failed to delete NSX SecurityPolicy in VPC", "nsxSecurityPolicyId", nsxSecurityPolicy.Id)
+		return err
+	}
+
+	return nil
+}
+
+// deleteNSXSecurityPolicyGroupShare deletes NSX SecurityPolicy associated the groups/shares for non-Default Project.
```
heypnus, line 1025: So you associated the deletions of groups, shares and shared groups... For example, if we want to delete a group A, a share B and a shared group C, but A is used in some user-created resource, then can B and C be deleted?

pkg/nsx/services/securitypolicy/store.go
```diff
@@ -122,7 +122,10 @@ func (securityPolicyStore *SecurityPolicyStore) Apply(i interface{}) error {
 	if i == nil {
 		return nil
 	}
-	sp := i.(*model.SecurityPolicy)
+	sp, ok := i.(*model.SecurityPolicy)
```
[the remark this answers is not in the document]
timdengyun: For this case, it will make sp.MarkedForDelete deference panic
```
var sp *model.SecurityPolicy = nil
Apply(sp) // i != nil, but i.(*model.SecurityPolicy) == nil
```
if you pass
Apply(nil) // i == nil is true

because an interface value is only nil if both the type and value are nil.
If the interface holds a type (even if the value is nil), the interface itself is not nil.

heypnus (approved):

PR #1222, opened 2025-08-29
Fix race in SubnetPort UT

yanjunz97:
The PR includes the 2 changes below:
- Avoid closing fatalErros channel before all go routines which may result in race in search calls.
- Replace the mock function for search without specifying number of calls.

Testing done:
After the fix, go test with -race can pass
```
go test -race github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/subnetport -gcflags=all=-l
ok  	github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/subnetport
```


yanjunz97:
> Do you think we can add `-race` parameter in the go test? It seems can find potential concurrent issue?

Sure, let me add the flag and run CI again

zhengxiexie (approved):
LGTM

heypnus (approved):

PR #1224, opened 2025-08-31
Merge ipRanges while incremental is true

TaoZou1:
While updating ipblocksinfo in incremental mode, operator needs to merge the ipRanges with old value

Test Done:
1. add a privdate_TGW ipblocks project-test-1
2. add the project-test-1 to the default connectivity profile
3. add a new namespace dev from VC
4. check the ipblocksinfo if it includes the right info


pkg/nsx/services/ipblocksinfo/ipblocksinfo.go
```diff
@@ -213,6 +256,8 @@ func (s *IPBlocksInfoService) createOrUpdateIPBlocksInfo(ctx context.Context, ip
 	if incremental {
 		ipBlocksInfo.ExternalIPCIDRs = s.mergeIPCidrs(ipBlocksInfoOld.ExternalIPCIDRs, ipBlocksInfo.ExternalIPCIDRs)
 		ipBlocksInfo.PrivateTGWIPCIDRs = s.mergeIPCidrs(ipBlocksInfoOld.PrivateTGWIPCIDRs, ipBlocksInfo.PrivateTGWIPCIDRs)
+		ipBlocksInfo.ExternalIPRanges = s.mergeIPRanges(ipBlocksInfoOld.ExternalIPRanges, ipBlocksInfo.ExternalIPRanges)
+		ipBlocksInfo.PrivateTGWIPRanges = s.mergeIPRanges(ipBlocksInfoOld.PrivateTGWIPRanges, ipBlocksInfo.PrivateTGWIPRanges)
```
yanjunz97, line 260: Just to confirm if MergeArraysWithoutDuplicate is enough or we need mergeIPRanges as we do not have subnet allocated from ip ranges?

pkg/nsx/services/ipblocksinfo/ipblocksinfo.go
```diff
@@ -137,6 +138,48 @@ func (s *IPBlocksInfoService) mergeIPCidrs(source []string, target []string) []s
 	return result
 }
 
+// mergeIPRanges merges target ranges into source CIDRs if not already covered by source.
+// Only considers IPv4, assumes no overlaps and all CIDRs are valid.
+// Assume there were no duplicate ranges in target,
+// None of the elements in target will be a subset of another element
```
yanjunz97, line 144: If needed, update the comments
```suggestion
// mergeIPRanges merges target ranges into source ranges if not already covered by source.
// Only considers IPv4, assumes no overlaps and all ranges are valid.
// Assume there were no duplicate ranges in target,
// None of the elements in target will be a subset of another element
```

pkg/nsx/services/ipblocksinfo/ipblocksinfo.go
TaoZou1: I did test and NSX has overlap checking while adding the ipblocks to the connectivity profile. so MergeArraysWithoutDuplicate should be enough

yanjunz97 (approved):
LGTM

PR #1226, opened 2025-09-01
Remove error annotation on ns if networkinfo exists

yanjunz97:
k8s sometimes return error even if the NetworkInfo CR is created successfully.
In this case, we need to clear error annotation on ns when networkinfo exists
to avoid false-positive error annotation.

Bug: 3556010


pkg/controllers/namespace/namespace_controller.go
```diff
@@ -67,6 +67,13 @@ func (r *NamespaceReconciler) createNetworkInfoCR(ctx context.Context, obj clien
 	if len(networkInfos.Items) > 0 {
 		// if there is already one networkInfo, return this networkInfo
 		log.Info("NetworkInfo already exists", "NetworkInfo", networkInfos.Items[0].Name, "Namespace", ns)
+		// Clear the error annotation on k8s ns
```
seanpang-vmware, line 70: nit: could you add some explanation here to clarify why we did this, because this scenario is quite corner, others may confuse why remove vpc error here.

seanpang-vmware (approved):

zhengxiexie (approved):
LGTM

PR #1227, opened 2025-09-01
Fix some logs issue

TaoZou1:
1. fix the caCount output
2. not shown access_token in the header
3. dump the response body before log


pkg/nsx/util/utils.go
```diff
@@ -140,19 +144,29 @@ func InitErrorFromResponse(host string, statusCode int, body []byte) NsxError {
 	return httpErrortoNSXError(&detail)
 }
 
+func dumpResponseBody(body []byte, statusCode int) {
+	if log.V(2).Enabled() {
```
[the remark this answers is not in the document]
TaoZou1: if not log.V(2).Enabled(), no need to Unmarshal the body.  The Unmarshal call is only for the log.V(2) output

pkg/nsx/util/utils.go
```diff
@@ -270,13 +283,13 @@ func HandleHTTPResponse(response *http.Response, result interface{}, debug bool)
...
 	}
 
-	if debug {
-		log.V(2).Info("Received HTTP response", "response", string(body))
-	}
 	if err := json.Unmarshal(body, result); err != nil {
 		log.Error(err, "Failed to convert HTTP response to result", "result type", result)
 		return err, body
 	}
+	if debug {
```
timdengyun, line 290: since we have debug control inside dumpResponseBody, there is no need to have this extra argument.
I suggest remove this parameter.

timdengyun: It's strange to have a parameter to controller debug behavior instead of config option or log level option.

TaoZou1: it's special for JWT token, when the function used for JWT, it should not output the body

timdengyun (approved):
LGTM

zhengxiexie (approved):
LGTM

PR #1233, opened 2025-09-08
Update the IP/MAC of realized binding in DHCP port

heypnus:
Previously we only update the IP/MAC in SubnetPort CR status for non-DHCP port. After the SubnetPort.Spec.AddressBindings is exposed, the DHCP port also has non-empty realized binding when SubnetPort.Spec.AddressBindings is specified. So this patch will also update the realized binding for that kind of DHCP port.

Testing done:
Step 1. create DHCP port with addressBindings:
```
kind: SubnetPort
metadata:
  name: debug-vm-22-s10-eth0
spec:
  subnet: s10
  addressBindings:
  - ipAddress: "192.168.15.34"
    macAddress: "00:50:56:ba:d6:7a"
```
Step 2: create VM attching to the SubnetPort, the VM is able to get the IP/MAC in the port's addressBindings:
```
    hostName: debug-vm-22
    interfaces:
    - deviceKey: 4000
      ip:
        addresses:
        - address: 192.168.15.34
          state: preferred
        - address: fe80::250:56ff:feba:d67a
          state: unknown
        macAddr: 00:50:56:ba:d6:7a
      name: eth0
```
Step 3. create a new DHCP port with addressBindings and the MAC is in NSX MAC pool's scope, the VM also can be able to get the IP/MAC.


pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -190,7 +190,7 @@ func (r *SubnetPortReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			},
 			DHCPDeactivatedOnSubnet: !enableDHCP,
 		}
-		if !enableDHCP && len(nsxSubnetPortState.RealizedBindings) > 0 {
+		if len(nsxSubnetPortState.RealizedBindings) > 0 {
```
yanjunz97, line 193: I recall we add this for another bug. Without checking the dhcp mode, we will set the ip in status for a dhcp subnetport without addressbinding when operator is restarted, which is not expected.
Shall we only update the CR status for the case that the spec.addressbinding is provided?

heypnus: Good point! Fixed, thanks!

yanjunz97 (approved):
LGTM

lxiaopei (approved):
LGTM

PR #1236, opened 2025-09-10
Update Timeout name to align with NCP

TaoZou1:
Rename DefaultTimeout to HttpTimeout to align with NCP, rename tag from default_timeout to http_timeout


zhengxiexie:

PR #1242, opened 2025-09-16
Use CR to check SubnetConnectionBindingMap dependency

yanjunz97:
Currently we check cache for SubnetConnectionBindingMap dependency.
But if two SubnetConnectionBindingMap CR are created at the same time; and the child of the
SubnetConnectionBindingMap is the parent of anther SubnetConnectionBindingMap. As at the
time of dependency check, the NSX BindingMaps are not created yet, both of the CRs will pass
the dependency check and NSX BindingMap with unexpected dependency will be created.
This PR checks the CR instead of cache for SubnetConnectionBindingMap dependency to avoid this race.

Testing done:

Created 2 SubnetConnectionBindingMap CRs where the parent subnet of scbm1 is the child of scbm2
```yaml
metadata:
  name: scbm1
spec:
  subnetName: subnet-3
  vlanTrafficTag: 201
  targetSubnetName: subnet-2
---
metadata:
  name: scbm2
spec:
  subnetName: subnet-2
  vlanTrafficTag: 202
  targetSubnetName: subnet-1
```

Observed the first SubnetConnectionBindingMap is realized and the second is failed

```yaml
apiVersion: v1
items:
- apiVersion: crd.nsx.vmware.com/v1alpha1
  metadata:
    annotations:
        {"apiVersion":"crd.nsx.vmware.com/v1alpha1","kind":"SubnetConnectionBindingMap","metadata":{"annotations":{},"name":"scbm1","namespace":"ns-1"},"spec":{"subnetName":"subnet-3","targetSubnetName":"subnet-2","vlanTrafficTag":201}}
    creationTimestamp: "2025-09-16T07:18:13Z"
    generation: 1
    name: scbm1
    namespace: ns-1
    resourceVersion: "4421205"
    uid: ebce79fc-73b9-46e1-9f83-8feebaeb0e60
  spec:
    subnetName: subnet-3
    targetSubnetName: subnet-2
    vlanTrafficTag: 201
  status:
    conditions:
    - lastTransitionTime: "2025-09-16T07:18:14Z"
      status: "True"
      type: Ready
- apiVersion: crd.nsx.vmware.com/v1alpha1
  metadata:
    annotations:
        {"apiVersion":"crd.nsx.vmware.com/v1alpha1","kind":"SubnetConnectionBindingMap","metadata":{"annotations":{},"name":"scbm2","namespace":"ns-1"},"spec":{"subnetName":"subnet-2","targetSubnetName":"subnet-1","vlanTrafficTag":202}}
    creationTimestamp: "2025-09-16T07:18:14Z"
    generation: 1
    name: scbm2
    namespace: ns-1
    resourceVersion: "4421211"
    uid: 2e607ce1-0ac2-4e03-a4f6-58d0be710dba
  spec:
    subnetName: subnet-2
    targetSubnetName: subnet-1
    vlanTrafficTag: 202
  status:
    conditions:
    - lastTransitionTime: "2025-09-16T07:18:14Z"
      message: Subnet CR subnet-2 is working as target by [ns-1/scbm1]
      reason: DependencyNotReady
      status: "False"
      type: Ready
kind: List
metadata:
  resourceVersion: ""
```


pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -338,22 +338,38 @@ func (r *Reconciler) validateVpcSubnetsBySubnetCR(ctx context.Context, namespace
 
 	// Check if the Subnet CR is nested.
 	if !isTarget {
-		bms := r.SubnetBindingService.GetSubnetConnectionBindingMapsByParentSubnet(subnetPaths[0])
+		bms, err := r.getSubnetConnectionBindingMapsByParentSubnet(ctx, namespace, name)
```
wenyingd, line 341: If the purpose is to check the CR instead of the NSX resource, should not move this validation to the beginning? I mean this validation has no dependency on the existence of `subnetPaths`, it can be performed earlier.

wenyingd (commented):
Actually, I don't get the description of the bug. We only have 1 worker to run subnetbinding_controller, right?  Then it should reconcile subnetConnectionBindingMap one by one. So even though two binding-map crs are created at the same time, I think one of the two should finally reconcile failed. Isn't it the observation?

yanjunz97:
> Actually, I don't get the description of the bug. We only have 1 worker to run subnetbinding_controller, right? Then it should reconcile subnetConnectionBindingMap one by one. So even though two binding-map crs are created at the same time, I think one of the two should finally reconcile failed. Isn't it the observation?

We have 8 workers for subnetbinding_controller. Thus they'll be realized at the same time.
https://github.com/vmware-tanzu/nsx-operator/blob/851c691fa3561985912a7351fec0fd8b74d17a0f/pkg/controllers/subnetbinding/subnetbinding_controller.go#L176
https://github.com/vmware-tanzu/nsx-operator/blob/851c691fa3561985912a7351fec0fd8b74d17a0f/pkg/controllers/common/utils.go#L124-L126
https://github.com/vmware-tanzu/nsx-operator/blob/851c691fa3561985912a7351fec0fd8b74d17a0f/pkg/controllers/common/types.go#L27

wenyingd (approved):
LGTM

lxiaopei:
Today I think 2 questions:
1. for shared subnet, if the shared subnet is with 2 NSes NS1 and NS2, if the NSX subnet is used as child in  BindingMap  in NS1, then in NS2 if the Subnet is used as parent, then it cannot detect that right? But on NSX it should not supported.
2. for SubnetSet CR, we just use all nsx subnets in SubnetSet as parent, user cannot use them as child since there is no Subnet CR for them, right?

lxiaopei (approved):
LGTM

PR #1270, opened 2025-10-24
Adjust log level and fix log format

yanjunz97:


pkg/nsx/services/inventory/builder.go
```diff
@@ -46,8 +46,10 @@ func (s *InventoryService) BuildPod(pod *corev1.Pod) (retry bool) {
 		if pod.Spec.NodeName != "" {
 			// retry when pod has Node but Node is missing in NodeInformer
 			retry = true
+			log.Error(err, "Cannot find node for Pod", "Pod", pod.Name, "Namespace", pod.Namespace, "Node", pod.Spec.NodeName, "retry", retry)
```
TaoZou1: Should Line 49/51 to be swapped?
Line 49 should be a Warn
Line 51 is a Error
Could you help to add return after Line 52?
Could you also help to update the Line40 with Warn?

pkg/nsx/services/staticroute/staticroute.go
```diff
@@ -92,7 +92,7 @@ func (service *StaticRouteService) CreateOrUpdateStaticRoute(namespace string, o
 
 	vpc := service.VPCService.ListVPCInfo(namespace)
 	if len(vpc) == 0 {
-		return fmt.Errorf("no vpc found for ns %s", namespace)
+		return fmt.Errorf("no vpc found for ns %s, will retry later", namespace)
```
TaoZou1: -> no VPC found for namespace
only here uses 'ns'

yanjunz97: updated, thanks!

TaoZou1 (dismissed):

PR #1296, opened 2025-11-27
Use status instead of annotation to show Pod restore

yanjunz97:
Spherelet does not have the permission to edit Pod annotation
thus we use Status instead to inform the Spherelet when the
SubnetPort for Pod is restored.
This PR adds a new Condition of type PodNetworkRestore in Pod Status
and after restore, this condition will be false with reason ReconfigureNic.

Testing done:
Simulated the Pod restore and observed after Operator restore the Pod status is updated
```
  status:
    conditions:
    - lastProbeTime: null
      lastTransitionTime: "2025-11-26T08:51:54Z"
      status: "True"
      type: PodScheduled
    - lastProbeTime: null
      lastTransitionTime: "2025-11-26T08:54:19Z"
      status: "True"
      type: Initialized
    - lastProbeTime: null
      lastTransitionTime: "2025-11-26T08:54:19Z"
      status: "True"
      type: ContainersReady
    - lastProbeTime: null
      lastTransitionTime: "2025-11-26T08:54:19Z"
      status: "True"
      type: Ready
    - lastProbeTime: null
      lastTransitionTime: "2025-11-27T02:46:24Z"
      reason: ReconfigureNic
      status: "False"
      type: PodNetworkRestore   <--------- this condition is added after restore
```

### DIFF
--- a/pkg/controllers/pod/pod_controller.go
+++ b/pkg/controllers/pod/pod_controller.go
@@ -8,12 +8,14 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"reflect"
 	"strings"
 	"time"
 
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apimachineryruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -127,7 +129,7 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 				log.Error(err, "Failed to update restore annotation on Pod", "Namespace", pod.Namespace, "Pod", pod.Name)
 				return err != nil
 			}, func() error {
-				return common.UpdateReconfigureNicAnnotation(r.Client, ctx, pod, "true")
+				return setReconfigureNicStatus(r.Client, ctx, pod)
 			})
 		}
 	} else {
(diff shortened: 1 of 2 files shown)


pkg/controllers/pod/pod_controller.go
```diff
@@ -127,7 +129,7 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 				log.Error(err, "Failed to update restore annotation on Pod", "Namespace", pod.Namespace, "Pod", pod.Name)
 				return err != nil
 			}, func() error {
-				return common.UpdateReconfigureNicAnnotation(r.Client, ctx, pod, "true")
+				return setReconfigureNicStatus(r.Client, ctx, pod)
```
heypnus: I saw you only modified the pod_controller, so the subnetport still holds the info in the annotation, right? I'm curious why we don't also move the subnetport's `ReconfigureNic` to status, to keep the pod and subnetport implementation consistent..

yanjunz97: The change from annotation to status is a fix for spherelet's permission issue. If we also change this for subnetport then we need to discuss it with vm operator/wcpsvc team again. I prefer to keep the current implementation unchanged as in 9.2 we will eventually replace it with the vif restore solution.

heypnus (dismissed):

pkg/controllers/pod/pod_controller.go
```diff
@@ -148,6 +150,67 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 	return common.ResultNormal, nil
 }
 
+// Set PodNetworkReady condition to false in Pod to inform spherelet the SubnetPort is restored with another vif ID
+func setReconfigureNicStatus(client client.Client, ctx context.Context, obj client.Object) error {
+	pod := obj.(*v1.Pod)
+	newConditions := []v1.PodCondition{
+		{
+			Type:               v1.PodConditionType("PodNetworkReady"),
```
timdengyun: This condition is only used for podRestore?  we may change  PodNetworkReady to PodNetworkRestore for more specific. 
Who is responsible to set this condition to true? We only see setting is as False. 

For this change, please add Spherelet team to have a review.

timdengyun (approved):
LGTM

heypnus (approved):

PR #1374, opened 2026-02-09
Add `--log-color` flag to control ANSI color log output

zhengxiexie:
### Log Color Control
- ANSI color output is now **disabled by default** in log output
- A new `--log-color` flag is added to explicitly enable colored log output
- Color is fully decoupled from log level — users opt in via `--log-color`

### Implementation Details
- Added `LogColor` config variable and `--log-color` CLI flag (default `false`) in `pkg/config/config.go`
- Updated `ZapCustomLogger` signature to accept `enableColor bool` parameter
- Set `ConsoleWriter.NoColor` based on the flag to control zerolog color output
- Guarded `FormatLevel` and `FormatCaller` color wrapping with `enableColor` check
- Widened log level switch from `== 2` to `>= 2` for `TraceLevel` mapping
- Registered `--log-color` flag in all entry points: `cmd/main.go`, `cmd_clean/main.go`, `test/e2e/main_test.go`
- Added unit tests `TestGetLogLevel` and `TestAnsiColorControlledByFlag`

## 🎯 Motivation
ANSI color escape codes were always emitted in log output regardless of log level. This causes garbled text in non-terminal environments such as `kubectl logs`, log aggregators, and CI pipelines where ANSI codes are not interpreted.

**Issue**: Log output contains raw ANSI escape sequences (e.g. `\033[32m`) when viewed in environments that don't support terminal colors, making logs harder to read and parse.

**Solution**: Make color output opt-in via a dedicated `--log-color` flag, keeping it independent from log verbosity levels. This is typically used only during local development/debugging where a real terminal is available.

### Unit Tests
- All existing unit tests pass
- Added new test coverage:
    - ✅ `TestGetLogLevel` - Validates log level calculation for all debug/logLevel combinations (default, debug_flag_only, log_level_1, log_level_3, debug_with_higher_level, debug_with_lower_level)
    - ✅ `TestAnsiColorControlledByFlag` - Captures log output to buffer and verifies:
        - ✅ `color_disabled` - No ANSI escape codes when `--log-color` is not set
        - ✅ `color_enabled` - ANSI escape codes present when `--log-color` is set

```
$ go test ./pkg/logger/ -v -run "TestGetLogLevel|TestAnsiColor"
=== RUN   TestGetLogLevel
=== RUN   TestGetLogLevel/default
=== RUN   TestGetLogLevel/debug_flag_only
=== RUN   TestGetLogLevel/log_level_1
=== RUN   TestGetLogLevel/log_level_3
=== RUN   TestGetLogLevel/debug_with_higher_level
=== RUN   TestGetLogLevel/debug_with_lower_level
--- PASS: TestGetLogLevel (0.00s)
=== RUN   TestAnsiColorControlledByFlag
=== RUN   TestAnsiColorControlledByFlag/color_disabled
=== RUN   TestAnsiColorControlledByFlag/color_enabled
--- PASS: TestAnsiColorControlledByFlag (0.00s)
PASS
```

## 🔄 Backward Compatibility

This change is fully backward compatible:
- Default behavior now outputs plain text instead of colored text, which is strictly better for production environments
- Users who want colored output can explicitly opt in with `--log-color`
- No configuration file changes or API changes required

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -88,7 +88,7 @@ func init() {
 		os.Exit(1)
 	}
 
-	log = logger.ZapCustomLogger(cf.DefaultConfig.Debug, config.LogLevel)
+	log = logger.ZapCustomLogger(cf.DefaultConfig.Debug, config.LogLevel, config.LogColor)
 	logger.Log = log
 	// Set the controller-runtime logger to prevent the warning about log.SetLogger(...) never being called
 	logf.SetLogger(log.Logger)
--- a/cmd_clean/main.go
+++ b/cmd_clean/main.go
@@ -61,6 +61,7 @@ func main() {
 	flag.StringVar(&envoyHost, "envoyhost", "", "envoy host")
 	flag.IntVar(&envoyPort, "envoyport", 0, "envoy port")
 	flag.IntVar(&config.LogLevel, "log-level", 2, "Use zap-core log system.")
+	flag.BoolVar(&config.LogColor, "log-color", false, "Enable ANSI color in log output.")
 	flag.Parse()
 
 	cf = config.NewNSXOpertorConfig()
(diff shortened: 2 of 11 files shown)


yanjunz97 (approved):
LGTM

PR #1383, opened 2026-03-02
Add sts pod support

TaoZou1:
Add sts pod support

    Not delete subnetport in pod_controller if it's a sts port.
    Rlease subnetport in statefulset_controller if index out of range.
    If the pod reschduled and the port found, reuse the port

    Test Done:

    Test case: upgrade
    1. Create the sts without patch
    2. Check subnet port created without sts related tag
    3. Patch the commit
    4. Check if subunit port has been added new tags

    Test Case: shrink
    1. Create the sts with replicas 2
    2. Update the replicas from 2 to 1
    3. Check if pod controller skip to delete the subnetport
    4. Check if statefulset controller has deleted the subnetport

    Test Case: delete
    1. Create the sts with replicas 2
    2. Delete the sts
    3. Check if subnetports have been deleted
       "statefulset_controller.go:124 Successfully deleted
       subnet port for StatefulSet StatefulSet"

    Test Case: gc
    1. Create a subnetport with sts tag
    2. Restart the nix-operator
    3. Check if the subnetport has been deleted
       "Found orphaned subnet port for deleted StatefulSet"

    Test Case: reuse, same node
    1. create the sts pod
    2. delete one pod
    3. check if the pod created with the same ip in the same node
    4. check if the subnetport id is the same as before
    
    Test Case: reuse, different node
    1. create the sts pod
    2. taint the node
    3. delete the pod
    4. check if the pod created with the same ip
    5. check if the subnetport id of the pod is the same as before
    
    Test Case: Update replicas
    1. create the sts pod with replicas=6
    2. update the replicas=20
    3. check if the pod has been created
    
    Test Case: Delete multiple pods
     1. create the sts with replicas=20
     2. delete some of pod like
        k delete pod web-sts-2,web-sts-6,web-sts-8,web-sts-10,web-sts-15 -nqe
     3. check if all the pod have been created with the same ip
     
    Test Case：NSX 9.1
    1. mock the nsx version 9.1.0.25317852
    2. restart the operator
    3. check if has log:
    2026-04-13 07:17:52.000 INFO main.go:254 NSX version does not yet support StatefulSet Pod feature...
    4. delete one of the pod
    5. check if the previous old port deleted
    6. check if new port allocated
    
    Test case: change nsx version from 9.1 to 9.2
    1. mock the nsx version 9.1.0.25317852
    2. restart the operator
    3. check if has log:
    2026-04-13 07:17:52.000 INFO main.go:254 NSX version does not yet support StatefulSet Pod feature...
    4, delete one of the pod
     5. check if pod controller handle the DELETE event
     6. mock the nsx version 9.2.0.25317852, wait for version checking interval
     7. check if statefulset GC log found
       
   Test case: check cm vpc_wcp_enhance
    1. set the vpc_wcp_enhance true
    2. check if the statefulset feature has been enabled
    3. update the vpc_wcp_enhance false 
    4. restart the nsx-operator
    5. check if the statefulset feature has been disabled
    StatefulSet Pod feature gated (NSX version or vpc_wcp_enhance); StatefulSet controller registered but replica/GC no-op until enabled

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -36,6 +36,7 @@ import (
 	"github.com/vmware-tanzu/nsx-operator/pkg/controllers/pod"
 	securitypolicycontroller "github.com/vmware-tanzu/nsx-operator/pkg/controllers/securitypolicy"
 	"github.com/vmware-tanzu/nsx-operator/pkg/controllers/service"
+	statefulsetcontroller "github.com/vmware-tanzu/nsx-operator/pkg/controllers/statefulset"
 	staticroutecontroller "github.com/vmware-tanzu/nsx-operator/pkg/controllers/staticroute"
 	"github.com/vmware-tanzu/nsx-operator/pkg/controllers/subnet"
 	subnetbindingcontroller "github.com/vmware-tanzu/nsx-operator/pkg/controllers/subnetbinding"
@@ -244,6 +245,15 @@ func startServiceController(mgr manager.Manager, nsxClient *nsx.Client) {
 			subnetbindingcontroller.NewReconciler(mgr, subnetService, subnetBindingService),
 			subnetipreservationcontroller.NewReconciler(mgr, subnetIPReservationService, subnetService),
 		)
+		// StatefulSet controller is always registered so that after NSX upgrades (e.g. to 9.2.0+)
+		// replica/GC logic can run without restarting the operator. Reconcile and CollectGarbage
+		// no-op until NSX version supports STS pods and vpc_wcp_enhance=true in config; delete cleanup still runs.
+		reconcilerList = append(reconcilerList, statefulsetcontroller.NewStatefulSetReconciler(mgr, subnetPortService))
+		if nsx.StatefulSetPodSubnetPortFeatureEnabled(commonService.NSXClient, commonService.NSXConfig) {
+			log.Info("NSX version and config allow StatefulSet Pod feature; StatefulSet controller will run replica/GC work")
+		} else {
+			log.Info("StatefulSet Pod feature gated (NSX version and/or vpc_wcp_enhance!=true); StatefulSet controller registered but replica/GC no-op until enabled")
+		}
 		if cf.EnableInventory {
 			reconcilerList = append(reconcilerList, inventory.NewInventoryController(mgr.GetClient(), inventoryService, cf))
 		}
--- a/pkg/controllers/pod/pod_controller.go
+++ b/pkg/controllers/pod/pod_controller.go
@@ -34,6 +34,7 @@ import (
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx"
 	servicecommon "github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/subnetport"
+	nsxutil "github.com/vmware-tanzu/nsx-operator/pkg/nsx/util"
 	"github.com/vmware-tanzu/nsx-operator/pkg/util"
 )
 
@@ -83,7 +84,7 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 		return common.ResultNormal, nil
 	}
 
-	if !podIsDeleted(pod) {
+	if !common.PodIsDeleted(pod) {
 		r.StatusUpdater.IncreaseUpdateTotal()
 		isExisting, nsxSubnetPath, subnetSetUID, subnetSetLock, err := r.GetSubnetPathForPod(ctx, pod)
 		if subnetSetLock != nil {
@@ -145,6 +146,11 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 			return common.ResultRequeue, err
 		}
 		if subnetPort != nil {
+			if nsx.StatefulSetPodSubnetPortFeatureEnabled(r.SubnetPortService.NSXClient, r.SubnetPortService.NSXConfig) && r.isStatefulSetSubnetPort(subnetPort) {
+				log.Info("Ignoring subnet port deletion for StatefulSet pod",
+					"pod", subnetPort.DisplayName, "statefulset-uid", r.getStsUID(subnetPort))
+				return common.ResultNormal, nil
+			}
 			if err := r.SubnetPortService.DeleteSubnetPort(subnetPort); err != nil {
 				r.StatusUpdater.DeleteFail(req.NamespacedName, pod, err)
 				return common.ResultRequeue, err
@@ -361,6 +367,18 @@ func (r *PodReconciler) CollectGarbage(ctx context.Context) error {
 	var errList []error
 	diffSet := nsxSubnetPortSet.Difference(PodSet)
 	for elem := range diffSet {
+		// StatefulSet pod ports are keyed in the pod-UID index but may briefly outlive the old Pod
+		// object (crash/replace window). STS SubnetPort lifecycle is owned by the StatefulSet
+		// reconciler when the feature is enabled; do not delete here (same as deleteSubnetPortByPodName).
+		store := r.SubnetPortService.SubnetPortStore
+		if store != nil && store.Indexer != nil {
+			if nsxSubnetPort := store.GetByKey(elem); nsxSubnetPort != nil &&
+				r.SubnetPortService.NSXClient != nil &&
+				nsx.StatefulSetPodSubnetPortFeatureEnabled(r.SubnetPortService.NSXClient, r.SubnetPortService.NSXConfig) && r.isStatefulSetSubnetPort(nsxSubnetPort) {
+				log.Info("Skipping pod GC for StatefulSet pod subnet port", "NSXSubnetPortID", elem, "statefulset-uid", r.getStsUID(nsxSubnetPort))
+				continue
+			}
+		}
 		log.Debug("GC collected Pod", "NSXSubnetPortID", elem)
 		r.StatusUpdater.IncreaseDeleteTotal()
 		err = r.SubnetPortService.DeleteSubnetPortById(elem)
@@ -418,21 +436,34 @@ func (r *PodReconciler) GetSubnetPathForPod(ctx context.Context, pod *v1.Pod) (b
 	return false, subnetPath, subnetSetUID, subnetSetLock, nil
 }
 
-func podIsDeleted(pod *v1.Pod) bool {
-	return !pod.ObjectMeta.DeletionTimestamp.IsZero() || pod.Status.Phase == "Succeeded" || pod.Status.Phase == "Failed"
-}
-
 func (r *PodReconciler) deleteSubnetPortByPodName(ctx context.Context, ns string, name string) error {
 	// NamespacedName is a unique identity in store as only one worker can deal with the NamespacedName at a time
 	nsxSubnetPorts := r.SubnetPortService.ListSubnetPortByPodName(ns, name)
 
 	for _, nsxSubnetPort := range nsxSubnetPorts {
+		// Check if this subnet port was created for StatefulSet
+		// Only skip if STS feature is enabled (NSX 9.2.0+)
+		if nsx.StatefulSetPodSubnetPortFeatureEnabled(r.SubnetPortService.NSXClient, r.SubnetPortService.NSXConfig) && r.isStatefulSetSubnetPort(nsxSubnetPort) {
+			log.Info("Ignoring subnet port deletion for StatefulSet pod",
+				"pod", name, "statefulset-uid", r.getStsUID(nsxSubnetPort))
+			continue // Skip deletion
+		}
+
+		// Normal pod: delete the subnet port
 		if err := r.SubnetPortService.DeleteSubnetPort(nsxSubnetPort); err != nil {
 			return err
 		}
 	}
 	log.Info("Successfully deleted nsxSubnetPort for Pod", "Namespace", ns, "Name", name)
 	return nil
+}
+
+func (r *PodReconciler) isStatefulSetSubnetPort(nsxSubnetPort *model.VpcSubnetPort) bool {
+	return r.getStsUID(nsxSubnetPort) != ""
+}
+
+func (r *PodReconciler) getStsUID(nsxSubnetPort *model.VpcSubnetPort) string {
+	return nsxutil.FindTag(nsxSubnetPort.Tags, servicecommon.TagScopeStatefulSetUID)
 }
 
 // PredicateFuncsPod filters out events where pod.Spec.HostNetwork is true
--- a/pkg/controllers/statefulset/statefulset_controller.go
+++ b/pkg/controllers/statefulset/statefulset_controller.go
@@ -1,0 +1,468 @@
+/* Copyright © 2026 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: Apache-2.0 */
+
+package statefulset
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/tools/record"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/controllers/common"
+	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx"
+	servicecommon "github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
+	subnetportservice "github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/subnetport"
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/util"
+)
+
+var (
+	log                      = logger.Log
+	MetricResTypeStatefulSet = common.MetricResTypeStatefulSet
+)
+
+// stsSubnetPortPendingRequeueAfter is used when StatefulSet-related subnet ports cannot be released yet
+// because the backing Pod is still running (scale-down/delete path). Requeue instead of relying only on GC.
+const stsSubnetPortPendingRequeueAfter = 10 * time.Second
+
+type releaseSubnetPortOutcome int
+
+const (
+	releaseSubnetPortNoop releaseSubnetPortOutcome = iota
+	releaseSubnetPortReleased
+	releaseSubnetPortSkippedRunningPod
+)
+
+// StatefulSetReconciler reconciles a StatefulSet object
+type StatefulSetReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+
+	SubnetPortService *subnetportservice.SubnetPortService
+	Recorder          record.EventRecorder
+	StatusUpdater     common.StatusUpdater
+}
+
+// StatefulSetPodFeatureEnabled reports whether the StatefulSet pod SubnetPort feature is active (NSX 9.2.0+ and vpc_wcp_enhance=true in config).
+func (r *StatefulSetReconciler) StatefulSetPodFeatureEnabled() bool {
+	return nsx.StatefulSetPodSubnetPortFeatureEnabled(r.SubnetPortService.NSXClient, r.SubnetPortService.NSXConfig)
+}
+
+func (r *StatefulSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log.Info("Reconciling StatefulSet", "StatefulSet", req.NamespacedName)
+	startTime := time.Now()
+	defer func() {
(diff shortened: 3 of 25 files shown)


dantingl:
Please test scale up case, and then restart several pods. 
Test replicas range.
And test with NSX 9.1 version?

pkg/controllers/pod/pod_controller.go
dantingl, line 439: Looks like this func is not used?

pkg/controllers/statefulset/statefulset_controller.go
```diff
@@ -0,0 +1,374 @@
...
+	startTime := time.Now()
+	defer func() {
+		log.Info("Finished reconciling StatefulSet", "StatefulSet", req.NamespacedName, "duration", time.Since(startTime))
+	}()
+
+	r.StatusUpdater.IncreaseSyncTotal()
+
+	sts := &appsv1.StatefulSet{}
+	if err := r.Client.Get(ctx, req.NamespacedName, sts); err != nil {
+		if apierrors.IsNotFound(err) {
```
dantingl, line 84: Is it possible when we get DELETED event, we can still find the STS? Do we need to check deleted property?

pkg/controllers/statefulset/statefulset_controller.go
```diff
@@ -0,0 +1,374 @@
...
+}
+
+func (r *StatefulSetReconciler) handleReplicaChange(ctx context.Context, sts *appsv1.StatefulSet) error {
+	start, end := r.GetOrdinalRange(sts)
+	log.Info("Checking replicas range", "start", start, "end", end)
+
+	existingPorts := r.SubnetPortService.ListSubnetPortByStsUid(sts.Namespace, string(sts.UID))
+
+	for _, port := range existingPorts {
+		if port.DisplayName == nil {
```
dantingl: Why use `DisplayName` to check index? it can be updated by user, and will we update DisplayName for upgrade case?

TaoZou1:
> Please test scale up case, and then restart several pods. Test replicas range. And test with NSX 9.1 version?

Update the test cases

pkg/nsx/services/subnetport/subnetport.go
```diff
@@ -419,6 +422,30 @@ func (service *SubnetPortService) ResetSubnetTotalIP(path string) {
 	info.totalIP = 0
 }
 
+func (service *SubnetPortService) ListSubnetPortByStsName(ns string, stsName string) []*model.VpcSubnetPort {
+	var result []*model.VpcSubnetPort
+	subnetports := service.SubnetPortStore.GetByIndex(servicecommon.TagScopeStatefulSetName, stsName)
```
dantingl, line 436: Can we get SubnetPort by `subnetPortIndexByStatefulSetUID`

pkg/util/utils.go
```diff
@@ -383,6 +383,15 @@ func BuildBasicTags(cluster string, obj interface{}, namespaceID types.UID) []mo
 		tags = append(tags, model.Tag{Scope: String(common.TagScopeNamespace), Tag: String(i.ObjectMeta.Namespace)})
 		tags = append(tags, model.Tag{Scope: String(common.TagScopePodName), Tag: String(i.ObjectMeta.Name)})
 		tags = append(tags, model.Tag{Scope: String(common.TagScopePodUID), Tag: String(string(i.UID))})
+		// Add StatefulSet tags if pod is owned by a StatefulSet
+		for _, ownerRef := range i.OwnerReferences {
+			if ownerRef.Kind == "StatefulSet" {
+				tags = append(tags, model.Tag{Scope: String(common.TagScopeStatefulSetName), Tag: String(ownerRef.Name)})
+				tags = append(tags, model.Tag{Scope: String(common.TagScopeStatefulSetUID), Tag: String(string(ownerRef.UID))})
+				tags = append(tags, model.Tag{Scope: String(common.TagScopeCreatedFor), Tag: String(common.TagValueSts)})
```
dantingl: Why we need tag TagScopeCreatedFor?

pkg/controllers/pod/pod_controller.go
TaoZou1: it's used in Line 72

pkg/nsx/services/subnetport/subnetport.go
TaoZou1: when the sts deleted, we don't know the UID

pkg/util/utils.go
TaoZou1: removed

pkg/controllers/statefulset/statefulset_controller.go
TaoZou1: Yes. Fixed it

pkg/controllers/statefulset/statefulset_controller.go
TaoZou1: OK, use the tags "nsx-op/pod_name" to get the index

pkg/controllers/statefulset/statefulset_controller.go
```diff
@@ -0,0 +1,407 @@
...
+	var errList []error
+	statefulSetUIDs := sets.New[string]()
+	for _, sts := range statefulSetList.Items {
+		existingPorts := r.SubnetPortService.SubnetPortStore.GetByIndex(servicecommon.TagScopeStatefulSetUID, string(sts.UID))
+		start, end := r.GetOrdinalRange(&sts)
+		log.Debug("StatefulSet garbage collector", "sts UID", sts.UID, "sts name", sts.Name, "start", start, "end", end)
+		for _, port := range existingPorts {
+			podName := util.FindTag(port.Tags, servicecommon.TagScopePodName)
+			if podName == "" {
+				continue
```
heypnus, line 291: In case which, the podName is ""?

pkg/controllers/statefulset/statefulset_controller.go
```diff
@@ -0,0 +1,407 @@
...
+			idx := stsPodOrdinalFromPort(port)
+			// Match handleReplicaChange: do not treat unparseable ordinals as out-of-range.
+			if idx == -1 {
+				continue
+			}
+			if idx < start || idx > end {
+				if podName != "" {
+					pod := &corev1.Pod{}
+					if err := r.Client.Get(ctx, types.NamespacedName{Namespace: sts.Namespace, Name: podName}, pod); err == nil {
+						log.Debug("GC: pod still exists, skipping port deletion", "pod", podName)
```
heypnus, line 302: The port deletion skipped for pod still existing, then when this port will be deleted? By whom?

pkg/controllers/statefulset/statefulset_controller.go
```diff
@@ -0,0 +1,407 @@
...
+			podName := util.FindTag(port.Tags, servicecommon.TagScopePodName)
+			namespace := util.FindTag(port.Tags, servicecommon.TagScopeNamespace)
+			if podName != "" && namespace != "" {
+				pod := &corev1.Pod{}
+				if err := r.Client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: podName}, pod); err == nil {
+					log.Debug("GC: pod still exists, skipping orphaned port deletion", "pod", podName)
+					continue
+				}
+			}
+			log.Debug("Found orphaned subnet port for deleted StatefulSet", "port", *port.Id, "stsID", stsID)
```
heypnus: Please add at least one INFO level log in the garbage collector to indicate the object is removed by GC, for the troubleshooting in the future.

pkg/controllers/statefulset/statefulset_controller.go
```diff
@@ -0,0 +1,407 @@
...
+	if replicas == 0 {
+		// No desired pods: empty ordinal range. Callers treat any idx>=0 as out-of-range
+		// so handleReplicaChange/GC release all STS-tagged ports for this StatefulSet.
+		return -1, -1
+	}
+
+	return start, start + replicas - 1
+}
+
+var PredicateFuncsForStatefulSet = predicate.Funcs{
```
heypnus, line 374: Could you please add a comment on the func `PredicateFuncsForStatefulSet` to explain how it filter the objects?

pkg/controllers/statefulset/statefulset_controller.go
```diff
@@ -0,0 +1,407 @@
...
+		if oldSts.Spec.Ordinals != nil {
+			oldStart = oldSts.Spec.Ordinals.Start
+		}
+		newStart := int32(0)
+		if newSts.Spec.Ordinals != nil {
+			newStart = newSts.Spec.Ordinals.Start
+		}
+
+		// If either has nil replicas, don't trigger (maintain backward compatibility)
+		if oldSts.Spec.Replicas == nil || newSts.Spec.Replicas == nil {
```
heypnus, line 393: What's the difference between `Replicas == nil` and `Replicas == 0`?

pkg/controllers/statefulset/statefulset_controller.go
TaoZou1: just a check, in case if tag missing

pkg/controllers/statefulset/statefulset_controller.go
TaoZou1: it's GC, if will be retried next round

pkg/controllers/statefulset/statefulset_controller.go
TaoZou1: Thanks. Update the Debug to Info

pkg/controllers/statefulset/statefulset_controller.go
TaoZou1: Done

pkg/controllers/statefulset/statefulset_controller.go
TaoZou1: Replicas is pointer type, Replicas == nil mean it didn't have a value. 
*Replicas ==0 mean its value is 0

pkg/controllers/statefulset/statefulset_controller.go
heypnus: The GC interval is 10min, right? So the stale resource may be collected after 10min, right? I'm not sure whether it's acceptable..

TaoZou1: yes. the GC interval is common. do you think it's not acceptable for all GC or just for the sts?

pkg/nsx/cluster.go
```diff
@@ -537,6 +537,9 @@ func (nsxVersion *NsxVersion) featureSupported(feature int) bool {
 	case StaticIPReservation:
 		minVersion = nsx920Version
 		validFeature = true
+	case StatefulSetPod:
+		minVersion = nsx920Version
```
dantingl: We will need to add NSX 9.1.2 for this feature

pkg/controllers/statefulset/statefulset_controller.go
```diff
@@ -0,0 +1,419 @@
...
+	log.Info("Releasing all subnet ports for StatefulSet", "StatefulSet", namespace+"/"+name)
+
+	subnetPorts := r.SubnetPortService.ListSubnetPortByStsName(namespace, name)
+
+	var errList []error
+	for _, subnetPort := range subnetPorts {
+		podName := util.FindTag(subnetPort.Tags, servicecommon.TagScopePodName)
+		if podName != "" {
+			pod := &corev1.Pod{}
+			if err := r.Client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: podName}, pod); err == nil {
```
dantingl, line 171: Please compare pod UID as well. Also check whether pod is in deletion status

pkg/controllers/statefulset/statefulset_controller.go
```diff
@@ -0,0 +1,419 @@
...
+	return nil
+}
+
+func (r *StatefulSetReconciler) releaseSubnetPortForPod(ctx context.Context, namespace, podName string) error {
+	log.Info("Releasing subnet port for pod", "pod", podName, "namespace", namespace)
+
+	pod := &corev1.Pod{}
+	if err := r.Client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: podName}, pod); err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Debug("Pod does not exist, skipping subnet port release", "pod", podName, "namespace", namespace)
```
dantingl: Why not deleting SubnetPort if Pod is not found?

pkg/controllers/statefulset/statefulset_controller.go
dantingl: Can we check Pod status as well, what is the GC logic for normal Pod?

pkg/nsx/services/common/types.go
```diff
@@ -125,6 +129,8 @@ const (
 	IndexKeySubnetPath          = "IndexKeySubnetPath"
 	IndexKeyNodeName            = "IndexKeyNodeName"
 	IndexKeyAttachmentID        = "IndexKeyAttachmentID"
+	IndexKeyAllStsPorts         = "IndexKeyAllStsPorts"
+	StsPortBucket               = "allStsPorts"
```
dantingl, line 133: Can you have an example of SubnetPort tags for statefulset pod

pkg/nsx/cluster.go
dantingl: Add TODO to include NSX 9.1.2 version

pkg/controllers/statefulset/statefulset_controller.go
TaoZou1: In fact, operator here will still release the subnetport, the log info is confused. Updated.

pkg/nsx/cluster.go
TaoZou1: update to NSX 9.1.2 since it's minimal version

pkg/controllers/statefulset/statefulset_controller.go
TaoZou1: add comparing the UID, not check the deletion status. the status is not useful here. In order to support graceful shutdown, it's better not to delete the port even the pod is in deletion status

pkg/controllers/pod/pod_controller.go
```diff
@@ -427,6 +433,15 @@ func (r *PodReconciler) deleteSubnetPortByPodName(ctx context.Context, ns string
 	nsxSubnetPorts := r.SubnetPortService.ListSubnetPortByPodName(ns, name)
 
 	for _, nsxSubnetPort := range nsxSubnetPorts {
+		// Check if this subnet port was created for StatefulSet
+		// Only skip if STS feature is enabled (NSX 9.2.0+)
+		if r.SubnetPortService.NSXClient.NSXCheckVersion(nsx.StatefulSetPod) && r.isStatefulSetSubnetPort(nsxSubnetPort) {
```
heypnus: I have a concern about this scenario:
1. the NSX version is < 9.2, then the statefulset controller not started
2. the NSX version is upgraded to 9.2 during NSX Operator runtime
3. the pod controller detects that the statefulset feature is supported, then skip the statefulset pod handling
4. but the statefulset controller didn't start, so the statefulset pod deletion cannot handled until the NSX Operator restarts
-- Is it possible to happen?

TaoZou1: good catch! I think it's possible. one way is that if the nsx-operator detected the NSX version changed, NSX-operator try to reboot itself.

pkg/nsx/services/common/types.go
TaoZou1: "tags" : [ {
      "scope" : "nsx-op/cluster",
      "tag" : "18f00b57-cb8e-411c-8cc1-21eedef12f2f"
    }, {
      "scope" : "nsx-op/version",
      "tag" : "1.0.0"
    }, {
      "scope" : "nsx-op/namespace",
      "tag" : "qe"
    }, {
      "scope" : "nsx-op/pod_name",
      "tag" : "web-sts-0"
    }, {
      "scope" : "nsx-op/pod_uid",
      "tag" : "3f21d12c-7bcc-4b75-8b2e-de37b6a0d71a"
    }, {
      "scope" : "nsx-op/sts_name",
      "tag" : "web-sts"
    }, {
      "scope" : "nsx-op/sts_uid",
      "tag" : "96cdbf08-4c63-48ba-9412-f91e11491124"
    }, {
      "scope" : "nsx-op/namespace_uid",
      "tag" : "57a0c845-20a3-46c5-862a-0238abb6e930"
    }, {
      "scope" : "app",
      "tag" : "nginx"
    }, {
      "scope" : "apps.kubernetes.io/pod-index",
      "tag" : "0"
    }, {
      "scope" : "controller-revision-hash",
      "tag" : "web-sts-869657df46"
    }, {
      "scope" : "statefulset.kubernetes.io/pod-name",
      "tag" : "web-sts-0"
    } ],

pkg/controllers/statefulset/statefulset_controller.go
TaoZou1: pod controller also check if the pod existing.
for graceful shutdown, it's better not to release the port when the pod existed. 
checking status doesn't help.
that may cause the subnetport leak if the pod not deleted.

pkg/controllers/pod/pod_controller.go
TaoZou1: Now i run the statefulset reconcile/GC even the nsx version not supported, but do nothing if StatefulSetPodFeatureEnabled is false.
Other option is to reboot the nsx-operator

dantingl: > the NSX version is upgraded to 9.2 during NSX Operator runtime

So nsx-operator detect NSX version at runtime, not just once when startup?

pkg/controllers/statefulset/statefulset_controller.go
```diff
@@ -0,0 +1,468 @@
...
+	SubnetPortService *subnetportservice.SubnetPortService
+	Recorder          record.EventRecorder
+	StatusUpdater     common.StatusUpdater
+}
+
+// StatefulSetPodFeatureEnabled reports whether NSX supports the StatefulSet pod SubnetPort feature (9.2.0+).
+func (r *StatefulSetReconciler) StatefulSetPodFeatureEnabled() bool {
+	if statefulSetPodFeatureTestOverride != nil {
+		return *statefulSetPodFeatureTestOverride
+	}
```
dantingl: Can we hack NSX version to test UT

dantingl (dismissed):

pkg/controllers/statefulset/statefulset_controller.go
```diff
@@ -0,0 +1,463 @@
...
+	if r.StatusUpdater.Client != nil {
+		r.StatusUpdater.DeleteSuccess(namespacedName, sts)
+	}
+	return common.ResultNormal, nil
+}
+
+func (r *StatefulSetReconciler) handleReplicaChange(ctx context.Context, sts *appsv1.StatefulSet) (ctrl.Result, error) {
+	start, end := r.GetOrdinalRange(sts)
+	log.Info("Checking replicas range", "start", start, "end", end)
```
heypnus: Please add identify info likes sts uid for toubleshooting.

pkg/controllers/statefulset/statefulset_controller.go
```diff
@@ -0,0 +1,463 @@
...
+
+	existingPorts := r.SubnetPortService.ListSubnetPortByStsUid(sts.Namespace, string(sts.UID))
+
+	pendingRunningPod := false
+	for _, port := range existingPorts {
+		podName := util.FindTag(port.Tags, servicecommon.TagScopePodName)
+		if podName == "" {
+			continue
+		}
```
heypnus, line 137: It seems that the function `stsPodOrdinalFromPort` also includes the logic to check podName == "", so L130-L133 is not needed? And the same question on L284-L287.

pkg/controllers/statefulset/statefulset_controller.go
```diff
@@ -0,0 +1,463 @@
...
+			}
+		}
+	}
+	if pendingRunningPod {
+		return ctrl.Result{RequeueAfter: replicaChangePendingRequeueAfter}, nil
+	}
+	return ctrl.Result{}, nil
+}
+
+func (r *StatefulSetReconciler) releaseSubnetPortsForStatefulSet(ctx context.Context, namespace, name string) error {
```
heypnus: Just curious that whether we can directly invoke `handleReplicaChange` to reuse the existing logic -- releaseSubnetPortsForStatefulSet is equal to "handle replica change to 0", right?

pkg/controllers/statefulset/statefulset_controller.go
TaoZou1: fixed

pkg/controllers/statefulset/statefulset_controller.go
TaoZou1: thanks. added

pkg/controllers/statefulset/statefulset_controller.go
TaoZou1: stsPodOrdinalFromPort not chek podName unless its can't get the index tag
Line 130-133 just protective code, normally the podName tag should be there.

pkg/controllers/statefulset/statefulset_controller.go
TaoZou1: not all the same. 
handleReplicaChange has the statefulset struct ( sts uid)
but when releaseSubnetPortsForStatefulSet, we only know the sts name, namespace name

heypnus (dismissed):
Thanks for addressing the comments. LTGM.

cmd/main.go
```diff
@@ -244,6 +245,15 @@ func startServiceController(mgr manager.Manager, nsxClient *nsx.Client) {
...
 			subnetipreservationcontroller.NewReconciler(mgr, subnetIPReservationService, subnetService),
 		)
+		// StatefulSet controller is always registered so that after NSX upgrades (e.g. to 9.2.0+)
+		// replica/GC logic can run without restarting the operator. Reconcile and CollectGarbage
+		// no-op until NSX version supports STS pods and vpc_wcp_enhance=true in config; delete cleanup still runs.
+		reconcilerList = append(reconcilerList, statefulsetcontroller.NewStatefulSetReconciler(mgr, subnetPortService))
+		if nsx.StatefulSetPodSubnetPortFeatureEnabled(commonService.NSXClient, commonService.NSXConfig) {
+			log.Info("NSX version and config allow StatefulSet Pod feature; StatefulSet controller will run replica/GC work")
+		} else {
+			log.Info("StatefulSet Pod feature gated (NSX version and/or vpc_wcp_enhance!=true); StatefulSet controller registered but replica/GC no-op until enabled")
```
dantingl, line 255: Remove `(NSX version and/or vpc_wcp_enhance!=true)` from log.
no-op: can you expand it?
How about: StatefulSet Pod IP persistent feature is not enabled

heypnus (approved):

dantingl (approved):

cmd/main.go
TaoZou1: will change it later when cm option vpc_wcp_enhance default value is true

PR #1396, opened 2026-03-19
Implement Mixed Mode State Detection

heypnus:
This patch introduces the Mixed Mode state detection logic for NSX Operator
to determine the network scope (T1, VPC, or VDS) it operates within. This
is critical for deciding which controllers should be activated.

Key features:
- Introduced an informer for the NetworkSettings CR in the Controller
  Runtime cache to efficiently determine the global network state.
- Added detection logic for VDS alongside T1 and VPC based on the
  NetworkSettings `provider` field.
- Implemented capability checking by reading the
  `supports_per_namespace_network_provider` key from the `status.supervisor`
  path in the SupervisorCapabilities CR.
- Incorporated the `EnableVPCNetwork` configuration override to act as an
  early trigger. This solves the "chicken-and-egg" problem by allowing
  VPC-specific controllers to start before the NetworkSettings CRs are
  updated to `provider: vpc`.
- Re-enabled the enable_vpc_network config override check in mixed mode
  initialization and state refresh, which acts as an early trigger for
  VPC-specific controllers.

After this change, the NSX-Operator will start VPC controllers as long as
enable_vpc_network is true. And the T1 controllers will be started when:
  (supports_per_namespace_network_provider is activated) AND (at least one NetworkingSettings CR has provider=="nsx-tier1")
OR
  (supports_per_namespace_network_provider is deactivated) AND (enable_vpc_network is false)

Testing done:
1. T1 pipeline with ID 5771
2. VPC pipeline with ID 19252

### DIFF
--- a/pkg/config/mixed_mode.go
+++ b/pkg/config/mixed_mode.go
@@ -1,0 +1,319 @@
+/* Copyright © 2026 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: Apache-2.0 */
+
+package config
+
+import (
+	"context"
+	"strings"
+	"sync"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
+)
+
+const (
+	VPCNetworkConfigAnnotation = "nsx.vmware.com/vpc_network_config"
+
+	supervisorCapabilitiesName = "supervisor-capabilities"
+)
+
+var (
+	supervisorCapabilitiesGVR = schema.GroupVersionResource{
+		Group:    "iaas.vmware.com",
+		Version:  "v1alpha1",
+		Resource: "supervisorcapabilities",
+	}
+
+	stateMu                        sync.RWMutex
+	hasT1Namespaces                bool
+	hasVPCNamespaces               bool
+	perNamespaceProvidersSupported *bool
+	stateInitialized               bool
+
+	// retryInitialInterval and retryMaxInterval control the exponential
+	// backoff used when a transient error prevents reading
+	// SupervisorCapabilities or listing namespaces. Overridable in tests.
+	retryInitialInterval = 2 * time.Second
+	retryMaxInterval     = 30 * time.Second
+
+	// storedClientset is kept from InitMixedMode so that RefreshMixedModeState
+	// can re-scan without requiring the caller to pass it each time.
+	storedClientset kubernetes.Interface
+
+	// namespaceRefreshReader, when non-nil, is used by RefreshMixedModeState to list
+	// namespaces from the controller-runtime cache (mgr.GetClient()) instead of
+	// a direct API list on storedClientset — reducing apiserver load on the 30s
+	// refresh ticker. Set via SetMixedModeNamespaceRefreshReader from cmd after
+	// controllers are registered on the manager.
+	namespaceRefreshReader client.Reader
+	refreshReaderMu        sync.RWMutex
+)
+
+var log = logger.Log
+
+// checkPerNamespaceProvidersSupported fetches the SupervisorCapabilities object and
+// returns whether per-namespace network providers are activated. It retries
+// all errors with exponential backoff (starting at retryInitialInterval,
+// doubling each attempt, capped at retryMaxInterval). The SupervisorCapabilities
+// CR is guaranteed to exist; all failures are treated as transient (e.g. API
+// server not yet ready at operator startup). Returns false only when the
+// context is cancelled.
+func checkPerNamespaceProvidersSupported(ctx context.Context, dynClient dynamic.Interface) bool {
+	interval := retryInitialInterval
+	for {
+		obj, err := dynClient.Resource(supervisorCapabilitiesGVR).Get(
+			ctx, supervisorCapabilitiesName, metav1.GetOptions{})
+		if err == nil {
+			return extractCapability(obj)
+		}
+		log.Info("Failed to get SupervisorCapabilities, will retry", "error", err, "retryIn", interval)
+		select {
+		case <-ctx.Done():
+			log.Info("Context cancelled while waiting for SupervisorCapabilities, falling back to legacy config")
+			return false
+		case <-time.After(interval):
+		}
+		interval = min(interval*2, retryMaxInterval)
+	}
+}
+
+func extractCapability(obj *unstructured.Unstructured) bool {
+	status, found, err := unstructured.NestedMap(obj.Object, "status")
+	if err != nil || !found {
+		return false
+	}
+	services, found, err := unstructured.NestedMap(status, "services")
+	if err != nil || !found {
+		return false
+	}
+	for _, svcCaps := range services {
+		capsMap, ok := svcCaps.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		cap, ok := capsMap["supports_per_namespace_network_providers"]
+		if !ok {
+			continue
+		}
+		capMap, ok := cap.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		activated, ok := capMap["activated"]
+		if ok {
+			if b, ok := activated.(bool); ok && b {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func namespaceHasVPCNetworkConfig(ns *v1.Namespace) bool {
+	if ns == nil {
+		return false
+	}
+	v := strings.TrimSpace(ns.Annotations[VPCNetworkConfigAnnotation])
+	return v != ""
+}
+
+func accumulateMixedModeFlagsFromNamespaces(items []v1.Namespace) (hasT1 bool, hasVPC bool) {
+	for i := range items {
+		if namespaceHasVPCNetworkConfig(&items[i]) {
+			hasVPC = true
+		} else {
+			hasT1 = true
+		}
+	}
+	return hasT1, hasVPC
+}
+
+func scanNamespaceProviders(ctx context.Context, clientset kubernetes.Interface) (hasT1 bool, hasVPC bool, err error) {
+	nsList, err := clientset.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return false, false, err
+	}
+	hasT1, hasVPC = accumulateMixedModeFlagsFromNamespaces(nsList.Items)
+	return hasT1, hasVPC, nil
+}
+
+func scanNamespaceProvidersWithClient(ctx context.Context, reader client.Reader) (hasT1 bool, hasVPC bool, err error) {
+	nsList := &v1.NamespaceList{}
+	if err := reader.List(ctx, nsList); err != nil {
+		return false, false, err
+	}
+	hasT1, hasVPC = accumulateMixedModeFlagsFromNamespaces(nsList.Items)
+	return hasT1, hasVPC, nil
+}
+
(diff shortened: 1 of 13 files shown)


pkg/config/mixed_mode.go
```diff
@@ -0,0 +1,215 @@
...
+	hasT1Namespaces                bool
+	hasVPCNamespaces               bool
+	perNamespaceProvidersSupported *bool
+	stateInitialized               bool
+	stateLog                       *zap.SugaredLogger
+)
+
+func init() {
+	zapLogger, _ := zap.NewProduction()
+	stateLog = zapLogger.Sugar()
```
yanjunz97: why not use the existing log like logger.Log as other components?

pkg/config/mixed_mode.go
```diff
@@ -0,0 +1,215 @@
...
+		}
+	}
+	return false
+}
+
+func scanNamespaceLabels(ctx context.Context, clientset kubernetes.Interface) (hasT1 bool, hasVPC bool) {
+	nsList, err := clientset.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		stateLog.Errorf("Failed to list namespaces for mixed-mode state detection: %v", err)
+		return false, false
```
yanjunz97: Shall we return an error here as list ns failure does not mean there is no vpc/t1 namespace?

pkg/config/mixed_mode.go
heypnus: This was copied from https://github.com/vmware-tanzu/nsx-operator/blob/main/pkg/config/config.go#L56 .. Changed it to `logger.Log`.

pkg/config/mixed_mode.go
```diff
@@ -0,0 +1,215 @@
...
+)
+
+func init() {
+	zapLogger, _ := zap.NewProduction()
+	stateLog = zapLogger.Sugar()
+}
+
+func checkPerNamespaceProvidersSupported(ctx context.Context, dynClient dynamic.Interface) bool {
+	obj, err := dynClient.Resource(supervisorCapabilitiesGVR).Get(
+		ctx, supervisorCapabilitiesName, metav1.GetOptions{})
```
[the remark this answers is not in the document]
heypnus: We don't need it, because the clusterrolebinding wcp:authenticated:capabilities binded the clusterrole wcp:capabilities-view to group system:authenticated, the NCP uses the serviceaccount so it has been able to access the supervisorcapibilities.

pkg/config/mixed_mode.go
```diff
@@ -0,0 +1,215 @@
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			stateLog.Info("SupervisorCapabilities CRD not found; " +
+				"falling back to legacy config")
+		} else {
+			stateLog.Infof("Failed to get SupervisorCapabilities: %v; "+
+				"falling back to legacy config", err)
+		}
```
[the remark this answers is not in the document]
heypnus: That's a good question, Tao also had the same question on the NCP patch. I changed the implementation to wait until we got the CR info, because in supervisor >= 9.2, we can assume that the CR always exists.

pkg/config/mixed_mode.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+	for _, ns := range nsList.Items {
+		provider := ns.Labels[NetworkProviderLabel]
+		switch provider {
+		case ProviderNSXT1:
+			hasT1 = true
+		case ProviderNSXVPC:
+			hasVPC = true
+		case ProviderVSphereNetwork:
+			hasVPC = true
```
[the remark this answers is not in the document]
heypnus: No, it's a code error. Thanks for finding it!

pkg/config/mixed_mode.go
heypnus: updated `scanNamespaceLabels` to return error

pkg/config/mixed_mode.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	stateInitialized = true
+	stateLog.Infof("Mixed-mode state initialized: HasT1Namespaces=%v, "+
+		"HasVPCNamespaces=%v", hasT1Namespaces, hasVPCNamespaces)
+}
+
+// RefreshMixedMode re-scans namespace labels and updates state.
+// Returns true if the state changed (caller should consider restarting).
+func RefreshMixedMode(ctx context.Context, clientset kubernetes.Interface) bool {
```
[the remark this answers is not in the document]
heypnus: Added the function `RefreshMixedModeState` as the invoker.

pkg/config/mixed_mode.go
```diff
@@ -0,0 +1,215 @@
...
+// IsMixedModeStateInitialized returns true after InitMixedMode has been called.
+func IsMixedModeStateInitialized() bool {
+	stateMu.RLock()
+	defer stateMu.RUnlock()
+	return stateInitialized
+}
+
+// SetMixedModeStateForTest sets hasT1Namespaces and hasVPCNamespaces for unit tests.
+// Must only be used from test code so production always goes through InitMixedMode.
+func SetMixedModeStateForTest(hasT1, hasVPC bool) {
```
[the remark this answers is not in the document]
heypnus: I'm afraid we cannot because this function was invoked by some other packages, e.g. `firewall_test.go`, `securitypolicy_controller_test.go`. If we put it in `mixed_mode_test.go`, then it only can be invoked inside the config package.

pkg/nsx/services/securitypolicy/parse.go
```diff
@@ -62,8 +63,11 @@ func getDefaultProjectDomain() string {
 	return "default"
 }
 
+// IsVPCEnabled returns whether VPC namespaces exist. Callers must ensure mixed-mode
+// state has been initialized (InitMixedMode in main; SetMixedModeStateForTest in tests).
+// The "no CR" fallback to config is handled inside InitMixedMode, not here.
 func IsVPCEnabled(service *SecurityPolicyService) bool {
```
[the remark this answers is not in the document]
heypnus: This function `IsVPCEnabled` is used in the security policy module, I want to put the further change in the follow-up patch specific for the security policy. This patch only wants to remove all codes to check `NSXConfig.EnableVPCNetwork`

pkg/nsx/util/license.go
```diff
@@ -69,8 +81,9 @@ func IsLicensed(feature string) bool {
 	return licenseMap[feature]
 }
 
+// SetEnableVpcNetwork is deprecated; mixed-mode state is now managed by
+// config.InitMixedMode. Kept for backward compatibility.
 func SetEnableVpcNetwork(vpcNetwork bool) {
```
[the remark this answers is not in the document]
heypnus: Yes. But I double evaluated and there's may be not outside call, so removed it in the new patch set. Thanks for pointing it out.

pkg/nsx/util/license_test.go
```diff
@@ -188,59 +188,55 @@ func TestUpdateFeatureLicense(t *testing.T) {
 }
```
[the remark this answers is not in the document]
heypnus: Removed the invokers.

pkg/config/mixed_mode.go
```diff
@@ -0,0 +1,319 @@
...
+	return v != ""
+}
+
+func accumulateMixedModeFlagsFromNamespaces(items []v1.Namespace) (hasT1 bool, hasVPC bool) {
+	for i := range items {
+		if namespaceHasVPCNetworkConfig(&items[i]) {
+			hasVPC = true
+		} else {
+			hasT1 = true
+		}
```
yanjunz97, line 114: Maybe we can break when both hasVPC and hasT1 are true

pkg/config/mixed_mode.go
```diff
@@ -0,0 +1,319 @@
...
+
+// RefreshMixedModeState re-scans namespaces using the clientset stored during
+// InitMixedMode and updates the global state. Returns true if the state
+// changed; the caller should then restart the operator so that VPC services
+// and controllers are initialized for the new mode.
+func RefreshMixedModeState(ctx context.Context) bool {
+	stateMu.Lock()
+	defer stateMu.Unlock()
+
+	if perNamespaceProvidersSupported == nil || !*perNamespaceProvidersSupported {
```
yanjunz97: Would you know if perNamespaceProvidersSupported support day2 update?

heypnus: Not sure for now because the capability's implementation is still in progress. I added a TODO for it.

pkg/config/mixed_mode.go
heypnus: Done, thanks.

yanjunz97 (dismissed):
LGTM

pkg/config/mixed_mode.go
```diff
@@ -0,0 +1,326 @@
...
+	v := strings.TrimSpace(ns.Annotations[VPCNetworkConfigAnnotation])
+	return v != ""
+}
+
+func accumulateMixedModeFlagsFromNamespaces(items []v1.Namespace) (hasT1 bool, hasVPC bool) {
+	for i := range items {
+		if namespaceHasVPCNetworkConfig(&items[i]) {
+			hasVPC = true
+		} else {
+			hasT1 = true
```
dantingl, line 124: For VDS case, nsx-operator will also be running. At the beginning, there is no VPC annotation, but it should not be T1 mode, right?

pkg/config/mixed_mode.go
```diff
@@ -0,0 +1,326 @@
...
+	stateMu                        sync.RWMutex
+	hasT1Namespaces                bool
+	hasVPCNamespaces               bool
+	perNamespaceProvidersSupported *bool
+	stateInitialized               bool
+
+	// retryInitialInterval and retryMaxInterval control the exponential
+	// backoff used when a transient error prevents reading
+	// SupervisorCapabilities or listing namespaces. Overridable in tests.
+	retryInitialInterval = 2 * time.Second
```
dantingl, line 48: Why don't we watch SupervisorCapability event but poll it every 2 seconds?

heypnus: This is for not the regular poll. It occurs during the init stage. If the NSX Operator can get the supervisor capability state successfully, then we won't retry; otherwise, retry in backoff. After init stage, we won't loop check the supervisor capability because I think it's not needed.

pkg/config/mixed_mode.go
heypnus: Added the VDS case handling in the new patchset.

pkg/config/mixed_mode.go
```diff
@@ -0,0 +1,376 @@
...
+		})
+		interval := retryInitialInterval
+		for {
+			if _, err := mgr.GetCache().GetInformer(context.TODO(), nsObj); err != nil {
+				log.Info("Failed to start informer for NetworkSettings, will retry", "error", err, "retryIn", interval)
+				time.Sleep(interval)
+				interval = min(interval*2, retryMaxInterval)
+			} else {
+				log.Info("Started informer for NetworkSettings successfully")
+				break
```
TaoZou1: should we query if the informer status before assign the  namespaceRefreshReader?
for {
					if informer.HasSynced() {
						log.Info("Informer for NetworkSettings synced successfully; enabling cache reader")
						refreshReaderMu.Lock()
						namespaceRefreshReader = mgr.GetClient()
						refreshReaderMu.Unlock()
						break
					}
					time.Sleep(1 * time.Second)
				}

heypnus: good catch! fixed, thanks.

TaoZou1 (approved):

pkg/config/mixed_mode.go
```diff
@@ -0,0 +1,383 @@
...
+func extractCapability(obj *unstructured.Unstructured) bool {
+	status, found, err := unstructured.NestedMap(obj.Object, "status")
+	if err != nil || !found {
+		return false
+	}
+	supervisor, found, err := unstructured.NestedMap(status, "supervisor")
+	if err != nil || !found {
+		return false
+	}
+	cap, ok := supervisor["supports_per_namespace_network_provider"]
```
dantingl, line 102: Add a constant for supports_per_namespace_network_provider?

dantingl (approved):

pkg/config/mixed_mode.go
heypnus: Thanks for the suggestion, will enhance it in the next patch.

PR #1409, opened 2026-04-15
Add per-namespace VPC filtering for VPC-only controllers

heypnus:
Introduce config.IsVPCNamespace() to decide whether a namespace is
served by VPC-only controllers: in mixed mode (when per-namespace
providers are supported) it checks for the presence of the
'nsx.vmware.com/vpc_network_config' annotation; in legacy mode,
it falls back to the cluster-wide HasVPCNamespaces() flag.

controllers/common: add VPCNamespacePredicate and register it with
WithEventFilter on VPC-only controllers so non-VPC namespace
creates and updates are dropped before the work queue; Delete events
stay allowed for cleanup

controllers/namespace: Reconcile skips non-VPC namespaces

Testing Done:
VPC pipeline id 20270
T1 pipeline id 5835

### DIFF
--- a/pkg/controllers/common/namespace_filter.go
+++ b/pkg/controllers/common/namespace_filter.go
@@ -1,0 +1,69 @@
+/* Copyright © 2026 Broadcom, Inc. All Rights Reserved.
+   SPDX-License-Identifier: Apache-2.0 */
+
+package common
+
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/config"
+)
+
+// isVPCNamespaceByName fetches the Namespace by name and calls config.IsVPCNamespace.
+// Returns true when the namespace cannot be fetched (transient error or already
+// gone) so the Reconcile loop can decide what to do.
+func isVPCNamespaceByName(c client.Reader, ns string) bool {
+	namespace := &corev1.Namespace{}
+	if err := c.Get(context.Background(), types.NamespacedName{Name: ns}, namespace); err != nil {
+		if !apierrors.IsNotFound(err) {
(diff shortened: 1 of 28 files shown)


pkg/controllers/common/namespace_filter.go
```diff
@@ -0,0 +1,72 @@
...
+//   - Delete: always allowed so the controller can clean up any existing NSX
+//     resources even if the namespace is already gone.
+//
+// The namespace check is skipped for cluster-scoped resources (empty namespace),
+// which are always allowed through.
+func VPCNamespacePredicate(c client.Reader) predicate.Funcs {
+	isVPCNs := func(ns string) bool {
+		if ns == "" {
+			// Cluster-scoped resource: no per-namespace filtering.
+			return true
```
dantingl: Do we need to check cluster scope config here?

heypnus: I think shouldn't check cluster scope. For example, if we check config.HasVPCNamespaces() here and return false when there are no VPC namespaces, we will drop events for cluster-scoped VPC prerequisites (like VPCNetworkConfiguration).
If we should check if a cluster scoped resource belongs to VPC or T1, we should check it in the reconcile loop.

dantingl: HasVPCNamespaces or cluster scope VPC should be true if vpc_network in NCP ConfigMap is true, right?

heypnus: Changed `return true` to `return config.HasVPCNamespaces()` as defensive programming.

pkg/controllers/common/namespace_filter.go
```diff
@@ -0,0 +1,75 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/config"
+)
+
+// isVPCNamespaceByName fetches the Namespace by name and calls config.IsVPCNamespace.
+// Returns true when the namespace cannot be fetched (transient error or already
+// gone) so the Reconcile loop can decide what to do.
+func isVPCNamespaceByName(c client.Reader, ns string) bool {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
dantingl: why add 30 seconds timeout

heypnus: Thanks for bringing this up! Since c.Get reads directly from the in-memory Informer cache and doesn't make network calls, I have simplified it to context.Background() to keep our code consistent with repo conventions.

dantingl (approved):
LGTM

TaoZou1 (approved):

PR #1418, opened 2026-04-27
CRD changes to support IPv6 part 2

yanjunz97:
The PR includes the CRD changes to support IPv6 for
SubnetPort, IPAddressAllocation and SubnetIPReservation.

### DIFF
--- a/build/yaml/crd/vpc/crd.nsx.vmware.com_subnetports.yaml
+++ b/build/yaml/crd/vpc/crd.nsx.vmware.com_subnetports.yaml
@@ -67,6 +67,29 @@ spec:
                       type: string
                   type: object
                 type: array
+              interfaceIPType:
+                description: |-
+                  InterfaceIPType decides the address families of static IP allocation, when
+                  DHCP or SLAAC is not activated on the Subnet. It will be ignored when
+                  StaticIPAllocationType is set.
+                enum:
+                - IPv4
+                - IPv6
+                - IPv4IPv6
+                type: string
+              staticIPAllocationType:
+                description: |-
+                  StaticIPAllocationType explicitly requests static IP allocation of the
+                  specified the address families. In a mixed-mode Subnet (where both DHCP
+                  and static allocation are enabled), use this to define which families
+                  should be allocated from the static IP pools. This field is only valid
+                  when static allocation is enabled on the Subnet.
+                enum:
+                - IPv4
+                - IPv6
+                - IPv4IPv6
+                - None
+                type: string
               subnet:
                 description: Subnet defines the parent Subnet name of the SubnetPort.
                 type: string
--- a/pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
+++ b/pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
@@ -6,11 +6,14 @@ package v1alpha1
 import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 type IPAddressVisibility string
+type IPAllocationAddressType string
 
 var (
-	IPAddressVisibilityExternal   IPAddressVisibility = "External"
-	IPAddressVisibilityPrivate    IPAddressVisibility = "Private"
-	IPAddressVisibilityPrivateTGW IPAddressVisibility = "PrivateTGW"
+	IPAddressVisibilityExternal   IPAddressVisibility     = "External"
+	IPAddressVisibilityPrivate    IPAddressVisibility     = "Private"
+	IPAddressVisibilityPrivateTGW IPAddressVisibility     = "PrivateTGW"
+	IPAllocationIPAddressTypeIPv4 IPAllocationAddressType = "IPv4"
+	IPAllocationIPAddressTypeIPv6 IPAllocationAddressType = "IPv6"
 )
 
 // +genclient
@@ -21,6 +24,7 @@ var (
 // IPAddressAllocation is the Schema for the IP allocation API.
 // +kubebuilder:printcolumn:name="IPAddressBlockVisibility",type=string,JSONPath=`.spec.ipAddressBlockVisibility`,description="IPAddressBlockVisibility of IPAddressAllocation"
 // +kubebuilder:printcolumn:name="AllocationIPs",type=string,JSONPath=`.status.allocationIPs`, description="AllocationIPs for the IPAddressAllocation"
+// +kubebuilder:printcolumn:name="IPv6AllocationPrefixLength",type=string,JSONPath=`.spec.ipv6AllocationPrefixLength`,description="IPv6 allocation prefix length of the IPAddressAllocation."
 type IPAddressAllocation struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata"`
@@ -40,16 +44,21 @@ type IPAddressAllocationList struct {
 
 // IPAddressAllocationSpec defines the desired state of IPAddressAllocation.
 // +kubebuilder:validation:XValidation:rule="!has(self.allocationSize) || !has(self.allocationIPs)", message="Only one of allocationSize or allocationIPs can be specified"
+// +kubebuilder:validation:XValidation:rule="!has(self.ipv6AllocationPrefixLength) || !has(self.allocationIPs)", message="Only one of ipv6AllocationPrefixLength or allocationIPs can be specified"
 // +kubebuilder:validation:XValidation:rule="!has(oldSelf.allocationSize) || has(self.allocationSize)", message="allocationSize is required once set"
 // +kubebuilder:validation:XValidation:rule="!has(oldSelf.allocationIPs) || has(self.allocationIPs)", message="allocationIPs is required once set"
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.ipv6AllocationPrefixLength) || has(self.ipv6AllocationPrefixLength)", message="ipv6AllocationPrefixLength is required once set"
+// +kubebuilder:validation:XValidation:rule="!has(self.allocationSize) || !has(self.ipAddressType) || self.ipAddressType == 'IPv4'", message="allocationSize can only be set when ipAddressType is IPv4"
+// +kubebuilder:validation:XValidation:rule="!has(self.ipv6AllocationPrefixLength) || self.ipAddressType == 'IPv6'", message="ipv6AllocationPrefixLength can only be set when ipAddressType is IPv6"
 type IPAddressAllocationSpec struct {
 	// IPAddressBlockVisibility specifies the visibility of the IPBlocks to allocate IP addresses. Can be External, Private or PrivateTGW.
+	// This field is not applicable if ipAddressType is IPv6.
 	// +kubebuilder:validation:Enum=External;Private;PrivateTGW
 	// +kubebuilder:default=Private
 	// +optional
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
 	IPAddressBlockVisibility IPAddressVisibility `json:"ipAddressBlockVisibility,omitempty"`
-	// AllocationSize specifies the size of allocationIPs to be allocated.
+	// AllocationSize specifies the size of IPv4 allocationIPs to be allocated.
 	// It should be a power of 2.
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
 	// +kubebuilder:validation:Minimum:=1
@@ -57,6 +66,16 @@ type IPAddressAllocationSpec struct {
 	// AllocationIPs specifies the Allocated IP addresses in CIDR or single IP Address format.
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
 	AllocationIPs string `json:"allocationIPs,omitempty"`
+	// IPv6AllocationPrefixLength specifies the prefix length of IPv6 addresses.
+	// +kubebuilder:validation:Minimum:=64
+	// +kubebuilder:validation:Maximum:=128
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	IPv6AllocationPrefixLength int `json:"ipv6AllocationPrefixLength,omitempty"`
+	// IPAddressType specifies the IP address type of the IPAddressAllocation.
+	// +kubebuilder:validation:Enum=IPv4;IPv6
+	// +kubebuilder:default=IPv4
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	IPAddressType IPAllocationAddressType `json:"ipAddressType,omitempty"`
 }
 
 // IPAddressAllocationStatus defines the observed state of IPAddressAllocation.
--- a/pkg/apis/vpc/v1alpha1/subnetport_types.go
+++ b/pkg/apis/vpc/v1alpha1/subnetport_types.go
@@ -7,6 +7,15 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type StaticIPAllocationType string
+
+const (
+	StaticIPAllocationTypeIPv4     StaticIPAllocationType = "IPv4"
+	StaticIPAllocationTypeIPv6     StaticIPAllocationType = "IPv6"
+	StaticIPAllocationTypeIPv4IPv6 StaticIPAllocationType = "IPv4IPv6"
+	StaticIPAllocationTypeNone     StaticIPAllocationType = "None"
+)
+
 // +kubebuilder:validation:XValidation:rule="!has(self.subnetSet) || !has(self.subnet)",message="Only one of subnet or subnetSet can be specified or both set to empty in which case default SubnetSet for VM will be used"
 // SubnetPortSpec defines the desired state of SubnetPort.
 type SubnetPortSpec struct {
@@ -16,6 +25,18 @@ type SubnetPortSpec struct {
 	SubnetSet string `json:"subnetSet,omitempty"`
 	// AddressBindings defines static address bindings used for the SubnetPort.
 	AddressBindings []PortAddressBinding `json:"addressBindings,omitempty"`
+	// InterfaceIPType decides the address families of static IP allocation, when
+	// DHCP or SLAAC is not activated on the Subnet. It will be ignored when
+	// StaticIPAllocationType is set.
+	// +kubebuilder:validation:Enum=IPv4;IPv6;IPv4IPv6
+	InterfaceIPType IPAddressType `json:"interfaceIPType,omitempty"`
+	// StaticIPAllocationType explicitly requests static IP allocation of the
+	// specified the address families. In a mixed-mode Subnet (where both DHCP
+	// and static allocation are enabled), use this to define which families
+	// should be allocated from the static IP pools. This field is only valid
+	// when static allocation is enabled on the Subnet.
+	// +kubebuilder:validation:Enum=IPv4;IPv6;IPv4IPv6;None
+	StaticIPAllocationType StaticIPAllocationType `json:"staticIPAllocationType,omitempty"`
 }
 
 // PortAddressBinding defines static addresses for the Port.
--- a/build/yaml/crd/vpc/crd.nsx.vmware.com_ipaddressallocations.yaml
+++ b/build/yaml/crd/vpc/crd.nsx.vmware.com_ipaddressallocations.yaml
@@ -23,6 +23,10 @@ spec:
       jsonPath: .status.allocationIPs
       name: AllocationIPs
       type: string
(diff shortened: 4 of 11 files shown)


build/yaml/crd/vpc/crd.nsx.vmware.com_subnetports.yaml
```diff
@@ -67,6 +67,24 @@ spec:
                       type: string
                   type: object
                 type: array
+              ipAddressType:
```
dantingl: Need Jianjun's review here if we should use 'ipfamilyPolicy' as VirtualMachine CRD

pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
```diff
@@ -44,19 +44,30 @@ type IPAddressAllocationList struct {
...
 	// +kubebuilder:validation:Minimum:=1
 	AllocationSize int `json:"allocationSize,omitempty"`
 	// AllocationIPs specifies the Allocated IP addresses in CIDR or single IP Address format.
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
 	AllocationIPs string `json:"allocationIPs,omitempty"`
+	// AllocationPrefixLength specifies the prefix length of IPv6 addresses.
+	// +kubebuilder:validation:Minimum:=64
+	// +kubebuilder:validation:Maximum:=128
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	AllocationPrefixLength int `json:"allocationPrefixLength,omitempty"`
```
jianjuns: In NSX API, it is just called AllocationPrefixLength, no v6 in the name?

pkg/apis/vpc/v1alpha1/subnetport_types.go
```diff
@@ -16,6 +16,14 @@ type SubnetPortSpec struct {
...
 	// AddressBindings defines static address bindings used for the SubnetPort.
 	AddressBindings []PortAddressBinding `json:"addressBindings,omitempty"`
+	// IPAddressType defines the IP address type of the SubnetPort.
+	// +kubebuilder:validation:Enum=IPV4;IPV6;IPV4IPV6
+	IPAddressType string `json:"ipAddressType,omitempty"`
+	// StaticIPAllocationType defines the address family for static IP allocation.
+	// Use this on dual-stack subnets or on subnets in mixed mode (DHCP and static allocation both enabled)
+	// to specify which IP families should be allocated from the static IP pool.
+	// +kubebuilder:validation:Enum=IPV4;IPV6;IPV4IPV6;None
+	StaticIPAllocationType string `json:"staticIPAllocationType,omitempty"`
```
jianjuns: Why we need this one?

pkg/apis/vpc/v1alpha1/subnetport_types.go
```diff
@@ -16,6 +16,14 @@ type SubnetPortSpec struct {
 	SubnetSet string `json:"subnetSet,omitempty"`
 	// AddressBindings defines static address bindings used for the SubnetPort.
 	AddressBindings []PortAddressBinding `json:"addressBindings,omitempty"`
+	// IPAddressType defines the IP address type of the SubnetPort.
+	// +kubebuilder:validation:Enum=IPV4;IPV6;IPV4IPV6
+	IPAddressType string `json:"ipAddressType,omitempty"`
```
jianjuns: It should be called "IPAddressTypes"?

In NSX APIs, we generally use "Type" or "Family" in field names?

pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
yanjunz97: The patch for vpcIPAddressAllocation change is not out yet. I use the name from their design doc
Currently they plan to use this approach 2.

pkg/apis/vpc/v1alpha1/subnetport_types.go
yanjunz97: This is needed for VCD migration to specify the exact static ip allocation type they need on a mixed mode subnet. This is also aligned with the NSX api.

pkg/apis/vpc/v1alpha1/subnetport_types.go
yanjunz97: NSX API uses ip_address_type for Subnet/VPCIPAddressAllocation/Subnet IPReservation...
So I use IPAddressType to align with other NSX resources.
As Danting mentioned above, VM operator ueses ipfamilyPolicy. Would you think it's better to align with NSX or VM operator?

pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
jianjuns: I think NSX API should either change the name or make it work for both v4 and v6 (the latter makes more sense to me). Let us monitor NSX API changes.

pkg/apis/vpc/v1alpha1/subnetport_types.go
jianjuns: Let us align with NSX then.

But here it should be called `IPAddressTypes`?

pkg/apis/vpc/v1alpha1/subnetport_types.go
jianjuns: Humm.. in this case, probably we should rename `IPAddressTypes` to something like "InterfaceIPTypePolicy" and also comment that decides the auto-allocation types when DHCP is not activated on the Subnet; and this one will override the auto-strategy.

yanjunz97: Sure, updated the name to InterfaceIPTypePolicy for IPAddressType. I also mention SLAAC in comment as when DHCP is de-activated on Subnet but SLAAC is enabled, this auto-allocation also does not work. Please help check if it's reasonable. Thanks!

pkg/apis/vpc/v1alpha1/subnetport_types.go
yanjunz97: Updated as InterfaceIPTypePolicy as discussed above, thanks

pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
jianjuns: I requested Policy folks to consider the adjustment. Again, let us monitor the final design.

pkg/apis/vpc/v1alpha1/subnetport_types.go
```diff
@@ -16,6 +16,17 @@ type SubnetPortSpec struct {
 	SubnetSet string `json:"subnetSet,omitempty"`
 	// AddressBindings defines static address bindings used for the SubnetPort.
 	AddressBindings []PortAddressBinding `json:"addressBindings,omitempty"`
+	// InterfaceIPTypePolicy decides the SubnetPort auto-allocation types when
+	// DHCP or SLAAC is not activated on the Subnet. It will be ignored when
+	// StaticIPAllocationType is set.
+	// +kubebuilder:validation:Enum=IPV4;IPV6;IPV4IPV6
+	InterfaceIPTypePolicy string `json:"interfaceIPTypePolicy,omitempty"`
+	// StaticIPAllocationType defines the address family for static IP allocation.
+	// Use this on dual-stack subnets or on subnets in mixed mode (DHCP and static allocation both enabled)
```
jianjuns: subnets -> Subnets

I think it is not really about "dual-stack Subnets", as that can be covered by InterfaceIPTypePolicy; only for mixed mode?

jianjuns: Could you wrap lines at a fixed length.

pkg/apis/vpc/v1alpha1/subnetport_types.go
```diff
@@ -16,6 +16,17 @@ type SubnetPortSpec struct {
...
+	// to specify which IP families should be allocated from the static IP pool.
```
jianjuns: pool(s)

pkg/apis/vpc/v1alpha1/subnetport_types.go
```diff
@@ -16,6 +16,17 @@ type SubnetPortSpec struct {
...
+	// If StaticIPAllocationType is set, InterfaceIPTypePolicy will be ignored.
```
jianjuns: Should we add it can only be set when the Subnet has static allocation enabled?

pkg/apis/vpc/v1alpha1/subnetport_types.go
yanjunz97: Updated, thanks!

pkg/apis/vpc/v1alpha1/subnetport_types.go
yanjunz97: Updated, thanks!
I've one question regarding this. Pre-created Subnets in SubnetSet can have different static allocation status. If the SubnetPort has StaticIPAllocationType set, but the first available Subnet on this SubnetSet has static ip allocation disabled. Shall we skip the Subnet to find the next one or shall we fail the SubnetPort creation?

pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
jianjuns: The NSX API spec is updated to add "v6" in the name. Could you check?

pkg/apis/vpc/v1alpha1/subnetport_types.go
```diff
@@ -16,6 +16,18 @@ type SubnetPortSpec struct {
...
+	// DHCP or SLAAC is not activated on the Subnet. It will be ignored when
+	// StaticIPAllocationType is set.
+	// +kubebuilder:validation:Enum=IPV4;IPV6;IPV4IPV6
+	InterfaceIPTypePolicy string `json:"interfaceIPTypePolicy,omitempty"`
+	// StaticIPAllocationType defines the address family for static IP allocation.
+	// Use this on Subnets in mixed mode (DHCP and static allocation both enabled)
+	// to specify which IP families should be allocated from the static IP pool(s).
+	// It can only be set when the Subnet has static allocation enabled.
+	// If StaticIPAllocationType is set, InterfaceIPTypePolicy will be ignored.
+	// +kubebuilder:validation:Enum=IPV4;IPV6;IPV4IPV6;None
```
jianjuns: I think the values should be IPv4, IPv6, IPv4IPv6

pkg/apis/vpc/v1alpha1/subnetport_types.go
```diff
@@ -16,6 +16,18 @@ type SubnetPortSpec struct {
 	SubnetSet string `json:"subnetSet,omitempty"`
 	// AddressBindings defines static address bindings used for the SubnetPort.
 	AddressBindings []PortAddressBinding `json:"addressBindings,omitempty"`
+	// InterfaceIPTypePolicy decides the SubnetPort auto-allocation types when
+	// DHCP or SLAAC is not activated on the Subnet. It will be ignored when
+	// StaticIPAllocationType is set.
+	// +kubebuilder:validation:Enum=IPV4;IPV6;IPV4IPV6
+	InterfaceIPTypePolicy string `json:"interfaceIPTypePolicy,omitempty"`
```
jianjuns: Question in VM Service API, the values of ipFamilyPolicy are just: IPv4, IPv6, DualStack; not something like: PreferDualStack right?

I am thinking should we just name the field: InterfaceIPType.

pkg/apis/vpc/v1alpha1/subnetport_types.go
```diff
@@ -16,6 +16,18 @@ type SubnetPortSpec struct {
 	SubnetSet string `json:"subnetSet,omitempty"`
 	// AddressBindings defines static address bindings used for the SubnetPort.
 	AddressBindings []PortAddressBinding `json:"addressBindings,omitempty"`
+	// InterfaceIPTypePolicy decides the SubnetPort auto-allocation types when
```
jianjuns: decides the IP families of static IP allocation, when DHCP or SLAAC is not activated on the Subnet.

pkg/apis/vpc/v1alpha1/subnetport_types.go
```diff
@@ -16,6 +16,18 @@ type SubnetPortSpec struct {
...
+	// DHCP or SLAAC is not activated on the Subnet. It will be ignored when
+	// StaticIPAllocationType is set.
+	// +kubebuilder:validation:Enum=IPV4;IPV6;IPV4IPV6
+	InterfaceIPTypePolicy string `json:"interfaceIPTypePolicy,omitempty"`
+	// StaticIPAllocationType defines the address family for static IP allocation.
```
jianjuns: family -> families

pkg/apis/vpc/v1alpha1/subnetport_types.go
jianjuns: SubnetSet supports only static allocation and DHCP server right?

Could we restrict all Subnets must share the same settings to make it simpler? Or if we do want to support different settings, we can go with what you described.

pkg/apis/vpc/v1alpha1/subnetport_types.go
```diff
@@ -16,6 +16,18 @@ type SubnetPortSpec struct {
...
+	// Use this on Subnets in mixed mode (DHCP and static allocation both enabled)
+	// to specify which IP families should be allocated from the static IP pool(s).
+	// It can only be set when the Subnet has static allocation enabled.
+	// If StaticIPAllocationType is set, InterfaceIPTypePolicy will be ignored.
```
jianjuns: I think no need to add a new line here.

pkg/apis/vpc/v1alpha1/subnetport_types.go
yanjunz97: It is to align with NSX. NSX supports IPV4/IPV6/IPV4_IPV6/NONE for static_ip_allocation_type. None refers to do not allocate from the static IP pools (e.g., rely entirely on DHCP).

pkg/apis/vpc/v1alpha1/subnetport_types.go
yanjunz97: Yes, the exact value proposed by vm operator is IPv4Only/IPv6Only/DualStack.
Sure, I'll rename it.

pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
yanjunz97: Thanks, updated.

pkg/apis/vpc/v1alpha1/subnetport_types.go
yanjunz97: > SubnetSet supports only static allocation and DHCP server right?

Yes, and we plan to support the mixed mode Subnet (static+DHCP Server) in SubnetSet in 9.2

> Could we restrict all Subnets must share the same settings to make it simpler? Or if we do want to support different settings, we can go with what you described.

Sure we'll restrict all Subnets must share the same settings to make it simpler

pkg/apis/vpc/v1alpha1/subnetport_types.go
jianjuns: We should follow K8s convention for constants, and I believe that means IPv4, IPv6, IPv4IPv6, None. If you do not believe so, could you check with other K8s APIs?

yanjunz97: Oh I misunderstood the previous comment. I see. IPv4, IPv6, IPv4IPv6, None, looks good to me. I also updated other occurrence of the ip address type value; and defines type instead string for these values. Please take another look. Thanks!

pkg/apis/vpc/v1alpha1/subnetport_types.go
```diff
@@ -16,6 +25,18 @@ type SubnetPortSpec struct {
 	SubnetSet string `json:"subnetSet,omitempty"`
 	// AddressBindings defines static address bindings used for the SubnetPort.
 	AddressBindings []PortAddressBinding `json:"addressBindings,omitempty"`
+	// InterfaceIPType decides the IP families of static IP allocation, when
```
jianjuns: Nit: you used "address families" in the comments of `StaticIPAllocationType`. Pls be consistent.

pkg/apis/vpc/v1alpha1/subnetport_types.go
```diff
@@ -16,6 +25,18 @@ type SubnetPortSpec struct {
...
+	// DHCP or SLAAC is not activated on the Subnet. It will be ignored when
+	// StaticIPAllocationType is set.
+	// +kubebuilder:validation:Enum=IPv4;IPv6;IPv4IPv6
+	InterfaceIPType IPAddressType `json:"interfaceIPType,omitempty"`
+	// StaticIPAllocationType defines the address families for static IP
```
jianjuns: How about: "explicitly requests static IP allocation of the specified the address families"

pkg/apis/vpc/v1alpha1/subnetport_types.go
```diff
@@ -16,6 +25,18 @@ type SubnetPortSpec struct {
...
+	// allocation. Use this on Subnets in mixed mode (DHCP and static allocation
```
jianjuns: Gemini suggests the following:

In a mixed-mode Subnet (where both DHCP and static allocation are enabled), use this to define which families should be allocated from the static IP pools. This field is only valid when static allocation is enabled on the Subnet.

jianjuns (commented):
LGTM

pkg/apis/vpc/v1alpha1/subnetport_types.go
yanjunz97: Updated to address families, thanks

pkg/apis/vpc/v1alpha1/subnetport_types.go
```diff
@@ -16,6 +25,18 @@ type SubnetPortSpec struct {
 	SubnetSet string `json:"subnetSet,omitempty"`
 	// AddressBindings defines static address bindings used for the SubnetPort.
 	AddressBindings []PortAddressBinding `json:"addressBindings,omitempty"`
+	// InterfaceIPType decides the address families of static IP allocation, when
+	// DHCP or SLAAC is not activated on the Subnet. It will be ignored when
+	// StaticIPAllocationType is set.
+	// +kubebuilder:validation:Enum=IPv4;IPv6;IPv4IPv6
+	InterfaceIPType IPAddressType `json:"interfaceIPType,omitempty"`
```
yanjunz97, line 32: Hi @jianjuns , I discussed with @dantingl and would like to confirm the workflow for SubnetPort on mixed-mode dual stack Subnet with you

For the case user would like to allocate IPv4 ip from static pool and IPv6 ip from dhcp pool on a Subnet with static ip allocation enabled + dhcp enabled + dhcpv6 enabled, I think the current workflow is:
- The user only specifies StaticIPAllocationType as IPv4 in SubnetPort
- NSX Operator sets both dhcp(v4) flag and dhcpv6 flag to true according to the Subnet
- VM Operator needs to check the ipFamily of the interface, the SubnetPort StaticIPAllocationType, and SubnetPort dhcp/dhcpv6 flag to determine if the guest needs to request dhcp ip for v4/v6

We are thinking about another workflow:
- The user sets StaticIPAllocationType to IPv4, and interfaceIPType to IPv4IPv6.
- NSX Operator sets dhcp(v4) flag to false and dhcpv6 flag to true on the SubnetPort status, as only v6 address needs to be requested from dhcp server 
- VM Operator only needs to check dhcp/dhcpv6 flag to determine if the guest needs to request dhcp ip for v4/v6

I think the difference is who is responsible to compare interface ip family with StaticIPAllocationType to determine if guest needs to be allocated ip from dhcp server. If we follow the later workflow, we may need to update the description of InterfaceIPType to just indicate it specifies the  address families of the interface.

Could you share you thoughts on this?

Besides, I just learned from vm operator team that the final name they plan to use for interface ip family would be ipamModes, accepting an array like [IPv4], [IPv6], [IPv4, IPv6]. Please let me know if you would like to update our field name InterfaceIPType according to this. Thanks!

jianjuns: I thought the DHCP and DHCPv6 flags on SubnetPort Status just means whether the Subnet has DHCP/DHCPv6 enabled, not about whether DHCP should be configured in the guest. VM Operator can decide itself whether DHCP should be configured based on the IPFamilyPolicy and whether there are static IP(s) allocated/specified.

Basically, I am saying why nsx-op cannot blindly pass the Subnet DHCP/DHCPv6 flags to SubnetPort Status, without checking any other fields/state?

yanjunz97: I see. We'll discuss with vm operator team with this. Let's keep the current CRD definition. Thanks!

dantingl (dismissed):

dantingl (dismissed):

lxiaopei (approved):

dantingl (approved):

PR #1420, opened 2026-04-28
nsxserviceaccount: backoff NSX calls after CCP connection capacity error 610139

liu4480:
When ClusterControlPlanesClient.Update returns NSX error 610139 (connection 
capacity full), record a global timestamp. While a 5-minute window from that 
timestamp is still in effect, NSXServiceAccount reconcile requeues without
calling NSX. After the window expires, reconcile proceeds normally; a
successful CCP update clears the timestamp, and another 610139 refreshes it.
Add IsCCPConnectionCapacityFullError in pkg/nsx/util. Fix controller tests that
assumed deletion state on objects returned from the fake client.

### DIFF
--- a/pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
+++ b/pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -33,7 +34,7 @@ import (
 	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
 	"github.com/vmware-tanzu/nsx-operator/pkg/nsx"
 	servicecommon "github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
-	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/nsxserviceaccount"
+	nsxsasvc "github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/nsxserviceaccount"
 )
 
 const proxyLabelKey = "mgmt-proxy.antrea-nsx.vmware.com"
@@ -114,6 +115,8 @@ var (
 )
 
 // NSXServiceAccountReconciler reconciles a NSXServiceAccount object.
+// After NSX error 610139 (cluster control plane connection capacity full), reconciles may
+// requeue without calling NSX until a shared backoff window expires.
 // Requires NSXT 4.0.1
 //
 // create/delete event will be processed by Reconcile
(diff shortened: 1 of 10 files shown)


pkg/metrics/metrics.go
Atish-iaf, line 1: Is this change related to CCP connection capacity limit ?

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -149,6 +152,12 @@ func (r *NSXServiceAccountReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ResultNormal, client.IgnoreNotFound(err)
 	}
 
+	if requeueAfter, skip := nsxsasvc.CCPConnectionCapacityFullBackoffRequeue(time.Now()); skip {
```
Atish-iaf: This skips realized NSXServiceAccount reconcile as well. Can we skip only for unrealized NSXServiceAccount by moving this block past `IsNSXServiceAccountRealized` ?

Atish-iaf (approved):

Atish-iaf (commented):
approved by mistake

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
liu4480: thanks, fixed

pkg/metrics/metrics.go
liu4480: Yes, nsx-operator call CCPN API for registration, if capacity is full, we need to avoid nsx-operator to access this NSX API too frequently

pkg/controllers/nsxserviceaccount/nsxserviceaccount_controller.go
```diff
@@ -186,6 +187,12 @@ func (r *NSXServiceAccountReconciler) Reconcile(ctx context.Context, req ctrl.Re
 			r.StatusUpdater.UpdateSuccess(ctx, obj, updateNSXServiceAccountStatus)
 			return ResultNormal, nil
 		}
+
+		if requeueAfter, skip := nsxsasvc.CCPConnectionCapacityFullBackoffRequeue(time.Now()); skip {
```
Atish-iaf, line 191: Suppose CCPn capacity is full and there are many realized NSXSAs and few unrealized NSXSAs due to CCPn capacity full.
Then NSX Restore happens, CCPn capacity is less than full, all NSXSAs reconcile. CCPn capacity becomes full again and few old realized NSXSAs reconcile last after CCPn capacity is full, they call the `RestoreRealizedNSXServiceAccount` method above which will hit NSX api to create CCP without checking CCPn capacity. They get error and reconcile again. They don't skip reconcile while CCPn capacity backoff is active.

liu4480: during restore, nsx-operator can not access NSX manager. Once restore finishes
1. If the restore does not revert any cluster, then no PI is deleted, NSX Operator will not think the realized NSXServiceAccount CRs are unrealized, nor will it fail to reconcile them. So the unrealized nsxserviceacocunt will not realize
2. If the store reverts some clusters, then the PI will be deleted. For realized nsxserviceaccount A(PI deleted) and unrealized nsxserviceaccount B, it depend which is enqueued first. We should ensure A will be enqueued first.

PR #1429, opened 2026-05-09
Add ipaddressallocation restore for tepless mode

TaoZou1:
To avoid the ipaddressallocation IP used by subnet, need to restore it first

Test Done
Test case 1:
1. delete the vpc qe using nsx API
2. backup from the nsx
3. stop the ncp
4. cleanup the nsx
5. Create ns test from VC side
qe
loadBalancerIPAddresses: 192.168.0.7 
test
loadBalancerIPAddresses: 192.168.0.33
6. restore from the nsx
7. restart the ncp
8. check if the vpc qe/test restored after restore complete, and if the ip is the same
9. Create a new svc under qe, check if the svc has been created and the vip could be accessed

### DIFF
--- a/pkg/nsx/services/common/types.go
+++ b/pkg/nsx/services/common/types.go
@@ -120,6 +120,11 @@ const (
 	SubnetPortGCInterval = 60 * time.Second
 	DefaultSNATID        = "DEFAULT"
 	AVISubnetLBID        = "_services"
+	// LBServiceIPAllocationID is the fixed NSX resource ID for the VPC-level LB service IP allocation used in tepless mode.
+	// It's only used for backup/restore
+	LBServiceIPAllocationID = "_DEFAULT--VPC_SERVICE_IP"
+	TagScopeVPCService      = "services/vpc_service"
+	TagValueUserSpecifiedIP = "__DEFAULT--VPC_SERVICE_IP__"
 
 	NSXServiceAccountFinalizerName = "nsxserviceaccount.nsx.vmware.com/finalizer"
 	T1SecurityPolicyFinalizerName  = "securitypolicy.nsx.vmware.com/finalizer"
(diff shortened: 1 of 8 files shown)


pkg/nsx/services/common/types.go
```diff
@@ -120,6 +120,10 @@ const (
 	SubnetPortGCInterval = 60 * time.Second
 	DefaultSNATID        = "DEFAULT"
 	AVISubnetLBID        = "_services"
+	// LBServiceIPAllocationID is the fixed NSX resource ID for the VPC-level LB service IP allocation used in tepless mode.
```
dantingl, line 123: explain it is only for B/R

pkg/nsx/services/common/types.go
```diff
@@ -120,6 +120,10 @@ const (
...
+	LBServiceIPAllocationID = "_DEFAULT--VPC_SERVICE_IP"
+	TagScopeVPCService      = "services/vpc_service"
+	TagValueUserSpecifiedIP = "__USER_SPECIFIED_IP__"
```
yanjunz97: Could you double check if the final choice is __DEFAULT--VPC_SERVICE_IP__ or __USER_SPECIFIED_IP__?

dantingl: __DEFAULT--VPC_SERVICE_IP__ for ipv4?

TaoZou1: Thanks. Updated

pkg/nsx/services/common/types.go
TaoZou1: Done

pkg/nsx/services/common/types.go
TaoZou1: yes, _DEFAULT–VPC_SERVICE_IP_V6 for v6

dantingl (approved):

yanjunz97 (approved):

PR #1431, opened 2026-05-11
Enable restorevif through config flag

yanjunz97:
This PR refers to a field `restore_vif` in configmap to determine
if restore the SubnetPort with unchanged vif.

Testing done:
Set restore_vif to true, observed the SubnetPort is restored with unchanged vif.
Set restore_vif to false, observed the SubnetPort is restored with changed vif.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -129,6 +129,8 @@ type NsxConfig struct {
 	// VpcWcpEnhance controls StatefulSet pod SubnetPort behavior together with NSX version.
 	// When omitted (nil), treated as false; only an explicit true enables the enhancement path.
 	VpcWcpEnhance *bool `ini:"vpc_wcp_enhance"`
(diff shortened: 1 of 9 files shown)


pkg/config/config.go
```diff
@@ -129,6 +129,8 @@ type NsxConfig struct {
 	// VpcWcpEnhance controls StatefulSet pod SubnetPort behavior together with NSX version.
 	// When omitted (nil), treated as false; only an explicit true enables the enhancement path.
 	VpcWcpEnhance *bool `ini:"vpc_wcp_enhance"`
+	// RestoreVif controls whether SubnetPort vif should be restored during restore mode.
+	RestoreVif *bool `ini:"restore_vif"`
```
heypnus, line 133: Should we also add the option in config.py of NCP?
And could you please explain a little about when (and how to) enable it?

dantingl (approved):

heypnus (approved):

PR #1432, opened 2026-05-11
NetworkInfo support for IPv6 and dual stack

poojav25:
The LoadBalancerBackendIPs field was declared in VPCState but never populated. This change wires it up for all three LB paths:

- NSX LB non-TEP-less: read all IPs from ExtendedAttributes["IpAddresses"].Values on the realized interface via new GetPolicyInterfaceIPs helper (both IPv4+IPv6).
- NSX LB TEP-less: best-effort fetch of both _DEFAULT--VPC_SERVICE_IP (IPv4) and _DEFAULT--VPC_SERVICE_IP_V6 (IPv6) allocations; NotFound is silently skipped so VNA-only, VNS-IPv4-only, VNS-IPv6-only, and VNS-dual-stack all work correctly.
- AVI: collect NetworkAddress from every VpcSubnetStatus result (one per IP family for dual-stack) instead of only Results[0].

Testing done:

→ LoadBalancerBackendIPs populated (NSX LB non-TEP-less, CTGW testbed)
```go
kubectl get networkinfo -n test-lb-backend test-lb-backend -o json | \
  python3 -c "
import json, sys
v = json.load(sys.stdin)['vpcs'][0]
print('LoadBalancerIPAddresses:', v.get('loadBalancerIPAddresses',''))
print('LoadBalancerBackendIPs: ', v.get('loadBalancerBackendIPs',[]))
"
LoadBalancerIPAddresses: 100.64.0.1
LoadBalancerBackendIPs:  ['100.64.0.1']
``` 
-> NSX API source confirmed — operator reads gateway-interface realized entity:
```go
curl -sku admin:"..." \
  "https://<NSX>/policy/api/v1/infra/realized-state/realized-entity?\
realized_path=/orgs/default/projects/.../realized-state/enforcement-points/\
default/vpcs/test-lb-backend_cmd98/gateway-interface" \
  | python3 -m json.tool | grep -A5 "IpAddresses"
    "key": "IpAddresses",
    "values": [
        "100.64.0.1/31"
    ]
```


dantingl:

pkg/nsx/services/realizestate/realize_state.go
yanjunz97: Maybe remove this function as it is not used any more?

yanjunz97 (commented):
Overall LGTM, just a nit

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -463,7 +471,26 @@ func (r *NetworkInfoReconciler) getNSXLBSNATIP(nc *v1alpha1.VPCNetworkConfigurat
...
+// primaryLBIP selects the canonical single IP/CIDR for LoadBalancerIPAddresses following
+// the Kubernetes dual-stack convention: prefer IPv4 for dual-stack, fall back to the first
+// entry for IPv6-only, and return "" for an empty slice.
+func primaryLBIP(ips []string) string {
+	for _, ip := range ips {
+		raw := ip
+		if idx := strings.Index(ip, "/"); idx != -1 {
+			raw = ip[:idx]
+		}
+		if net.ParseIP(raw) != nil && net.ParseIP(raw).To4() != nil {
```
wenqiq: Nit: Can we store the result of calling `net.ParseIP(raw)` in a local variable instead of calling it repeatedly here?

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -463,7 +471,26 @@ func (r *NetworkInfoReconciler) getNSXLBSNATIP(nc *v1alpha1.VPCNetworkConfigurat
...
+			return ip
+		}
+	}
+	if len(ips) > 0 {
+		return ips[0]
```
wenqiq: The IPv6-only fallback uses `ips[0]` without checking if it is a valid IP address. If the list contains invalid data, this could potentially write an invalid string into `LoadBalancerIPAddresses`?

yanjunz97 (approved):
LGTM

wenqiq (approved):
LGTM

PR #1433, opened 2026-05-15
Downgrade golang version

TaoZou1:
Downgrade golang version to align with nsx-ujo 1.25.8 
Will upgrade it to 1.26.2 later

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/vmware-tanzu/nsx-operator
 
-go 1.26.2
+go 1.25.8
 
 replace (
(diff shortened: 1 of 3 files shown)


dantingl (approved):

seanpang-vmware (changes requested):
I will submit patch on NCP side to upgrade golang since this is also a CVE.

if we revert golang here and then upgrade NCP golang first, it will also cause conflict anyway

TaoZou1:
no need to downgrade golang version

TaoZou1:
reopen for align with VKS golang version

seanpang-vmware (approved):

PR #1439, opened 2026-05-21
Support IPv6 for SubnetPort 

yanjunz97:
This PR focuses on happy path for IPv6/Dual Stack in SubnetPort
It contains the following changes
- Support multiple status.networkInterfaceConfig.ipAddresses and spec.addressBindings in SubnetPort
- Support StaticIPAllocationType in SubnetPort
- Support AllocatePortFromSubnet for IPv6 and dual stack

Testing done:
Tested regression case for Pod/VM with IPv4
Tested SubnetPort/VM creation in dual stack testbed

### DIFF
--- a/pkg/controllers/subnetport/subnetport_controller.go
+++ b/pkg/controllers/subnetport/subnetport_controller.go
@@ -106,7 +106,7 @@ func (r *SubnetPortReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		r.StatusUpdater.IncreaseUpdateTotal()
 
 		old_status := subnetPort.Status.DeepCopy()
-		isExisting, isParentResourceTerminating, nsxSubnetPath, subnetSetUID, subnetSetLock, err := r.CheckAndGetSubnetPathForSubnetPort(ctx, subnetPort)
+		isExisting, isParentResourceTerminating, nsxSubnetPath, subnetSetUID, subnetSetLock, interfaceIPType, err := r.CheckAndGetSubnetPathForSubnetPort(ctx, subnetPort)
 		if subnetSetLock != nil {
 			defer common.RUnlockSubnetSet(*subnetSetUID, subnetSetLock)
 		}
@@ -120,7 +120,7 @@ func (r *SubnetPortReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 			return common.ResultRequeue, err
 		}
 		if !isExisting {
-			defer r.SubnetPortService.ReleasePortInSubnet(nsxSubnetPath)
+			defer r.SubnetPortService.ReleasePortInSubnet(nsxSubnetPath, interfaceIPType)
 		}
 
 		var labels *map[string]string
@@ -152,6 +152,10 @@ func (r *SubnetPortReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 			r.StatusUpdater.UpdateFail(ctx, subnetPort, err, fmt.Sprintf("Failed to get Subnet by path: %s", nsxSubnetPath), setSubnetPortReadyStatusFalse, r.SubnetPortService, r.restoreMode)
 			return common.ResultRequeue, err
 		}
+		err = r.updateSubnetPortIPType(ctx, subnetPort, interfaceIPType, nsxSubnet)
+		if err != nil {
+			return common.ResultNormal, err
+		}
 
 		isVmSubnetPort := true
 		if value, exists := subnetPort.Labels[servicecommon.LabelImageFetcher]; exists && value == "true" {
@@ -167,7 +171,7 @@ func (r *SubnetPortReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 			r.StatusUpdater.UpdateFail(ctx, subnetPort, err, "Failed to create NSX IPAddressAllocation for AddressBinding restore", setSubnetPortReadyStatusFalse, r.SubnetPortService, r.restoreMode)
 			return common.ResultRequeue, err
 		}
-		nsxSubnetPortState, enableDHCP, err := r.SubnetPortService.CreateOrUpdateSubnetPort(subnetPort, nsxSubnet, "", labels, isVmSubnetPort, r.restoreMode)
+		nsxSubnetPortState, err := r.SubnetPortService.CreateOrUpdateSubnetPort(subnetPort, nsxSubnet, "", labels, isVmSubnetPort, r.restoreMode, interfaceIPType)
 		if err != nil {
 			r.StatusUpdater.UpdateFail(ctx, subnetPort, err, "", setSubnetPortReadyStatusFalse, r.SubnetPortService, r.restoreMode)
 			if nsxutil.IsRealizeStateError(err) {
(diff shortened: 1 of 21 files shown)


pkg/nsx/services/subnet/subnet.go
```diff
@@ -528,11 +539,21 @@ func (service *SubnetService) UpdateSubnetSet(ns string, vpcSubnets []*model.Vpc
 		// is only updated after the updating succeeds.
 		updatedSubnet := *vpcSubnets[i] // #nosec G602
 		updatedSubnet.Tags = newTags
+		updatedSubnet.IpAddressType = String(controllercommon.ConvertCRIPAddressTypeToNSX(subnetsetCR.Spec.IPAddressType))
```
poojav25, line 558: IpAddressType is overwritten unconditionally on every UpdateSubnetSet call. If spec.ipAddressType is not set (empty string), ConvertCRIPAddressTypeToNSX defaults to "IPV4", silently changing the address type of existing IPv6 or dual-stack NSX subnets ?

yanjunz97: I think this shall be fine as here we only update Subnet created by NSX Operator.
And before we introduce ipAddressType to Subnet CRD, NSX Operator only creates IPv4 Subnet, so there is no existing IPv6 or dual-stack NSX subnets will be updated to IPv4 in this case?

pkg/nsx/services/subnetport/subnetport.go
```diff
@@ -461,16 +460,68 @@ func (service *SubnetPortService) ListSubnetPortByStsUid(ns string, stsUid strin
...
+			return false, err
+		}
+	}
+
+	// Increment counters for successful allocation
+	switch subnetIPAddressType {
+	case model.VpcSubnet_IP_ADDRESS_TYPE_IPV4_IPV6:
+		info.dirtyCount += 1
+		info.dirtyCountIPv6 += 1
+		log.Trace("Allocate Subnetport to dual-stack Subnet", "Subnet", *subnet.Path, "dirtyPortCount", info.dirtyCount, "dirtyPortCountIPv6", info.dirtyCountIPv6)
```
wenqiq: When using dual-stack allocation, both dirtyCount and dirtyCountIPv6 will be incremented simultaneously. However, it seems that ReleasePortInSubnet only decrements dirtyCount?

pkg/nsx/services/subnetport/subnetport.go
wenqiq, line 662: Does `ReleasePortInSubnet` require to know the `IPAddressType`, and perform the corresponding count for IPv6-only / dual-stack decrement?

pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -191,13 +191,34 @@ func (r *SubnetPortReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+					// The MAC address is updated here when the SubnetPort's StaticIPAllocation
+					// is enabled or spec.AddressBindings is specific. For the other cases, the MAC
+					// address will be updated in the VIF polling.
+					macAddress := ""
+					for i, binding := range nsxSubnetPortState.RealizedBindings {
+						if binding.Binding != nil && binding.Binding.IpAddress != nil {
+							if macAddress == "" && binding.Binding.MacAddress != nil {
+								macAddress = strings.Trim(*binding.Binding.MacAddress, "\"")
+							}
+							subnetPort.Status.NetworkInterfaceConfig.IPAddresses[i].IPAddress = *binding.Binding.IpAddress
```
wenqiq: Writes values by index here, Could it possibly panic or use the wrong address family?

pkg/nsx/services/subnetport/subnetport.go
yanjunz97: Good caught. Yes, ReleasePortInSubnet needs to consume ipAddressType and decrement dirtyCount accordingly.

pkg/controllers/subnetport/subnetport_controller.go
yanjunz97: By design we should not get out of index error, but it makes sense to defense here. I changed it to append when it is out of index and record in log.

poojav25:
LGTM

poojav25 (dismissed):

wenqiq:
The e2e test failed.

pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -191,13 +195,43 @@ func (r *SubnetPortReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+							}
+							if len(subnetPort.Status.NetworkInterfaceConfig.IPAddresses) <= i {
+								// TODO: revisit this when supporting multiple addressbindings per IPAddressType
+								log.Warn("More IPs are realized on SubnetPort", "Namespace", subnetPort.Namespace, "SubnetPort", subnetPort.Name, "RealizedBindings", nsxSubnetPortState.RealizedBindings)
+								subnetPort.Status.NetworkInterfaceConfig.IPAddresses = append(
+									subnetPort.Status.NetworkInterfaceConfig.IPAddresses,
+									v1alpha1.NetworkInterfaceIPAddress{Gateway: "", IPAddress: *binding.Binding.IpAddress},
+								)
+							} else {
+								subnetPort.Status.NetworkInterfaceConfig.IPAddresses[i].IPAddress = *binding.Binding.IpAddress
```
wenqiq, line 229: Dose the NSX API guarantee the order of bindings (IPv4 vs IPv6)? If array order of RealizedBindings is unstable, the array order might flip between reconciles, triggering unnecessary CR status updates.

pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -924,19 +1002,82 @@ func (r *SubnetPortReconciler) CheckAndGetSubnetPathForSubnetPort(ctx context.Co
...
+					break
+				}
+			}
+		}
+	}
+
+	// Second pass: Process empty IPs and empty Gateways
+	// For DHCP case, the IPAddress will be empty. We set the unused Gateway in IPAddresses
+	for _, ipConfig := range subnetPort.Status.NetworkInterfaceConfig.IPAddresses {
+		if len(ipConfig.IPAddress) == 0 && ipConfig.Gateway == "" {
```
wenqiq, line 1076: Is there such a scenario where ipConfig.IPAddress is empty, but ipConfig.Gateway is not empty?

yanjunz97: Yes, when DHCP is enabled on Subnet, the SubnetPort does not have the IPAddress when created and we still need to update the gateway on the SubnetPort.

pkg/controllers/subnetport/subnetport_controller.go
yanjunz97: We only fetch SubnetPort state when NSX SubnetPort config is update so there might not be many unnecessary update. But yes, we're not sure if the array order may change. Sort the IPAddresses in updateSubnetStatusOnSubnetPort to keep the order.

pkg/controllers/subnetport/subnetport_controller.go
wenqiq: My concern is that, in that case, this will lead a reduction in the length of the IPAddresses array (for example, in dual-stack mode, there should have been 2 elements, but ended up with only 1, and the IPv6 Gateway was lost)? During the next Reconcile process, the Controller will notice that the length is insufficient and will append an empty IPAddress, which will then be discarded again, thus causing an infinite Reconcile loop (Infinite Update Loop).

yanjunz97: I see, you're right. That's a missed corner case, updated, thanks!

wenqiq (dismissed):

poojav25 (approved):

wenqiq (approved):

PR #1440, opened 2026-05-22
Support ExternalIPBlocksConfigured check for IPv6

yanjunz97:
ExternalIPBlocksConfigured should be checked based on Supervisor IPFamily.
- For IPv4, checks ExternalIpBlocks
- For IPv6, checks Ipv6Blocks
- For dualstack, checks both ExternalIpBlocks and Ipv6Blocks 

Testing done:
Set Supervisor IPFamily to dualstack and IPv6, observed the default VPC Connectivity Profile results false in ExternalIPBlocksConfigured.

```
  - lastTransitionTime: "2026-05-22T07:32:43Z"
    message: No External IP Blocks exist in VPC Connectivity Profile
    reason: ExternalIPBlockMissingInProfile
    status: "False"
    type: ExternalIPBlocksConfigured
```

Add ipv6 blocks to default VPC Connectivity Profile, observed the ExternalIPBlocksConfigured is updated
```
  - lastTransitionTime: "2026-05-22T08:43:54Z"
    status: "True"
    type: ExternalIPBlocksConfigured
```

### DIFF
--- a/pkg/controllers/networkinfo/networkinfo_controller.go
+++ b/pkg/controllers/networkinfo/networkinfo_controller.go
@@ -358,7 +358,14 @@ func (r *NetworkInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 		// Check external IP blocks on system VPC network config.
 		if ncName == commonservice.SystemVPCNetworkConfigurationName {
-			hasExternalIPs := len(vpcConnectivityProfile.ExternalIpBlocks) > 0
+			ipFamily := r.Service.NSXConfig.K8sConfig.GetIPAddressType()
+			hasExternalIPs := true
+			if util.IPAddressTypeIncludesIPv4(ipFamily) {
+				hasExternalIPs = len(vpcConnectivityProfile.ExternalIpBlocks) > 0
+			}
+			if util.IPAddressTypeIncludesIPv6(ipFamily) {
+				hasExternalIPs = hasExternalIPs && len(vpcConnectivityProfile.Ipv6Blocks) > 0
+			}
 			setVPCNetworkConfigurationStatusWithNoExternalIPBlock(ctx, r.Client, systemVpcNetCfg, hasExternalIPs)
 			if !hasExternalIPs && !retryWithSystemVPC {
 				log.Error(err, "There is no ExternalIPBlock in VPC ConnectivityProfile", "NetworkInfo", req.NamespacedName)
--- a/pkg/controllers/networkinfo/networkinfo_controller_test.go
+++ b/pkg/controllers/networkinfo/networkinfo_controller_test.go
@@ -782,91 +782,6 @@ func TestNetworkInfoReconciler_Reconcile(t *testing.T) {
 			wantErr: false,
 		},
 		{
-			name: "VPCNetworkConfigurationStatusWithNoExternalIPBlockInSystemVPC",
-			prepareFunc: func(t *testing.T, r *NetworkInfoReconciler, ctx context.Context) (patches *gomonkey.Patches) {
-				assert.NoError(t, r.Client.Create(ctx, &v1alpha1.NetworkInfo{
-					ObjectMeta: metav1.ObjectMeta{
-						Namespace: requestArgs.req.Namespace,
-						Name:      requestArgs.req.Name,
-					},
-				}))
-				assert.NoError(t, r.Client.Create(ctx, &v1alpha1.VPCNetworkConfiguration{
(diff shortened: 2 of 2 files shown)


pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -358,7 +358,16 @@ func (r *NetworkInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			ipFamily := r.Service.NSXConfig.K8sConfig.GetIPAddressType()
+			hasExternalIPs := false
+			switch ipFamily {
+			case v1alpha1.IPAddressTypeIPv4:
+				hasExternalIPs = len(vpcConnectivityProfile.ExternalIpBlocks) > 0
+			case v1alpha1.IPAddressTypeIPv6:
+				hasExternalIPs = len(vpcConnectivityProfile.Ipv6Blocks) > 0
+			case v1alpha1.IPAddressTypeIPv4IPv6:
+				hasExternalIPs = len(vpcConnectivityProfile.ExternalIpBlocks) > 0 && len(vpcConnectivityProfile.Ipv6Blocks) > 0
+			}
```
wenqiq, line 368: How about:
```
hasExternalIPs := true
if util.IPAddressTypeIncludesIPv4(ipFamily) {
    hasExternalIPs = len(vpcConnectivityProfile.ExternalIpBlocks) > 0
}
if util.IPAddressTypeIncludesIPv6(ipFamily) {
    hasExternalIPs = hasExternalIPs && len(vpcConnectivityProfile.Ipv6Blocks) > 0
}
```
The same file `updateDefaultSubnetSet` has used `util.IPAddressTypeIncludesIPv4/IPv6`; it can be modified to follow the same pattern to avoid redundant switch statements and align with the default behavior of ParseIPFamily.
Additionally, the current switch keeps `hasExternalIPs == false` for unlisted IPAddressTypes, which may lead to false positives?Using the helper treats unknown types as IPv4, consistent with `config.ParseIPFamily`.

pkg/controllers/networkinfo/networkinfo_controller_test.go
```diff
@@ -866,6 +866,118 @@ func TestNetworkInfoReconciler_Reconcile(t *testing.T) {
...
+				patches.ApplyFunc(setVPCNetworkConfigurationStatusWithSnatEnabled, func(_ context.Context, _ client.Client, _ *v1alpha1.VPCNetworkConfiguration, _ bool) {})
+				patches.ApplyFunc(setNSNetworkReadyCondition, func(ctx context.Context, client client.Client, nsName string, condition *corev1.NamespaceCondition) {
+					require.True(t, nsConditionEquals(*condition, *nsMsgVPCNoExternalIPBlock.getNSNetworkCondition()))
+				})
+				return patches
+			},
+			args:    requestArgs,
+			want:    common.ResultRequeueAfter60sec,
+			wantErr: false,
+		},
```
wenqiq: These two newly added test cases have some code duplication. Could we extract some common code into functions to reduce the maintenance cost?

pkg/controllers/networkinfo/networkinfo_controller.go
yanjunz97: ParseIPFamily can always convert the unlisted ipFamily to IPv4 so it should be fine with switch.
But I do not see drawback of this if condition solution. Updated, thanks!

pkg/controllers/networkinfo/networkinfo_controller_test.go
yanjunz97: I put all cases related to external ipblock check in a separate ut. It shall be clearer now. Thanks!

wenqiq (approved):
LGTM. I have a question about dual-stack mode, which requires both `ExternalIpBlocks` and `Ipv6Blocks`. If the product allows for a gradual deployment process where v4 is configured first and then v6, will the current implementation keep `ExternalIPBlocksConfigured=False` and continually requeue?

yanjunz97:
> LGTM. I have a question about dual-stack mode, which requires both `ExternalIpBlocks` and `Ipv6Blocks`. If the product allows for a gradual deployment process where v4 is configured first and then v6, will the current implementation keep `ExternalIPBlocksConfigured=False` and continually requeue?

I think currently we require both v4 and v6 for dual stack. Maybe we can do this update when there is initiative to support this gradual deployment process

poojav25:
LGTM

poojav25 (approved):

PR #1443, opened 2026-05-27
Support cross-VPC SubnetConnectionBindingMap

wenqiq:
Support cross-VPC SubnetConnectionBindingMap via Trunk-Branch association

This PR introduces support for cross-VPC SubnetConnectionBindingMap
by leveraging the Trunk-Branch Association API. It allows a CHILD
(Branch) Subnet in one VPC to extend a PARENT (Trunk) Subnet's VLAN
network into another VPC, enabling cross-VPC connectivity.

TestDone:

1. Prepare Namespaces and VPCs Create two Namespaces: ns-vpc-a (for the Parent/Trunk) and ns-vpc-b (for the Child/Branch). Ensure the Operator has created or associated VPCs for them on the system side.

2. Prepare the Parent Subnet Create the Parent Subnet in ns-vpc-a. It should attach to an existing Distributed VLAN Connection. 
```
kind: Subnet
metadata:
  name: parent-vlan-subnet
  namespace: ns-vpc-a
spec:
  accessMode: Private
  ipv4SubnetSize: 16
  vlanConnectionName: gatewayconnection1
```
3. Prepare the Child Subnet (Cross-VPC) Create the Child Subnet in ns-vpc-b.
```
kind: Subnet
metadata:
  name: child-vlan-subnet
  namespace: ns-vpc-b
spec:
  accessMode: Private
  ipv4SubnetSize: 16
  # vlanConnectionName: gatewayconnection1
```
4. Create the SubnetConnectionBindingMap CR in the Parent Subnet's Namespace (ns-vpc-a), specifying subnetAssociation: BRANCH.
```
metadata:
  name: cross-vpc-binding
  namespace: ns-vpc-a
spec:
  subnetName: parent-vlan-subnet         
  targetSubnetName: child-vlan-subnet        
  subnetAssociation: Branch               
  vlanTrafficTag: 201
```
5. Check CR Status

```
root@422c72f507918f80fe61efcdc8407eb3 [ ~ ]# k get subnetconnectionbindingmap cross-vpc-binding -n ns-vpc-a
NAME                SUBNET               TARGETSUBNET        TARGETSUBNETSET   VLANTRAFFICTAG   SUBNETASSOCIATION
cross-vpc-binding   parent-vlan-subnet   child-vlan-subnet                     201              Branch
k get subnetconnectionbindingmap cross-vpc-binding -n ns-vpc-a -oyaml
metadata:
  annotations:
      {"apiVersion":"crd.nsx.vmware.com/v1alpha1","kind":"SubnetConnectionBindingMap","metadata":{"annotations":{},"name":"cross-vpc-binding","namespace":"ns-vpc-a"},"spec":{"subnetAssociation":"Branch","subnetName":"parent-vlan-subnet","targetSubnetName":"child-vlan-subnet","vlanTrafficTag":201}}
  creationTimestamp: "2026-07-08T13:28:24Z"
  generation: 1
  name: cross-vpc-binding
  namespace: ns-vpc-a
  resourceVersion: "7637782"
  uid: e217e213-32ca-4870-9888-6cfd92435986
spec:
  subnetAssociation: Branch
  subnetName: parent-vlan-subnet
  targetSubnetName: child-vlan-subnet
  vlanTrafficTag: 201
status:
  conditions:
  - lastTransitionTime: "2026-07-08T13:28:25Z"
    status: "True
```

6. Verify Resources in system

```
curl -k -u "${NSX_USER}:${NSX_PASS}"   -X GET   -H "Content-Type: application/json"   "${NSX_MANAGER}policy/api/v1/orgs/default/projects/project-quality/vpcs/ns-vpc-a_i5e9h/subnets/parent-vlan-subnet_i5e9h/subnet-connection-binding-maps"
{
  "results" : [ {
    "vlan_traffic_tag" : 201,
    "subnet_path" : "/orgs/default/projects/project-quality/vpcs/ns-vpc-b_ld78t/subnets/child-vlan-subnet_ld78t",
    "subnet_association" : "BRANCH",
    "resource_type" : "SubnetConnectionBindingMap",
    "id" : "cross-vpc-binding_c0uz1",
    "display_name" : "cross-vpc-binding",
    "tags" : [ {
      "scope" : "nsx-op/cluster",
      "tag" : "9995e6cf-e8fa-4970-b9a1-30edb607a709"
    }, {
      "scope" : "nsx-op/version",
      "tag" : "1.0.0"
    }, {
      "scope" : "nsx-op/namespace",
      "tag" : "ns-vpc-a"
    }, {
      "scope" : "nsx-op/subnetbinding_name",
      "tag" : "cross-vpc-binding"
    }, {
      "scope" : "nsx-op/subnetbinding_uid",
      "tag" : "e217e213-32ca-4870-9888-6cfd92435986"
    } ],
    "path" : "/orgs/default/projects/project-quality/vpcs/ns-vpc-a_i5e9h/subnets/parent-vlan-subnet_i5e9h/subnet-connection-binding-maps/cross-vpc-binding_c0uz1",
    "relative_path" : "cross-vpc-binding_c0uz1",
    "parent_path" : "/orgs/default/projects/project-quality/vpcs/ns-vpc-a_i5e9h/subnets/parent-vlan-subnet_i5e9h",
    "remote_path" : "",
    "unique_id" : "57de1a3c-8b4f-48a4-802d-1605c334fee7",
    "realization_id" : "57de1a3c-8b4f-48a4-802d-1605c334fee7",
    "owner_id" : "7de74739-f9fa-442d-b3d2-8bc33b171db9",
    "marked_for_delete" : false,
    "overridden" : false,
    "_system_owned" : false,
    "_protection" : "REQUIRE_OVERRIDE",
    "_create_time" : 1783517305113,
    "_create_user" : "wcp-cluster-user-9995e6cf-e8fa-4970-b9a1-30edb607a709-b3816e2f-166a-43e6-9a3e-013de621c9af",
    "_last_modified_time" : 1783517305113,
    "_last_modified_user" : "wcp-cluster-user-9995e6cf-e8fa-4970-b9a1-30edb607a709-b3816e2f-166a-43e6-9a3e-013de621c9af",
    "_revision" : 0
  } ],
  "result_count" : 1,
  "sort_by" : "display_name",
  "sort_ascending" : true
}
```


pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
```diff
@@ -17,6 +26,17 @@ type SubnetConnectionBindingMapSpec struct {
 	// TargetSubnetName specifies the target Subnet which a Subnet is connected to.
 	// +kubebuilder:validation:Optional
 	TargetSubnetName string `json:"targetSubnetName,omitempty"`
+	// TargetSubnetNamespace is the Namespace of targetSubnetName. When unset, the SubnetConnectionBindingMap
+	// Namespace is used. Required for cross-VPC bindings when the target Subnet lives in another Namespace.
+	// +kubebuilder:validation:Optional
+	TargetSubnetNamespace string `json:"targetSubnetNamespace,omitempty"`
```
dantingl: I think the target Subnet CR should also under the same namespace, a NSX subnet under a different VPC can be shared to this namespace

wenqiq: updated

wenyingd: Do we need this field `targetSubnetNamespace` in the SubnetConnectionBindingMap CR?  @dantingl @wenqiq

pkg/controllers/namespace/subnet_share_controller.go
```diff
@@ -369,17 +369,22 @@ func (r *NamespaceReconciler) checkSubnetReferences(ctx context.Context, ns stri
 		return true, nil
 	}
 
-	// Check if there are any SubnetConnectionBindingMap CRs referencing this Subnet CR
+	// Check if there are any SubnetConnectionBindingMap CRs referencing this Subnet CR as host or target.
 	subnetBindingList := &v1alpha1.SubnetConnectionBindingMapList{}
-	err = r.Client.List(ctx, subnetBindingList, client.InNamespace(ns), client.MatchingFields{"spec.subnetName": subnet.Name})
-	if err != nil {
+	if err = r.Client.List(ctx, subnetBindingList); err != nil {
```
wenyingd: This change is not needed, as we should use list SubnetConnectionBindingMaps in the subnet's namespace?

pkg/controllers/namespace/subnet_share_controller.go
```diff
@@ -369,17 +369,22 @@ func (r *NamespaceReconciler) checkSubnetReferences(ctx context.Context, ns stri
...
 		return false, fmt.Errorf("failed to list SubnetConnectionBindingMap CRs: %w", err)
 	}
-
-	if len(subnetBindingList.Items) > 0 {
-		log.Info("Cannot delete Subnet CR for shared subnet because it is referenced by a SubnetConnectionBindingMap CR",
-			"Namespace", ns, "Name", subnet.Name, "SubnetBinding", subnetBindingList.Items[0].Name)
-		return true, nil
+	for _, bm := range subnetBindingList.Items {
+		if bm.Spec.SubnetName == subnet.Name && bm.Namespace == ns {
```
wenyingd: no need to compare bindingmap's namespace?

pkg/controllers/namespace/subnet_share_controller.go
```diff
@@ -369,17 +369,22 @@ func (r *NamespaceReconciler) checkSubnetReferences(ctx context.Context, ns stri
...
+			log.Info("Cannot delete Subnet CR for shared subnet because it is referenced by a SubnetConnectionBindingMap CR",
+				"Namespace", ns, "Name", subnet.Name, "SubnetBinding", bm.Name)
+			return true, nil
+		}
+		if bm.Spec.TargetSubnetName == subnet.Name && bm.Namespace == ns {
```
wenyingd: ditto

pkg/controllers/namespace/subnet_share_controller_test.go
```diff
@@ -442,7 +466,11 @@ func TestCheckSubnetReferences(t *testing.T) {
 				// Check if there are any SubnetConnectionBindingMap CRs referencing this Subnet CR
 				for _, obj := range tt.existingResources {
 					if binding, ok := obj.(*v1alpha1.SubnetConnectionBindingMap); ok {
-						if binding.Namespace == ns && binding.Spec.TargetSubnetName == subnet.Name {
+						if binding.Spec.SubnetName == subnet.Name && binding.Namespace == ns {
```
wenyingd: ditto

pkg/controllers/subnet/subnetbinding_handler.go
```diff
@@ -42,7 +43,8 @@ func enqueueSubnets(ctx context.Context, c client.Client, bindingMap *v1alpha1.S
 	}
 
 	if bindingMap.Spec.TargetSubnetName != "" {
-		_ = enqueue(ctx, c, bindingMap.Namespace, bindingMap.Spec.TargetSubnetName, q)
+		targetNs := bindingMap.Namespace
```
wenyingd: not needed?

pkg/controllers/subnet/subnetbinding_handler.go
```diff
@@ -89,13 +91,17 @@ func (r *SubnetReconciler) getNSXSubnetBindingsBySubnet(subnetCRUID string) []*v
 func (r *SubnetReconciler) getSubnetBindingCRsBySubnet(ctx context.Context, subnetCR *v1alpha1.Subnet) []v1alpha1.SubnetConnectionBindingMap {
 	validBindings := make([]v1alpha1.SubnetConnectionBindingMap, 0)
 	bindingList := &v1alpha1.SubnetConnectionBindingMapList{}
-	err := r.Client.List(ctx, bindingList, client.InNamespace(subnetCR.Namespace))
+	err := r.Client.List(ctx, bindingList)
```
wenyingd: ditto

pkg/controllers/subnet/subnetbinding_handler.go
```diff
@@ -89,13 +91,17 @@ func (r *SubnetReconciler) getNSXSubnetBindingsBySubnet(subnetCRUID string) []*v
...
 	if err != nil {
-		log.Error(err, "Unable to list SubnetConnectionBindingMaps", "Namespace", subnetCR.Namespace)
+		log.Error(err, "Unable to list SubnetConnectionBindingMaps")
 		return validBindings
 	}
 	for _, bm := range bindingList.Items {
-		if bm.Spec.SubnetName == subnetCR.Name || bm.Spec.TargetSubnetName == subnetCR.Name {
+		if bm.Spec.SubnetName == subnetCR.Name && bm.Namespace == subnetCR.Namespace {
```
wenyingd: ditto

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
```diff
@@ -17,6 +26,13 @@ type SubnetConnectionBindingMapSpec struct {
 	// TargetSubnetName specifies the target Subnet which a Subnet is connected to.
 	// +kubebuilder:validation:Optional
 	TargetSubnetName string `json:"targetSubnetName,omitempty"`
+	// SubnetAssociation indicates the role of targetSubnetName in the binding.
+	// TRUNK: targetSubnetName is the parent Subnet (default, legacy behavior).
+	// BRANCH: targetSubnetName is the child Subnet; subnetName is the parent and hosts the NSX binding map.
+	// +kubebuilder:validation:Enum=TRUNK;BRANCH
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="subnetAssociation is immutable"
+	SubnetAssociation string `json:"subnetAssociation,omitempty"`
```
yanjunz97: Maybe define it as a type like we do for access mode or ip address type?

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
```diff
@@ -17,6 +26,13 @@ type SubnetConnectionBindingMapSpec struct {
 	// TargetSubnetName specifies the target Subnet which a Subnet is connected to.
 	// +kubebuilder:validation:Optional
 	TargetSubnetName string `json:"targetSubnetName,omitempty"`
+	// SubnetAssociation indicates the role of targetSubnetName in the binding.
+	// TRUNK: targetSubnetName is the parent Subnet (default, legacy behavior).
+	// BRANCH: targetSubnetName is the child Subnet; subnetName is the parent and hosts the NSX binding map.
+	// +kubebuilder:validation:Enum=TRUNK;BRANCH
```
yanjunz97: TRUNK -> Trunk
BRANCH -> Branch
I remember we are asked to not capitalize all the chars to follow k8s convention. That's why some fields like access mode / ip address type are different from NSX ones.

pkg/controllers/subnet/subnetbinding_handler.go
```diff
@@ -26,13 +26,14 @@ func requeueSubnetBySubnetBindingUpdate(ctx context.Context, c client.Client, ob
 	}
 
 	if newBM.Spec.TargetSubnetName != "" {
-		if err := enqueue(ctx, c, newBM.Namespace, newBM.Spec.TargetSubnetName, q); err != nil {
-			log.Error(err, "Failed to enqueue the new target Subnet after SubnetConnectionBindingMap updates", "Namespace", newBM.Namespace, "Subnet", newBM.Spec.TargetSubnetName, "SubnetConnectionBindingMap", newBM.Name)
+		targetNs := newBM.Namespace
+		if err := enqueue(ctx, c, targetNs, newBM.Spec.TargetSubnetName, q); err != nil {
+			log.Error(err, "Failed to enqueue the new target Subnet after SubnetConnectionBindingMap updates", "Namespace", targetNs, "Subnet", newBM.Spec.TargetSubnetName, "SubnetConnectionBindingMap", newBM.Name)
```
yanjunz97: Just for curiosity, what's the purpose of this change?

pkg/controllers/subnet/subnetbinding_handler.go
yanjunz97: It looks like the behavior is the same after the change, then maybe we can keep with the previous implementation to only list bindingmap in this ns?

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
wenqiq: Done. Forgot to delete related field in the sample yamls.

pkg/controllers/subnetbinding/subnets_handler.go
```diff
@@ -64,21 +64,24 @@ func requeueBindingMapsBySubnetUpdate(ctx context.Context, c client.Client, _, o
 
 func requeueSubnetConnectionBindingMapsBySubnet(ctx context.Context, c client.Client, namespace string, subnet string, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
 	bindingMapList := &v1alpha1.SubnetConnectionBindingMapList{}
-	err := c.List(ctx, bindingMapList, client.InNamespace(namespace))
+	err := c.List(ctx, bindingMapList)
```
yanjunz97: ditto

pkg/nsx/services/subnetbinding/builder.go
```diff
@@ -18,22 +18,27 @@ var (
...
-			Id:             String(s.buildSubnetBindingID(binding, vpcSubnetInfo.ID)),
-			DisplayName:    String(binding.Name),
-			VlanTrafficTag: Int64(binding.Spec.VLANTrafficTag),
-			SubnetPath:     &path,
-			Tags:           tags,
+			Id:                String(s.buildSubnetBindingID(binding, vpcSubnetInfo.ID)),
+			DisplayName:       String(binding.Name),
+			VlanTrafficTag:    Int64(binding.Spec.VLANTrafficTag),
+			SubnetPath:        &path,
+			SubnetAssociation: String(subnetAssociation),
```
yanjunz97: Maybe not set subnetAssociation when it is not set in CR to avoid issue when running with NSX 9.1

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -217,66 +217,81 @@ func getVpcPath(subnetPath string) (string, *errorWithRetry) {
...
+// (spec.subnetName, where the binding map is created) and peer Subnet path(s) (target Subnet or SubnetSet members).
 func (r *Reconciler) validateDependency(ctx context.Context, bindingMap *v1alpha1.SubnetConnectionBindingMap) (string, []string, *errorWithRetry) {
-	childSubnetPaths, childSubnetCR, err := r.validateVpcSubnetsBySubnetCR(ctx, bindingMap.Namespace, bindingMap.Spec.SubnetName, false)
+	isBranch := bindingMap.Spec.IsBranchAssociation()
+	targetNamespace := bindingMap.Namespace
+
+	var hostCheckUsedAsTarget, hostCheckUsedAsHost bool
+	if isBranch {
+		// Parent (trunk) host may be referenced by multiple branch bindings.
+		hostCheckUsedAsTarget, hostCheckUsedAsHost = false, false
```
wenyingd: In your current implementation, hostCheckUsedAsHost is always false in the two mode, is this a typo? If no, then variable hostCheckUsedAsHost is not needed indeed?

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
```diff
@@ -17,6 +26,13 @@ type SubnetConnectionBindingMapSpec struct {
 	// TargetSubnetName specifies the target Subnet which a Subnet is connected to.
 	// +kubebuilder:validation:Optional
 	TargetSubnetName string `json:"targetSubnetName,omitempty"`
+	// SubnetAssociation indicates the role of targetSubnetName in the binding.
+	// TRUNK: targetSubnetName is the parent Subnet (default, legacy behavior).
+	// BRANCH: targetSubnetName is the child Subnet; subnetName is the parent and hosts the NSX binding map.
```
yanjunz97: Would you know when SubnetAssociation is Branch, can TargetSubnetSetName be specified?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -107,7 +107,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	// Create or update SubnetConnectionBindingMap
 	r.StatusUpdater.IncreaseUpdateTotal()
-	childSubnetPath, parentSubnetPaths, err := r.validateDependency(ctx, bindingMapCR)
+	hostSubnetPath, peerSubnetPaths, err := r.validateDependency(ctx, bindingMapCR)
```
wenyingd: why do you want to rename as "host" and "peer"? Are you using "host" to refer the "child" while using the "peer" to refer the parent?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -217,66 +217,81 @@ func getVpcPath(subnetPath string) (string, *errorWithRetry) {
...
-//     Subnet/SubnetSet readiness update will actively trigger a requeue event
-//  2. the associated Subnet is already used as a target Subnet in another SubnetConnectionBindingMap CR, or the target
-//     Subnet already has associated SubnetConnectionBindingMap CR. In this case, a retry error is returned.
-//  3. the target Subnet is a pre-created Subnet or target SubnetSet is a pre-created SubnetSet.
-//     In this case, a not-retry error is returned.
-//  4. the associated Subnet is a pre-created Subnet in a VPC different from the target Subnet Namespace VPC
-//     In this case, not-retry error is returned.
+// validateDependency validates dependent Subnets/SubnetSets and returns the NSX path of the host Subnet
+// (spec.subnetName, where the binding map is created) and peer Subnet path(s) (target Subnet or SubnetSet members).
 func (r *Reconciler) validateDependency(ctx context.Context, bindingMap *v1alpha1.SubnetConnectionBindingMap) (string, []string, *errorWithRetry) {
```
wenyingd, line 287: I am a bit confused on this function, could you help to clarify which cases are supported for cross-vpc bindingmap and which cases are not? Please also specify the combinations of Subnet, TargetSubnet, TargetSubnetSet, shared Subnet

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -309,13 +322,13 @@ func (r *Reconciler) validateVpcSubnetsBySubnetCR(ctx context.Context, namespace
 				retry:   true,
 			}
 		}
-	} else {
-		bms, err := r.getSubnetConnectionBindingMapsByChildSubnet(ctx, namespace, name)
+	}
+	if checkNotUsedAsHost {
```
wenyingd: In my understanding, a NSX VpcSubnet is not working as target and host at the same time, no matter what mode it is configured with. But the current implementation seems to allow such case. Could you double confirm if this is a new requirement ?

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Before this commit, the SubnetConnectionBindingMap defaulted to and only supported one network topology mode (namely the Trunk mode). In the old logic: 

Spec.SubnetName (the subnet where the binding relationship is located) is regarded as the Child (sub-subnet).
Spec.TargetSubnetName (the subnet being pointed to) is regarded as the Parent (parent-subnet/Trunk subnet).

However, this commit introduces the Cross-VPC Branch mode (SubnetAssociationBranch). In this mode, the network topology is completely reversed: 

Spec.SubnetName (the subnet where the binding is located) actually plays the role of the Parent (Trunk) Subnet.
Spec.TargetSubnetName plays the role of the Child (Branch) Subnet, which is responsible for extending the VLANs of the parent and child networks to another VPC.

If the variables continue to be named with childSubnet and parentSubnet, when dealing with the Branch mode, the code will be completely illogical in terms of logical semantics (that is, referring to the object that is actually the Parent as Child).

By renaming these variables, the controller's logic works seamlessly for both TRUNK and BRANCH.

pkg/controllers/subnet/subnetbinding_handler.go
wenqiq: The main advantage of this modification lies in the improved readability and consistency of the code, as well as the reduction of redundant attribute access. Just a programming habit. I noticed it and changed it right there. Restored it already. Thanks for reviewing.

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
wenqiq: done

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
wenqiq: done

pkg/controllers/subnet/subnetbinding_handler.go
wenqiq: It was mixed up and now it has been updated. Thanks.

pkg/controllers/namespace/subnet_share_controller.go
wenqiq: Updated.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Typo. Updated. Thanks for reviewing.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: 1. Trunk Mode (Legacy/Default: empty or TRUNK association)
In this mode, the SubnetName acts as the Child, and the Target acts as the Parent Trunk.

Supported:
SubnetName -> TargetSubnetName (within the same VPC).
SubnetName -> TargetSubnetSetName (within the same VPC).
Shared (pre-created) SubnetName -> Target (only if both are in the same VPC).
Unsupported:
Cross-VPC bindings.
Using a Shared (pre-created) Subnet as the TargetSubnetName.
Using the SubnetName as a Host or Target in any other bindings (1:1 relationship enforced).
2. Branch Mode (SubnetAssociation: BRANCH)
In this mode, the SubnetName acts as the Parent Trunk, and the Target acts as the Child Branch pulling the VLAN. This mode is specifically designed for Cross-VPC connectivity.

Supported:
SubnetName -> TargetSubnetName (Cross-VPC is allowed and supported).
Using a Shared (pre-created) Subnet as the TargetSubnetName.
One SubnetName acting as the Parent for multiple branch bindings (1:N relationship allowed).
Unsupported:
SubnetName -> TargetSubnetSetName (Sets are explicitly rejected for Branch mode).
Chaining bindings (the TargetSubnetName cannot act as a Host in another binding).

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
wenqiq: In my understanding, TargetSubnetSetName cannot be specified when SubnetAssociation is set to Branch. A "Branch" requires a strict 1:1 point-to-point connection. Binding to a dynamic "SubnetSet" creates a 1:N mapping, causing VLAN sharing conflicts and unpredictable network states.

pkg/nsx/services/subnetbinding/builder.go
wenqiq: updated.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Yes, the code allows a subnet to be a host in one CR and a target in another, provided it acts as a Parent (Trunk) in both.However, it strictly forbids a subnet from acting as both Parent and Child to prevent cascading.

wenqiq:
@dantingl @yanjunz97 @wenyingd I have addressed the comments and updated the testdone in the description. Please take another look. Thanks!

pkg/controllers/subnetbinding/subnetbinding_controller.go
dantingl: Do you think we can use trunk/branch subnet

pkg/controllers/namespace/subnet_share_controller.go
```diff
@@ -369,17 +369,22 @@ func (r *NamespaceReconciler) checkSubnetReferences(ctx context.Context, ns stri
 		return true, nil
 	}
 
-	// Check if there are any SubnetConnectionBindingMap CRs referencing this Subnet CR
+	// Check if there are any SubnetConnectionBindingMap CRs referencing this Subnet CR as host or target.
 	subnetBindingList := &v1alpha1.SubnetConnectionBindingMapList{}
-	err = r.Client.List(ctx, subnetBindingList, client.InNamespace(ns), client.MatchingFields{"spec.subnetName": subnet.Name})
-	if err != nil {
+	if err = r.Client.List(ctx, subnetBindingList, client.InNamespace(ns)); err != nil {
```
dantingl, line 374: This may also need to be updated in WCPSVC validation @lxiaopei @TaoZou1

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
```diff
@@ -5,6 +5,17 @@ package v1alpha1
...
+	// SubnetAssociationTrunk means targetSubnetName is the parent (trunk) Subnet in the binding.
+	// This is the default when subnetAssociation is unset (legacy same-VPC workflow).
+	SubnetAssociationTrunk SubnetAssociation = "Trunk"
+	// SubnetAssociationBranch means targetSubnetName is the child (branch) Subnet in the binding.
+	// Used for cross-VPC VLAN extension; NSX SubnetConnectionBindingMap is created under subnetName (parent).
+	SubnetAssociationBranch SubnetAssociation = "Branch"
+)
+
 // +kubebuilder:validation:XValidation:rule="has(self.targetSubnetSetName) && !has(self.targetSubnetName) || !has(self.targetSubnetSetName) && has(self.targetSubnetName)",message="Only one of targetSubnetSetName or targetSubnetName can be specified"
 // +kubebuilder:validation:XValidation:rule="!has(self.targetSubnetName) || (self.subnetName != self.targetSubnetName)",message="subnetName and targetSubnetName must be different"
```
yanjunz97, line 20: Would you think we shall a validation rule to block setting TargetSubnetSetName with SubnetAssociation?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -217,66 +217,90 @@ func getVpcPath(subnetPath string) (string, *errorWithRetry) {
...
-//     In this case, not-retry error is returned.
+// validateDependency validates dependent Subnets/SubnetSets and returns the NSX path of the host Subnet
+// (spec.subnetName, where the binding map is created) and peer Subnet path(s) (target Subnet or SubnetSet members).
 func (r *Reconciler) validateDependency(ctx context.Context, bindingMap *v1alpha1.SubnetConnectionBindingMap) (string, []string, *errorWithRetry) {
-	childSubnetPaths, childSubnetCR, err := r.validateVpcSubnetsBySubnetCR(ctx, bindingMap.Namespace, bindingMap.Spec.SubnetName, false)
+	isBranch := bindingMap.Spec.IsBranchAssociation()
+	targetNamespace := bindingMap.Namespace
+
+	// Any Subnet cannot be used as a child in another binding (to prevent multiple parents and cascading).
+	const checkNotUsedAsChild = true
```
yanjunz97: It looks like checkNotUsedAsChild is always true, do we need this value?
Actually I've another question, do you mean a subnet cannot be child of multiple parents? But I think it works today for truck. For example:

```yaml
metadata:
  name: scbm1
spec:
  subnetName: subnet-1
  vlanTrafficTag: 201
  targetSubnetName: subnet-2
---
metadata:
  name: scbm2
spec:
  subnetName: subnet-1
  vlanTrafficTag: 202
  targetSubnetName: subnet-3
```

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -290,40 +314,65 @@ func (r *Reconciler) validateVpcSubnetsBySubnetCR(ctx context.Context, namespace
...
 			}
 		}
-		if len(bms) > 0 {
+		bmsAsHost, err := r.getSubnetConnectionBindingMapsByHostSubnet(ctx, namespace, name)
+		if err != nil {
+			log.Error(err, "Failed to get SubnetConnectionBindingMaps with Subnet as host Subnet", "Subnet", subnetKey.String())
 			return nil, subnetCR, &errorWithRetry{
-				message: fmt.Sprintf("Subnet CR %s is working as target by %s", name, bms),
-				error:   fmt.Errorf("Subnet %s already works as target in SubnetConnectionBindingMap %s", name, bms),
+				message: fmt.Sprintf("Failed to get SubnetConnectionBindingMaps with Subnet as host Subnet %s", name),
```
yanjunz97: Maybe still use associated instead of host? I feel the term host might be confused to users

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
wenqiq: done

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Updated.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Thanks for pointing this out. I have updated it.

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -217,66 +217,138 @@ func getVpcPath(subnetPath string) (string, *errorWithRetry) {
...
+// |  |  +--------------------------+                                +-------------------------+  |  |
+// |  |  |      Parent Subnet       |                                |      Child Subnet       |  |  |
+// |  |  |         (Trunk)          |                                |        (Branch)         |  |  |
+// |  |  |                          |                                |                         |  |  |
+// |  |  |    [targetSubnetName]    <--------------------------------+      [subnetName]       |  |  |
+// |  |  |                          |                                | (Host of BindingMap CR) |  |  |
+// |  |  +--------------------------+                                +-------------------------+  |  |
+// |  +-------------------------------------------------------------------------------------------+  |
+// |                                                                                                 |
+// |  Cross-VPC Branch Mode (subnetAssociation: Branch)                                              |
```
wenyingd: Do you mean only "Branch" mode works for cross-VPC case; and for the case parent-child are in the same VPC, we can only use trunk (legacy) mode?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -217,66 +217,138 @@ func getVpcPath(subnetPath string) (string, *errorWithRetry) {
...
+// |  |  +---------------------------------+  |       |  +---------------------------------+  |      |
+// |  +---------------------------------------+       +---------------------------------------+      |
+// |                                                                                                 |
+// +-------------------------------------------------------------------------------------------------+
 func (r *Reconciler) validateDependency(ctx context.Context, bindingMap *v1alpha1.SubnetConnectionBindingMap) (string, []string, *errorWithRetry) {
-	childSubnetPaths, childSubnetCR, err := r.validateVpcSubnetsBySubnetCR(ctx, bindingMap.Namespace, bindingMap.Spec.SubnetName, false)
+	isBranch := bindingMap.Spec.IsBranchAssociation()
+	targetNamespace := bindingMap.Namespace
+
+	var subnetCheckNotUsedAsChild, subnetCheckNotUsedAsParent bool
```
wenyingd: if `subnetCheckNotUsedAsChild` == `!subnetCheckNotUsedAsParent` can we simply the variables and corresponding logic? The same as `targetCheckNotUsedAsChild v.s. targetCheckNotUsedAsParent`

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -217,66 +217,138 @@ func getVpcPath(subnetPath string) (string, *errorWithRetry) {
...
-		parentSubnetPaths, parentSubnetCR, err = r.validateVpcSubnetsBySubnetCR(ctx, bindingMap.Namespace, bindingMap.Spec.TargetSubnetName, true)
+		var targetSubnetCR *v1alpha1.Subnet
+		targetSubnetPaths, targetSubnetCR, err = r.validateVpcSubnetsBySubnetCR(ctx, targetNamespace, bindingMap.Spec.TargetSubnetName, targetCheckNotUsedAsChild, targetCheckNotUsedAsParent, bindingMap.Name)
 		if err != nil {
 			return "", nil, err
 		}
-		// Check if the target Subnet is pre-created Subnet
-		if _, ok := parentSubnetCR.GetAnnotations()[servicecommon.AnnotationAssociatedResource]; ok {
+		if !isBranch {
+			if _, ok := targetSubnetCR.GetAnnotations()[servicecommon.AnnotationAssociatedResource]; ok {
```
wenyingd: If `pre-created Subnet` can't act as the target subnet in TRUNK/legacy mode, shall we the same check on the Subnet CR in Branch mode?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -217,66 +217,138 @@ func getVpcPath(subnetPath string) (string, *errorWithRetry) {
...
+				return "", nil, &errorWithRetry{
+					message: fmt.Sprintf("Target Subnet %s/%s is a pre-created Subnet", targetNamespace, bindingMap.Spec.TargetSubnetName),
+					error:   fmt.Errorf("pre-created Subnet %s/%s cannot be a target Subnet", targetNamespace, bindingMap.Spec.TargetSubnetName),
+					retry:   false,
+				}
+			}
+		}
+	} else {
+		if isBranch {
```
wenyingd: this if condition is not needed as you add validations in the CRD?

pkg/nsx/services/subnetbinding/subnetbinding.go
```diff
@@ -54,12 +54,12 @@ func InitializeService(service servicecommon.Service) (*BindingService, error) {
 }
 
 // CreateOrUpdateSubnetConnectionBindingMap creates or updates the SubnetConnectionBindingMaps with the given
-// SubnetConnectionBindingMap CR and attaches it to the parentSubnets.
+// SubnetConnectionBindingMap CR under subnetPath (spec.subnetName) and targetSubnetPaths (target Subnet/SubnetSet).
```
wenyingd, line 57: do you need to update the index function `parentSubnetIndexFunc` and `childSubnetIndexFunc` since the role is different according to `SubnetAssociation` value ?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -217,66 +217,138 @@ func getVpcPath(subnetPath string) (string, *errorWithRetry) {
...
-		parentVpcPath, err := getVpcPath(parentSubnetPaths[0])
-		if err != nil {
-			return "", nil, err
-		}
-		if childVpcPath != parentVpcPath {
-			return "", nil, &errorWithRetry{
-				message: fmt.Sprintf("Subnet %s and target Subnet %s are in different VPCs", childSubnetPath, parentSubnetPaths[0]),
-				retry:   false,
-				error:   fmt.Errorf("Subnet and target Subnet are in different VPCs"),
+	// Legacy Trunk workflow: a shared (pre-created) child must stay in the same VPC as the parent target.
```
wenyingd: Is this requirement still true that a shared child must stay in the same VPC as the parent in Trank/legacy mode after this change?  cc @dantingl

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -290,40 +362,65 @@ func (r *Reconciler) validateVpcSubnetsBySubnetCR(ctx context.Context, namespace
 		}
 	}
 
-	// Check if the Subnet CR is nested.
-	if !isTarget {
-		bms, err := r.getSubnetConnectionBindingMapsByParentSubnet(ctx, namespace, name)
+	if checkNotUsedAsChild || checkNotUsedAsParent {
```
wenyingd: This check always returns true, as either `checkNotUsedAsChild` or `checkNotUsedAsParent` is always true..

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -217,66 +217,138 @@ func getVpcPath(subnetPath string) (string, *errorWithRetry) {
...
+				}
 			}
 		}
 	}
-	return childSubnetPath, parentSubnetPaths, nil
+	return subnetPath, targetSubnetPaths, nil
 }
 
-func (r *Reconciler) validateVpcSubnetsBySubnetCR(ctx context.Context, namespace, name string, isTarget bool) ([]string, *v1alpha1.Subnet, *errorWithRetry) {
+func (r *Reconciler) validateVpcSubnetsBySubnetCR(ctx context.Context, namespace, name string, checkNotUsedAsChild, checkNotUsedAsParent bool, currentBindingMapName string) ([]string, *v1alpha1.Subnet, *errorWithRetry) {
```
wenyingd: ditto, a single flag for checkNotUsedAsChild/checkNotUsedAsParent is enough, no need to have both.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Thanks for the review! You are right that in the current validateDependency caller, one of them is always true and they act mutually exclusive.

The rationale for keeping two separate flags here is for the extensibility and explicit intent of this utility function:

1. Future-proofing: If a future caller only needs to validate the Subnet CR's existence and realization without checking its binding usage, they can pass false, false to safely skip the expensive API calls inside this block. A single boolean flag (e.g., checkAsChild) wouldn't allow this 'skip both' state.
2. Readability: Having two explicit flags makes the caller's intent very clear. If we used a single flag, passing false would implicitly mean 'check as parent', which is slightly less readable.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Good catch. I think we can delete this.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: The same question as https://github.com/vmware-tanzu/nsx-operator/pull/1443#discussion_r3602907421 ?

pkg/nsx/services/subnetbinding/subnetbinding.go
wenqiq: done

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: updated

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: ~~1. Only "Branch" mode works for the cross-VPC case (True)
This is correct. In Trunk mode (default/legacy), there is a strict validation enforcing that both the parent and child subnets must reside in the same VPC. For cross-VPC scenarios, we must use Branch mode because:
It skips the same-VPC check.
It places the SubnetConnectionBindingMap CR on the consumer's side (under the Trunk/Parent subnet), resolving the cross-project write permission issue.~~

2. For the same-VPC case, both Trunk and Branch modes are supported (Not limited to Trunk only)

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: ~~Yes, I think this requirement is still true.~~

~~Even after introducing the new association mode, if a connection binding map is created in Trunk (legacy) mode, the child subnet and the parent subnet must stay in the same VPC.~~

After confirmation, this requirement is no longer true and I have changed the code. A shared child subnet no longer needs to stay in the same VPC as the parent in Trunk/legacy mode.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenyingd: We should not have this assumption that a Subnet needs the validation that it can be both parent and child at the same time. This case is not reasonable.

wenyingd (commented):
The change overall looks good to me except for the following two:
1. https://github.com/vmware-tanzu/nsx-operator/pull/1443/changes#r3602907421
2. https://github.com/vmware-tanzu/nsx-operator/pull/1443/changes#r3490810230

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: I have refactored the code in the latest commit:

1. Replaced the two separate boolean flags with a single, clear parameter: `isParent bool`.
2. Inside validateVpcSubnetsBySubnetCR, the validation logic is now determined exclusively by this flag, making it impossible to check both cases at the same time:
`checkNotUsedAsChild := isParent`
`checkNotUsedAsParent := !isParent`
3. Refactored the callers in `validateDependency` accordingly to pass subnetIsParent and targetIsParent based on the association mode, which makes the overall logic clean and self-documenting.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Updated.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Updated the code and comments.

The relationships are much more flexible:
1. For the Cross-VPC Case
Both "Branch" and "Trunk" modes work.
2. For the Same-VPC Case
Both "Trunk" and "Branch" modes can be used.

pkg/controllers/namespace/subnet_share_controller.go
```diff
@@ -369,17 +369,35 @@ func (r *NamespaceReconciler) checkSubnetReferences(ctx context.Context, ns stri
 		return true, nil
 	}
 
-	// Check if there are any SubnetConnectionBindingMap CRs referencing this Subnet CR
+	// Check if there are any SubnetConnectionBindingMap CRs referencing this Subnet CR as host or target.
```
wenyingd: no need to highlight `as host or target`.

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -205,78 +205,100 @@ func (r *Reconciler) listBindingMapIDsFromCRs(ctx context.Context) (sets.Set[str
...
-	var parentSubnetPaths []string
-	if bindingMap.Spec.TargetSubnetName != "" {
-		var parentSubnetCR *v1alpha1.Subnet
-		parentSubnetPaths, parentSubnetCR, err = r.validateVpcSubnetsBySubnetCR(ctx, bindingMap.Namespace, bindingMap.Spec.TargetSubnetName, true)
-		if err != nil {
-			return "", nil, err
-		}
-		// Check if the target Subnet is pre-created Subnet
-		if _, ok := parentSubnetCR.GetAnnotations()[servicecommon.AnnotationAssociatedResource]; ok {
+	if isBranch {
```
wenyingd: Here should check a "parent" subnet can not be a pre-created Subnet? e.g., 
```
if subnetIsParent {
....
}
```

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -205,78 +205,100 @@ func (r *Reconciler) listBindingMapIDsFromCRs(ctx context.Context) (sets.Set[str
...
-			return "", nil, err
-		}
 	}
 
-	// If child Subnet is a pre-created Subnet, check if it is in the same vpc as parent Subnet
-	if _, ok := childSubnetCR.GetAnnotations()[servicecommon.AnnotationAssociatedResource]; ok {
-		childVpcPath, err := getVpcPath(childSubnetPath)
+	var targetSubnetPaths []string
+	if bindingMap.Spec.TargetSubnetName != "" {
+		// targetSubnetName is Parent: it can have multiple children, but CANNOT be a child in any binding.
```
wenyingd: Did you forget checking targetSubnet is not a pre-created subnet when targetIsParent==True ?
I would suggest adding an inner function, e.g.,
```
	checkPreCreatesSubnetAsParent := func(subnet *v1alpha1.Subnet) error {
		if _, ok := subnetCR.GetAnnotations()[servicecommon.AnnotationAssociatedResource]; ok {
			return &errorWithRetry{
				message: fmt.Sprintf("Subnet %s/%s is a pre-created Subnet", bindingMap.Namespace, bindingMap.Spec.SubnetName),
				error:   fmt.Errorf("pre-created Subnet %s/%s cannot be a parent Subnet", bindingMap.Namespace, bindingMap.Spec.SubnetName),
				retry:   false,
			}
		}
		return nil
	}
	if subnetIsParent {
		if err := checkPreCreatesSubnetAsParent(subnetCR); err != nil {
			return "", nil, err
		}
	}
	if bindingMap.Spec.TargetSubnetName != "" {
		targetSubnetPaths, targetSubnetCR, err = r.validateVpcSubnetsBySubnetCR(ctx, targetNamespace, bindingMap.Spec.TargetSubnetName, targetIsParent, bindingMap.Name)
		if targetIsParent {
			if err := checkPreCreatesSubnetAsParent(targetSubnetCR); err != nil {
				return "", nil, err
			}
		}
	} else {
		...
	}
```

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -205,78 +205,100 @@ func (r *Reconciler) listBindingMapIDsFromCRs(ctx context.Context) (sets.Set[str
...
-			}
-		}
 	}
-	return childSubnetPath, parentSubnetPaths, nil
+
+	return subnetPath, targetSubnetPaths, nil
 }
 
-func (r *Reconciler) validateVpcSubnetsBySubnetCR(ctx context.Context, namespace, name string, isTarget bool) ([]string, *v1alpha1.Subnet, *errorWithRetry) {
+func (r *Reconciler) validateVpcSubnetsBySubnetCR(ctx context.Context, namespace, name string, isParent bool, currentBindingMapName string) ([]string, *v1alpha1.Subnet, *errorWithRetry) {
```
wenyingd: Since the caller already knows whether the current subnet isParent or child, we can check if if there exists another bindingmap using the current subnet as the other role, if yes, return fail.
e.g, if `isParent==true`, then the expectation is `r.getSubnetConnectionBindingMapsByParentSubnet(ctx, namespace, name)` returns no bindinmaps, if `len(bindingmaps) > 0`,  a conflict exits. It also works in the reverse direction, if `isParent==false`, ``r.getSubnetConnectionBindingMapsByChildSubnet(ctx, namespace, name)` should return 0 bindingmaps.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenyingd: Maybe you can move this check into `validateVpcSubnetsBySubnetCR`

```
		if _, ok := subnetCR.GetAnnotations()[servicecommon.AnnotationAssociatedResource]; ok {
			return &errorWithRetry{
				message: fmt.Sprintf("Subnet %s/%s is a pre-created Subnet", bindingMap.Namespace, bindingMap.Spec.SubnetName),
				error:   fmt.Errorf("pre-created Subnet %s/%s cannot be a parent Subnet", bindingMap.Namespace, bindingMap.Spec.SubnetName),
				retry:   false,
			}
		}
```

pkg/controllers/namespace/subnet_share_controller.go
wenqiq: done

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Updated.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Good point.Updated.

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
```diff
@@ -5,8 +5,20 @@ package v1alpha1
 
 import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+type SubnetAssociation string
+
+const (
+	// SubnetAssociationTrunk means targetSubnetName is the parent (trunk) Subnet in the binding.
```
jianjuns: Probably go this way "trunk (parent)". Parent/child Subnet are not official terms.

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
```diff
@@ -5,8 +5,20 @@ package v1alpha1
...
+	// This is the default when subnetAssociation is unset.
+	SubnetAssociationTrunk SubnetAssociation = "Trunk"
+	// SubnetAssociationBranch means targetSubnetName is the child (branch) Subnet in the binding.
+	// NSX SubnetConnectionBindingMap is created under subnetName (parent).
```
jianjuns: Again and again - pls do not mention NSX or NSX internal terminologies (Policy, Project, ID, path, etc.) in Namespace API/CRD comments.

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
```diff
@@ -17,6 +29,13 @@ type SubnetConnectionBindingMapSpec struct {
 	// TargetSubnetName specifies the target Subnet which a Subnet is connected to.
 	// +kubebuilder:validation:Optional
 	TargetSubnetName string `json:"targetSubnetName,omitempty"`
+	// SubnetAssociation indicates the role of targetSubnetName in the binding.
+	// Trunk: targetSubnetName is the parent Subnet (default behavior).
```
jianjuns: parent -> trunk

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
```diff
@@ -17,6 +29,13 @@ type SubnetConnectionBindingMapSpec struct {
...
+	// Branch: targetSubnetName is the child Subnet; subnetName is the parent and hosts the NSX binding map.
```
jianjuns: child -> branch

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
```diff
@@ -17,6 +29,13 @@ type SubnetConnectionBindingMapSpec struct {
 	// TargetSubnetName specifies the target Subnet which a Subnet is connected to.
 	// +kubebuilder:validation:Optional
 	TargetSubnetName string `json:"targetSubnetName,omitempty"`
+	// SubnetAssociation indicates the role of targetSubnetName in the binding.
```
jianjuns: Could we add the VCF version that starts supporting the field? And better to validate the NSX version and return an error earlier? @dantingl

jianjuns (commented):
For new

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
wenqiq: done

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
wenqiq: Got it. I've checked, and the newly added code will no longer appear this in this commit.

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
wenqiq: Updated.

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
```diff
@@ -5,8 +5,20 @@ package v1alpha1
...
 import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+type SubnetAssociation string
+
+const (
+	// SubnetAssociationTrunk means targetSubnetName is the trunk (parent) Subnet in the binding.
+	// This is the default when subnetAssociation is unset.
+	SubnetAssociationTrunk SubnetAssociation = "Trunk"
+	// SubnetAssociationBranch means targetSubnetName is the branch (child) Subnet in the binding.
+	// SubnetConnectionBindingMap is created under subnetName (trunk (parent) Subnet).
```
jianjuns: Remove (parent).

wenqiq: Should we clean up (parent) and (child) across all API comments in this file as well? They were added to help users who habitually think in terms of 'parent/child' Subnets, but removing them keeps the docs strictly aligned with trunk and branch.

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -495,38 +512,50 @@ func updateBindingMapCondition(c client.Client, ctx context.Context, bindingMap
...
 		}
-		return []string{binding.Spec.TargetSubnetName}
+		return []string{binding.Spec.SubnetName}
 	}
 }
 
 // SetupFieldIndexers sets up the field indexers for SubnetConnectionBindingMap
 func (r *Reconciler) SetupFieldIndexers(mgr ctrl.Manager) error {
-	if err := mgr.GetFieldIndexer().IndexField(context.TODO(), &v1alpha1.SubnetConnectionBindingMap{}, "spec.subnetName", subnetConnectionBindingMapSubnetNameIndexFunc); err != nil {
+	if err := mgr.GetFieldIndexer().IndexField(context.TODO(), &v1alpha1.SubnetConnectionBindingMap{}, "parentSubnet", subnetConnectionBindingMapParentSubnetIndexFunc); err != nil {
```
wenyingd: I don't get why you change this indexer. This is to help setup an local index between the SubnetConnectionBindingMap and Subnet CRs to speed up getting Subnets via SubnetConnectionBindingMap. If you update the field with a static word which is not a path of the field in the API, the indexer does not work.

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -414,30 +437,24 @@ func (r *Reconciler) validateVpcSubnetsBySubnetSetCR(ctx context.Context, namesp
 	return subnetPaths, nil
 }
 
-func (r *Reconciler) getSubnetConnectionBindingMapsByParentSubnet(ctx context.Context, ns, name string) ([]types.NamespacedName, error) {
-	bmKeys := []types.NamespacedName{}
+// getSubnetConnectionBindingMapsByParentSubnet lists bindings that reference the Subnet as trunk (parent) Subnet
```
wenyingd: Not get why we need this function in CR level, my original comment is to call the index func provided in bindingstore:  https://github.com/vmware-tanzu/nsx-operator/blob/main/pkg/nsx/services/subnetbinding/store.go

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
jianjuns: Yes, pls do that.

wenqiq: Done.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Got it. Removed these changes.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: I have updated the validation logic to use the index function, as we discussed offline.

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
dantingl: I saw Wenqi already check NSX version in bindingmap controller, maybe also add VCF version requirement in CRD description

yanjunz97 (commented):
Overall LGTM, 
Just double check would you plan to add description for subnetAssociation is supported after 9.2?
https://github.com/vmware-tanzu/nsx-operator/pull/1443#discussion_r3677342320

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
wenqiq: @jianjuns Should the term "parent/child" also be excluded from error messages/log? My concern is that it might impact other components or ops.

jianjuns: Yes, ideally we should change logs too but that is less important. By error messages, do you mean errors in logs, or Status? Status is more important to change.

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
```diff
@@ -17,6 +29,14 @@ type SubnetConnectionBindingMapSpec struct {
 	// TargetSubnetName specifies the target Subnet which a Subnet is connected to.
 	// +kubebuilder:validation:Optional
 	TargetSubnetName string `json:"targetSubnetName,omitempty"`
+	// SubnetAssociation indicates the role of targetSubnetName in the binding.
+	// Trunk: targetSubnetName is the trunk Subnet (default behavior).
+	// Branch: targetSubnetName is the branch Subnet; subnetName is the trunk Subnet and hosts the binding map.
+	// Supported starting from VCF 9.2.0.
```
jianjuns: from -> with

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
```diff
@@ -17,6 +29,14 @@ type SubnetConnectionBindingMapSpec struct {
 	// TargetSubnetName specifies the target Subnet which a Subnet is connected to.
 	// +kubebuilder:validation:Optional
 	TargetSubnetName string `json:"targetSubnetName,omitempty"`
+	// SubnetAssociation indicates the role of targetSubnetName in the binding.
+	// Trunk: targetSubnetName is the trunk Subnet (default behavior).
```
jianjuns: TargetSubnetName

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
```diff
@@ -17,6 +29,14 @@ type SubnetConnectionBindingMapSpec struct {
...
+	// Branch: targetSubnetName is the branch Subnet; subnetName is the trunk Subnet and hosts the binding map.
```
jianjuns: SubnetName

jianjuns: "that hosts the..." or "and it hosts the..."

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
jianjuns: Ok. Could we make sure other new features like IPv6 also have the VCF version requirement described in the API comments?

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
wenqiq: Done.

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
wenqiq: Updated.

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
wenqiq: I will implement this in the PR https://github.com/vmware-tanzu/nsx-operator/pull/1523

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -479,14 +465,49 @@ func (r *Reconciler) validateVpcSubnetsBySubnetCR(ctx context.Context, namespace
...
+		bindingMaps := make([]*model.SubnetConnectionBindingMap, 0)
+		for _, subnetPath := range subnetPaths {
+			bindings := r.SubnetBindingService.GetSubnetConnectionBindingMapsByChildSubnet(subnetPath)
+			if len(bindings) > 0 {
+				bindingMaps = append(bindingMaps, bindings...)
+			}
+		}
+		if len(bindingMaps) > 0 {
+			bmName := getBindingMapName(bindingMaps[0])
+			return nil, &errorWithRetry{
```
wenyingd: why not directly returni the retry-error  in line 481? It is not necessary to introduce an aggregation .

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -479,14 +465,49 @@ func (r *Reconciler) validateVpcSubnetsBySubnetCR(ctx context.Context, namespace
...
+	} else {
+		bindingMaps := make([]*model.SubnetConnectionBindingMap, 0)
+		for _, subnetPath := range subnetPaths {
+			bindings := r.SubnetBindingService.GetSubnetConnectionBindingMapsByParentSubnet(subnetPath)
+			if len(bindings) > 0 {
+				bindingMaps = append(bindingMaps, bindings...)
+			}
+		}
+		if len(bindingMaps) > 0 {
+			bmName := getBindingMapName(bindingMaps[0])
```
wenyingd: ditto

pkg/controllers/subnetbinding/subnets_handler.go
```diff
@@ -70,15 +70,18 @@ func requeueSubnetConnectionBindingMapsBySubnet(ctx context.Context, c client.Cl
...
+		if !isSubnet && !isTarget {
+			continue
 		}
+		log.Info("Requeue SubnetConnectionBindingMap because the dependent Subnet realization state is changed", "Namespace", namespace, "Name", bm.Name)
+		q.Add(reconcile.Request{
+			NamespacedName: types.NamespacedName{
+				Name:      bm.Name,
+				Namespace: bm.Namespace,
+			},
+		})
```
wenyingd: It looks this change is performing the same behavior as the original code, so is it an unnecessary change?

dantingl (dismissed):
LGTM

wenyingd (approved):

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Make sense, updated. Originally, the goal was to aggregate and print all conflicting SubnetConnectionBindingMap instances in the logs/error message, which is why the slice collection logic was initially introduced.

dantingl (approved):

PR #1445, opened 2026-06-02
Support auto-allocate VLAN for SubnetConnectionBindingMap

wenqiq:
* Support auto-allocate VLAN for SubnetConnectionBindingMap

Previously, users were required to manually specify a unique `vlanTrafficTag`
when creating a `SubnetConnectionBindingMap`. This update introduces a VLAN
auto-allocation feature that assigns an available VLAN ID automatically when
the field is omitted.

CRD API Changes:
- Spec: `VLANTrafficTag` is now an optional pointer (`*int64`) rather than
  a required integer, allowing it to be omitted. Its valid minimum value is
  also updated from 0 to 1.
- Status: Introduced a new `VLANTrafficTag` field to reflect the realized VLAN ID,
  representing either the user-specified value or the system-allocated one.
- PrintColumns: Updated the `kubectl get` output column to display the realized
  VLAN ID from `.status.vlanTrafficTag` instead of `.spec.vlanTrafficTag`.

TestDone:

1. Ensure the Testbed has the latest nsx-operator deployed with the VLAN auto-allocation feature.
before deploy the new `manager`:
0e09e45f585481698f5895b569d57ed2

2. Create a YAML file test-vlan-auto.yaml containing a parent subnet, two child subnets, and a binding map:

```
kind: Subnet
metadata:
  name: parent-subnet
  namespace: default
spec:
  ipv4SubnetSize: 16
---
kind: Subnet
metadata:
  name: child-subnet-1
  namespace: default
spec:
  ipv4SubnetSize: 16
---
kind: Subnet
metadata:
  name: child-subnet-2
  namespace: default
spec:
  ipv4SubnetSize: 16
---
metadata:
  name: binding-auto-1
  namespace: default
spec:
  subnetName: child-subnet-1
  targetSubnetName: parent-subnet
  # Intentionally omit vlanTrafficTag to trigger auto-allocation
```
a7874d761ca643942e507aa66be222fd

3. Verify Auto-allocation

Run kubectl apply -f test-vlan-auto.yaml.
Wait for the resources to be ready: kubectl get subnetconnectionbindingmap binding-auto-1.
Check the allocation result: kubectl get subnetconnectionbindingmap binding-auto-1 -o yaml.
Expected Result: spec.vlanTrafficTag is automatically injected (e.g., 0, 100, or another value within the [0, 4094] range).

7a1fe4de34621bfbd7cfde509268004e

c75f5a5bfc8dcf83ccedba3fe5dcf779


pkg/nsx/services/vlanpool/vlanpool.go
```diff
@@ -0,0 +1,60 @@
...
+type Service struct {
+	collector VlanCollector
+}
+
+func NewService(collector *subnetbinding.BindingService) *Service {
+	return &Service{collector: collector}
+}
+
+// Allocate picks an available VLAN. When preferred is valid and unused on the target, it is returned;
+// otherwise the smallest free ID from [0, 4094] is used.
```
wenyingd: Better to not use 0

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -276,6 +287,70 @@ func (r *Reconciler) validateDependency(ctx context.Context, bindingMap *v1alpha
...
+		log.Error(err, "Failed to get Subnet CR for VLAN auto allocation", "Subnet", childSubnetKey.String())
+		return &errorWithRetry{
+			message: fmt.Sprintf("Unable to get Subnet CR %s for VLAN auto allocation", bindingMap.Spec.SubnetName),
+			error:   fmt.Errorf("failed to get Subnet %s in Namespace %s: %w", bindingMap.Spec.SubnetName, bindingMap.Namespace, err),
+			retry:   false,
+		}
+	}
+
+	preferred := int64(-1)
+	if childSubnet.Status.VLANExtension.VLANID != 0 {
```
dantingl: Why we get VLAN ID from child Subnet? The vlan ID of vlan extention subnet is a different use case

dantingl (commented):
Could you add more local test cases that some existing NSX BindingMap already allocate some VLAN ID

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -276,6 +287,70 @@ func (r *Reconciler) validateDependency(ctx context.Context, bindingMap *v1alpha
...
+	vlan, err := r.VlanPoolService.Allocate(parentSubnetPaths, string(bindingMap.UID), preferred)
+	if err != nil {
+		return &errorWithRetry{
+			message: err.Error(),
+			error:   err,
+			retry:   true,
+		}
+	}
+
+	if err := r.patchSpecVlanTrafficTag(ctx, bindingMap, vlan); err != nil {
```
wenyingd: Firstly, it is not suggested to let the reconciler update `SubnetConnectionBindingMap` spec, which is possible to break user's intent, e.g., we may not differentiate if this VLAN is set by user or by nsx-operator. Updating spec is different from updating status. So I would prefer to introduce a field in SubnetConnectionBindingMap.status to record the final VLAN, no matter it is from user's intent (spec) or from the auto-calculation of nsx-operator.

Secondly, the current implementation is 1) patch the CR first, and then 2) continue the reconciliation (sync to nsx subnet-connection-binding-map) without a re-queue. It makes the reconciler usage mixed. As we have patched the CR, another update event will be triggered, in which the CR could have all required configurations. So I would prefer to split the "VLAN allocation" and "nsx resource create/update" into two independenty reconciliations, e.g., 1) if VLAN is not specified in one event handling cycle, the reconciler only allocates VLAN, 2) if everything is ready, do the reconciliation to sync to NSX.

Even more, if we must go with the route to patch the CR spec rather than introducing a field in the status, what do you think to use webhook for allocating VLAN?

Personally, I prefer to use the status to show the auto-allcated VLAN rather than patching spec..

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -276,6 +287,70 @@ func (r *Reconciler) validateDependency(ctx context.Context, bindingMap *v1alpha
...
+		return &errorWithRetry{
+			message: fmt.Sprintf("Unable to get Subnet CR %s for VLAN auto allocation", bindingMap.Spec.SubnetName),
+			error:   fmt.Errorf("failed to get Subnet %s in Namespace %s: %w", bindingMap.Spec.SubnetName, bindingMap.Namespace, err),
+			retry:   false,
+		}
+	}
+
+	preferred := int64(-1)
+	if childSubnet.Status.VLANExtension.VLANID != 0 {
+		preferred = int64(childSubnet.Status.VLANExtension.VLANID)
```
wenyingd: Why we prefer to use child subnet's vlan extension tag? I don't see the relationship between subnet-connection-binding-map with vlan-extension subnet's tag.

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -276,6 +287,70 @@ func (r *Reconciler) validateDependency(ctx context.Context, bindingMap *v1alpha
 	return childSubnetPath, parentSubnetPaths, nil
 }
 
+func (r *Reconciler) reconcileVlanTrafficTag(ctx context.Context, bindingMap *v1alpha1.SubnetConnectionBindingMap, parentSubnetPaths []string) *errorWithRetry {
+	if bindingMap.Spec.HasVlanTrafficTag() {
+		vlan := *bindingMap.Spec.VLANTrafficTag
+		if err := r.VlanPoolService.ValidateManualVlan(parentSubnetPaths, vlan, string(bindingMap.UID)); err != nil {
```
wenyingd: Why we need to verify the user input vlan tag?

wenyingd: What if this vlan tag is allocated by the controller? e.g., in round 1, controller successfully reconciles the CR, then user has made update, the resource enqueue again, then this function may break the controller to do the expected reconciliation.

pkg/nsx/services/vlanpool/vlanpool.go
```diff
@@ -0,0 +1,60 @@
...
+func (s *Service) Allocate(parentSubnetPaths []string, excludeCRUID string, preferred int64) (int64, error) {
```
wenyingd: Can we try a valid tag (not used) from the cache first, then query from NSX only when NSX returns error that the value is already used?

pkg/nsx/services/vlanpool/vlanpool.go
```diff
@@ -0,0 +1,60 @@
...
+	used, err := s.collector.CollectUsedVlansOnParentSubnets(parentSubnetPaths, excludeCRUID)
+	if err != nil {
+		return 0, err
+	}
+
+	if preferred >= poolStart && preferred <= poolEnd && !used.Has(int(preferred)) {
+		return preferred, nil
+	}
+
+	for id := poolStart; id <= poolEnd; id++ {
```
wenyingd, line 98: You may need to think about the race conditions in parallel.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Updated

pkg/nsx/services/vlanpool/vlanpool.go
wenqiq: done

pkg/nsx/services/vlanpool/vlanpool.go
wenqiq: updated

pkg/nsx/services/vlanpool/vlanpool.go
wenqiq: updated

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Thanks for reviewing. I've updated the implementation accordingly.

On splitting reconciliation: Agreed. VLAN allocation and NSX sync are now handled in separate reconcile cycles. When `vlanTrafficTag` is omitted, the reconciler only allocates a VLAN, persists it to the CR spec, and returns. A subsequent reconcile (triggered by the spec update) validates the VLAN and syncs to NSX.

On webhook-based allocation: I don't think a mutating webhook is a good fit here. VLAN allocation depends on dependent Subnets being ready and NSX state (cache/NSX query, pending pool, overlap retry), which fits the controller lifecycle better than admission.

On persisting to spec vs. status: We still persist the auto-allocated VLAN in spec.vlanTrafficTag (The API definition still needs to be confirmed by @jianjuns). To distinguish operator-allocated vs. user-provided VLANs after persistence, we set annotation nsx.vmware.com/auto_allocated_vlan_traffic_tag when writing the spec. This keeps pending VLAN pool commit/release correct in the second reconcile cycle.

wenqiq:
> Could you add more local test cases that some existing NSX BindingMap already allocate some VLAN ID

Updated the E2E test cases to include this.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: updated

build/yaml/crd/vpc/crd.nsx.vmware.com_subnetconnectionbindingmaps.yaml
```diff
@@ -79,14 +79,13 @@ spec:
               vlanTrafficTag:
                 description: |-
                   VLANTrafficTag is the VLAN tag configured in the binding. Note, the value of VLANTrafficTag should be
-                  unique on the target Subnet or SubnetSet.
+                  unique on the target Subnet or SubnetSet. When omitted, the operator auto-allocates a VLAN ID.
                 format: int64
                 maximum: 4094
-                minimum: 0
```
jianjuns, line 85: Please update the commit description (it now says [0-4094]) too.

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -276,6 +328,103 @@ func (r *Reconciler) validateDependency(ctx context.Context, bindingMap *v1alpha
...
+	})
+	if err != nil {
+		log.Error(err, "Failed to persist auto-allocated vlanTrafficTag on SubnetConnectionBindingMap", "SubnetConnectionBindingMap", key.String())
+		return err
+	}
+	log.Debug("Persisted auto-allocated vlanTrafficTag on SubnetConnectionBindingMap", "SubnetConnectionBindingMap", key.String(), "vlanTrafficTag", vlan)
+	return nil
+}
+
+func (r *Reconciler) reconcileVlanTrafficTag(ctx context.Context, bindingMap *v1alpha1.SubnetConnectionBindingMap, parentSubnetPaths []string, fromNSX bool) *errorWithRetry {
```
jianjuns: For a VLAN ext Subnet, could we first try using the Subnet's VLAN ID, when it is not used yet?

build/yaml/crd/vpc/crd.nsx.vmware.com_subnetconnectionbindingmaps.yaml
wenqiq: done

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: It seems there is no architectural relationship between the VLAN extension Subnet's VLAN ID and the `vlanTrafficTag`, correct? This aligns with previous comments that mentioned: https://github.com/vmware-tanzu/nsx-operator/pull/1445#discussion_r3354987284 https://github.com/vmware-tanzu/nsx-operator/pull/1445#discussion_r3355274183

wenqiq: cc @dantingl @wenyingd

build/yaml/crd/vpc/crd.nsx.vmware.com_subnetconnectionbindingmaps.yaml
```diff
@@ -79,14 +79,13 @@ spec:
               vlanTrafficTag:
```
wenyingd, line 79: @jianjuns could you help share your thoughts regarding how to show the "auto-allocated" vlan in the CR?

We have two options,
option 1: Re-use SubnetConnectionBindingMap CR spec.vlanTrafficTag for the allocated VLAN,
option 2: keep spec.vlanTrafficTag as empty (the same as user input), and introduce a field in status to show the final VLAN configured in NSX.

The current implementation is option1, the advantage is we don't need to update the APIs. In the meanwhile, it is impossible to know whether the VLAN is allocated by system or provided by the user (user input spec is modified). For nsx-operator logic, it may have two request to kube-apiserver : 1) `update` the CR first and 2) `update status` with the conditions. 

For option 2, the advantage is the original user intent is kept, and nsx-operator only send request  to apiserver to "update status". In the meanwhile, we may need to introduce a new field in status to record the final VLAN used in NSX.

wenqiq: Thanks @wenyingd, I`d like to add a point is that, during the reconcile process we set annotation nsx.vmware.com/auto_allocated_vlan_traffic_tag when writing the spec, this helps distinguish operator-allocated vs. user-provided VLANs.  see https://github.com/vmware-tanzu/nsx-operator/pull/1445#discussion_r3379620060

build/yaml/crd/vpc/crd.nsx.vmware.com_subnetconnectionbindingmaps.yaml
dantingl: @jianjuns NSX API supports VLAN [0-4094], so do you think we should keep the same range in CRD definition including VLAN 0, but do auto-allocation from VLAN 1?

jianjuns: VLAN 0 maps to the PARENT (VNIC) port. How can it be used for CHILD ports?

pkg/controllers/subnetbinding/subnetbinding_controller.go
jianjuns: While what you said is true, from VLAN network consumer perspective - esp. for those who used to consume VLAN DVPG or SV VDS network directly, but migrate to VPC and VLAN extension - it would be much easier to keep using the same VLAN IDs in the guest.
Think about the major use case of Telco.

build/yaml/crd/vpc/crd.nsx.vmware.com_subnetconnectionbindingmaps.yaml
jianjuns: In my understanding, K8s API design does not reuse the Spec field to populate system generated values, while NSX Policy API has such cases. Could you also check with @tnqn on this one?

wenqiq: I checked, and it appears these two examples are similar to our situation, right? K8s dynamically allocates values for omitted spec fields (e.g., [ClusterIP](https://kubernetes.io/docs/concepts/services-networking/cluster-ip-allocation/), [NodePort](https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport)) and writes them back to the spec.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Good point. Updated.

build/yaml/crd/vpc/crd.nsx.vmware.com_subnetconnectionbindingmaps.yaml
jianjuns: Sure, but there can also be an argument that Service v1 API is a legacy design pattern. Could we understand the NSX Policy API pattern for such cases to decide?

tnqn: It's recommended to store allocated value in `status` for new APIs: https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#representing-allocated-values.

Yes, there are examples in K8s API that allocated values are stored in `spec` but K8s apiserver can reject invalid input synchronously so the value in `spec` can be consumed safely. For the PR's case, nsx-operator cannot confirm/reject the user requested vlan ID synchronously, which could cause potential problems for consumption.

I think NSX reuses the property for user to provide the desired value and the system to write the allocated value, but it's also able to reject the request synchronously if the user desired value cannot be allocated. And it doesn't have another good place to store the allocated value.

jianjuns: > nsx-operator cannot confirm/reject the user requested vlan ID synchronously, which could cause potential problems for consumption.

@tnqn What problems you mean here?

tnqn: If there is a consumer using the vlan ID in spec as source of truth to configure the datapath (antrea in this case?), user may provide invalid or conflicting vlan ID to disrupt the datapath intentionally or unintentionally.

jianjuns: But that can anyway happen since we will continue to support user specified VLAN ID in the Spec. The discussion here is just that when user does not specify a VLAN ID, should system auto-allocated VLAN ID be set in the same Spec field.

tnqn: If `status.vlanID` is introduced, the consumer should use `status.vlanID` only, which resolves the above problem.
- When user specify `spec.vlanID`, if it cannot be realized in NSX, `status.vlanID` will not be set.
- When user specify `spec.vlanID`, if it can be realized in NSX, `status.vlanID` will be set to `spec.vlanID`.
- When user omit `spec.vlanID`, `status.vlanID` will be set to the allocated VLAN ID.

jianjuns: I see what you mean now. However, even without VLAN ID in status, the consumer should check whether the resource is realized from the status before using it?

tnqn: It could but it's more complex, especially if the vlan ID is mutable: consumer needs to know the realized status is for the previous ID or the current ID. There could be solutions anyway even if reusing the field in spec. It's just a recommendation to separate the user input and the actual status to make it simpler.

jianjuns: Make sense. When NSX supports VLAN ID auto-allocation and we also expose SubnetConnectionBindingMap in CCI, in your mind how should we define the CCI API - auto-allocated VLAN ID in Status? @tnqn

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
```diff
@@ -18,18 +18,20 @@ type SubnetConnectionBindingMapSpec struct {
 	// +kubebuilder:validation:Optional
 	TargetSubnetName string `json:"targetSubnetName,omitempty"`
 	// VLANTrafficTag is the VLAN tag configured in the binding. Note, the value of VLANTrafficTag should be
-	// unique on the target Subnet or SubnetSet.
+	// unique on the target Subnet or SubnetSet. When omitted, the operator auto-allocates a VLAN ID.
```
jianjuns: Probably: "When omitted, the system automatically allocates a VLAN ID."

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
```diff
@@ -18,18 +18,20 @@ type SubnetConnectionBindingMapSpec struct {
...
+	// +kubebuilder:validation:Optional
+	VLANTrafficTag *int64 `json:"vlanTrafficTag,omitempty"`
 }
 
 // SubnetConnectionBindingMapStatus defines the observed state of SubnetConnectionBindingMap.
 type SubnetConnectionBindingMapStatus struct {
 	// Conditions described if the SubnetConnectionBindingMaps is configured on NSX or not.
 	// Condition type ""
 	Conditions []Condition `json:"conditions,omitempty"`
+	// VLANID is the VLAN ID realized in NSX for this binding.
```
jianjuns: Do not use "NSX" in Supervisor API definition and comments.

Just say" the realized VLAN ID for this binding".

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
wenqiq: Done.

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
wenqiq: Thanks for the reminder. I have updated it.

wenqiq:
The related e2e test case passed:
```
└── ✅ TestSubnetSet (2m24.873s)
    └── ✅ SequentialTests (2m10.585s)
        ├── ✅ case=DefaultSubnetSet (24.165s)
        ├── ✅ case=NoIPSubnet (4.057s)
        ├── ✅ case=SharedSubnetSet (7.518s)
        ├── ✅ case=SubnetBindingAutoVLANAllocation (17.333s)
        ├── ✅ case=SubnetBindingAutoVLANWithExistingBinding (14.504s)
```
https://10.191.159.144/job/nsx_operator/2489

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
```diff
@@ -18,18 +18,20 @@ type SubnetConnectionBindingMapSpec struct {
...
+	VLANTrafficTag *int64 `json:"vlanTrafficTag,omitempty"`
 }
 
 // SubnetConnectionBindingMapStatus defines the observed state of SubnetConnectionBindingMap.
 type SubnetConnectionBindingMapStatus struct {
 	// Conditions described if the SubnetConnectionBindingMaps is configured on NSX or not.
 	// Condition type ""
 	Conditions []Condition `json:"conditions,omitempty"`
+	// VLANID is the realized VLAN ID for this binding.
+	VLANID *int64 `json:"vlanID,omitempty"`
```
wenyingd: What's the use case when `VLANID` is empty? For the brown field (upgrade case), would the controller automatically update this field?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -118,13 +122,44 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	if vlanErr != nil {
+		r.releaseAutoAllocatedVlan(parentSubnetPaths, bindingMapCR, vlanID)
+		r.StatusUpdater.UpdateFail(ctx, bindingMapCR, vlanErr, "failed to reconcile VLAN traffic tag", updateBindingMapStatusWithUnreadyCondition, "VlanAllocationFailed", vlanErr.message)
+		if !vlanErr.retry {
+			return common.ResultNormal, nil
+		}
 		return common.ResultRequeue, nil
 	}
+
+	autoAllocatedVlan := !bindingMapCR.Spec.HasVlanTrafficTag()
```
wenyingd: Shall we return `autoAllocatedVlan` from function`reconcileVlanTrafficTag`? And we can always return "err==nil" in case "reconcileVlanTrafficTag == False" ?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -118,13 +122,44 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+		if autoAllocatedVlan && nsxutil.IsOverlapVlanError(err) {
+			log.Info("VLAN allocation conflict with NSX cache, fallback to query NSX", "SubnetConnectionBindingMap", req.NamespacedName)
+			r.releaseAutoAllocatedVlan(parentSubnetPaths, bindingMapCR, vlanID)
+			vlanID2, vlanErr2 := r.reconcileVlanTrafficTag(ctx, bindingMapCR, parentSubnetPaths, true)
+			if vlanErr2 != nil {
+				r.releaseAutoAllocatedVlan(parentSubnetPaths, bindingMapCR, vlanID2)
+				r.StatusUpdater.UpdateFail(ctx, bindingMapCR, vlanErr2, "failed to reconcile VLAN traffic tag from NSX", updateBindingMapStatusWithUnreadyCondition, "VlanAllocationFailed", vlanErr2.message)
+				return common.ResultRequeue, nil
+			}
+			if err2 := r.SubnetBindingService.CreateOrUpdateSubnetConnectionBindingMap(bindingMapCR, vlanID2, childSubnetPath, parentSubnetPaths); err2 != nil {
```
wenyingd: Do you mean we should have a second retry on the auto-allocated VLAN after receiving the `nsxutil.IsOverlapVlanError(err)` ? What if we still have conflict on the second auto-allocated value?

wenyingd: What do you think if we move the retry logic to the nsx services layer? Even more, maybe we can should also move the auto-allocation logic down to the nsx services? And we can leave the CR level validations inside the controller.

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -276,6 +311,64 @@ func (r *Reconciler) validateDependency(ctx context.Context, bindingMap *v1alpha
 	return childSubnetPath, parentSubnetPaths, nil
 }
 
+func (r *Reconciler) releaseAutoAllocatedVlan(parentSubnetPaths []string, bindingMap *v1alpha1.SubnetConnectionBindingMap, vlanID int64) {
+	if bindingMap.Spec.HasVlanTrafficTag() || vlanID == 0 {
```
wenyingd: We should let the caller to check `bindingMap.Spec.HasVlanTrafficTag()` and let this function focusing on the release logic itself?

Why we have check `vlanID == 0` ? 0 is not a valid value from the CRD?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -276,6 +311,64 @@ func (r *Reconciler) validateDependency(ctx context.Context, bindingMap *v1alpha
...
+		return
+	}
+	r.VlanPoolService.ReleasePending(parentSubnetPaths, vlanID)
+}
+
+func (r *Reconciler) commitAutoAllocatedVlan(parentSubnetPaths []string, bindingMap *v1alpha1.SubnetConnectionBindingMap, vlanID int64) {
+	if bindingMap.Spec.HasVlanTrafficTag() || vlanID == 0 {
```
wenyingd: ditto

pkg/nsx/services/subnetbinding/vlan_allocation.go
```diff
@@ -0,0 +1,136 @@
...
+			break
+		}
+		cursor = response.Cursor
+	}
+
+	return bindings, nil
+}
+
+// ValidateVlanOnParentSubnets checks that vlan is not already used on any of the parent Subnet paths.
+func (s *BindingService) ValidateVlanOnParentSubnets(parentSubnetPaths []string, vlan int64, excludeCRUID string, fromNSX bool) error {
```
wenyingd: I didn't find consumption on this function.

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -276,6 +311,64 @@ func (r *Reconciler) validateDependency(ctx context.Context, bindingMap *v1alpha
...
+			return 0, &errorWithRetry{
+				message: err.Error(),
+				error:   err,
+				retry:   true,
+			}
+		}
+		return vlan, nil
+	}
+
+	// Try to reuse already allocated VLAN from NSX cache
```
wenyingd: Isn't the allocated VLAN the one in bindingMap.status.VlanID ?

wenyingd: Thinking more, we should use this order: use existing vlanIDs (succeeded synced value) -> use the intent value from spec -> allocate a value

pkg/nsx/services/subnetbinding/vlan_allocation.go
wenqiq: You are absolutely right. It appears that its logic was moved to `ValidateManualVlan` in `pkg/nsx/services/vlanpool/vlanpool.go`, but the original function was left behind. I have removed the unused function.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: 1. Regarding HasVlanTrafficTag(): releaseAutoAllocatedVlan is called in multiple error-handling branches (at least 5 places in the Reconcile loop). If we move the check to the caller, we would have to duplicate this if condition in every error branch, making the main reconciliation logic quite verbose. Keeping it inside makes the function a centralized "safe-cleanup" helper (it simply becomes a no-op if the VLAN is manually specified), which keeps the caller's code much cleaner.

2. Regarding the `vlanID == 0` check: You are absolutely right that 0 is not a valid value from the CRD. However, this check is not for validating CRD input. When reconcileVlanTrafficTag() fails (e.g., VLAN pool is exhausted), it returns the default Go zero value 0 for vlanID alongside the error. Since the caller still triggers the release function in the error branch to ensure rollback, the `vlanID == 0` check acts as an internal safety guard to prevent the underlying VlanPoolService from attempting to release an unallocated/invalid VLAN 0.

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
wenqiq: 1. Use case when VLANID is empty: The VLANID is empty (nil) upon the initial creation of the CR before the controller picks it up and processes it. It will also remain empty if the controller fails to allocate a VLAN (e.g., the VLAN pool is exhausted) or if dependencies (like parent/child subnets) are not ready. It essentially represents the "not yet allocated or realized" state.

2. For the brown field (upgrade case): Yes, the controller will automatically update this field. During an operator upgrade, when the new controller starts up, it will enqueue and reconcile all existing SubnetConnectionBindingMap CRs. Since the existing CRs already have spec.vlanTrafficTag specified (as it was a required field previously), the controller will read that value, realize it, and automatically populate .status.vlanID during the status update step without any manual intervention.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: I assume by `reconcileVlanTrafficTag == False` you meant `autoAllocatedVlan == False` ?

Returning autoAllocatedVlan from reconcileVlanTrafficTag: Yes, this makes sense. I updated this.

Always returning `err == nil` when `autoAllocatedVlan == False`: If we always return `err == nil` for manually specified VLANs, it means we can remove the local `ValidateManualVlan` check and rely entirely on NSX to validate overlaps?

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Yes, exactly.

Why we have a second retry: The first `reconcileVlanTrafficTag` call (with `fromNSX=false`) allocates a VLAN based on our local cache. Due to potential race conditions or cache delays, this local VLAN might already be taken in NSX, which triggers the `nsxutil.IsOverlapVlanError`. In this fallback block, we retry `reconcileVlanTrafficTag(..., fromNSX=true)`. By passing true, the `VlanPoolService` ignores the local cache and strictly queries the NSX API to find a real-time available VLAN.

What if it still conflicts on the second attempt: If the second attempt (vlanID2) still results in a conflict (or any other API error), the code falls through to update the CR status to ConfigureFailed and returns `common.ResultRequeue`. Returning Requeue means the controller will continuously re-queue and re-trigger the entire Reconcile loop. So it will keep trying to allocate a new VLAN and realize it on NSX until it eventually succeeds.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Why use NSX Cache instead of Status.VlanID: In crash-recovery scenarios, if the operator fails after configuring NSX but before updating the CR status, Status.VlanID will be empty. The NSX cache reflects the true remote state, which prevents duplicate allocations.

Regarding the evaluation order: I think the Spec (user intent) must always take precedence. If we put Existing first, user updates to Spec.vlanTrafficTag would be ignored because the controller would just blindly return the old existing value. For example, if a binding has an existing auto-allocated VLAN (e.g., 10), and the user later decides to manually set spec.vlanTrafficTag to 20, the controller would return the existing 10 and silently ignore the user's new intent. Therefore, the current order Spec (intent) -> Existing -> Allocate is necessary to ensure the controller correctly reacts to user updates in the Spec, while still reusing existing VLANs when auto-allocation is desired.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: > What do you think if we move the retry logic to the nsx services layer? Even more, maybe we can should also move the auto-allocation logic down to the nsx services? And we can leave the CR level validations inside the controller.

This is a very good architectural point. I completely agree that moving it to the NSX services layer provides better separation of concerns and a cleaner controller. However, since this involves refactoring the service interfaces and state management, I'd prefer to address it in a separate follow-up PR to keep this feature PR focused.

wenyingd: 1. Please add comments to explain that why we may have a second try for `reconcileVlanTrafficTag` .
2. Would you create an issue to track the follow-up change?
3. Extract the retry logic ( line 138-150 into a dedicated function) to make this reconcile logic clear.

pkg/nsx/services/subnetbinding/vlan_allocation.go
```diff
@@ -0,0 +1,118 @@
...
+	return ""
+}
+
+func (s *BindingService) listBindingMapsByParentSubnetPath(parentSubnetPath string) ([]*model.SubnetConnectionBindingMap, error) {
+	if s == nil || s.NSXClient == nil || s.NSXClient.QueryClient == nil {
+		return nil, fmt.Errorf("NSX query client is not initialized")
+	}
+
+	pathEscaped := strings.ReplaceAll(parentSubnetPath, "/", "\\/")
+	queryParam := fmt.Sprintf("%s:%s AND marked_for_delete:false AND subnet_path:%s",
```
wenyingd, line 77: Please add a comment here, once https://github.com/vmware-tanzu/nsx-operator/pull/1443 is merged, we need a update here to support bridge mode.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenyingd: please add comment to explain the case about vlan==0

wenqiq: done

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Done. Thanks for reviewing.

pkg/nsx/services/subnetbinding/vlan_allocation.go
wenqiq: Updated.

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
```diff
@@ -18,18 +18,20 @@ type SubnetConnectionBindingMapSpec struct {
...
 	// VLANTrafficTag is the VLAN tag configured in the binding. Note, the value of VLANTrafficTag should be
-	// unique on the target Subnet or SubnetSet.
+	// unique on the target Subnet or SubnetSet. When omitted, the system automatically allocates a VLAN ID.
 	// +kubebuilder:validation:Maximum:=4094
-	// +kubebuilder:validation:Minimum:=0
-	// +kubebuilder:validation:Required
-	VLANTrafficTag int64 `json:"vlanTrafficTag"`
+	// +kubebuilder:validation:Minimum:=1
+	// +kubebuilder:validation:Optional
+	VLANTrafficTag *int64 `json:"vlanTrafficTag,omitempty"`
```
dantingl, line 25: Can we keep the original type, if VKS consumes this property, it would be a breaking API change, right?

build/yaml/crd/vpc/crd.nsx.vmware.com_subnetconnectionbindingmaps.yaml
```diff
@@ -33,8 +33,8 @@ spec:
       name: targetSubnetSet
       type: string
     - description: Vlan used in the NSX SubnetConnectionBindingMap
-      jsonPath: .spec.vlanTrafficTag
-      name: vlanTrafficTag
```
dantingl, line 37: I think we should keep the vlanTrafficTag in spec which is the existing way for user to set it manually, just add vlanID in status for auto-allocated one?

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
jianjuns: I think it does not really break API compatibility, and the old client will get value 0 from the VLANTrafficTag field?

Do you mean any consumer will assume non-zero VLAN ID here? Like UI? Do we support old version UI work with new version CRDs?

wenqiq: I think using the `int64` type won't meet our current feature requirements. Assuming we still use the `int64` type, if the user omits this value, it will be serialized to JSON as `"vlanTrafficTag": 0`.

The core difference between `VLANTrafficTag int64` and `VLANTrafficTag *int64` with omitempty lies in zero-value semantics and JSON serialization behavior; since we added the validation `Minimum:=1`, there is no need to worry about this.

build/yaml/crd/vpc/crd.nsx.vmware.com_subnetconnectionbindingmaps.yaml
wenqiq: The removal of `vlanTrafficTag` you may have noticed in the diff only affects the `kubectl get` print columns (additionalPrinterColumns). I updated the print column to display `status.vlanID` instead of `spec.vlanTrafficTag`.

The reason we only show the status field in print columns is to align with standard Kubernetes best practices. The design philosophy of `kubectl get` is to provide a "summary view," while leaving complete data for other commands and output formats. Consequently, I think we should not display values from both the spec and status in full.

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -276,6 +301,92 @@ func (r *Reconciler) validateDependency(ctx context.Context, bindingMap *v1alpha
...
+				message: err.Error(),
+				error:   err,
+				retry:   true,
+			}
+		}
+		return vlan, false, nil
+	}
+
+	// Try to reuse already allocated VLAN from NSX cache
+	existingBMs := r.SubnetBindingService.BindingStore.GetByIndex("bindingMapCRUID", string(bindingMap.UID))
```
wenyingd: Shall we have a pre-check that line 360 is executed only when `fromNSX==false`? otherwise, it may bypass the allocation logic based on NSX resources.

A potential failure workflow is like this,
1. user create a SubnetConnectionBindingMap with VLAN auto-allocation -> it succeeded -> a bindingmap resource is added in the store with index of the CR UUID
2. user updates parent subnet in the SubnetConnectionBindingMap, -> the existing VLAN ID is returned from the store with the first call on `reconcileVlanTrafficTag`  -> NSX returns `IsOverlapVlanError` ->  a second call on `reconcileVlanTrafficTag` still returns the existing VLAN ID in the store.

pkg/nsx/services/subnetbinding/vlan_allocation.go
```diff
@@ -0,0 +1,119 @@
...
+			}
+			if bm.VlanTrafficTag != nil {
+				used.Insert(int(*bm.VlanTrafficTag))
+			}
+		}
+	}
+	return used, nil
+}
+
+func bindingMapCRUID(bm *model.SubnetConnectionBindingMap) string {
```
wenyingd, line 53: Is this function the same as `bindingMapCRUIDIndexFunc` ?

pkg/nsx/services/subnetbinding/vlan_allocation.go
```diff
@@ -0,0 +1,119 @@
...
+	// TODO：update here to support bridge mode.
+	queryParam := fmt.Sprintf("%s:%s AND marked_for_delete:false AND subnet_path:%s",
+		servicecommon.ResourceType, ResourceTypeSubnetConnectionBindingMap, pathEscaped)
+
+	converter := servicecommon.NewConverter()
+	bindings := make([]*model.SubnetConnectionBindingMap, 0)
+	var cursor *string
+	pageSize := searchPageSize
+
+	for {
```
wenyingd: can we call `func (service *Service) SearchResource(resourceTypeValue string, queryParam string, store Store, filter Filter) (uint64, error) {` to avoid duplicated code ?

pkg/nsx/util/utils.go
```diff
@@ -785,3 +785,20 @@ func ParseDHCPMode(mode string) string {
...
+
+func IsOverlapVlanError(err error) bool {
+	if err == nil {
+		return false
+	}
+	if nsxApiErr, ok := err.(*NSXApiError); ok {
+		if nsxApiErr.ErrorCode != nil && *nsxApiErr.ErrorCode == 8327 {
+			return true
+		}
+		for _, relatedErr := range nsxApiErr.RelatedErrors {
```
wenyingd, line 798: Could you explain more about the cases NSX policy API returns errorCode 8327 v.s.  NSX policy API returns relateErroes with code 8327?

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Good catch, updated it.

pkg/nsx/services/subnetbinding/vlan_allocation.go
wenqiq: I think keeping them separate is better, it`s standard practice in Kubernetes operator development to maintain a clean separation between cache-indexing logic and business-logic helpers. There is no need to modify or merge these two functions.

pkg/nsx/services/subnetbinding/vlan_allocation.go
wenqiq: Make sense, I updated it.

pkg/nsx/util/utils.go
wenqiq: I checked the API in a local testbed, confirmed the error code is 640873, and have updated the function.
```
curl -k -X PUT "https://${NSX_MGR_IP}/policy/api/v1/orgs/${ORG_ID}/projects/${PROJECT_ID}/vpcs/${VPC_ID}/subnets/child-subnet-c/subnet-connection-binding-maps/binding-c" \
  -u "admin:${NSX_PASSWORD}" \
  -H "Content-Type: application/json" \
  -d '{
    "resource_type": "SubnetConnectionBindingMap",
    "id": "binding-c",
    "display_name": "Binding-C",
    "vlan_traffic_tag": 100,
    "subnet_path": "/orgs/default/projects/proj-b/vpcs/vpc-b/subnets/parent-subnet-a" 
  }'
{
  "httpStatus" : "BAD_REQUEST",
  "error_code" : 640873,
  "module_name" : "policy",
  "error_message" : "Cannot create path=[/orgs/default/projects/proj-b/vpcs/vpc-b/subnets/child-subnet-c/subnet-connection-binding-maps/binding-c] under path=[/orgs/default/projects/proj-b/vpcs/vpc-b/subnets/child-subnet-c] because VLAN tag 100 is already used by path=[/orgs/default/projects/proj-b/vpcs/vpc-b/subnets/child-subnet-b/subnet-connection-binding-maps/binding-b] with the same association type."
}% 
```

pkg/nsx/services/subnetbinding/vlan_allocation.go
wenyingd: I mean no need to write the same logic again. You can directly call the existing function?

pkg/nsx/util/utils.go
```diff
@@ -785,3 +786,20 @@ func ParseDHCPMode(mode string) string {
 		return ""
 	}
 }
+
+func IsVpcOverlapVlanError(err error) bool {
+	if err == nil {
+		return false
+	}
+	if nsxApiErr, ok := err.(*NSXApiError); ok {
+		if nsxApiErr.ErrorCode != nil && *nsxApiErr.ErrorCode == 640873 {
```
wenyingd: Better to define a constant for the known error code.

pkg/nsx/util/utils.go
wenyingd: I didn't see a "relatedErrors" in this example, did you check that line 798 to 802 is a valid logic path?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -276,6 +301,94 @@ func (r *Reconciler) validateDependency(ctx context.Context, bindingMap *v1alpha
...
+			return 0, false, &errorWithRetry{
+				message: err.Error(),
+				error:   err,
+				retry:   true,
+			}
+		}
+		return vlan, false, nil
+	}
+
+	// Try to reuse already allocated VLAN from NSX cache
```
wenyingd:
```suggestion
	// Try to reuse already allocated VLAN from cache
```

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -276,6 +301,94 @@ func (r *Reconciler) validateDependency(ctx context.Context, bindingMap *v1alpha
...
+
+func (r *Reconciler) commitAutoAllocatedVlan(parentSubnetPaths []string, bindingMap *v1alpha1.SubnetConnectionBindingMap, vlanID int64) {
+	// vlanID == 0 means the VLAN auto-allocation failed or hasn't occurred, so there is nothing to commit.
+	if bindingMap.Spec.HasVlanTrafficTag() || vlanID == 0 {
+		return
+	}
+	r.VlanPoolService.CommitPending(parentSubnetPaths, vlanID)
+}
+
+func (r *Reconciler) reconcileVlanTrafficTag(ctx context.Context, bindingMap *v1alpha1.SubnetConnectionBindingMap, parentSubnetPaths []string, fromNSX bool) (int64, bool, *errorWithRetry) {
```
wenyingd, line 345: It looks this function always return `errorWithRetry.retry == true` when error != nil. If that's the intent, shall we move the retry to the caller?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -118,13 +122,34 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 	}
 
-	if err := r.SubnetBindingService.CreateOrUpdateSubnetConnectionBindingMap(bindingMapCR, childSubnetPath, parentSubnetPaths); err != nil {
-		// Update SubnetConnectionBindingMap with not-ready condition
-		r.StatusUpdater.UpdateFail(ctx, bindingMapCR, err, "failure to configure SubnetConnectionBindingMaps on NSX", updateBindingMapStatusWithUnreadyCondition, "ConfigureFailed", fmt.Sprintf("Failed to realize SubnetConnectionBindingMap %s on NSX", req.Name))
+	vlanID, autoAllocatedVlan, vlanErr := r.reconcileVlanTrafficTag(ctx, bindingMapCR, parentSubnetPaths, false)
+	if vlanErr != nil {
+		r.releaseAutoAllocatedVlan(parentSubnetPaths, bindingMapCR, vlanID)
+		r.StatusUpdater.UpdateFail(ctx, bindingMapCR, vlanErr, "failed to reconcile VLAN traffic tag", updateBindingMapStatusWithUnreadyCondition, "VlanAllocationFailed", vlanErr.message)
+		if !vlanErr.retry {
```
wenyingd, line 129: This is meaningless since it is always true ?

pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -118,13 +122,34 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			return common.ResultNormal, nil
+		}
 		return common.ResultRequeue, nil
 	}
+
+	if err := r.SubnetBindingService.CreateOrUpdateSubnetConnectionBindingMap(bindingMapCR, vlanID, childSubnetPath, parentSubnetPaths); err != nil {
+		if autoAllocatedVlan && nsxutil.IsVpcOverlapVlanError(err) {
+			var returnEarly bool
+			var res ctrl.Result
+			vlanID, res, returnEarly = r.handleVlanAllocationFallback(ctx, bindingMapCR, parentSubnetPaths, childSubnetPath, req, vlanID)
```
wenyingd: No need to introduce `returnEarly` , it can execute `return common.ResultRequeue, nil` when vlanID == 0?

pkg/nsx/util/utils.go
wenqiq: Yes, absolutely.
```
{
  "httpStatus" : "BAD_REQUEST",
  "error_code" : 500157,
  "module_name" : "Policy",
  "error_message" : "Error while creating objects of type:SubnetConnectionBindingMap",
  "related_errors" : [ {
    "httpStatus" : "BAD_REQUEST",
    "error_code" : 640873,
    "module_name" : "policy",
    "error_message" : "Cannot create path=[/orgs/default/projects/proj-b/vpcs/vpc-b/subnets/child-subnet-c/subnet-connection-binding-maps/binding-c] under path=[/orgs/default/projects/proj-b/vpcs/vpc-b/subnets/child-subnet-c] because VLAN tag 100 is already used by path=[/orgs/default/projects/proj-b/vpcs/vpc-b/subnets/child-subnet-b/subnet-connection-binding-maps/binding-b] with the same association type."
  } ]
}
```

pkg/nsx/util/utils.go
wenqiq: done

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: done

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Keeping errorWithRetry maintains consistency with other controller helpers, encapsulates retry logic, and avoids breaking existing unit test mocks. We can leave it as is.

pkg/controllers/subnetbinding/subnetbinding_controller.go
wenqiq: Good idea. Updated.

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
jianjuns: Let us name the field VLANTrafficTag to be consistent with NSX API.

wenyingd (dismissed):
LGTM

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
jianjuns: A reminder.

wenyingd (dismissed):

pkg/nsx/services/vlanpool/vlanpool.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	return 0, fmt.Errorf("no available VLAN in pool [%d, %d] for target Subnet or SubnetSet", poolStart, poolEnd)
+}
+
+// CommitPending removes a VLAN from the in-flight allocation set after it is realized on NSX.
+func (s *Service) CommitPending(parentSubnetPaths []string, vlan int64) {
+	state := s.getPoolState(parentSubnetPaths)
+	state.mu.Lock()
+	defer state.mu.Unlock()
+	state.pending.Delete(int(vlan))
```
dantingl, line 112: Do we need to delete the key from s.poolStates somewhere?

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
dantingl: +1 to use name `VLANTrafficTag` in CRD status as well

pkg/nsx/services/vlanpool/vlanpool.go
wenqiq: I think deleting the key from s.poolStates is not needed and should be avoided. Keeping keys in s.poolStates ensures concurrent reconciles targeting the same parent subnets share the exact same lock instance instead of re-creating new state objects.An idle poolState with an empty pending set consumes negligible memory.

pkg/nsx/services/vlanpool/vlanpool.go
```diff
@@ -0,0 +1,134 @@
...
+	state := s.getPoolState(parentSubnetPaths)
+	state.mu.Lock()
+	defer state.mu.Unlock()
+
+	used, err := s.collectUsed(parentSubnetPaths, excludeCRUID, fromNSX)
+	if err != nil {
+		return err
+	}
+	cleanupCommittedPending(used, state.pending)
+	if unavailableVlans(used, state.pending).Has(int(vlan)) {
```
dantingl, line 132: Should we insert manually assigned vlan ID to used cache, or we always sync with NSX to get used vlan?

pkg/nsx/services/vlanpool/vlanpool.go
```diff
@@ -0,0 +1,134 @@
...
+
+// Allocate picks an available VLAN. When preferred is valid and unused on the target, it is returned;
+// otherwise the smallest free ID from [1, 4094] is used.
+// Pending allocations are tracked per parent Subnet path set so parallel reconciles do not pick the same VLAN.
+func (s *Service) Allocate(parentSubnetPaths []string, excludeCRUID string, preferred int64, fromNSX bool) (int64, error) {
+	state := s.getPoolState(parentSubnetPaths)
+	state.mu.Lock()
+	defer state.mu.Unlock()
+
+	used, err := s.collectUsed(parentSubnetPaths, excludeCRUID, fromNSX)
```
dantingl, line 81: should we move collectUsed() before state.mu.Lock()? in case there is NSX API call for fallback case

wenqiq: Great catch! Moved collectUsed() before state.mu.Lock(). Holding state.mu.Lock() during network I/O may create unnecessary lock contention for concurrent allocations on the same VLAN pool.

pkg/nsx/services/vlanpool/vlanpool.go
wenqiq: Querying NSX on every reconcile pass introduces performance bottlenecks/latency and still doesn't prevent in-flight concurrent races within the operator.

pkg/apis/vpc/v1alpha1/subnetconnectionbindingmap_types.go
wenqiq: Done.

dantingl (approved):
LGTM

wenyingd (approved):

PR #1451, opened 2026-06-05
SubnetPort Status CRD update for IPv6

yanjunz97:
Update the SubnetPort Status to show DHCPv6 and RA mode status.
Update some description for interfaceIPType and staticIPAllocationType for better understanding

### DIFF
--- a/pkg/apis/vpc/v1alpha1/subnetport_types.go
+++ b/pkg/apis/vpc/v1alpha1/subnetport_types.go
@@ -26,15 +26,16 @@ type SubnetPortSpec struct {
 	// AddressBindings defines static address bindings used for the SubnetPort.
 	AddressBindings []PortAddressBinding `json:"addressBindings,omitempty"`
 	// InterfaceIPType decides the address families of static IP allocation, when
-	// DHCP or SLAAC is not activated on the Subnet. It will be ignored when
-	// StaticIPAllocationType is set.
+	// DHCP or SLAAC is not activated on the Subnet. When StaticIPAllocationType
+	// is set, IP families of InterfaceIPType should be a superset of
+	// StaticIPAllocationType.
 	// +kubebuilder:validation:Enum=IPv4;IPv6;IPv4IPv6
 	InterfaceIPType IPAddressType `json:"interfaceIPType,omitempty"`
 	// StaticIPAllocationType explicitly requests static IP allocation of the
 	// specified the address families. In a mixed-mode Subnet (where both DHCP
 	// and static allocation are enabled), use this to define which families
-	// should be allocated from the static IP pools. This field is only valid
-	// when static allocation is enabled on the Subnet.
+	// should be allocated from the static IP pools. If not specified, this field
+	// will be back-filled based on InterfaceIPType and Subnet configuration.
 	// +kubebuilder:validation:Enum=IPv4;IPv6;IPv4IPv6;None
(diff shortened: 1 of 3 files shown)


yanjunz97:
@jianjuns Based on the discussions with vm operator team, I make some changes in the SubnetPort CRD. Could you help take a look? Thanks!

- Add dhcpv6DeactivatedOnSubnet in status: this is the same as we discussed before to show if dhcpv6 is enabled on the Subnet
- Add raDeactivated in status: instead of exposing the RA Mode as a string, VM operator said cloudinit only needs a flag to indicate if RA mode is enabled, so we just add a flag for RA Mode
- vm operator team has some confusion for some description of InterfaceIPType/StaticIPAllocationType. I tried to update to make it less ambiguous.
  - The words "InterfaceIPType will be ignored when StaticIPAllocationType is set" seems confusing as we plan to verify that StaticIPAllocationType shall be a subset of interfaceIPType. 
  - The words "StaticIPAllocationType is only valid when static allocation is enabled on the Subnet" seems contradict with our plan to always back-fill the StaticIPAllocationType based on InterfaceIPType.

pkg/apis/vpc/v1alpha1/subnetport_types.go
```diff
@@ -26,15 +26,15 @@ type SubnetPortSpec struct {
 	// AddressBindings defines static address bindings used for the SubnetPort.
 	AddressBindings []PortAddressBinding `json:"addressBindings,omitempty"`
 	// InterfaceIPType decides the address families of static IP allocation, when
-	// DHCP or SLAAC is not activated on the Subnet. It will be ignored when
-	// StaticIPAllocationType is set.
+	// DHCP or SLAAC is not activated on the Subnet. When StaticIPAllocationType
+	// is set, InterfaceIPType should be a superset of StaticIPAllocationType.
```
jianjuns: "IP families of InterfaceIPType"

pkg/apis/vpc/v1alpha1/subnetport_types.go
```diff
@@ -26,15 +26,15 @@ type SubnetPortSpec struct {
...
 	// +kubebuilder:validation:Enum=IPv4;IPv6;IPv4IPv6
 	InterfaceIPType IPAddressType `json:"interfaceIPType,omitempty"`
 	// StaticIPAllocationType explicitly requests static IP allocation of the
 	// specified the address families. In a mixed-mode Subnet (where both DHCP
 	// and static allocation are enabled), use this to define which families
-	// should be allocated from the static IP pools. This field is only valid
-	// when static allocation is enabled on the Subnet.
+	// should be allocated from the static IP pools. This field will be back-filled
```
jianjuns: "back-filled" is only when the client does not specify a value?

pkg/apis/vpc/v1alpha1/subnetport_types.go
yanjunz97: Updated, thanks

pkg/apis/vpc/v1alpha1/subnetport_types.go
yanjunz97: Yes, updated to explicit mention this. Please help take another look, thanks!

jianjuns (approved):

dantingl (approved):

PR #1457, opened 2026-06-08
Retry on empty bindingmap in scaled case

yanjunz97:
In scaled case, after HAPI succeeds, NSX Subnet ConnectionBindingMap may not exist in the return of list API.
Currently we always read it and save to store. It will hit a nil pointer dereference failure.
This PR
- checks if the Subnet ConnectionBindingMap exists and adds a retry logic to get the NSX Subnet ConnectionBindingMap list
- adds a loop to get all the NSX Subnet ConnectionBindingMap through list API

Testing done:
Tested regression case for bindingmap creation

```
metadata:
  annotations:
      {"apiVersion":"crd.nsx.vmware.com/v1alpha1","kind":"SubnetConnectionBindingMap","metadata":{"annotations":{},"name":"scbm1","namespace":"ns-1"},"spec":{"subnetName":"subnet-2","targetSubnetName":"subnet-1","vlanTrafficTag":201}}
  creationTimestamp: "2026-06-08T04:09:34Z"
  generation: 1
  name: scbm1
  namespace: ns-1
  resourceVersion: "3156967"
  uid: 00b9bdcf-200d-484d-822f-0c97ce2c4b24
spec:
  subnetName: subnet-2
  targetSubnetName: subnet-1
  vlanTrafficTag: 201
status:
  conditions:
  - lastTransitionTime: "2026-06-08T04:09:34Z"
    status: "True"
    type: Ready
```

### DIFF
--- a/pkg/nsx/services/subnetbinding/subnetbinding.go
+++ b/pkg/nsx/services/subnetbinding/subnetbinding.go
@@ -2,15 +2,18 @@ package subnetbinding
 
 import (
 	"context"
+	"fmt"
 	"sync"
 
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/util/retry"
 
 	"github.com/vmware-tanzu/nsx-operator/pkg/apis/vpc/v1alpha1"
 	"github.com/vmware-tanzu/nsx-operator/pkg/logger"
 	servicecommon "github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
 	nsxutil "github.com/vmware-tanzu/nsx-operator/pkg/nsx/util"
+	"github.com/vmware-tanzu/nsx-operator/pkg/util"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/nsx/services/subnetbinding/subnetbinding.go
```diff
@@ -182,30 +185,46 @@ func (s *BindingService) Apply(subnetPath string, bindingMaps []*model.SubnetCon
...
-		if bm.MarkedForDelete != nil && *bm.MarkedForDelete {
-			s.BindingStore.Apply(bm)
-		} else {
-			nsxBindingMap := nsxBindingMaps[*bm.Id]
-			s.BindingStore.Apply(&nsxBindingMap)
+		var missingBindingMaps []*model.SubnetConnectionBindingMap
+
+		for i := range pendingBindingMaps {
+			bm := pendingBindingMaps[i]
+			if bm.MarkedForDelete != nil && *bm.MarkedForDelete {
```
wenyingd: Use this format?

```suggestion
			if bm.MarkedForDelete != nil && *bm.MarkedForDelete {
				s.BindingStore.Apply(bm)
				continue
			}
			nsxBindingMap, ok := nsxBindingMaps[*bm.Id]
			if ok {
				s.BindingStore.Apply(&nsxBindingMap)
				continue
			}
			missingBindingMaps = append(missingBindingMaps, bm)
```

pkg/nsx/services/subnetbinding/subnetbinding.go
```diff
@@ -182,30 +185,46 @@ func (s *BindingService) Apply(subnetPath string, bindingMaps []*model.SubnetCon
...
-		return err
-	}
+	// In scaled case, when HAPI call succeeds, the SubnetConnectionBindingMaps may not be immediately available in NSX.
+	// So we need to retry to list the SubnetConnectionBindingMaps from NSX until they are available.
+	pendingBindingMaps := bindingMaps
+	err = retry.OnError(util.NSXTRealizeRetry, func(err error) bool { return err != nil }, func() error {
+		subnetBindingListResult, listErr := s.NSXClient.SubnetConnectionBindingMapsClient.List(vpcInfo.OrgID, vpcInfo.ProjectID, vpcInfo.VPCID, subnetID, nil, nil, nil, nil, nil, nil)
+		if listErr != nil {
+			log.Error(listErr, "Failed to list SubnetConnectionBindingMaps from NSX under Subnet", "orgID", vpcInfo.OrgID, "projectID", vpcInfo.ProjectID, "vpcID", vpcInfo.VPCID, "subnetID", subnetID, "subnetConnectionBindingMaps", pendingBindingMaps)
+			return nsxutil.TransNSXApiError(listErr)
```
wenyingd: Do we need retry if the error is NSX list API error rather than the NSX resource full-synced delay?

wenyingd (commented):
Overall LGTM, two minor comments

pkg/nsx/services/subnetbinding/subnetbinding.go
yanjunz97: Updated, thanks

pkg/nsx/services/subnetbinding/subnetbinding.go
yanjunz97: I feel both shall work, but let me change it to not retry for list API error to have the same behavior as before.
I also notice another issue. The list API may not return the full list of the bindingmap as the default page size is 1000. So I also add a loop to fetch all the bindingmap. Please help take another loop, thanks!

pkg/nsx/services/subnetbinding/subnetbinding.go
```diff
@@ -182,30 +185,58 @@ func (s *BindingService) Apply(subnetPath string, bindingMaps []*model.SubnetCon
...
+		if err == nil {
+			return false
+		}
+		if _, ok := err.(*nsxutil.NSXApiError); ok {
+			return false
+		}
+		return true
+	}, func() error {
+		for {
+			subnetBindingListResult, listErr := s.NSXClient.SubnetConnectionBindingMapsClient.List(vpcInfo.OrgID, vpcInfo.ProjectID, vpcInfo.VPCID, subnetID, cursor, nil, nil, nil, nil, nil)
```
wenyingd, line 213: Double check NSX API, we may need to enforce `includeMarkForDeleteObjectsParam` as true, as the resources marked for deletion are not included by default, and the current implementation is possible to cause the for-loop never complete, because line 219 is never hit and the pendingBindingMaps is never cleaned up. 

```
	// @param includeMarkForDeleteObjectsParam If true, resources that are marked for deletion will be included in the results. By default, these resources are not included. (optional, default to false)
```
An alternative is we move the line 219 outside of the for-loop, and only check the create/update bindingmaps in this realization check, then we could ignore the mark-for-deletion resources in NSX API call. A pre-process is filter-out the deleting resources in the pending list.

seanpang-vmware (dismissed):

pkg/nsx/services/subnetbinding/subnetbinding.go
yanjunz97: moved the MarkedForDelete logic outside of the retry as discussed.

wenyingd (approved):
LGTM

seanpang-vmware (approved):

PR #1461, opened 2026-06-09
Expose LBCapability condition on system VPCNetworkConfiguration for VNA+IPv6

poojav25:
VNA (VLANBackedVPC) does not support L4 IPv6 Load Balancing when NSX-LB
is in use. NSX Operator now sets an LBCapability status condition on the
system VPCNetworkConfiguration so WCPSVC can observe this limitation
without probing NSX directly.

Setup: TEP-less (VNA) Supervisor with use_native_loadbalancer=True and ip_family=IPv6 or ip_family=DualStack in the nsx-operator config.

Verify condition is set:
```
kubectl get vpcnetworkconfiguration system -o jsonpath='{.status.conditions[?(@.type=="LBCapability")]}'
```

Expected:
```
{
  "lastTransitionTime": "...",
  "reason": "IPv6 LB is not supported as VPCs are hosted on VNA",
  "status": "False",
  "type": "LBCapability"
}
```

### DIFF
--- a/pkg/apis/vpc/v1alpha1/condition_types.go
+++ b/pkg/apis/vpc/v1alpha1/condition_types.go
@@ -14,6 +14,7 @@ const (
 	ServiceClusterReady        ConditionType = "ServiceClusterReady"
 	AutoSnatEnabled            ConditionType = "AutoSnatEnabled"
 	ExternalIPBlocksConfigured ConditionType = "ExternalIPBlocksConfigured"
+	LBCapability               ConditionType = "LBCapability"
 	DeleteFailure              ConditionType = "DeletionFailed"
 	UpdateFailure              ConditionType = "UpdateFailed"
 )
--- a/pkg/controllers/networkinfo/networkinfo_controller.go
+++ b/pkg/controllers/networkinfo/networkinfo_controller.go
@@ -463,9 +463,9 @@ func (r *NetworkInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	// LBCapability=False is informational only — namespace readiness is not blocked.
 	if ncName == commonservice.SystemVPCNetworkConfigurationName {
 		ipFamily := r.Service.NSXConfig.K8sConfig.GetIPAddressType()
-		lbCapable := !(networkStack == v1alpha1.VLANBackedVPC &&
-			lbProvider == vpc.NSXLB &&
-			util.IPAddressTypeIncludesIPv6(ipFamily))
+		lbCapable := networkStack != v1alpha1.VLANBackedVPC ||
+			lbProvider != vpc.NSXLB ||
+			!util.IPAddressTypeIncludesIPv6(ipFamily)
 		// When LBCapability is False, read NSX LBS realization alarms (best-effort) so
 		// manual testing can observe what NSX actually reports. The hardcoded reason string
 		// is always used for machine-readability; the NSX alarm goes into the message field.
--- a/pkg/controllers/networkinfo/networkinfo_controller.go
+++ b/pkg/controllers/networkinfo/networkinfo_controller.go
@@ -329,15 +329,22 @@ func (r *NetworkInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		log.Error(err, "Failed to get LB Provider")
 		return common.ResultRequeue, nil
 	}
-	// For non-pre-created system VPC, set LBCapability before VPC creation.
+	// Set LBCapability before VPC creation for the system VPC.
 	// CreateOrUpdateVPC fails in VNA + NSX-LB + IPv6/DualStack, so without this early
 	// check the condition would never be reported when VPC creation errors out.
-	if ncName == commonservice.SystemVPCNetworkConfigurationName && !vpc.IsPreCreatedVPC(nc) && len(nc.Spec.VPCConnectivityProfile) > 0 {
-		if earlyProfile, profileErr := r.Service.GetVpcConnectivityProfile(nc, nc.Spec.VPCConnectivityProfile); profileErr == nil {
+	if ncName == commonservice.SystemVPCNetworkConfigurationName && len(nc.Spec.VPCConnectivityProfile) > 0 {
+		earlyProfile, profileErr := r.Service.GetVpcConnectivityProfile(nc, nc.Spec.VPCConnectivityProfile)
+		if profileErr != nil {
+			log.Error(profileErr, "Requeue NetworkInfo CR because failed to get VPC connectivity profile for LBCapability check", "req", req)
+			if !retryWithSystemVPC {
+				retryWithSystemVPC = true
+				systemNSCondition = nsMsgVPCConnProfileNotReady.getNSNetworkCondition()
+			}
+		} else {
 			ipFamily := r.Service.NSXConfig.K8sConfig.GetIPAddressType()
 			isVNA, vnaErr := vpc.IsVNAMode(earlyProfile)
 			if vnaErr != nil {
-				log.Info("Requeue NetworkInfo CR because VPC connectivity profile EdgeClusterPaths are not available", "req", req)
+				log.Info("Requeue NetworkInfo CR because VPC connectivity profile ServiceGateway/EdgeClusterPaths not available", "req", req)
 				if !retryWithSystemVPC {
 					retryWithSystemVPC = true
 					systemNSCondition = nsMsgVPCConnProfileNotReady.getNSNetworkCondition()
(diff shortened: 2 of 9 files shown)


pkg/apis/vpc/v1alpha1/condition_types.go
```diff
@@ -14,6 +14,7 @@ const (
 	ServiceClusterReady        ConditionType = "ServiceClusterReady"
 	AutoSnatEnabled            ConditionType = "AutoSnatEnabled"
 	ExternalIPBlocksConfigured ConditionType = "ExternalIPBlocksConfigured"
+	LBCapability               ConditionType = "LBCapability"
```
yanjunz97, line 17: Run `make changecrd`?

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -458,6 +458,27 @@ func (r *NetworkInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 	}
 
+	// Set LBCapability condition on the system VPC config.
+	// Condition is False when VNA (VLANBackedVPC) + NSX-LB + IPv6/DualStack, True otherwise.
```
yanjunz97: VNA is not equivalent to VLANBackedVPC.
You will need to check it based on the vpc connectivity profile.
If all edge_cluster_paths in vpc connectivity profile are in the format of `/infra/sites/{{site-id}}/enforcement-points/{{ep-id}}/virtual-network-appliance-clusters/xxx`, then it means the VPC only has VNA and we need to set this condition to false (for NSX-LB + IPv6/DualStack).

pkg/nsx/services/vpc/vpc.go
```diff
@@ -1263,3 +1263,26 @@ func (s *VPCService) IsDefaultNSXProject(orgID, projectID string) (bool, error)
 	s.defaultProjectCache.Store(cacheKey, isDefault)
 	return isDefault, nil
 }
+
+// GetLBSRealizationAlarmMessages reads the realization state of the given LBS policy path
+// and collects any alarm messages reported by NSX. Returns "" when no alarms are present
+// or when the realization state cannot be fetched (errors are logged, not propagated).
+// This is used for diagnostic logging when LBCapability is False, so callers can observe
+// what NSX actually reports rather than only seeing the hardcoded condition reason.
+func (s *VPCService) GetLBSRealizationAlarmMessages(lbsPath string) string {
```
yanjunz97: I might miss something. I have 2 questions regarding this:
- If lbs can be created under a vpc with ipv6 block only?
- As we call this function after CreateOrUpdateVPC and system namespace is auto-created, I suppose the lbs should already be created without realization otherwise we will get error from CreateOrUpdateVPC?
https://github.com/vmware-tanzu/nsx-operator/blob/1dc78848df722d412c6ab6adbbafc16299031a7a/pkg/controllers/networkinfo/networkinfo_controller.go#L331

pkg/nsx/services/vpc/vpc.go
```diff
@@ -1051,6 +1051,25 @@ func IsEnableAutoSNAT(vpcConnectivityProfile *model.VpcConnectivityProfile) bool
...
+// (/infra/sites/.../virtual-network-appliance-clusters/...). Returns false when the
+// profile or its paths are absent — safe default, avoids false positives on LBCapability.
+func IsVNAMode(profile *model.VpcConnectivityProfile) bool {
+	if profile == nil || profile.ServiceGateway == nil {
+		return false
+	}
+	paths := profile.ServiceGateway.EdgeClusterPaths
+	if len(paths) == 0 {
+		return false
+	}
```
yanjunz97, line 1065: I'm think about if we should distinguish the edge cluster path missing and not VNAMode. As currently if IsVNAMode returns false, we will set LBCapability to true, which seems not accurate. Maybe return an error in this case, not set LBCapability and set retryWithSystemVPC?

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -458,6 +458,18 @@ func (r *NetworkInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	// Set LBCapability condition on the system VPC config.
+	// Condition is False when VNA (all ServiceGateway.EdgeClusterPaths are VNA clusters) +
+	// NSX-LB + IPv6/DualStack, True otherwise.
+	// LBCapability=False is informational only — namespace readiness is not blocked.
+	if ncName == commonservice.SystemVPCNetworkConfigurationName {
+		ipFamily := r.Service.NSXConfig.K8sConfig.GetIPAddressType()
+		lbCapable := !vpc.IsVNAMode(vpcConnectivityProfile) ||
+			lbProvider != vpc.NSXLB ||
+			!util.IPAddressTypeIncludesIPv6(ipFamily)
+		setVPCNetworkConfigurationStatusWithLBCapability(ctx, r.Client, systemVpcNetCfg, lbCapable)
```
yanjunz97: > If lbs can be created under a vpc with ipv6 block only?

Hi Pooja, could you help test and answer this? If lbs cannot be created under a vna + ipv6 only environment, we may need to move this above the CreateOrUpdateVPC. Otherwise we will always get error at CreateOrUpdateVPC and never reach here

yanjunz97: I checked with NSX side, they said vpc lbs creation will be failed in tepless + vna + ipv6 only. So we may need to move this above vpc creation so that the condition will always be reported.

pkg/nsx/services/vpc/vpc.go
poojav25: Done. IsVNAMode now returns (bool, error). ServiceGateway != nil but EdgeClusterPaths empty → returns an error (profile not fully provisioned, should retry). The controller sets retryWithSystemVPC = true and reports nsMsgVPCConnProfileNotReady on the namespace condition in that case, rather than incorrectly treating it as non-VNA.

pkg/controllers/networkinfo/networkinfo_controller.go
poojav25: Fixed by adding an early LBCapability check before CreateOrUpdateVPC for non-pre-created system VPCs — it fetches the profile from nc.Spec.VPCConnectivityProfile (available before VPC creation) and sets the condition immediately. A test case LBCapabilitySetBeforeVPCCreationFailure was added to verify the condition is set even when CreateOrUpdateVPC returns an error.

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -328,6 +329,26 @@ func (r *NetworkInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		log.Error(err, "Failed to get LB Provider")
 		return common.ResultRequeue, nil
 	}
+	// For non-pre-created system VPC, set LBCapability before VPC creation.
+	// CreateOrUpdateVPC fails in VNA + NSX-LB + IPv6/DualStack, so without this early
+	// check the condition would never be reported when VPC creation errors out.
+	if ncName == commonservice.SystemVPCNetworkConfigurationName && !vpc.IsPreCreatedVPC(nc) && len(nc.Spec.VPCConnectivityProfile) > 0 {
```
yanjunz97: The system vpc is always not pre-created. So maybe we can skip the check for `!vpc.IsPreCreatedVPC(nc)` and also line 482-501 is not needed?

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -328,6 +329,26 @@ func (r *NetworkInfoReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		if earlyProfile, profileErr := r.Service.GetVpcConnectivityProfile(nc, nc.Spec.VPCConnectivityProfile); profileErr == nil {
```
yanjunz97: I think we shall requeue if fails to get the profile?

pkg/nsx/services/vpc/vpc.go
```diff
@@ -1051,6 +1051,26 @@ func IsEnableAutoSNAT(vpcConnectivityProfile *model.VpcConnectivityProfile) bool
...
 }
 
+// IsVNAMode reports whether all ServiceGateway.EdgeClusterPaths are VNA cluster paths
+// (/infra/sites/.../virtual-network-appliance-clusters/...).
+// Returns (false, nil) when profile or ServiceGateway is nil — not a VNA deployment.
+// Returns (false, error) when EdgeClusterPaths is empty — ServiceGateway exists but paths
+// are not yet populated; callers should not set LBCapability and should retry.
+func IsVNAMode(profile *model.VpcConnectivityProfile) (bool, error) {
+	if profile == nil || profile.ServiceGateway == nil {
+		return false, nil
```
yanjunz97: I think no profile/ServiceGateway shall also be considered as error like we do for empty EdgeClusterPaths?

poojav25: Done. IsVNAMode now returns (false, error) for both nil profile and nil ServiceGateway — treating all three "can't determine VNA" cases consistently: nil profile, nil ServiceGateway, and empty EdgeClusterPaths all return an error and trigger a retry.

pkg/controllers/networkinfo/networkinfo_controller.go
poojav25: Done. Removed !vpc.IsPreCreatedVPC(nc) — system VPC is always non-pre-created so the guard was redundant. Also removed the late LBCapability block (L482-501); the early check now unconditionally covers the system VPC case.

pkg/controllers/networkinfo/networkinfo_controller.go
poojav25: Done. Changed from if profileErr == nil { ... } (silent skip on error) to an explicit if profileErr != nil branch that sets retryWithSystemVPC = true and systemNSCondition = nsMsgVPCConnProfileNotReady, consistent with how other transient fetch failures are handled.

yanjunz97 (dismissed):
LGTM

pkg/nsx/services/vpc/vpc.go
```diff
@@ -1051,6 +1051,26 @@ func IsEnableAutoSNAT(vpcConnectivityProfile *model.VpcConnectivityProfile) bool
...
 }
 
+// IsVNAMode reports whether all ServiceGateway.EdgeClusterPaths are VNA cluster paths
+// (/infra/sites/.../virtual-network-appliance-clusters/...).
+// Returns (false, nil) when profile or ServiceGateway is nil — not a VNA deployment.
+// Returns (false, error) when EdgeClusterPaths is empty — ServiceGateway exists but paths
+// are not yet populated; callers should not set LBCapability and should retry.
+func IsVNAMode(profile *model.VpcConnectivityProfile) (bool, error) {
+	if profile == nil || profile.ServiceGateway == nil {
+		return false, fmt.Errorf("VPC connectivity profile or ServiceGateway is nil, cannot determine VNA mode")
```
wenqiq, line 1061: The logic here doesn't seem to match the comments in L1056?

pkg/controllers/networkinfo/networkinfo_controller.go
dantingl, line 407: Should we check `earlyProfile` before getting profile again from NSX side?

pkg/nsx/services/vpc/vpc.go
poojav25: fixed it

pkg/controllers/networkinfo/networkinfo_controller.go
poojav25: Fixed — reused earlyProfile instead of fetching again.

yanjunz97 (approved):

dantingl (approved):

PR #1467, opened 2026-06-11
Define ServiceEndpoint and VPCEndpoint CRDs

dantingl:


pkg/apis/vpc/v1alpha1/vpcendpoint_types.go
```diff
@@ -0,0 +1,53 @@
...
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// VPCEndpointSpec defines the desired state of VPCEndpoint
+type VPCEndpointSpec struct {
+	// ServiceEndpointPath is NSX path of the VPC service endpoint being consumed.
```
jianjuns: I feel I commented on this one many times - please do not use "NSX", "path", "Policy", etc. in any API for Namespace/Tenant users, as they do not understand these terms.

Let us call this field ServiceEndpointName, and change the comments too.

dantingl: Thanks Jianjun for your comment. Previously Quan mentioned to me that we need to use service endpoint path in VPC endpoint API, do you mean we call it ServiceEndpointName, the value is NSX path instead of ServiceEndpoint CR name, right?

jianjuns: In CCI API, a shared resource will be referenced by <org-name>:<vpc-name>:<resource-name>, and we still call it "name" not "path". Could you check with @tnqn, what is the plan for ServiceEndpoint reference in CCI API?

dantingl: Thanks Jianjun and Quan, as discussed offline, ServiceEndpointName will use the name from CCI API, with orgID:vpcID:serviceEndpointID format (orgID is NSX project ID)

pkg/apis/vpc/v1alpha1/vpcendpoint_types.go
```diff
@@ -0,0 +1,53 @@
...
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// VPCEndpointSpec defines the desired state of VPCEndpoint
+type VPCEndpointSpec struct {
+	// ServiceEndpointName is VPC service endpoint name being consumed.
```
jianjuns: the VPC service endpoint name

jianjuns (commented):
A nit. Otherwise LGTM

pkg/apis/vpc/v1alpha1/serviceendpoint_types.go
```diff
@@ -0,0 +1,50 @@
...
+type ServiceEndpointSpec struct {
+	// ServiceEndpointIP is the IP address of the VPC service endpoint.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=`^([0-9]{1,3}\.){3}[0-9]{1,3}$`
+	ServiceEndpointIP string `json:"serviceEndpointIP"`
+}
+
+// ServiceEndpointStatus defines the observed state of ServiceEndpoint
+type ServiceEndpointStatus struct {
+	// Conditions describe the current state of the ServiceEndpoint.
```
jianjuns: describes

pkg/apis/vpc/v1alpha1/vpcendpoint_types.go
```diff
@@ -0,0 +1,53 @@
...
+	ServiceEndpointName string `json:"serviceEndpointName"`
+
+	// IPAllocationName defines the IPAddressAllocation CR name that supplies the IP of VPC endpoint.
+	// +kubebuilder:validation:Required
+	IPAllocationName string `json:"ipAllocationName"`
+}
+
+// VPCEndpointStatus defines the observed state of VPCEndpoint
+type VPCEndpointStatus struct {
+	// Conditions describe the current state of the VPCEndpoint.
```
jianjuns: describes

pkg/apis/vpc/v1alpha1/serviceendpoint_types.go
```diff
@@ -0,0 +1,50 @@
+/* Copyright © 2026 VMware, Inc. All Rights Reserved.
+   SPDX-License-Identifier: Apache-2.0 */
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ServiceEndpointSpec defines the desired state of ServiceEndpoint
```
jianjuns: Add "." at the end of a sentence.

jianjuns (commented):
Nits. LGTM

pkg/apis/vpc/v1alpha1/serviceendpoint_types.go
```diff
@@ -0,0 +1,50 @@
...
+type ServiceEndpointSpec struct {
+	// ServiceEndpointIP is the IP address of the VPC service endpoint.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=`^([0-9]{1,3}\.){3}[0-9]{1,3}$`
```
yanjunz97: Shall we support ipv6 for this? I think this format only works for v4 ip

poojav25:
LGTM

pkg/apis/vpc/v1alpha1/serviceendpoint_types.go
jianjuns: In 9.2 Endpoint supports only IPv4.

dantingl: Thanks Jianjun, yes, I added this because NSX has description in API yaml that 'Must be a valid IPv4 address.'
I think Yanjun has a good point, it's better we remove IP family validation in CRD definition, to avoid incompatibility case for old Supervisor and new NSX.

dantingl:
@jianjuns do we need to support 'service_endpoint_ip_type' in Supervisor CRD now? right now it only has WORKLOAD

pkg/apis/vpc/v1alpha1/serviceendpoint_types.go
jianjuns: > I think Yanjun has a good point, it's better we remove IP family validation in CRD definition, to avoid incompatibility case for old Supervisor and new NSX.

Why this is an issue? It just means to support IPv6 endpoints, you need to upgrade Supervisor too. I think it is generally the case for other features too, no?

dantingl: This is not an issue. I'm thinking if we remove this validation from CRD definition, after NSX supports IPv6 for ServiceEndpoint, we should be able to use IPv6 with old Supervisor and new NSX. And as we can rely on NSX validation and error message, maybe we do not need to add extra validation?

jianjuns:
> service_endpoint_ip_type

No, let us not add it.

yanjunz97 (approved):
LGTM

poojav25 (approved):

PR #1468, opened 2026-06-12
Add loadBalancerVPC/ipblock for vpcnetworkconfiguration/IPAddressAllocation CRD

TaoZou1:
Introduce the `loadBalancerVPC` field to the VPCNetworkConfiguration Spec. 
This allows users to specify an alternative NSX Policy path for a dedicated VPC 
to host load balancer resources (such as VIP and LB pools) instead of using the 
default/current VPC associated with the Namespace.

Introduce the `ipBlock` field to the IPAddressAllocation Spec.
This allows users to specify an alternative NSX Policy path for an IP block
to be used for the IP address allocation.

### DIFF
--- a/build/yaml/crd/vpc/crd.nsx.vmware.com_ipaddressallocations.yaml
+++ b/build/yaml/crd/vpc/crd.nsx.vmware.com_ipaddressallocations.yaml
@@ -90,6 +90,13 @@ spec:
                 x-kubernetes-validations:
                 - message: Value is immutable
                   rule: self == oldSelf
+              ipBlockName:
+                description: IPBlockName specifies name of the IPBlock to allocate
+                  IP addresses.
+                type: string
+                x-kubernetes-validations:
+                - message: Value is immutable
+                  rule: self == oldSelf
               ipv6AllocationPrefixLength:
                 description: |-
                   IPv6AllocationPrefixLength specifies the prefix length of IPv6 addresses.
--- a/pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
+++ b/pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
@@ -77,6 +77,10 @@ type IPAddressAllocationSpec struct {
 	// +kubebuilder:default=IPv4
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
 	IPAddressType IPAllocationAddressType `json:"ipAddressType,omitempty"`
+	// IPBlockName specifies name of the IPBlock to allocate IP addresses.
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	IPBlockName string `json:"ipBlockName,omitempty"`
 }
 
 // IPAddressAllocationStatus defines the observed state of IPAddressAllocation.
--- a/pkg/apis/vpc/v1alpha1/vpcnetworkconfiguration_types.go
+++ b/pkg/apis/vpc/v1alpha1/vpcnetworkconfiguration_types.go
@@ -47,6 +47,10 @@ type VPCNetworkConfigurationSpec struct {
 	// +kubebuilder:validation:Minimum:=2
 	// +kubebuilder:validation:Maximum:=127
 	DefaultIPv6PrefixLength int `json:"defaultIPv6PrefixLength,omitempty"`
+
+	// NSX Policy path of the Load Balancer VPC. If set, load balancer resources (such as virtual servers and LB pools) of the Namespace will be created on this VPC's load balancer, instead of the Namespace's primary VPC.
+	// +optional
+	LoadBalancerVPC string `json:"loadBalancerVPC,omitempty"`
 }
 
 // SharedSubnet defines the information for a Subnet shared with vSphere Namespace.
--- a/build/yaml/crd/vpc/crd.nsx.vmware.com_vpcnetworkconfigurations.yaml
+++ b/build/yaml/crd/vpc/crd.nsx.vmware.com_vpcnetworkconfigurations.yaml
@@ -71,6 +71,12 @@ spec:
                 items:
                   type: string
                 type: array
+              loadBalancerVPC:
+                description: NSX Policy path of the Load Balancer VPC. If set, load
+                  balancer resources (such as virtual servers and LB pools) of the
+                  Namespace will be created on this VPC's load balancer, instead of
+                  the Namespace's primary VPC.
+                type: string
               nsxProject:
                 description: NSX Project the Namespace is associated with.
                 type: string
--- a/docs/ref/apis/vpc.md
+++ b/docs/ref/apis/vpc.md
@@ -771,7 +771,6 @@ _Appears in:_
 | `enabled` _boolean_ | Activate or deactivate static IP allocation for VPC Subnet Ports.<br />If the DHCP mode is DHCPDeactivated or not set, its default value is true.<br />If the DHCP mode is DHCPServer, its default value is false.<br />If the DHCP mode is DHCPRelay, its default value is false. |  |  |
 | `poolRanges` _string array_ | PoolRanges specifies the IP address ranges for static IP allocation.<br />Each entry is either a single IP address (e.g. "192.168.1.5") or a<br />dash-separated range (e.g. "192.168.1.10-192.168.1.20"). Both IPv4 and<br />IPv6 entries may appear in a single list.<br />Example value: ["192.168.1.1", "192.168.1.3-192.168.1.100"] |  | Optional: \{\} <br /> |
 
-
 #### StaticIPAllocationType
 
 _Underlying type:_ _string_
@@ -1412,6 +1411,7 @@ _Appears in:_
 | `defaultSubnetSize` _integer_ | Default size of IPv4 Subnets.<br />Defaults to 32. | 32 | Maximum: 65536 <br /> |
 | `dnsZones` _string array_ | DNSZones specifies the list of permitted DNS zones, identified by their NSX paths. |  |  |
 | `defaultIPv6PrefixLength` _integer_ | Default prefix length of IPv6 Subnets.<br />Defaults to 64. | 64 | Maximum: 127 <br />Minimum: 2 <br /> |
+| `loadBalancerVPC` _string_ | NSX Policy path of the Load Balancer VPC. If set, load balancer resources (such as virtual servers and LB pools) of the Namespace will be created on this VPC's load balancer, instead of the Namespace's primary VPC. |  |  |
 
 
 #### VPCNetworkConfigurationStatus
--- a/docs/ref/apis/vpc.md
+++ b/docs/ref/apis/vpc.md
@@ -121,7 +121,7 @@ _Appears in:_
 | --- | --- | --- | --- |
 | `type` _[ConditionType](#conditiontype)_ | Type defines condition type. |  |  |
 | `status` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#conditionstatus-v1-core)_ | Status of the condition, one of True, False, Unknown. |  |  |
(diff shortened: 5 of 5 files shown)


pkg/apis/vpc/v1alpha1/vpcnetworkconfiguration_types.go
```diff
@@ -47,6 +47,10 @@ type VPCNetworkConfigurationSpec struct {
 	// +kubebuilder:validation:Minimum:=2
 	// +kubebuilder:validation:Maximum:=127
 	DefaultIPv6PrefixLength int `json:"defaultIPv6PrefixLength,omitempty"`
+
+	// NSX Policy path of the Load Balancer VPC. If set, this VPC will be used to create load balancer resources (such as VIP and LB pools) instead of the current VPC.
```
jianjuns: How about: "If set, load balancer resources (such as virtual servers and LB pools) of the Namespace will be created on this VPC's load balancer, instead of the Namespace's primary VPC."

TaoZou1: updated

jianjuns (commented):
LGTM

build/yaml/crd/vpc/crd.nsx.vmware.com_ipaddressallocations.yaml
```diff
@@ -90,6 +90,15 @@ spec:
                 x-kubernetes-validations:
                 - message: Value is immutable
                   rule: self == oldSelf
+              ipBlock:
```
dantingl: This should be the IPBlock name from eas API? cc @jianjuns

pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
```diff
@@ -77,6 +77,11 @@ type IPAddressAllocationSpec struct {
 	// +kubebuilder:default=IPv4
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
 	IPAddressType IPAllocationAddressType `json:"ipAddressType,omitempty"`
+	// IPBlock specifies the Policy path for IP Block for the allocation.
```
jianjuns: Could you clarify if the change is needed because an IPBlock in the LB VPC may be specified to allocate the LB VIPs? Are such allocations by end-users - then how users know which IPBlocks to use? Is it for the VDS migration case only?
@lxiaopei

jianjuns: This CRD is Namespace user API - pls do not add NSX terms like "Policy path" to Namespace API comments, which users wont understand.

Let us call the field IPBlockName, and say "the name of the IP Block"

pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
```diff
@@ -77,6 +77,11 @@ type IPAddressAllocationSpec struct {
...
+	// If set, IP addresses of this allocation will be allocated from this IP Block, instead of the default IP Block associated with the Namespace's primary VPC.
```
jianjuns: Remove "this allocation"

this IP Block -> the IP Block

Remove "instead of..."

pkg/apis/vpc/v1alpha1/vpcnetworkconfiguration_types.go
```diff
@@ -47,6 +47,10 @@ type VPCNetworkConfigurationSpec struct {
 	// +kubebuilder:validation:Minimum:=2
 	// +kubebuilder:validation:Maximum:=127
 	DefaultIPv6PrefixLength int `json:"defaultIPv6PrefixLength,omitempty"`
+
+	// NSX Policy path of the Load Balancer VPC. If set, load balancer resources (such as virtual servers and LB pools) of the Namespace will be created on this VPC's load balancer, instead of the Namespace's primary VPC.
+	// +optional
+	LoadBalancerVPC string `json:"loadBalancerVPC,omitempty"`
```
jianjuns, line 53: Thought over - is it better to pass the NSX load balancer path (rather than VPC path) here? @lxiaopei

TaoZou1: ipaddressallocation and virtual servers are created under the VPC path instead of the Load Balancer path in NSX-T. Passing the load balancer path here won't work because virtual servers and load balancers are sibling resources  under the VPC, rather than nested under the load balancer itself.

pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
TaoZou1: It's ipblock id, shold we call it IPBlockID?
if under project , the value should be project-id:ipblock-id
else ipblock-id

pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
TaoZou1: Done

pkg/apis/vpc/v1alpha1/vpcnetworkconfiguration_types.go
jianjuns: You can get the VPC from the LB path anyway. My point is when we support multiple LBs per VPC, it might be possible to assign a single LB to a Namespace?
Anyway, I do not have a strong opinion to change this one. This is not end-user facing API, so we can still change later when needed.

pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
```diff
@@ -77,6 +77,14 @@ type IPAddressAllocationSpec struct {
...
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
 	IPAddressType IPAllocationAddressType `json:"ipAddressType,omitempty"`
+	// IPBlockID specifies the IP Block identifier for the allocation.
+	// It supports two formats:
+	// 1. "ipblock-id" - for IP Blocks under "/infra" (e.g., "ipblock-192.168.0.0-netmask-16").
+	// 2. "project-id:ipblock-id" - for IP Blocks under "/projects" (e.g., "project-quality:ipblock-10.0.0.0").
+	// If set, IP addresses will be allocated from the IP Block
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	IPBlockID string `json:"ipBlockID,omitempty"`
```
jianjuns: We do not use "ID" in CCI and Namespace APIs. Policy ID is a NSX term. CCI and Namespace should always use "name" to refer to a resource. In CCI and SV extension apiserver APIs, resource name is set to Policy ID.

pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
```diff
@@ -77,6 +77,14 @@ type IPAddressAllocationSpec struct {
 	// +kubebuilder:default=IPv4
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
 	IPAddressType IPAllocationAddressType `json:"ipAddressType,omitempty"`
+	// IPBlockID specifies the IP Block identifier for the allocation.
+	// It supports two formats:
+	// 1. "ipblock-id" - for IP Blocks under "/infra" (e.g., "ipblock-192.168.0.0-netmask-16").
```
jianjuns: Let us not explain all these NSX internal implementation that is not exposed to the Namespace users. Just say something like: IPBlockName specifies name of the IPBlock to allocate IP addresses.

jianjuns: BTW, our convention is to add a `:` to infra resource names like: `:ipblock-192.168.0.0-netmask-16`. Please check the CCI API definitions.

@dantingl @lxiaopei @yanjunz97 @timdengyun : let us follow the CCI API conventions in Supervisor CRDs and EAS APIs.

pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
```diff
@@ -77,6 +77,14 @@ type IPAddressAllocationSpec struct {
 	// +kubebuilder:default=IPv4
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
 	IPAddressType IPAllocationAddressType `json:"ipAddressType,omitempty"`
```
jianjuns, line 79: Add my previous question back:

> Could you clarify if the change is needed because an IPBlock in the LB VPC may be specified to allocate the LB VIPs? Are such allocations by end-users - then how users know which IPBlocks to use? Is it for the VDS migration case only?

@lxiaopei

pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
TaoZou1: Done

pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
TaoZou1: Got it

build/yaml/crd/vpc/crd.nsx.vmware.com_ipaddressallocations.yaml
TaoZou1: Done

pkg/apis/vpc/v1alpha1/vpcnetworkconfiguration_types.go
TaoZou1: OK. Change it later if multiple lbs needed

pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
lxiaopei: "IPBlock" is introduced to specify the old IPBlock which used by existing Namespaces during SV migration.
This is optional. End users can also specify it if needed, User can use "VPCIPAddressUsage” API described in nsx-op EAS server to get the IPBlock.

BTW, @TaoZou1  could you use IPBlock ID here instead of policy path?

TaoZou1: Yes，  updated

pkg/apis/vpc/v1alpha1/ipaddressallocation_types.go
jianjuns: @lxiaopei : could you clarify if the IPBlock can be from the LB VPC or not? In that case, will EAS API support such Blocks or not, and how users know which IPBlocks should be used for LB?

lxiaopei: I checked with Sean, Currently EAS "IPBlock" list API will list all IPBlocks from Namespace' default VPC, it will not list IPBlocks from LB VPC and LB VPC is not displayed in NetworkInfo CR.
Could we set IPBlock from LB VPC just for migration and Namespace user can optionally select IPBlock from Namespace default VPC when creating IPAddressAllocation CR? Could you share your opinion on this @jianjuns ?

jianjuns: I was asking you if you have a use case for end-users to select IPBlocks in LB VPCs. If not, we can start from exposing only default/primary VPC's Blocks in EAS APIs (I guess finally we may need to support LB VPC Blocks though).

lxiaopei: Currently I only have use case to set IPBlocks in LB VPC for migration to avoid re-allocate VIP. I do not know other use case for end-users now. I agree in future we need to also expose IPBlocks in LB VPC. Thanks @jianjuns

jianjuns: Ok, let us revisit this one when needed.

lxiaopei (approved):

yanjunz97 (approved):
LGTM

PR #1470, opened 2026-06-15
Fix default SubnetSet IPAddressType calculation

yanjunz97:
The default SubnetSet should not be created when NSX VPC does not have IPBlocks.

Testing done:
Create a pre-created namespace without private and privatetgw cidr
Observed no vm-default or pod-default SubnetSet is created

### DIFF
--- a/pkg/controllers/networkinfo/networkinfo_controller.go
+++ b/pkg/controllers/networkinfo/networkinfo_controller.go
@@ -142,9 +142,11 @@ func (r *NetworkInfoReconciler) computeSubnetSetIPAddressType(hasIPv4CIDR, hasIP
 	} else if hasIPv4CIDR {
 		vpcIPFamily = v1alpha1.IPAddressTypeIPv4
 	}
-	if vpcIPFamily != "" {
-		ipTypes = append(ipTypes, vpcIPFamily)
+	if vpcIPFamily == "" {
+		log.Info("VPC has no IPv4 or IPv6 blocks for default SubnetSet")
+		return IPAddressTypeNone
 	}
+	ipTypes = append(ipTypes, vpcIPFamily)
 
 	// Compute intersection using common utility
 	intersectedType, err := common.IntersectIPAddressTypes(ipTypes)
--- a/pkg/controllers/networkinfo/networkinfo_controller_test.go
+++ b/pkg/controllers/networkinfo/networkinfo_controller_test.go
@@ -2449,7 +2449,6 @@ func TestNetworkInfoReconciler_ComputeSubnetSetIPAddressType(t *testing.T) {
 		hasIPv4CIDR        bool
 		hasIPv6CIDR        bool
 		expected           v1alpha1.IPAddressType
-		expectedErr        bool
 	}{
 		{
 			name:               "Both supervisor and VPC support IPv4 and IPv6",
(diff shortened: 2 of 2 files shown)


poojav25:
LGTM

pkg/controllers/networkinfo/networkinfo_controller_test.go
```diff
@@ -2449,79 +2449,69 @@ func TestNetworkInfoReconciler_ComputeSubnetSetIPAddressType(t *testing.T) {
...
-			expectedErr:        false,
 		},
 		{
 			name:               "Supervisor IPv4 with VPC IPv6 only - no intersection",
 			supervisorIPFamily: v1alpha1.IPAddressTypeIPv4,
 			hasIPv4CIDR:        false,
 			hasIPv6CIDR:        true,
-			expected:           "",
-			expectedErr:        true,
+			expected:           v1alpha1.IPAddressType(""),
```
wenqiq: It can utilize constants `IPAddressTypeNone` here rather than hard-coded empty strings?

pkg/controllers/networkinfo/networkinfo_controller_test.go
```diff
@@ -2449,79 +2449,69 @@ func TestNetworkInfoReconciler_ComputeSubnetSetIPAddressType(t *testing.T) {
...
 		},
 		{
 			name:               "Supervisor IPv6 with VPC IPv4 only - no intersection",
 			supervisorIPFamily: v1alpha1.IPAddressTypeIPv6,
 			hasIPv4CIDR:        true,
 			hasIPv6CIDR:        false,
-			expected:           "",
-			expectedErr:        true,
+			expected:           v1alpha1.IPAddressType(""),
```
wenqiq: ditto

pkg/controllers/networkinfo/networkinfo_controller_test.go
```diff
@@ -2449,79 +2449,69 @@ func TestNetworkInfoReconciler_ComputeSubnetSetIPAddressType(t *testing.T) {
...
-			expectedErr:        false,
 		},
 		{
 			name:               "VPC has no CIDR blocks",
 			supervisorIPFamily: v1alpha1.IPAddressTypeIPv4IPv6,
 			hasIPv4CIDR:        false,
 			hasIPv6CIDR:        false,
-			expected:           v1alpha1.IPAddressTypeIPv4IPv6,
-			expectedErr:        false,
+			expected:           v1alpha1.IPAddressType(""),
```
wenqiq: ditto

pkg/controllers/networkinfo/networkinfo_controller.go
```diff
@@ -142,6 +142,10 @@ func (r *NetworkInfoReconciler) computeSubnetSetIPAddressType(hasIPv4CIDR, hasIP
 	} else if hasIPv4CIDR {
 		vpcIPFamily = v1alpha1.IPAddressTypeIPv4
 	}
+	if vpcIPFamily == "" {
+		log.Info("VPC has no IPv4 or IPv6 blocks for default SubnetSet")
+		return IPAddressTypeNone
+	}
 	if vpcIPFamily != "" {
 		ipTypes = append(ipTypes, vpcIPFamily)
```
wenqiq: I think we can just append directly here. After adding the early return in L147.

pkg/controllers/networkinfo/networkinfo_controller_test.go
yanjunz97: Updated, thanks

pkg/controllers/networkinfo/networkinfo_controller.go
yanjunz97: Updated, thanks!

wenqiq (approved):

poojav25 (approved):

PR #1471, opened 2026-06-16
Fix node TN ID lookup failure due to FQDN case sensitivity

heypnus:
Fix node TN ID lookup failure due to FQDN case sensitivity and optimize sync strategy

When the ESX host FQDN in NSX contains uppercase letters (e.g., Host-A.domain.local),
nsx-operator failed to match it with the Kubernetes node name which is typically lowercase,
resulting in "Failed to get Node ID for Pod" errors and pods stuck in ProviderFailed state.

This commit fixes the issue and optimizes the synchronization strategy by:
1. Making the NodeStore index key case-insensitive by converting FQDNs to lowercase.
2. Implementing a background goroutine that periodically performs a full `List` of NSX
   HostTransportNodes to keep the local cache fresh. The interval is configurable via
   `tn_id_check_interval` in `ncp.ini` (default 5 minutes).
3. Preventing store pollution by ensuring the background sync only updates nodes that
   are already known to Kubernetes (i.e., existing keys in the local NodeStore).
4. Removing the `InitializeResourceStore` logic for nodes on startup. This was a dry run
   that always returned 0 results (since HostTransportNodes lack cluster tags), resulting
   in an unnecessary NSX Search API call.

The refactored workflow:
1. Background Periodic Sync Workflow
```
+-----------------------------------------------------------------------------------------+
| NodeService.startPeriodicSync()                                                         |
|                                                                                         |
|  1. Ticker triggers every `tn_id_check_interval` (default 5 mins)                       |
|          |                                                                              |
|          v                                                                              |
|  2. Trigger full query: NSXClient.HostTransPortNodesClient.List()                       |
|          |                                                                              |
|          v                                                                              |
|    +-------------+                                                                      |
|    | NSX Manager |                                                                      |
|    +-------------+                                                                      |
|          |                                                                              |
|          | 3. Return all HostTransportNodes list                                        |
|          v                                                                              |
|  4. Build `nsxNodeMap` from results (Key: lowercase FQDN, Value: Node)                  |
|          |                                                                              |
|          v                                                                              |
|  5. Get all keys currently in local NodeStore (Nodes known to K8s)                      |
|          |                                                                              |
|          v                                                                              |
|  6. Iterate over local keys:                                                            |
|     For each `key`:                                                                     |
|       If `nsxNodeMap` has `key`:                                                        |
|           NodeStore.Apply(nsxNode)  <-- Update local cache with fresh NSX data          |
|       Else:                                                                             |
|           NodeStore.Delete(key)     <-- Node deleted in NSX, remove from local cache    |
|                                                                                         |
|  7. Sleep until next tick                                                               |
+-----------------------------------------------------------------------------------------+
```
2. Node Add Workflow (Triggered by K8s Node Event)
```
+----------------+
| K8s API Server |
+----------------+
        |
        | 1. Node Event (Add / Update)
        v
+----------------+
| NodeController |
+----------------+
        |
        | 2. Reconcile(nodeName)
        |    Calls SyncNodeStore(nodeName, deleted=false)
        v
+-----------------------------------------------------------------------------------------+
| NodeService.SyncNodeStore                                                               |
|                                                                                         |
|  3. Convert to lowercase: nodeNameLower = strings.ToLower(nodeName)                     |
|                                                                                         |
|  4. Query local cache: NodeStore.GetByIndex(IndexKeyNodeName, nodeNameLower)            |
|          |                                                                              |
|    +-----+-----+                                                                        |
|    |           |                                                                        |
| [Cache Hit] [Cache Miss]                                                                |
|    |           |                                                                        |
|    v           v                                                                        |
| [Return]    5. NSXClient.HostTransPortNodesClient.List()                                |
|                |                                                                        |
|                v                                                                        |
|             6. Iterate list & case-insensitive match (EqualFold)                        |
|                |                                                                        |
|             7. Write matched node to NodeStore                                          |
|                |                                                                        |
|                v                                                                        |
|             8. Return nil (Success)                                                     |
+-----------------------------------------------------------------------------------------+
```

Testing done:
1. update the node fqdn to upper case, then check the NSX Operator can reconcile the pod in the node
2. remove one node, the NSX Operator can delete the node from store.
3. add the node back, the NSX Operator can add the node to the store.

### DIFF
--- a/pkg/nsx/services/node/node.go
+++ b/pkg/nsx/services/node/node.go
@@ -2,7 +2,8 @@ package node
 
 import (
 	"fmt"
-	"sync"
+	"strings"
+	"time"
 
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 	"k8s.io/client-go/tools/cache"
@@ -21,15 +22,10 @@ var (
 type NodeService struct {
 	servicecommon.Service
 	NodeStore *NodeStore
+	stopChan  chan struct{}
 }
 
 func InitializeNode(service servicecommon.Service) (*NodeService, error) {
-	wg := sync.WaitGroup{}
-	wgDone := make(chan bool)
-	fatalErrors := make(chan error)
-
-	wg.Add(1)
-
 	nodeService := &NodeService{
 		Service: service,
 		NodeStore: &NodeStore{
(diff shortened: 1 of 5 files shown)


pkg/nsx/services/node/node.go
```diff
@@ -59,16 +61,43 @@ func InitializeNode(service servicecommon.Service) (*NodeService, error) {
...
+func (service *NodeService) syncAllNodes() {
+	nodeResults, err := service.NSXClient.HostTransPortNodesClient.List("default", "default", nil, nil, nil, nil, nil, nil, nil, nil, nil, nil)
+	err = nsxutil.TransNSXApiError(err)
+	if err != nil {
+		log.Error(err, "failed to list HostTransportNodes during periodic sync")
+		return
+	}
+	for _, node := range nodeResults.Results {
+		node := node
+		if err := service.NodeStore.Apply(&node); err != nil {
```
yanjunz97: Shall we only update the node already in the store, otherwise it will add the tn node that not related to the cluster in the store?

yanjunz97 (commented):
Overall LGTM, just one nit.

pkg/nsx/services/node/node.go
```diff
@@ -59,16 +61,43 @@ func InitializeNode(service servicecommon.Service) (*NodeService, error) {
...
-	return service.NodeStore.GetByIndex(servicecommon.IndexKeyNodeName, nodeName)
+	return service.NodeStore.GetByIndex(servicecommon.IndexKeyNodeName, strings.ToLower(nodeName))
 }
 
 func (service *NodeService) SyncNodeStore(nodeName string, deleted bool) error {
-	nodes := service.NodeStore.GetByIndex(servicecommon.IndexKeyNodeName, nodeName)
+	nodes := service.NodeStore.GetByIndex(servicecommon.IndexKeyNodeName, strings.ToLower(nodeName))
 	if len(nodes) > 1 {
 		return fmt.Errorf("multiple nodes found for node name %s", nodeName)
 	}
```
yanjunz97, line 127: If the node id may change, is that possible that we get more than 1 nodes here in some corner case, i.e. both node id and nodeName are changed?
Maybe when syncAllNodes, we should also delete the node in store if it does not present on NSX?

pkg/nsx/services/node/node.go
yanjunz97: I see there's an issue related to tn id may change when rejoining. Please feel free to resolve this comment if that's a reason to always add all tn node

dantingl (dismissed):
LGTM

pkg/nsx/services/node/node.go
heypnus: Thanks for pointing it out, added the cleanup logic.

pkg/nsx/services/node/node.go
heypnus: Good point, I added a node name check for the nodeResults.Results.

yanjunz97 (approved):
LGTM

seanpang-vmware (approved):

hickeng:
You probably don't want to use strings.EqualFold in a comparison if you're using strings.ToLower to normalize map keys as they process things differently and handle runes differently.

I'd recommend either using ToLower consistently, or using `cases.Fold()` to fold strings for map key use, eg.

```
package main

import (
	"fmt"
	"golang.org/x/text/cases"
)

func main() {
	// Create a stateless, concurrency-safe folder
	folder := cases.Fold()

	key1 := folder.String("GoLang")
	key2 := folder.String("golang")
	
	fmt.Println(key1 == key2) // true
}
```

I don't know how you handle unicode in FQDNs - the correct approach probably depends on whether it's blocked, uses Punycode encoding, or something else.

PR #1479, opened 2026-06-22
Truncate NSX DFW rule ID to 255 characters

heypnus:
When a SecurityPolicy contains many ports in a rule, the generated NSX Rule ID can exceed the maximum limit of 255 characters, leading to NSX API errors during policy creation.

This change uses `util.NormalizeId` in `buildExpandedRuleID` to properly truncate and hash the rule ID, ensuring it remains within the 255-character limit while avoiding name collisions.


dantingl (approved):

timdengyun (approved):
LGTM

PR #1480, opened 2026-06-22
Avoid update status if no change

yanjunz97:
Observed NSX Operator may update the CR status even if it is not changed.
This will result in a log of API calls in scaled testbed.
This PR reduce the number of update times for CR status.

Testing done:

Observe status update will be skipped when nothing change on status

```
2026-06-26 05:50:18.000 TRACE subnetset_controller.go:420 Conditions already match Existing Condition={"lastTransitionTime":"2026-06-18T07:24:50Z","message":"SubnetSet CR has been successfully created/updated","reason":"SubnetSetReady","status":"True","type":"Ready"} New Condition={"lastTransitionTime":"2026-06-26T05:50:18Z","message":"SubnetSet CR has been successfully created/updated","reason":"SubnetSetReady","status":"True","type":"Ready"}
```

### DIFF
--- a/pkg/controllers/ipaddressallocation/ipaddressallocation_controller.go
+++ b/pkg/controllers/ipaddressallocation/ipaddressallocation_controller.go
@@ -47,7 +47,7 @@ type IPAddressAllocationReconciler struct {
 
 func setReadyStatusFalse(client client.Client, ctx context.Context, obj client.Object, transitionTime metav1.Time, err error, _ ...interface{}) {
 	ipaddressallocation := obj.(*v1alpha1.IPAddressAllocation)
-	conditions := []v1alpha1.Condition{
+	newConditions := []v1alpha1.Condition{
 		{
 			Type:   v1alpha1.Ready,
 			Status: v1.ConditionFalse,
@@ -59,16 +59,12 @@ func setReadyStatusFalse(client client.Client, ctx context.Context, obj client.O
 			LastTransitionTime: transitionTime,
 		},
 	}
-	ipaddressallocation.Status.Conditions = conditions
-	e := client.Status().Update(ctx, ipaddressallocation)
-	if e != nil {
-		log.Error(e, "Unable to update IPAddressAllocation status", "IPAddressAllocation", ipaddressallocation)
-	}
+	updateIPAddressAllocationStatusConditions(client, ctx, ipaddressallocation, newConditions)
 }
 
 func setReadyStatusTrue(client client.Client, ctx context.Context, obj client.Object, transitionTime metav1.Time, _ ...interface{}) {
 	ipaddressallocation := obj.(*v1alpha1.IPAddressAllocation)
-	conditions := []v1alpha1.Condition{
+	newConditions := []v1alpha1.Condition{
 		{
 			Type:               v1alpha1.Ready,
 			Status:             v1.ConditionTrue,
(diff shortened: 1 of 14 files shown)


pkg/controllers/ipaddressallocation/ipaddressallocation_controller.go
```diff
@@ -77,11 +73,50 @@ func setReadyStatusTrue(client client.Client, ctx context.Context, obj client.Ob
...
+
+	if matchedCondition != nil && matchedCondition.Status == newCondition.Status && matchedCondition.Reason == newCondition.Reason && matchedCondition.Message == newCondition.Message {
+		log.Trace("Conditions already match", "New Condition", newCondition, "Existing Condition", matchedCondition)
+		return false
+	}
+
+	if matchedCondition != nil {
+		matchedCondition.Reason = newCondition.Reason
+		matchedCondition.Message = newCondition.Message
+		matchedCondition.Status = newCondition.Status
```
heypnus, line 106: Should we also update the `LastTransitionTime`?

pkg/controllers/networkinfo/networkinfo_utils.go
```diff
@@ -192,7 +188,6 @@ func mergeStatusCondition(conditions *[]v1alpha1.Condition, newCondition *v1alph
 		existingCondition.Reason = newCondition.Reason
 		existingCondition.Message = newCondition.Message
 		existingCondition.Status = newCondition.Status
-		existingCondition.LastTransitionTime = metav1.Now()
```
heypnus, line 195: ditto. Not sure why the `LastTransitionTime` was removed.

pkg/controllers/ipaddressallocation/ipaddressallocation_controller.go
```diff
@@ -77,11 +73,50 @@ func setReadyStatusTrue(client client.Client, ctx context.Context, obj client.Ob
...
 		},
 	}
-	ipaddressallocation.Status.Conditions = conditions
-	e := client.Status().Update(ctx, ipaddressallocation)
-	if e != nil {
-		log.Error(e, "Unable to update IPAddressAllocation status", "IPAddressAllocation", ipaddressallocation)
+	updateIPAddressAllocationStatusConditions(client, ctx, ipaddressallocation, newConditions)
+}
+
+func updateIPAddressAllocationStatusConditions(client client.Client, ctx context.Context, ipaddressallocation *v1alpha1.IPAddressAllocation, newConditions []v1alpha1.Condition) {
```
heypnus, line 79: Could you please add more UT for the changed logic?

pkg/controllers/ipaddressallocation/ipaddressallocation_controller.go
yanjunz97: It's to follow the other controller. But I think previously the condition is always considered as not match so that we skip the LastTransitionTime update to avoid infinite update. Now we correctly get the condition matched result. It should be fine for use to update the LastTransitionTime when status change.

pkg/controllers/networkinfo/networkinfo_utils.go
yanjunz97: Added as above.

pkg/controllers/ipaddressallocation/ipaddressallocation_controller.go
yanjunz97: Added, thanks!

heypnus (dismissed):

pkg/controllers/subnetset/subnetset_controller.go
```diff
@@ -413,7 +416,7 @@ func updateSubnetSetStatusConditions(client client.Client, ctx context.Context,
 func mergeSubnetSetStatusCondition(subnetSet *v1alpha1.SubnetSet, newCondition *v1alpha1.Condition) bool {
 	matchedCondition := getExistingConditionOfType(newCondition.Type, subnetSet.Status.Conditions)
 
-	if reflect.DeepEqual(matchedCondition, newCondition) {
+	if matchedCondition != nil && matchedCondition.Status == newCondition.Status && matchedCondition.Reason == newCondition.Reason && matchedCondition.Message == newCondition.Message {
```
TaoZou1: could you have a common help function like
 func IsConditionSemanticEqual(existing, newCond *v1alpha1.Condition) bool {
} ?

yanjunz97: Updated, thanks!

TaoZou1 (approved):

heypnus (approved):

PR #1481, opened 2026-06-22
[EAS] Generate CRD definition files for EAS API

seanpang-vmware:
Although eas server is using generic api which do not require CRD definitions, but the CRD file could clearly shows the data structure and other teams may use the CRDs to define their work


timdengyun (approved):
LGTM

dantingl (approved):

PR #1487, opened 2026-06-26
Restart if service lb controller failed to get k8s version

yanjunz97:
If k8s version is failed to get when creating the Service lb controller,
Service lb controller may not be start and result in lb service traffic issue.

We shall restart the nsx operator in this case.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -253,10 +253,12 @@ func startServiceController(mgr manager.Manager, nsxClient *nsx.Client) {
 			subnetport.NewSubnetPortReconciler(mgr, subnetPortService, subnetService, vpcService, ipAddressAllocationService),
 			pod.NewPodReconciler(mgr, subnetPortService, subnetService, vpcService, nodeService),
 			networkpolicycontroller.NewNetworkPolicyReconciler(mgr, commonService, vpcService),
-			service.NewServiceLbReconciler(mgr, commonService, dnsRecordService),
 			subnetbindingcontroller.NewReconciler(mgr, subnetService, subnetBindingService),
 			subnetipreservationcontroller.NewReconciler(mgr, subnetIPReservationService, subnetService),
 		)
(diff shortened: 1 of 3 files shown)


timdengyun (approved):
LGTM

cmd/main.go
```diff
@@ -253,10 +253,12 @@ func startServiceController(mgr manager.Manager, nsxClient *nsx.Client) {
 			subnetport.NewSubnetPortReconciler(mgr, subnetPortService, subnetService, vpcService, ipAddressAllocationService),
 			pod.NewPodReconciler(mgr, subnetPortService, subnetService, vpcService, nodeService),
 			networkpolicycontroller.NewNetworkPolicyReconciler(mgr, commonService, vpcService),
-			service.NewServiceLbReconciler(mgr, commonService, dnsRecordService),
 			subnetbindingcontroller.NewReconciler(mgr, subnetService, subnetBindingService),
 			subnetipreservationcontroller.NewReconciler(mgr, subnetIPReservationService, subnetService),
 		)
+		if lbReconciler := service.NewServiceLbReconciler(mgr, commonService, dnsRecordService); lbReconciler != nil {
+			reconcilerList = append(reconcilerList, lbReconciler)
+		}
```
TaoZou1, line 261: you skip the service_lb_controller if the isServiceLbStatusIpModeSupported is false, but it sill need to support the dns.
is it the right logic?
should we launch the service_lb_controller with IpModeSupported = false?

yanjunz97: The code change here is just a nil pointer protection. Without the change, service_lb_controller also does not run when IpModeSupported is false. Let me sync with Wenying later to see if we need to add a new mode for service_lb_controller to support DNS feature in 9.2 with lower k8s version. Thanks!

TaoZou1 (approved):

cmd/main.go
yanjunz97: Synced with Wenying, dns feature is added for 9.2. As wcp 9.2 does not compatible with k8s 1.29 and before, so it should be fine to not run dns logic in this scenario.

PR #1499, opened 2026-07-06
[Migration] Implement mixed-mode discovery using explicit Namespace annotations per upstream design

heypnus:
According to the new choreographed Namespace bootstrapping design from the upstream components, both NCP and nsx-operator must now discover active T1 and VPC namespaces using explicit Namespace annotations. To ensure decoupling and avoid obsolete dependencies, this commit completely removes the dependency on the NetworkSettings CRD lookup, replacing it entirely with scanning core Namespace annotations, and removes the legacy NetworkSettingsWatcher and NetworkSettingsStore entirely.

Specifically, this commit:
1. Replaces the legacy NetworkSettings CRD lookup with scanning core Namespace annotations.
2. Cleans up the legacy NetworkSettings CRD watcher, store, and test mock references.
3. Implements a "self-blocking" window during Namespace creation where standard Namespaces that are "bare" (lacking explicit labels or annotations) do not trigger physical Tier-1 router or subnet provisioning.
4. Detects active Tier-1 networking scope when any of the three explicit upstream activation annotations is present on a Namespace:
   - Case 1: 'nsx.vmware.com/t1_default_config: "true"' (Default T1)
   - Case 2: 'vmware-system-shared-t1: "true"' (Shared T1)
   - Case 3: 'ncp/nsx_network_config_crd: <NNC-Name>' (Override T1)
5. Detects active VPC networking scope when the 'nsx.vmware.com/vpc_network_config' annotation is present.

Testing Done:
1. T1 pipeline with id 5895
2. VPC pipeline with id 21028

### DIFF
--- a/pkg/config/mixed_mode.go
+++ b/pkg/config/mixed_mode.go
@@ -303,8 +303,6 @@ func initMixedModeWithClients(ctx context.Context, clientset kubernetes.Interfac
 	// the mutex to avoid holding the lock during potentially many retries.
 	supported, _ := checkPerNamespaceProvidersSupported(ctx, dynClient)
 
-	storedEnableVPCNetwork = enableVPCNetwork
-
 	var t1, vpc bool
 	if supported {
 		log.Info("Per-namespace network providers are supported, scanning namespaces for mixed-mode")
--- a/pkg/config/mixed_mode.go
+++ b/pkg/config/mixed_mode.go
@@ -5,13 +5,16 @@ package config
 
 import (
 	"context"
+	"strings"
 	"sync"
 	"time"
 
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -20,8 +23,32 @@ import (
 )
 
 const (
+	VPCNetworkConfigAnnotation     = "nsx.vmware.com/vpc_network_config"
+	T1DefaultConfigAnnotation      = "nsx.vmware.com/t1_default_config"
+	T1VmwareSystemSharedAnnotation = "vmware-system-shared-t1"
+	T1NsxNetworkConfigAnnotation   = "ncp/nsx_network_config_crd"
+
 	capabilitiesName = "supervisor-capabilities"
 )
+
+func hasT1ActivationAnnotation(annotations map[string]string) bool {
+	if val, ok := annotations[T1DefaultConfigAnnotation]; ok {
+		if strings.ToLower(strings.TrimSpace(val)) == "true" {
+			return true
+		}
+	}
+	if val, ok := annotations[T1VmwareSystemSharedAnnotation]; ok {
+		if strings.ToLower(strings.TrimSpace(val)) == "true" {
+			return true
+		}
+	}
+	if val, ok := annotations[T1NsxNetworkConfigAnnotation]; ok {
+		if strings.TrimSpace(val) != "" {
+			return true
+		}
+	}
+	return false
+}
 
 var (
 	capabilitiesGVR = schema.GroupVersionResource{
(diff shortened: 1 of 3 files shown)


pkg/config/mixed_mode.go
```diff
@@ -220,84 +256,89 @@ func currentNamespaceRefreshReader() client.Reader {
...
+	log.Info("Mixed-mode state initialized", "hasT1Namespaces", t1, "hasVPCNamespaces", vpc)
 }
 
-// RefreshMixedModeState re-scans namespaces using the dynamic client stored during
+// RefreshMixedModeState re-scans namespaces using the clientset stored during
 // InitMixedMode and updates the global state. Returns true if the state
 // changed; the caller should then restart the operator so that VPC services
 // and controllers are initialized for the new mode.
 func RefreshMixedModeState(ctx context.Context) bool {
 	stateMu.Lock()
```
TaoZou1: if we use the handler pattern to listen the namespace event, we don't need to RefreshMixedModeState? just a suggestion

pkg/config/mixed_mode.go
```diff
@@ -116,85 +137,100 @@ func extractCapability(obj *unstructured.Unstructured) bool {
...
+				annotations := item.GetAnnotations()
+				if _, ok := annotations[VPCNetworkConfigAnnotation]; ok {
+					hasVPC = true
+				}
+				if hasT1ActivationAnnotation(annotations) {
+					hasT1 = true
+				}
+			}
+		} else {
+			log.V(1).Info("Failed to list Namespaces during mixed-mode scan", "error", err)
```
TaoZou1: log.Debug?

pkg/config/mixed_mode.go
dantingl: can we hold the lock before line 365?

pkg/config/mixed_mode.go
```diff
@@ -321,39 +362,40 @@ func RefreshMixedModeState(ctx context.Context) bool {
...
+// HasVPCNamespaces returns true when enable_vpc_network is true or at least one Namespace has the VPC config annotation.
 func HasVPCNamespaces() bool {
 	stateMu.RLock()
 	defer stateMu.RUnlock()
 	return hasVPCNamespaces
 }
 
-// HasVDSNamespaces returns true when at least one NetworkSettings CR uses VDS.
+// HasVDSNamespaces returns true when at least one Namespace uses VDS.
 func HasVDSNamespaces() bool {
```
dantingl: Should we remove `HasVDSNamespaces`?

heypnus: good catch, removed

pkg/config/mixed_mode.go
heypnus: done, thanks for the clever suggestion

pkg/config/mixed_mode.go
```diff
@@ -220,84 +260,86 @@ func currentNamespaceRefreshReader() client.Reader {
...
 }
 
-func initMixedModeWithClients(ctx context.Context, dynClient dynamic.Interface, enableVPCNetwork bool) {
+func initMixedModeWithClients(ctx context.Context, clientset kubernetes.Interface, dynClient dynamic.Interface, enableVPCNetwork bool) {
 	// checkPerNamespaceProvidersSupported retries on transient errors; runs outside
 	// the mutex to avoid holding the lock during potentially many retries.
 	supported, _ := checkPerNamespaceProvidersSupported(ctx, dynClient)
 
-	var t1, vpc, vds bool
+	storedEnableVPCNetwork = enableVPCNetwork
```
dantingl: duplicate with line 326?

heypnus: good catch! removed

dantingl (dismissed):

TaoZou1 (dismissed):

pkg/config/mixed_mode.go
heypnus: That is a great suggestion, and using an event handler is definitely a more elegant, resource-efficient approach for the long term. However, the current 30-second ticker is extremely simple, robust, and already provides a natural debounce mechanism to prevent rapid, repetitive restarts during bulk namespace updates. Since mixed-mode state changes are low-frequency Day-2 operations, the performance overhead of polling the local cache is negligible. To ensure absolute stability and avoid introducing new concurrency complexities right before check-in, we prefer to keep the existing implementation for now. We can definitely log this as a valuable tech debt item for future optimization.

pkg/config/mixed_mode.go
heypnus: Done, thanks.

pkg/config/mixed_mode.go
heypnus: I added a gate for the function `RefreshMixedModeState` -- if the supervisor capability is not activated, the loop won't start, to ensure that there won't be any regression issues caused by it.

TaoZou1 (approved):

dantingl (approved):

PR #1506, opened 2026-07-16
Update golang package version to fix CVEs for eas

seanpang-vmware:
Update golang package version to fix CVEs for eas


dantingl (approved):

yanjunz97 (approved):
LGTM

PR #1509, opened 2026-07-21
Support Subnetport multi ip addressbindings 

poojav25:
Support for multiple IP addresses sharing one MAC address on a single SubnetPort (needed for Blueprint VM scenarios where one VM interface needs several IPs).

1. SubnetPort status now correctly shows all the IPs NSX assigns to a port, no matter how many there are — before, the code only expected 1 IP (or 2 for dual-stack) and silently mishandled anything beyond that.
2. Double-checked that the existing logic deciding how NSX allocates IPs/MACs doesn't need any changes — it already works correctly regardless of how many bindings a port has.
3. Updated docs, regenerated the CRD, and added a sample YAML showing multiple IPs on one port.

### Setup — Subnet

```bash
kubectl apply -f - <<EOF
kind: Subnet
metadata:
  name: subnet-multi-ip-test
  namespace: subnetport-multi-ip-test
spec:
  accessMode: Private
  ipv4SubnetSize: 64
  subnetDHCPConfig:
    mode: DHCPDeactivated
  advancedConfig:
    staticIPAllocation:
      enabled: true
EOF
kubectl get subnet subnet-multi-ip-test -n subnetport-multi-ip-test -o yaml
```

```yaml
status:
  conditions:
  - message: 'NSX Subnet with DHCPv4: DHCPDeactivated has been successfully created/updated'
    reason: SubnetReady
    status: "True"
    type: Ready
  gatewayAddresses:
  - 172.26.0.1/26
  networkAddresses:
  - 172.26.0.0/26
```

---

**T1 — Multiple explicit IPs, one shared MAC: NSX realizes all 3, status reflects all 3**

```bash
kubectl apply -f - <<EOF
kind: SubnetPort
metadata:
  name: sp-multi-ip-test1
  namespace: subnetport-multi-ip-test
spec:
  subnet: subnet-multi-ip-test
  addressBindings:
  - ipAddress: 172.26.0.10
    macAddress: "00:11:22:33:44:55"
  - ipAddress: 172.26.0.11
    macAddress: "00:11:22:33:44:55"
  - ipAddress: 172.26.0.12
    macAddress: "00:11:22:33:44:55"
EOF
kubectl get subnetport sp-multi-ip-test1 -n subnetport-multi-ip-test -o yaml
```

```yaml
status:
  conditions:
  - message: NSX subnet port has been successfully created/updated
    reason: SubnetPortReady
    status: "True"
    type: Ready
  networkInterfaceConfig:
    ipAddresses:
    - gateway: 172.26.0.1
      ipAddress: 172.26.0.10/26
    - gateway: 172.26.0.1
      ipAddress: 172.26.0.11/26
    - gateway: 172.26.0.1
      ipAddress: 172.26.0.12/26
    macAddress: "00:11:22:33:44:55"
```

```bash
curl -sk -u 'admin:<password>' \\
  "https://<nsx-manager>/policy/api/v1/orgs/default/projects/project-quality/vpcs/subnetport-multi-ip-test_jm54x/subnets/subnet-multi-ip-test_jm54x/ports" \\
  | python3 -m json.tool | grep -E '"display_name"|"allocate_addresses"|"ip_address"|"mac_address"'
```

```json
"display_name": "sp-multi-ip-test1",
"allocate_addresses": "IP_POOL",
"ip_address": "172.26.0.10",
"mac_address": "00:11:22:33:44:55",
"ip_address": "172.26.0.11",
"mac_address": "00:11:22:33:44:55",
"ip_address": "172.26.0.12",
"mac_address": "00:11:22:33:44:55",
```

---

**T2 — Multiple pool-allocated IPs (MAC-only bindings), one shared MAC: NSX allocates distinct IPs**

```bash
kubectl apply -f - <<EOF
kind: SubnetPort
metadata:
  name: sp-multi-ip-test3
  namespace: subnetport-multi-ip-test
spec:
  subnet: subnet-multi-ip-test
  addressBindings:
  - macAddress: "00:11:22:33:44:77"
  - macAddress: "00:11:22:33:44:77"
EOF
kubectl get subnetport sp-multi-ip-test3 -n subnetport-multi-ip-test -o yaml
```

```yaml
status:
  conditions:
  - message: NSX subnet port has been successfully created/updated
    reason: SubnetPortReady
    status: "True"
    type: Ready
  networkInterfaceConfig:
    ipAddresses:
    - gateway: 172.26.0.1
      ipAddress: 172.26.0.34/26
    - gateway: 172.26.0.1
      ipAddress: 172.26.0.35/26
    macAddress: "00:11:22:33:44:77"
```

```bash
curl -sk -u 'admin:<password>' \\
  "https://<nsx-manager>/policy/api/v1/orgs/default/projects/project-quality/vpcs/subnetport-multi-ip-test_jm54x/subnets/subnet-multi-ip-test_jm54x/ports/sp-multi-ip-test3_jm54x/state"
```

```json
"realized_bindings" : [
  { "binding": { "ip_address": "172.26.0.34", "mac_address": "00:11:22:33:44:77" } },
  { "binding": { "ip_address": "172.26.0.35", "mac_address": "00:11:22:33:44:77" } }
]
```

---

**T3 — Dual-stack regression check: plain dual-stack SubnetPort still works, no `addressBindings`**

```bash
kubectl apply -f - <<EOF
kind: SubnetPort
metadata:
  name: sp-dualstack-plain
  namespace: subnetport-multi-ip-test
spec:
  subnet: subnet-dualstack-test
  interfaceIPType: IPv4IPv6
EOF
kubectl get subnetport sp-dualstack-plain -n subnetport-multi-ip-test -o yaml
```

```yaml
status:
  conditions:
  - message: NSX subnet port has been successfully created/updated
    reason: SubnetPortReady
    status: "True"
    type: Ready
  networkInterfaceConfig:
    ipAddresses:
    - gateway: 172.26.0.65
      ipAddress: 172.26.0.66/26
    - gateway: fd12:1234::1
      ipAddress: fd12:1234::ffff:ffff:ffff:fc30/64
    macAddress: "04:50:56:00:80:00"
```

Confirmed unchanged dual-stack behavior — one IPv4 + one IPv6, correct per-family gateway, one shared MAC.

---

### Known limitations discovered during testing

**1. Mixing explicit-IP and pool-allocated bindings in one port fails — NSX platform defect, not fixable in nsx-operator**

```bash
kubectl apply -f - <<EOF
kind: SubnetPort
metadata:
  name: sp-multi-ip-test2
  namespace: subnetport-multi-ip-test
spec:
  subnet: subnet-multi-ip-test
  addressBindings:
  - macAddress: "00:11:22:33:44:66"
  - ipAddress: 172.26.0.20
    macAddress: "00:11:22:33:44:66"
  - macAddress: "00:11:22:33:44:66"
EOF
kubectl get subnetport sp-multi-ip-test2 -n subnetport-multi-ip-test -o yaml
```

```yaml
status:
  conditions:
  - message: 'error occurred while processing the SubnetPort CR. Error: .../ports/sp-multi-ip-test2_jm54x
      realized with errors: [IpAddressPool path=[.../ip-pools/static-ipv4-default]
      is exhausted, no free IPs available for allocation.]'
    reason: SubnetPortNotReady
    status: "False"
    type: Ready
```

Reproduced independently of nsx-operator via a direct `curl` PATCH to NSX with the identical binding shape — confirms NSX itself, not the operator, is the cause:

```bash
curl -sk -u 'admin:<password>' \\
  "https://<nsx-manager>/policy/api/v1/infra/realized-state/realized-entities?intent_path=.../ports/repro-mixed"
```

```json
"alarms": [{
  "message": "TX ABORT | ... | Conflict Stream = nsx$AllocationMarker | Cause = CONFLICT",
  "id": "PROVIDER_INVOCATION_FAILURE"
}]
```

The pool was not actually exhausted (`available_ips: 30` of 61 at the time). This is a false-exhaustion error from an internal NSX allocation-transaction conflict when a port mixes explicit-IP and auto-allocated bindings. Each failed attempt also leaks the pool's `allocated_ip_allocations` usage counter without releasing it (confirmed: `61 → 8` available across ~5 failed attempts, while real allocations stayed at 7).

**Workaround:** keep all `addressBindings` entries in a given SubnetPort uniformly explicit-IP or uniformly pool-allocated — don't mix within one port. Filed as an NSX platform issue separately.

**2. Dual-stack + multi-IP pool-allocated bindings always resolve to IPv4 only — NSX platform behavior, not a bug**

```bash
kubectl apply -f - <<EOF
kind: SubnetPort
metadata:
  name: sp-dualstack-multiip2
  namespace: subnetport-multi-ip-test
spec:
  subnet: subnet-dualstack-test
  interfaceIPType: IPv4IPv6
  addressBindings:
  - macAddress: "00:11:22:33:44:13"
  - macAddress: "00:11:22:33:44:13"
  - macAddress: "00:11:22:33:44:13"
EOF
kubectl get subnetport sp-dualstack-multiip2 -n subnetport-multi-ip-test -o yaml
```

```yaml
status:
  conditions:
  - message: NSX subnet port has been successfully created/updated
    reason: SubnetPortReady
    status: "True"
    type: Ready
  networkInterfaceConfig:
    ipAddresses:
    - gateway: 172.26.0.65
      ipAddress: 172.26.0.67/26
    - gateway: 172.26.0.65
      ipAddress: 172.26.0.68/26
    - gateway: 172.26.0.65
      ipAddress: 172.26.0.69/26
    macAddress: "00:11:22:33:44:13"
```

Even with `interfaceIPType`/`staticIPAllocationType: IPv4IPv6`, all 3 bindings were allocated as IPv4 — no IPv6 address at all. Reproduced identically via a direct `curl` PATCH to NSX, confirming this is genuine NSX platform behavior, not caused by how nsx-operator builds the request:

```bash
curl -sk -u 'admin:<password>' \\
  "https://<nsx-manager>/policy/api/v1/orgs/default/projects/project-quality/vpcs/subnetport-multi-ip-test_jm54x/subnets/subnet-dualstack-test_jm54x/ports/repro-dualstack-pool/state"
```

```json
"realized_bindings": [
  {"ip_address": "172.26.0.72", "mac_address": "00:11:22:33:44:14"},
  {"ip_address": "172.26.0.70", "mac_address": "00:11:22:33:44:14"},
  {"ip_address": "172.26.0.71", "mac_address": "00:11:22:33:44:14"}
]
```

**Root cause:** NSX's `PortAddressBindingEntry` schema has no per-binding address-family field, so a pool-allocated (IP-omitted) binding gives NSX no way to know which family to allocate from — it defaults to IPv4 for all of them.

multi-IP `addressBindings` should be treated as effectively single-family when relying on pool allocation. To combine families in one port, every binding's IP must be given explicitly.

### DIFF
--- a/pkg/apis/vpc/v1alpha1/subnetport_types.go
+++ b/pkg/apis/vpc/v1alpha1/subnetport_types.go
@@ -23,7 +23,11 @@ type SubnetPortSpec struct {
 	Subnet string `json:"subnet,omitempty"`
 	// SubnetSet defines the parent SubnetSet name of the SubnetPort.
 	SubnetSet string `json:"subnetSet,omitempty"`
-	// AddressBindings defines static address bindings used for the SubnetPort.
+	// AddressBindings defines static address bindings used for the SubnetPort. Multiple
+	// bindings sharing one MAC address are supported (NSX 9.2+) for use cases such as
+	// Blueprint VM creation where a single interface needs several IP addresses. NSX
+	// requires all non-empty MACAddress values across entries to match, and rejects
+	// changing the number of entries once the SubnetPort has been created.
 	AddressBindings []PortAddressBinding `json:"addressBindings,omitempty"`
 	// InterfaceIPType decides the address families of static IP allocation, when
 	// DHCP or SLAAC is not activated on the Subnet. When StaticIPAllocationType
--- a/pkg/apis/vpc/v1alpha1/subnetport_types.go
+++ b/pkg/apis/vpc/v1alpha1/subnetport_types.go
@@ -24,7 +24,7 @@ type SubnetPortSpec struct {
 	// SubnetSet defines the parent SubnetSet name of the SubnetPort.
 	SubnetSet string `json:"subnetSet,omitempty"`
 	// AddressBindings defines static address bindings used for the SubnetPort.
-	// Multiple bindings sharing one MAC address are supported from VCF 9.2.
+	// Multiple bindings sharing one MAC address are supported starting with VCF 9.2.0.
 	AddressBindings []PortAddressBinding `json:"addressBindings,omitempty"`
 	// InterfaceIPType decides the address families of static IP allocation, when
 	// DHCP or SLAAC is not activated on the Subnet. When StaticIPAllocationType
--- a/pkg/apis/vpc/v1alpha1/subnetport_types.go
+++ b/pkg/apis/vpc/v1alpha1/subnetport_types.go
@@ -23,11 +23,8 @@ type SubnetPortSpec struct {
 	Subnet string `json:"subnet,omitempty"`
 	// SubnetSet defines the parent SubnetSet name of the SubnetPort.
 	SubnetSet string `json:"subnetSet,omitempty"`
-	// AddressBindings defines static address bindings used for the SubnetPort. Multiple
-	// bindings sharing one MAC address are supported (NSX 9.2+) for use cases such as
-	// Blueprint VM creation where a single interface needs several IP addresses. NSX
-	// requires all non-empty MACAddress values across entries to match, and rejects
-	// changing the number of entries once the SubnetPort has been created.
+	// AddressBindings defines static address bindings used for the SubnetPort.
+	// Multiple bindings sharing one MAC address are supported from VCF 9.2.
 	AddressBindings []PortAddressBinding `json:"addressBindings,omitempty"`
 	// InterfaceIPType decides the address families of static IP allocation, when
 	// DHCP or SLAAC is not activated on the Subnet. When StaticIPAllocationType
--- a/pkg/controllers/pod/pod_controller.go
+++ b/pkg/controllers/pod/pod_controller.go
@@ -85,7 +85,7 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 
 	if !common.PodIsDeleted(pod) {
 		r.StatusUpdater.IncreaseUpdateTotal()
-		isExisting, nsxSubnetPath, subnetSetUID, subnetSetLock, interfaceIPType, err := r.GetSubnetPathForPod(ctx, pod)
+		isExisting, nsxSubnetPath, subnetSetUID, subnetSetLock, interfaceIPType, staticIPAllocationType, err := r.GetSubnetPathForPod(ctx, pod)
 		if subnetSetLock != nil {
 			defer common.RUnlockSubnetSet(*subnetSetUID, subnetSetLock)
 		}
@@ -94,15 +94,8 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 			r.StatusUpdater.UpdateFail(ctx, pod, err, "", nil)
 			return common.ResultRequeue, err
 		}
-		// nsxSubnet is resolved below; the Release closure reads it lazily so it sees the
-		// resolved Subnet even though it's registered before the Subnet is fetched.
-		var nsxSubnet *model.VpcSubnet
 		if !isExisting {
 			defer func() {
-				staticIPAllocationType := v1alpha1.StaticIPAllocationTypeNone
-				if nsxSubnet != nil {
-					staticIPAllocationType = util.ComputeDefaultStaticIPAllocationType(nsxSubnet, interfaceIPType)
-				}
 				r.SubnetPortService.ReleasePortInSubnet(nsxSubnetPath, interfaceIPType, staticIPAllocationType, nil)
 			}()
 		}
@@ -119,7 +112,7 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 			log.Error(err, "Failed to check if the Subnet is a shared Subnet", "path", nsxSubnetPath)
 			return common.ResultNormal, err
 		}
-		nsxSubnet, err = r.SubnetService.GetSubnetByPath(nsxSubnetPath, inSharedSubnet)
+		nsxSubnet, err := r.SubnetService.GetSubnetByPath(nsxSubnetPath, inSharedSubnet)
 		if err != nil {
 			r.StatusUpdater.UpdateFail(ctx, pod, err, "", nil)
 			return common.ResultRequeue, err
@@ -418,12 +411,16 @@ func (r *PodReconciler) getSubnetByPod(pod *v1.Pod, subnetSet *v1alpha1.SubnetSe
 	return common.GetSubnetByIP(subnets, net.ParseIP(pod.Status.PodIP))
 }
 
-func (r *PodReconciler) GetSubnetPathForPod(ctx context.Context, pod *v1.Pod) (bool, string, *types.UID, *sync.RWMutex, v1alpha1.IPAddressType, error) {
+// GetSubnetPathForPod also returns the StaticIPAllocationType actually resolved when
+// allocating the Subnet (only meaningful when existing is false), so the caller can release
+// the same pool capacity it reserved without recomputing it later from state that may no
+// longer be available (e.g. if a later step in Reconcile fails).
+func (r *PodReconciler) GetSubnetPathForPod(ctx context.Context, pod *v1.Pod) (bool, string, *types.UID, *sync.RWMutex, v1alpha1.IPAddressType, v1alpha1.StaticIPAllocationType, error) {
 	var subnetSetLock *sync.RWMutex
 	var subnetSetUID *types.UID
 	subnetSet, err := common.GetDefaultSubnetSetByNamespace(r.SubnetPortService.Client, pod.Namespace, servicecommon.DefaultPodNetwork)
 	if err != nil {
-		return false, "", subnetSetUID, subnetSetLock, "", err
+		return false, "", subnetSetUID, subnetSetLock, "", "", err
 	}
 	// Resolve the interface IP type from the SubnetSet up front so it stays correct on every
 	// reconcile, not just the one that first creates the SubnetPort - buildSubnetPort's static
@@ -435,7 +432,7 @@ func (r *PodReconciler) GetSubnetPathForPod(ctx context.Context, pod *v1.Pod) (b
 	subnetPath := r.SubnetPortService.GetSubnetPathForSubnetPortFromStore(pod.GetUID())
 	if len(subnetPath) > 0 {
 		log.Debug("NSX SubnetPort had been created, returning the existing NSX Subnet path", "pod.UID", pod.UID, "subnetPath", subnetPath)
-		return true, subnetPath, subnetSetUID, subnetSetLock, interfacetype, nil
+		return true, subnetPath, subnetSetUID, subnetSetLock, interfacetype, "", nil
 	}
 	log.Info("Got default SubnetSet for Pod, allocating the NSX Subnet", "subnetSet.Name", subnetSet.Name, "subnetSet.UID", subnetSet.UID, "pod.Name", pod.Name, "pod.UID", pod.UID)
 	if r.restoreMode {
@@ -444,22 +441,22 @@ func (r *PodReconciler) GetSubnetPathForPod(ctx context.Context, pod *v1.Pod) (b
 			subnetPath, err = r.getSubnetByPod(pod, subnetSet)
 			if err != nil {
 				log.Error(err, "Failed to find Subnet for restored Pod", "Pod", pod)
-				return false, "", subnetSetUID, subnetSetLock, "", err
+				return false, "", subnetSetUID, subnetSetLock, "", "", err
 			}
 			log.Debug("NSX SubnetPort will be restored on the existing NSX Subnet", "pod.UID", pod.UID, "subnetPath", subnetPath)
-			return true, subnetPath, subnetSetUID, subnetSetLock, interfacetype, nil
+			return true, subnetPath, subnetSetUID, subnetSetLock, interfacetype, "", nil
 		}
 	}
 	// SubnetSet can be created without IPAddressType, we need to wait for the value initialized by subnetset controller
 	if subnetSet.Spec.IPAddressType == "" {
-		return false, "", subnetSetUID, subnetSetLock, "", fmt.Errorf("default Pod SubnetSet IPAddressType is under calculation")
+		return false, "", subnetSetUID, subnetSetLock, "", "", fmt.Errorf("default Pod SubnetSet IPAddressType is under calculation")
 	}
-	subnetPath, subnetSetUID, subnetSetLock, err = common.AllocateSubnetFromSubnetSet(r.Client, r.APIReader, subnetSet, r.VPCService, r.SubnetService, r.SubnetPortService, interfacetype, "", nil)
+	subnetPath, staticIPAllocationType, subnetSetUID, subnetSetLock, err := common.AllocateSubnetFromSubnetSet(r.Client, r.APIReader, subnetSet, r.VPCService, r.SubnetService, r.SubnetPortService, interfacetype, "", nil)
 	if err != nil {
-		return false, subnetPath, subnetSetUID, subnetSetLock, interfacetype, err
+		return false, subnetPath, subnetSetUID, subnetSetLock, interfacetype, staticIPAllocationType, err
 	}
 	log.Info("Allocated NSX Subnet for Pod", "nsxSubnetPath", subnetPath, "pod.Name", pod.Name, "pod.UID", pod.UID)
-	return false, subnetPath, subnetSetUID, subnetSetLock, interfacetype, nil
+	return false, subnetPath, subnetSetUID, subnetSetLock, interfacetype, staticIPAllocationType, nil
 }
 
 func (r *PodReconciler) deleteSubnetPortByPodName(ctx context.Context, ns string, name string) error {
--- a/pkg/controllers/pod/pod_controller.go
+++ b/pkg/controllers/pod/pod_controller.go
@@ -94,8 +94,17 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 			r.StatusUpdater.UpdateFail(ctx, pod, err, "", nil)
 			return common.ResultRequeue, err
 		}
+		// nsxSubnet is resolved below; the Release closure reads it lazily so it sees the
+		// resolved Subnet even though it's registered before the Subnet is fetched.
+		var nsxSubnet *model.VpcSubnet
 		if !isExisting {
-			defer r.SubnetPortService.ReleasePortInSubnet(nsxSubnetPath, interfaceIPType)
+			defer func() {
+				staticIPAllocationType := v1alpha1.StaticIPAllocationTypeNone
+				if nsxSubnet != nil {
+					staticIPAllocationType = util.ComputeDefaultStaticIPAllocationType(nsxSubnet, interfaceIPType)
+				}
+				r.SubnetPortService.ReleasePortInSubnet(nsxSubnetPath, interfaceIPType, staticIPAllocationType, nil)
+			}()
 		}
 		log.Info("Got NSX Subnet for Pod", "NSX Subnet path", nsxSubnetPath, "pod.Name", pod.Name, "pod.UID", pod.UID)
 		node, err := r.GetNodeByName(pod.Spec.NodeName)
@@ -110,7 +119,7 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 			log.Error(err, "Failed to check if the Subnet is a shared Subnet", "path", nsxSubnetPath)
 			return common.ResultNormal, err
 		}
-		nsxSubnet, err := r.SubnetService.GetSubnetByPath(nsxSubnetPath, inSharedSubnet)
+		nsxSubnet, err = r.SubnetService.GetSubnetByPath(nsxSubnetPath, inSharedSubnet)
 		if err != nil {
 			r.StatusUpdater.UpdateFail(ctx, pod, err, "", nil)
 			return common.ResultRequeue, err
@@ -445,7 +454,7 @@ func (r *PodReconciler) GetSubnetPathForPod(ctx context.Context, pod *v1.Pod) (b
 	if subnetSet.Spec.IPAddressType == "" {
 		return false, "", subnetSetUID, subnetSetLock, "", fmt.Errorf("default Pod SubnetSet IPAddressType is under calculation")
 	}
-	subnetPath, subnetSetUID, subnetSetLock, err = common.AllocateSubnetFromSubnetSet(r.Client, r.APIReader, subnetSet, r.VPCService, r.SubnetService, r.SubnetPortService, interfacetype)
+	subnetPath, subnetSetUID, subnetSetLock, err = common.AllocateSubnetFromSubnetSet(r.Client, r.APIReader, subnetSet, r.VPCService, r.SubnetService, r.SubnetPortService, interfacetype, "", nil)
 	if err != nil {
 		return false, subnetPath, subnetSetUID, subnetSetLock, interfacetype, err
 	}
--- a/pkg/controllers/subnetport/subnetport_controller.go
+++ b/pkg/controllers/subnetport/subnetport_controller.go
@@ -198,51 +198,28 @@ func (r *SubnetPortReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 				}
 			}
 			subnetPort.Status.Attachment = v1alpha1.PortAttachment{ID: *nsxSubnetPortState.Attachment.Id}
+			// Placeholder sizing until RealizedBindings arrives below and replaces it outright.
+			preSeedCount := 1
+			if subnetPort.Spec.InterfaceIPType == v1alpha1.IPAddressTypeIPv4IPv6 {
+				preSeedCount = 2
+			}
+			if len(subnetPort.Spec.AddressBindings) > preSeedCount {
+				preSeedCount = len(subnetPort.Spec.AddressBindings)
+			}
 			subnetPort.Status.NetworkInterfaceConfig = v1alpha1.NetworkInterfaceConfig{
-				IPAddresses: []v1alpha1.NetworkInterfaceIPAddress{
-					{
-						Gateway: "",
-					},
-				},
+				IPAddresses:               make([]v1alpha1.NetworkInterfaceIPAddress, preSeedCount),
 				DHCPDeactivatedOnSubnet:   !util.NSXSubnetDHCPEnabled(nsxSubnet),
 				DHCPv6DeactivatedOnSubnet: !util.NSXSubnetDHCPv6Enabled(nsxSubnet),
 			}
-			// Append one more ipaddress for dual stack SubnetPort
-			if subnetPort.Spec.InterfaceIPType == v1alpha1.IPAddressTypeIPv4IPv6 {
-				subnetPort.Status.NetworkInterfaceConfig.IPAddresses = append(
-					subnetPort.Status.NetworkInterfaceConfig.IPAddresses,
-					v1alpha1.NetworkInterfaceIPAddress{Gateway: ""},
-				)
-			}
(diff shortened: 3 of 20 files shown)


pkg/apis/vpc/v1alpha1/subnetport_types.go
```diff
@@ -23,7 +23,11 @@ type SubnetPortSpec struct {
 	Subnet string `json:"subnet,omitempty"`
 	// SubnetSet defines the parent SubnetSet name of the SubnetPort.
 	SubnetSet string `json:"subnetSet,omitempty"`
-	// AddressBindings defines static address bindings used for the SubnetPort.
+	// AddressBindings defines static address bindings used for the SubnetPort. Multiple
+	// bindings sharing one MAC address are supported (NSX 9.2+) for use cases such as
+	// Blueprint VM creation where a single interface needs several IP addresses. NSX
+	// requires all non-empty MACAddress values across entries to match, and rejects
+	// changing the number of entries once the SubnetPort has been created.
```
yanjunz97: Maybe no need to explain a lot in the CRD description? Just `Multiple bindings sharing one MAC address are supported from VCF 9.2` cc @dantingl

yanjunz97:
There's one more thing

https://github.com/vmware-tanzu/nsx-operator/blob/a4b9079eb23c27ba968a16da4cc862d25811ad02/pkg/nsx/services/subnetport/subnetport.go#L495-L503
https://github.com/vmware-tanzu/nsx-operator/blob/a4b9079eb23c27ba968a16da4cc862d25811ad02/pkg/nsx/services/subnetport/subnetport.go#L686-L693

Currently we always consider only one ip will be used for each ipaddresstype when evaluate the subnet v4/v6 capacity. For multiple addressbinding we will need to reserve/release the correct number of IP.

yanjunz97:
I just notice we also miss the changes in AllocatePortFromSubnet for mixed mode subnet, i.e. currently we will only check DHCP pool for Subnet capacity.

https://github.com/vmware-tanzu/nsx-operator/blob/d1e1a9c67e5614380ac77138fa3b9bd778fc976a/pkg/nsx/services/subnetport/subnetport.go#L532-L544

But we shall check DHCP pool when Subnetport IP is allocated from dhcp pool, and check static ip pool when Subnetport IP is allocated from static pool

Maybe for mixed mode Subnet we should have one totalIP for static pool and another totalIP for dhcp pool. And this also applies to IPv6.

Hi Pooja, could you check and include this change in this PR or a following up PR?

pkg/apis/vpc/v1alpha1/subnetport_types.go
dantingl: Yeah, no need to explain the details

poojav25: Done — simplified to: "AddressBindings defines static address bindings used for the SubnetPort. Multiple bindings sharing one MAC address are supported from VCF 9.2."

poojav25:
> There's one more thing
> 
> https://github.com/vmware-tanzu/nsx-operator/blob/a4b9079eb23c27ba968a16da4cc862d25811ad02/pkg/nsx/services/subnetport/subnetport.go#L495-L503
> 
> 
> https://github.com/vmware-tanzu/nsx-operator/blob/a4b9079eb23c27ba968a16da4cc862d25811ad02/pkg/nsx/services/subnetport/subnetport.go#L686-L693
> 
> Currently we always consider only one ip will be used for each ipaddresstype when evaluate the subnet v4/v6 capacity. For multiple addressbinding we will need to reserve/release the correct number of IP.

Fixed — AllocatePortFromSubnet/ReleasePortInSubnet now count the actual number of IPv4/IPv6 addresses requested via addressBindings (via a new CountAddressBindingsByFamily helper) instead of assuming 1 per family. A SubnetPort requesting N addresses in one family now reserves/releases N slots of capacity. Added unit tests covering exact-fit and over-capacity multi-IP allocation, plus an e2e case in SubnetMixedMode that creates a 2-IP SubnetPort and asserts both IPs are realized.

poojav25:
> I just notice we also miss the changes in AllocatePortFromSubnet for mixed mode subnet, i.e. currently we will only check DHCP pool for Subnet capacity.
> 
> https://github.com/vmware-tanzu/nsx-operator/blob/d1e1a9c67e5614380ac77138fa3b9bd778fc976a/pkg/nsx/services/subnetport/subnetport.go#L532-L544
> 
> But we shall check DHCP pool when Subnetport IP is allocated from dhcp pool, and check static ip pool when Subnetport IP is allocated from static pool
> 
> Maybe for mixed mode Subnet we should have one totalIP for static pool and another totalIP for dhcp pool. And this also applies to IPv6.
> 
> Hi Pooja, could you check and include this change in this PR or a following up PR?


Addressed in this PR. CountInfo now tracks separate dirty/total counters for the static pool and the DHCP pool, per family (totalStaticIP/totalDhcpIP, plus IPv6 equivalents). Each SubnetPort's capacity is checked against whichever pool it actually allocates from — resolved per-family via StaticIPAllocationType — instead of always checking DHCP pool stats.

This also covers IPv6: dhcp-server-config-stats already returns real DHCPv6 pool usage (dhcp_ipv6.ip_pool_stats) on NSX versions with VpcIpv6 enabled, so I wired that up too (gated by NSXCheckVersion(nsx.IPv6), falling back to the previous unconditional-allow behavior on older NSX).

While testing this live against NSX I found and fixed a related bug: existingPortCount was counting every realized port on the Subnet regardless of which pool it came from, so a DHCP-sourced port could falsely eat into static-pool capacity accounting (and vice versa) on a mixed-mode Subnet. Filtered by each port's Attachment.AllocateAddresses now.

Added unit tests for pool selection on a genuinely mixed-mode Subnet (DHCP+static both enabled), the existing-port pool-filtering fix, and multi-IP capacity accounting. Verified live against NSX with a rebuilt image — confirmed the client-side pre-check no longer falsely rejects legitimate requests, and correctly rejects (client-side, before ever calling NSX) requests that genuinely exceed a pool's capacity.

pkg/apis/vpc/v1alpha1/subnetport_types.go
```diff
@@ -24,6 +24,7 @@ type SubnetPortSpec struct {
 	// SubnetSet defines the parent SubnetSet name of the SubnetPort.
 	SubnetSet string `json:"subnetSet,omitempty"`
 	// AddressBindings defines static address bindings used for the SubnetPort.
+	// Multiple bindings sharing one MAC address are supported from VCF 9.2.
```
yanjunz97: Could you run `make changecrd`?

pkg/controllers/pod/pod_controller.go
```diff
@@ -94,8 +94,17 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
...
+		// nsxSubnet is resolved below; the Release closure reads it lazily so it sees the
+		// resolved Subnet even though it's registered before the Subnet is fetched.
+		var nsxSubnet *model.VpcSubnet
 		if !isExisting {
-			defer r.SubnetPortService.ReleasePortInSubnet(nsxSubnetPath, interfaceIPType)
+			defer func() {
+				staticIPAllocationType := v1alpha1.StaticIPAllocationTypeNone
+				if nsxSubnet != nil {
+					staticIPAllocationType = util.ComputeDefaultStaticIPAllocationType(nsxSubnet, interfaceIPType)
+				}
```
yanjunz97: If `r.SubnetService.GetSubnetByPath(nsxSubnetPath, inSharedSubnet)` fails, `nsxSubnet` remains `nil`. Consequently, the deferred function executes with `staticIPAllocationType = StaticIPAllocationTypeNone`.
But static ip may be reserved in `GetSubnetPathForPod`. Maybe let `GetSubnetPathForPod` return `staticIPAllocationType`?

pkg/nsx/services/subnetport/subnetport.go
```diff
@@ -97,12 +97,32 @@ func setupStore() *SubnetPortStore {
...
+	return true
+}
+
 func (service *SubnetPortService) portAlreadyRealized(obj interface{}, nsxSubnetPort *model.VpcSubnetPort) bool {
 	switch o := obj.(type) {
 	case *v1alpha1.SubnetPort:
+		expectedIPCount := 1
+		if o.Spec.InterfaceIPType == v1alpha1.IPAddressTypeIPv4IPv6 {
+			expectedIPCount = 2
+		}
```
yanjunz97: Maybe we need to use staticIPAllocation? e.g. for mixed mode subnet, Subnetport with InterfaceIPType IPv4IPv6, and staticIPAllocation IPv4, we should only expect 1 ip in status.

Could you also add cases for subnetport with multiple bindings or without bindings for both dhcp subnet and mixed modes subnet?

pkg/nsx/services/subnetport/subnetport.go
```diff
@@ -368,6 +388,33 @@ func (service *SubnetPortService) GetPortsOfSubnet(subnetPath string) (ports []*
...
+
+// countExistingPortsForPool counts only realized ports sourced from the pool being checked,
+// so a DHCP-sourced port on a mixed-mode Subnet doesn't count against static capacity (or
+// vice versa).
+func (service *SubnetPortService) countExistingPortsForPool(subnetPath string, useStaticPool bool) int {
+	count := 0
+	for _, port := range service.GetPortsOfSubnet(subnetPath) {
+		if isPortStaticAllocated(port) == useStaticPool {
+			count++
+		}
```
yanjunz97: Shall we count for IPs instead of ports here? For Ports with 2 bindings, it should be counted for 2 IPs are used?

yanjunz97:
Hi Pooja, could you also add test for day 2 update in addressbinding? i.e. Add more addressbindings or remove addressbinding after the SubnetPort is created.

pkg/apis/vpc/v1alpha1/subnetport_types.go
poojav25: Done. The generated YAMLs under build/yaml/crd have been updated

pkg/controllers/pod/pod_controller.go
poojav25: Done.

pkg/nsx/services/subnetport/subnetport.go
poojav25: Done. countExistingPortsForPool has been updated to count actual IP addresses consumed across all bindings (via countNSXAddressBindingsByFamily) rather than just counting the ports.

pkg/nsx/services/subnetport/subnetport.go
poojav25: Done. expectedIPCount now correctly derives from StaticIPAllocationType. I've also added the requested test cases for DHCP and mixed-mode subnets (with and without explicit bindings) in   the portAlreadyRealized tests.

poojav25:
> Hi Pooja, could you also add test for day 2 update in addressbinding? i.e. Add more addressbindings or remove addressbinding after the SubnetPort is created.

Done. added "UpdateAddAddressBindings" and "UpdateRemoveAddressBindings" test cases to cover both adding and removing bindings during a Day 2 update

build/yaml/crd/eas/eas.nsx.vmware.com_vpcipaddressusages.yaml
dantingl, line 79: Looks like this is added after rebase

pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -1038,6 +1020,25 @@ func (r *SubnetPortReconciler) CheckAndGetSubnetPathForSubnetPort(ctx context.Co
...
 }
 
+// networkInterfaceIPAddressesFromRealizedBindings builds one entry per realized
+// binding with an IP, plus the shared MAC address, independent of binding count.
+func networkInterfaceIPAddressesFromRealizedBindings(realizedBindings []model.AddressBindingEntry) ([]v1alpha1.NetworkInterfaceIPAddress, string) {
+	macAddress := ""
+	ipAddresses := make([]v1alpha1.NetworkInterfaceIPAddress, 0, len(realizedBindings))
+	for _, binding := range realizedBindings {
+		if binding.Binding != nil && binding.Binding.IpAddress != nil {
+			if macAddress == "" && binding.Binding.MacAddress != nil {
```
dantingl, line 1082: Do we support a case that realized binding has mac address but does not have IP address, for DHCP case we can also get IP address from realized binding?

pkg/nsx/services/subnetport/subnetport.go
```diff
@@ -97,12 +98,45 @@ func setupStore() *SubnetPortStore {
...
+		// comes from DHCP/SLAAC and isn't tracked here.
+		expectedIPCount := 1
+		switch o.Spec.StaticIPAllocationType {
+		case v1alpha1.StaticIPAllocationTypeIPv4IPv6:
+			expectedIPCount = 2
+		case "", v1alpha1.StaticIPAllocationTypeNone:
+			// Not backfilled yet, or genuinely no static allocation (the AllocateAddresses
+			// check below already ensures we only reach here when static allocation is
+			// actually active) - fall back to InterfaceIPType as a best-effort default.
+			if o.Spec.InterfaceIPType == v1alpha1.IPAddressTypeIPv4IPv6 {
```
yanjunz97, line 130: Just to double check if this is a defense code or you do observe case like this. Shall the StaticIPAllocationType shall always be backfilled before this function is called?

pkg/nsx/services/subnetport/subnetport.go
```diff
@@ -368,6 +402,63 @@ func (service *SubnetPortService) GetPortsOfSubnet(subnetPath string) (ports []*
...
+		}
+	}
+	return ipv4Count, ipv6Count
+}
+
+// countExistingPortsForPool counts IPs (not ports) consumed from the pool being checked, for
+// the given family. Only ports sourced from that pool are counted, so a DHCP-sourced port on
+// a mixed-mode Subnet doesn't count against static capacity (or vice versa). A port with
+// multiple bindings in the requested family counts for all of them, not just one.
+func (service *SubnetPortService) countExistingPortsForPool(subnetPath string, useStaticPool bool, useIPv6 bool) int {
```
yanjunz97: countExistingPortsForPool -> countExistingIPsForPool?

pkg/nsx/services/subnetport/subnetport.go
```diff
@@ -576,14 +680,14 @@ func (service *SubnetPortService) checkIPv4Capacity(subnet *model.VpcSubnet, sha
...
-		info.totalIP = totalIP - 4
+		info.totalDhcpIP = totalIP - 4
 	}
 
 	if time.Since(info.exhaustedCheckTime) < IPReleaseTime {
 		return false, nil
 	}
 
-	existingPortCount := len(service.GetPortsOfSubnet(*subnet.Path))
+	existingPortCount := service.countExistingPortsForPool(*subnet.Path, useStaticPool, false)
```
yanjunz97: existingPortCount -> existingIPCount

pkg/nsx/services/subnetport/subnetport.go
```diff
@@ -368,6 +402,63 @@ func (service *SubnetPortService) GetPortsOfSubnet(subnetPath string) (ports []*
...
+	case "BOTH", "IP_POOL":
+		return true
+	default:
+		return false
+	}
+}
+
+// countNSXAddressBindingsByFamily is the NSX-model equivalent of
+// util.CountAddressBindingsByFamily, used for a realized port's AddressBindings.
+func countNSXAddressBindingsByFamily(bindings []model.PortAddressBindingEntry) (ipv4Count, ipv6Count int) {
```
yanjunz97, line 421: I notice the coverage of this file is low, could you also add some ut to cover new lines in countNSXAddressBindingsByFamily/checkIPv6Capacity/ReleasePortInSubnet

pkg/controllers/subnetport/subnetport_controller.go
yanjunz97: This remind me of one thing, for VM attached to DHCP subnetport, the port will get both IP address and mac from realized binding. In this case, we shall not update the IP address on SubnetPort, as DHCP IP can change and we do not want SubnetPort has an IP different with the real one on VM. You can also skip the MAC for this case as we have vif poll logic to update the MAC.

For DHCP subnetport with MAC specified and not attached to VM, I think it is possible that realized binding has mac address but does not have IP address. We need to have the MAC updated at the SubnetPort realized time. Previously we rely on this condition to get the MAC filled from AddressBindings
```
else if !util.NSXSubnetStaticIPAllocationEnabled(nsxSubnet) && len(subnetPort.Spec.AddressBindings) > 0 {
					// If StaticIPAllocation is disabled, propagate the MAC from spec.addressBinding to status
					// StaticIPAllocation disabled: propagate MAC from spec since there's no realized binding yet.
					subnetPort.Status.NetworkInterfaceConfig.MACAddress = subnetPort.Spec.AddressBindings[0].MACAddress
				}
```
but now we have mixed mode subnet so we need to improve the logic to catch it.

pkg/nsx/services/subnetport/subnetport.go
poojav25: Yes, this handles the case documented in Reconcile where the VM Operator can briefly overwrite the backfilled StaticIPAllocationType during a race condition, so I added this as defensive code to handle that scenario.

pkg/nsx/services/subnetport/subnetport.go
poojav25: done

pkg/nsx/services/subnetport/subnetport.go
poojav25: done

pkg/nsx/services/subnetport/subnetport.go
poojav25: Done

pkg/controllers/subnetport/subnetport_controller.go
poojav25: I have updated the logic to address this. Instead of relying on util.NSXSubnetStaticIPAllocationEnabled(nsxSubnet) (which evaluates the global subnet setting), the logic now checks staticIPAllocationType == v1alpha1.StaticIPAllocationTypeNone at the individual port level.

This ensures that even on a mixed-mode subnet, a DHCP-sourced SubnetPort with a specified MAC address will correctly catch this condition and proactively propagate the MAC from the spec to the status before the realized binding exists

build/yaml/crd/eas/eas.nsx.vmware.com_vpcipaddressusages.yaml
poojav25: Yes, that's correct. Those changes were automatically brought in when running make changecrd after rebasing the  branch

pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -232,14 +232,14 @@ func (r *SubnetPortReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 				DHCPv6DeactivatedOnSubnet: !util.NSXSubnetDHCPv6Enabled(nsxSubnet),
 				RADeactivated:             raDeactivated,
 			}
-			if util.NSXSubnetStaticIPAllocationEnabled(nsxSubnet) || len(subnetPort.Spec.AddressBindings) > 0 {
+			if staticIPAllocationType != v1alpha1.StaticIPAllocationTypeNone || len(subnetPort.Spec.AddressBindings) > 0 {
 				if len(nsxSubnetPortState.RealizedBindings) > 0 {
 					realizedIPAddresses, macAddress := networkInterfaceIPAddressesFromRealizedBindings(nsxSubnetPortState.RealizedBindings)
-					if len(realizedIPAddresses) > 0 {
+					if staticIPAllocationType != v1alpha1.StaticIPAllocationTypeNone && len(realizedIPAddresses) > 0 {
```
yanjunz97: For mixed mode Subnet case like 
InterfaceIPType = IPv4IPv6, StaticIPAllocationType = IPv6
networkInterfaceIPAddressesFromRealizedBindings may return the DHCP IPv4 IP 
Maybe we shall compare the staticIPAllocationType with the realized IP family in networkInterfaceIPAddressesFromRealizedBindings?

poojav25: Good catch! I've updated the logic.

pkg/apis/vpc/v1alpha1/subnetport_types.go
wenqiq: Better update the comment to `Supported starting with VCF 9.2.0.` to align with the descriptions of the other features of version 9.2.

pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -206,60 +213,39 @@ func (r *SubnetPortReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			}
 			// In restore mode, we do not update the SubnetPort status to retian the status, so no need to check the realized bindings
 			// Check StaticIPAllocationType as mixed mode Subnet also have static ip allocation enabled but SubnetPort may be created in dhcp pool
-			if !r.restoreMode && util.NSXSubnetStaticIPAllocationEnabled(nsxSubnet) && subnetPort.Spec.StaticIPAllocationType != "None" {
-				if len(nsxSubnetPortState.RealizedBindings) == 0 {
+			if !r.restoreMode && util.NSXSubnetStaticIPAllocationEnabled(nsxSubnet) && staticIPAllocationType != v1alpha1.StaticIPAllocationTypeNone {
```
yanjunz97: Use subnetPort.Spec.StaticIPAllocationType for safety? As staticIPAllocationType can be empty for existing subnetport, while nsxSubnetPortState can be nil for update on the existing subnetport so code can reach here with empty staticIPAllocationType.

pkg/controllers/subnetport/subnetport_controller.go
yanjunz97: Have you updated this? I think some change in networkInterfaceIPAddressesFromRealizedBindings will need to fix this. Currently dhcp ipv4 address in RealizedBindings can still be shown on SubnetPort status?

pkg/apis/vpc/v1alpha1/subnetport_types.go
poojav25: fixed

pkg/controllers/subnetport/subnetport_controller.go
poojav25: done

pkg/controllers/subnetport/subnetport_controller.go
poojav25: I looked into this, and I don't think we should add that comparison — filtering by staticIPAllocationType inside networkInterfaceIPAddressesFromRealizedBindings would end up hiding a real address.

In that exact scenario (InterfaceIPType=IPv4IPv6, StaticIPAllocationType=IPv6), yes — the DHCP IPv4 address will show up in RealizedBindings alongside the static IPv6 one, and today's code does return both. But that's because the port genuinely has both addresses at the same time: DHCP keeps assigning IPv4 since it isn't statically pinned, independent of the IPv6 static allocation. I checked this against the NSX source (DHCP and static allocation are separate, independent mechanisms).

If we filtered networkInterfaceIPAddressesFromRealizedBindings down to only the family in staticIPAllocationType, we'd drop the DHCP IPv4 entry from status — even though the port's interface really does have that address. That would make NetworkInterfaceConfig.IPAddresses report less than the truth.

The place that does need to compare against staticIPAllocationType is realizedBindingsCoverStaticFamilies — deciding "has the family I actually requested arrived" — and that's already implemented correctly. networkInterfaceIPAddressesFromRealizedBindings's job is different: report what the port actually has, not just what was requested.

I added a test case covering exactly this mixed DHCP+static scenario, plus a comment on the function, so this is explicit rather than implicit going forward. Let me know your thoughts

poojav25: thanks for the clarification on why — that's the piece I was missing. Implemented: networkInterfaceIPAddressesFromRealizedBindings now takes staticIPAllocationType and drops any realized address whose family isn't covered by it, so a DHCP-realized address (e.g. IPv4 on a dual-stack port statically allocating only IPv6) no longer shows up in SubnetPort status.

dantingl (dismissed):
LGTM

yanjunz97 (dismissed):
LGTM

yanjunz97 (approved):

dantingl (approved):

PR #1512, opened 2026-07-24
Support RA mode status on SubnetPort.

poojav25:
SubnetPort.status.networkInterfaceConfig.raDeactivated now correctly reflects whether IPv6 RA is deactivated on the port's VPC — the CRD field already existed but was never populated.

To figure out whether RA is on or off, the operator looks up the SubnetPort's Subnet, finds its VPC, checks which service profile that VPC uses, and reads the ra_mode from the IPv6 NDRA profile attached to it. If the VPC has no service profile, or that profile doesn't have an NDRA profile attached, RA is treated as deactivated.

**Setup — Subnet**
```bash
kubectl get subnet subnet-dualstack-test -n subnetport-multi-ip-test -o yaml

status:
  conditions:
  - message: 'NSX Subnet with DHCPv4: DHCPDeactivated has been successfully created/updated'
    reason: SubnetReady
    status: "True"
    type: Ready
  gatewayAddresses:
  - 172.26.0.65/26
  - fd12:1234::1/64
  networkAddresses:
  - 172.26.0.64/26
  - fd12:1234::/64
```
**T1 — No NDRA profile attached to the VPC's VpcServiceProfile: RA reported as deactivated**

```bash
curl -sk -u 'admin:<password>' \
  "https://<nsx-manager>/policy/api/v1/orgs/default/projects/project-quality/vpc-service-profiles/default--46bd56af-40f2-4730-a127-baa70f304477" \
  | python3 -c "import json,sys; d=json.load(sys.stdin); print('ipv6_profile_paths:', d.get('ipv6_profile_paths'))"

ipv6_profile_paths: None
```
```bash
kubectl apply -f - <<EOF
kind: SubnetPort
metadata:
  name: sp-ra-test
  namespace: subnetport-multi-ip-test
spec:
  subnet: subnet-dualstack-test
  interfaceIPType: IPv4IPv6
EOF
kubectl get subnetport sp-ra-test -n subnetport-multi-ip-test -o yaml
------------
status:
  conditions:
  - message: NSX subnet port has been successfully created/updated
    reason: SubnetPortReady
    status: "True"
    type: Ready
  networkInterfaceConfig:
    ipAddresses:
    - gateway: 172.26.0.65
      ipAddress: 172.26.0.70/26
    - gateway: fd12:1234::1
      ipAddress: fd12:1234::ffff:ffff:ffff:f830/64
    macAddress: "04:50:56:00:20:00"
    raDeactivated: true
  ```
  
  **T2 — Active NDRA profile attached (ra_mode: SLAAC_DNS_THROUGH_RA): RA reported as not deactivated**
  
  ```bash
  curl -sk -u 'admin:<password>' -X PUT \
  "https://<nsx-manager>/policy/api/v1/orgs/default/projects/project-quality/vpc-service-profiles/default--46bd56af-40f2-4730-a127-baa70f304477" \
  -H "Content-Type: application/json" \
  -d @vsp-with-default-ndra.json

"ipv6_profile_paths" : [ "/infra/ipv6-ndra-profiles/default" ]
```
```bash
kubectl apply -f - <<EOF
kind: SubnetPort
metadata:
  name: sp-ra-test-round2
  namespace: subnetport-multi-ip-test
spec:
  subnet: subnet-dualstack-test
  interfaceIPType: IPv4IPv6
EOF
kubectl get subnetport sp-ra-test-round2 -n subnetport-multi-ip-test -o jsonpath='{.status.networkInterfaceConfig}'
----------
{"dhcpDeactivatedOnSubnet":true,"dhcpv6DeactivatedOnSubnet":true,"ipAddresses":[...],"logicalSwitchUUID":"...","macAddress":"04:50:56:00:bc:00"}
```

```bash
raDeactivated is correctly false here, but is omitted from the object entirely — the field has omitempty and false is the Go zero value, so it's indistinguishable from "never set" in kubectl output by design (same as dhcpDeactivatedOnSubnet/dhcpv6DeactivatedOnSubnet would be). Confirmed directly via the operator's debug log instead:


DEBUG vpc.go:1267 Resolved RA mode for VPC Ipv6NdraProfile=/infra/ipv6-ndra-profiles/default RADeactivated=false RaMode=SLAAC_DNS_THROUGH_RA VPC=/orgs/default/projects/project-quality/vpcs/subnetport-multi-ip-test_jm54x
```
**T3 — Explicit ra_mode: DISABLED NDRA profile attached: RA reported as deactivated**

```bash
curl -sk -u 'admin:<password>' -X PUT \
  "https://<nsx-manager>/policy/api/v1/infra/ipv6-ndra-profiles/default" \
  -H "Content-Type: application/json" \
  -d @ndra-disabled.json

"ra_mode" : "DISABLED"
```
```bash
kubectl apply -f - <<EOF
kind: SubnetPort
metadata:
  name: sp-ra-test-round3
  namespace: subnetport-multi-ip-test
spec:
  subnet: subnet-dualstack-test
  interfaceIPType: IPv4IPv6
EOF
-------
kubectl get subnetport sp-ra-test-round3 -n subnetport-multi-ip-test -o jsonpath='{.status.networkInterfaceConfig}'
------------
{"dhcpDeactivatedOnSubnet":true,"dhcpv6DeactivatedOnSubnet":true,"ipAddresses":[...],"logicalSwitchUUID":"...","macAddress":"04:50:56:00:10:00","raDeactivated":true}

```

### DIFF
--- a/pkg/controllers/subnetport/subnetport_controller.go
+++ b/pkg/controllers/subnetport/subnetport_controller.go
@@ -207,6 +207,11 @@ func (r *SubnetPortReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 				DHCPDeactivatedOnSubnet:   !util.NSXSubnetDHCPEnabled(nsxSubnet),
 				DHCPv6DeactivatedOnSubnet: !util.NSXSubnetDHCPv6Enabled(nsxSubnet),
 			}
+			if raDeactivated, err := r.VPCService.IsRADeactivatedByVPCPath(nsxSubnetPath); err != nil {
+				log.Error(err, "Failed to determine RA mode for SubnetPort's VPC", "SubnetPort", subnetPort, "nsxSubnetPath", nsxSubnetPath)
+			} else {
+				subnetPort.Status.NetworkInterfaceConfig.RADeactivated = raDeactivated
+			}
 			// Append one more ipaddress for dual stack SubnetPort
 			if subnetPort.Spec.InterfaceIPType == v1alpha1.IPAddressTypeIPv4IPv6 {
 				subnetPort.Status.NetworkInterfaceConfig.IPAddresses = append(
--- a/pkg/controllers/subnetport/subnetport_controller.go
+++ b/pkg/controllers/subnetport/subnetport_controller.go
@@ -160,7 +160,7 @@ func (r *SubnetPortReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		if err != nil {
 			log.Error(err, "Failed to determine RA mode for SubnetPort's VPC", "SubnetPort", subnetPort, "nsxSubnetPath", nsxSubnetPath)
 			r.StatusUpdater.UpdateFail(ctx, subnetPort, err, "Failed to determine RA mode for SubnetPort's VPC", setSubnetPortReadyStatusFalse, r.SubnetPortService, r.restoreMode)
-			return common.ResultRequeue, err
+			return common.ResultNormal, err
 		}
 
 		isVmSubnetPort := true
--- a/pkg/controllers/subnetport/subnetport_controller.go
+++ b/pkg/controllers/subnetport/subnetport_controller.go
@@ -156,6 +156,12 @@ func (r *SubnetPortReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		if err != nil {
 			return common.ResultNormal, err
 		}
+		raDeactivated, err := r.VPCService.IsRADeactivatedByVPCPath(nsxSubnetPath)
+		if err != nil {
+			log.Error(err, "Failed to determine RA mode for SubnetPort's VPC", "SubnetPort", subnetPort, "nsxSubnetPath", nsxSubnetPath)
+			r.StatusUpdater.UpdateFail(ctx, subnetPort, err, "Failed to determine RA mode for SubnetPort's VPC", setSubnetPortReadyStatusFalse, r.SubnetPortService, r.restoreMode)
+			return common.ResultRequeue, err
+		}
 
 		isVmSubnetPort := true
 		if value, exists := subnetPort.Labels[servicecommon.LabelImageFetcher]; exists && value == "true" {
@@ -206,11 +212,7 @@ func (r *SubnetPortReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 				},
 				DHCPDeactivatedOnSubnet:   !util.NSXSubnetDHCPEnabled(nsxSubnet),
 				DHCPv6DeactivatedOnSubnet: !util.NSXSubnetDHCPv6Enabled(nsxSubnet),
-			}
-			if raDeactivated, err := r.VPCService.IsRADeactivatedByVPCPath(nsxSubnetPath); err != nil {
-				log.Error(err, "Failed to determine RA mode for SubnetPort's VPC", "SubnetPort", subnetPort, "nsxSubnetPath", nsxSubnetPath)
-			} else {
-				subnetPort.Status.NetworkInterfaceConfig.RADeactivated = raDeactivated
+				RADeactivated:             raDeactivated,
 			}
 			// Append one more ipaddress for dual stack SubnetPort
 			if subnetPort.Spec.InterfaceIPType == v1alpha1.IPAddressTypeIPv4IPv6 {
--- a/pkg/nsx/services/vpc/vpc.go
+++ b/pkg/nsx/services/vpc/vpc.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"path"
 	"strings"
 	"sync"
 
@@ -54,9 +55,10 @@ var (
 
 type VPCService struct {
 	common.Service
-	VpcStore            *VPCStore
-	LbsStore            *LBSStore
-	defaultProjectCache sync.Map // cache for IsDefaultNSXProject results, key: "orgID/projectID", value: bool
+	VpcStore             *VPCStore
+	LbsStore             *LBSStore
+	defaultProjectCache  sync.Map // cache for IsDefaultNSXProject results, key: "orgID/projectID", value: bool
+	raDeactivatedCache   sync.Map // cache for IsRADeactivatedByVPCPath results, key: VPC path, value: bool
 }
 
 func (s *VPCService) GetDefaultNetworkConfig() (*v1alpha1.VPCNetworkConfiguration, error) {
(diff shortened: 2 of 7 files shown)


pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -207,6 +207,11 @@ func (r *SubnetPortReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 				DHCPDeactivatedOnSubnet:   !util.NSXSubnetDHCPEnabled(nsxSubnet),
 				DHCPv6DeactivatedOnSubnet: !util.NSXSubnetDHCPv6Enabled(nsxSubnet),
 			}
+			if raDeactivated, err := r.VPCService.IsRADeactivatedByVPCPath(nsxSubnetPath); err != nil {
+				log.Error(err, "Failed to determine RA mode for SubnetPort's VPC", "SubnetPort", subnetPort, "nsxSubnetPath", nsxSubnetPath)
```
yanjunz97: I think we should `return common.ResultNormal, err` if failed to get RA mode
As it is only determined by nsxSubnetPath, maybe we can get raDeactivated at line 125 or 155; and assign it to subnetPort.Status.NetworkInterfaceConfig.RADeactivated here?

pkg/nsx/services/vpc/vpc.go
```diff
@@ -1213,6 +1214,71 @@ func (s *VPCService) GetVPCFromNSXByPath(vpcPath string) (*model.Vpc, error) {
 	return &vpc, nil
 }
 
+// IsRADeactivatedByVPCPath determines whether IPv6 Router Advertisement is deactivated for the
+// VPC that owns the given NSX path (a VPC path or any path nested under a VPC, e.g. a Subnet
+// path). It resolves the VPC's VpcServiceProfile, finds the referenced Ipv6NdraProfile, and
+// checks its RaMode. If no VpcServiceProfile or no Ipv6NdraProfile is configured, RA is
+// considered deactivated: NSX does not push any ra_config to the dataplane in that case, so no
+// Router Advertisements are actually sent on the wire, which is dataplane-equivalent to an
+// explicit ra_mode of DISABLED. Not cached: an admin can change VPC RA config at any time, and a
```
yanjunz97, line 1223: May I learn how do you determine this?
`NSX does not push any ra_config to the dataplane in that case`

I observed if ndra profile is not explicitly set in vpc service profile, ra mode will always be SLAAC_DNS_THROUGH_RA, which refers to RAMode enabled.

pkg/nsx/services/vpc/vpc.go
```diff
@@ -1213,6 +1214,71 @@ func (s *VPCService) GetVPCFromNSXByPath(vpcPath string) (*model.Vpc, error) {
...
+		return false, err
+	}
+
+	ndraProfilePath := findIpv6NdraProfilePath(profile.Ipv6ProfilePaths)
+	if ndraProfilePath == "" {
+		log.Debug("No Ipv6NdraProfile referenced by VpcServiceProfile, treating RA as deactivated", "VpcServiceProfile", *vpc.VpcServiceProfile)
+		return true, nil
+	}
+
+	ndraProfile, err := s.NSXClient.Ipv6NdraProfileClient.Get(path.Base(ndraProfilePath))
```
yanjunz97: I find ndra profile can also be created under /orgs/default/projects/project-quality/infra/ipv6-ndra-profiles/default
could you double check if we also need to introduce the project client for ipv6Ndra?

poojav25: Confirmed — added ProjectIpv6NdraProfileClient (orgs/projects/infra.Ipv6NdraProfilesClient) and now pick between the infra-level and project-scoped client based on whether the NDRA profile path starts with /orgs/

pkg/controllers/subnetport/subnetport_controller.go
poojav25: Good catch — updated.

pkg/nsx/services/vpc/vpc.go
poojav25: Thanks for flagging this — I traced it through the NSX server source to confirm (VpcProviderNsxT.java, the code that builds a VPC's logical router).

When VpcServiceProfile.ipv6_profile_paths has no NDRA reference, NSX sets LrIpv6Profiles = null on the VPC's router, with no fallback to a default profile — so no RA config is pushed to the VPC in that case. This matches T1 in the PR description exactly: ipv6_profile_paths: None → raDeactivated: true, observed live.

There is a seeded default Ipv6NdraProfile (/infra/ipv6-ndra-profiles/default) with ra_mode = SLAAC_DNS_THROUGH_RA, which is likely what you saw. Its only consumer, though, is TransitGatewayProvider.getNdraDefaultProfile(), which applies it to the Transit Gateway/uplink logical router — a separate object from the VPC's own VpcServiceProfile that this code resolves.

pkg/nsx/services/vpc/vpc.go
```diff
@@ -1213,6 +1214,82 @@ func (s *VPCService) GetVPCFromNSXByPath(vpcPath string) (*model.Vpc, error) {
...
+	if err != nil {
+		log.Error(err, "Failed to read VPC object from NSX", "VPC", vpcResInfo.GetVPCPath())
+		return false, err
+	}
+	if vpc.VpcServiceProfile == nil {
+		log.Debug("No VpcServiceProfile configured on VPC, treating RA as deactivated", "VPC", vpcResInfo.GetVPCPath())
+		return true, nil
+	}
+
+	profile, err := s.NSXClient.VpcServiceProfileClient.Get(vpcResInfo.OrgID, vpcResInfo.ProjectID, path.Base(*vpc.VpcServiceProfile))
```
yanjunz97, line 1246: Maybe also add a defense here for `*vpc.VpcServiceProfile != ""`

yanjunz97 (commented):
Overall LGTM, just a nit.

I've checked the default RA mode value when ndra profile path is not set with NSX team and they say that is not fully determined. As NSX does not have an ETA for the final decision, maybe you can create an issue to track so that we can revisit it in the future?

pkg/nsx/services/vpc/vpc.go
poojav25: done

yanjunz97 (approved):
LGTM

pkg/nsx/services/vpc/vpc.go
```diff
@@ -1213,6 +1214,82 @@ func (s *VPCService) GetVPCFromNSXByPath(vpcPath string) (*model.Vpc, error) {
...
+// IsRADeactivatedByVPCPath determines whether IPv6 Router Advertisement is deactivated for the
+// VPC that owns the given NSX path (a VPC path or any path nested under a VPC, e.g. a Subnet
+// path). It resolves the VPC's VpcServiceProfile, finds the referenced Ipv6NdraProfile, and
+// checks its RaMode. If no VpcServiceProfile or no Ipv6NdraProfile is configured, RA is
+// considered deactivated: NSX does not push any ra_config to the dataplane in that case, so no
+// Router Advertisements are actually sent on the wire, which is dataplane-equivalent to an
+// explicit ra_mode of DISABLED. Not cached: an admin can change VPC RA config at any time, and a
+// stale positive would silently mislead every SubnetPort in the VPC until the operator restarts.
+func (s *VPCService) IsRADeactivatedByVPCPath(vpcPath string) (bool, error) {
+	vpcResInfo, err := common.ParseVPCResourcePath(vpcPath)
```
dantingl, line 1226: We should not have a runtime store for VPC and service profile, right? it may be changed in day2

dantingl (approved):

pkg/nsx/services/vpc/vpc.go
poojav25: Yes, exactly! That's why this method calls the NSX API directly.

PR #1515, opened 2026-07-29
Handle SubnetDHCPConfig.Mode or SubnetDHCPv6Config.Mode omitted case …

lxiaopei:
…when in SubnetSet webhook

Root Cause
In pkg/controllers/subnetset/subnetset_webhook.go, method validateSubnets validates that all member Subnets in a SubnetSet have identical DHCPConfigMode and DHCPv6ConfigMode.

However, the webhook was reading raw strings directly from crdSubnet.Spec.SubnetDHCPConfig.Mode and crdSubnet.Spec.SubnetDHCPv6Config.Mode:

In Subnet CRD definition (pkg/apis/vpc/v1alpha1/subnet_types.go), DHCPDeactivated is the default behavior when subnet.Spec.SubnetDHCPConfig.Mode (or SubnetDHCPv6Config.Mode) is omitted / empty (""). subnet-01 did not explicitly specify subnetDHCPConfig.mode in its spec, so its mode string evaluated to "". vlan-subnet-2 explicitly specified subnetDHCPConfig.mode: DHCPDeactivated. validateSubnets compared string "" with "DHCPDeactivated" and failed validation even though both subnets effectively operate in DHCPDeactivated mode.

Fix Applied
1.Helper Functions: Added getEffectiveDHCPMode and getEffectiveDHCPv6Mode in pkg/controllers/subnetset/subnetset_webhook.go to return v1alpha1.DHCPConfigModeDeactivated / v1alpha1.DHCPv6ConfigModeDeactivated when the mode field is empty (""). 2.Webhook Validation: Updated validateSubnets to evaluate effective DHCP modes before checking relay constraints and mode equality across subnets in a SubnetSet. 3.Unit Tests: Added test cases in pkg/controllers/subnetset/subnetset_webhook_test.go to cover combinations of subnets with implicit (empty) and explicit DHCPDeactivated modes for IPv4 and IPv6.

### DIFF
--- a/pkg/controllers/subnetset/subnetset_webhook.go
+++ b/pkg/controllers/subnetset/subnetset_webhook.go
@@ -329,6 +329,20 @@ func getEffectiveStaticIPAllocation(subnet *v1alpha1.Subnet) bool {
 	return util.GetDefaultStaticIPAllocation(subnet)
 }
 
+func getEffectiveDHCPMode(subnet *v1alpha1.Subnet) v1alpha1.DHCPConfigMode {
+	if subnet.Spec.SubnetDHCPConfig.Mode == "" {
+		return v1alpha1.DHCPConfigMode(v1alpha1.DHCPConfigModeDeactivated)
+	}
+	return subnet.Spec.SubnetDHCPConfig.Mode
+}
+
+func getEffectiveDHCPv6Mode(subnet *v1alpha1.Subnet) v1alpha1.DHCPv6ConfigMode {
+	if subnet.Spec.SubnetDHCPv6Config.Mode == "" {
+		return v1alpha1.DHCPv6ConfigModeDeactivated
+	}
+	return subnet.Spec.SubnetDHCPv6Config.Mode
+}
+
 func (v *SubnetSetValidator) validateSubnets(ctx context.Context, ns string, subnetNames *[]string, subnetSet string) (bool, error) {
 	var namespaceVpc string
 	var existingVPC string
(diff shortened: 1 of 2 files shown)


pkg/controllers/subnetset/subnetset_webhook.go
```diff
@@ -346,27 +360,29 @@ func (v *SubnetSetValidator) validateSubnets(ctx context.Context, ns string, sub
 		if err != nil {
 			return false, fmt.Errorf("failed to get Subnet %s/%s: %v", ns, subnetName, err)
 		}
-		if crdSubnet.Spec.SubnetDHCPConfig.Mode == v1alpha1.DHCPConfigMode(v1alpha1.DHCPConfigModeRelay) {
+		dhcpMode := getEffectiveDHCPMode(crdSubnet)
+		if dhcpMode == v1alpha1.DHCPConfigMode(v1alpha1.DHCPConfigModeRelay) {
```
yanjunz97, line 364: No need to convert the type?
```suggestion
		if dhcpMode == v1alpha1.DHCPConfigModeRelay {
```

yanjunz97 (commented):
LGTM, only a nit

pkg/controllers/subnetset/subnetset_webhook.go
lxiaopei: Thanks @yanjunz97 for the quick review.
There is a slight inconsistency in how the constants were originally defined
DHCPConfigModeRelay is explicitly defined as a string.
DHCPv6ConfigModeRelay is explicitly defined as a DHCPv6ConfigMode.

So line 364 we still need type conversion while line 368 no type conversion.

TaoZou1 (approved):

yanjunz97 (approved):

PR #1518, opened 2026-07-31
feat(dns): handle contributing owners tag overflow issue and the race condition when sharing a  single FQDN

wenyingd:
This PR stabilizes shared DNS record reconciliation (used by Gateway API) by fixing concurrency issues that caused NSX updates to overwrite each other, and by handling NSX tag length limits for heavily shared records.

**Key Changes**
1. Fix TOCTOU Race Conditions: 
1.1 Implemented a local lock (recordAccessArbitrator) to serialize concurrent updates to the same DNS record.
1.2 Moved conflict validation inside the lock to ensure reads aren't stale.
1.3 Fixed a bug where the primary owner would mistakenly drop existing contributing_owners during its own reconciliation.
2. Fix Tag Overflow: 
2.1 Introduced an extension tag (nsx-op/dns_contributing_owners_ext) to handle cases where the joined dns_contributing_owners string exceeds the NSX 255-character limit.
3. Corrected misleading "delete resource" logs for PATCH operations.

**Test Done**
1. Prepare 1 gateway and 10 HTTPRoutes, all HTTPRoutes are configured with the same FQDN and refer to the same gateway. Each HTTPRoute.spec.rule is configured with a unique path

2. Apply the manifest and check the status 
```
root@420aec474057587e08ae81c7f6ce8984 [ ~/wenying ]# kubectl -n test1 apply -f gateway-multiple-routes.yaml 
gateway.gateway.networking.k8s.io/gateway created
httproute.gateway.networking.k8s.io/http-1 created
httproute.gateway.networking.k8s.io/http-2 created
httproute.gateway.networking.k8s.io/http-3 created
httproute.gateway.networking.k8s.io/http-4 created
httproute.gateway.networking.k8s.io/http-5 created
httproute.gateway.networking.k8s.io/http-6 created
httproute.gateway.networking.k8s.io/http-7 created
httproute.gateway.networking.k8s.io/http-8 created
httproute.gateway.networking.k8s.io/http-9 created
httproute.gateway.networking.k8s.io/http-10 created

root@420aec474057587e08ae81c7f6ce8984 [ ~/wenying ]# kubectl get gateway,httproute -n test1 -owide
NAME                                        CLASS   ADDRESS       PROGRAMMED   AGE
gateway.gateway.networking.k8s.io/gateway   istio   192.168.0.7   True         32s

NAME                                          HOSTNAMES                 AGE
httproute.gateway.networking.k8s.io/http-1    ["httpbin.example.com"]   32s
httproute.gateway.networking.k8s.io/http-10   ["httpbin.example.com"]   31s
httproute.gateway.networking.k8s.io/http-2    ["httpbin.example.com"]   32s
httproute.gateway.networking.k8s.io/http-3    ["httpbin.example.com"]   31s
httproute.gateway.networking.k8s.io/http-4    ["httpbin.example.com"]   31s
httproute.gateway.networking.k8s.io/http-5    ["httpbin.example.com"]   31s
httproute.gateway.networking.k8s.io/http-6    ["httpbin.example.com"]   31s
httproute.gateway.networking.k8s.io/http-7    ["httpbin.example.com"]   31s
httproute.gateway.networking.k8s.io/http-8    ["httpbin.example.com"]   31s
httproute.gateway.networking.k8s.io/http-9    ["httpbin.example.com"]   31s
```
3. Check all HTTPRoutes are configured with DNSRecordReady==True condition
```
root@420aec474057587e08ae81c7f6ce8984 [ ~/wenying ]# for i in {1..10}; do
  status=$(kubectl get httproute -n test1 "http-$i" -o jsonpath='{.status.parents[*].conditions[?(@.type=="DNSRecordReady")].status}' 2>/dev/null)
  echo "http-$i: ${status:-Not Found / No Status}"
done
http-1: True
http-2: True
http-3: True
http-4: True
http-5: True
http-6: True
http-7: True
http-8: True
http-9: True
http-10: True
```
4. Check NSX DNSRecord before applying this change, a single DNS Record is created, and tags for  `dns_gateway_index_list` and `dns_contributing_owners` are encoded.
```
root@420aec474057587e08ae81c7f6ce8984 [ ~/wenying ]# curl -s -k -u ${cred} https://10.192.49.176/policy/api/v1/orgs/default/projects/project-quality/dns-records | jq '{
  result_count,
  results: [.results[] | {zone_path, record_name, record_type, record_values, tags}]
}'
{
  "result_count": 1,
  "results": [
    {
      "zone_path": "/orgs/default/projects/project-quality/dns-services/default-dns-service/zones/zone1",
      "record_name": "httpbin",
      "record_type": "A",
      "record_values": [
        "192.168.0.7"
      ],
      "tags": [
        {
          "scope": "nsx-op/cluster",
          "tag": "6edf5c92-2967-41a3-9b7c-a3d50fca857b"
        },
        {
          "scope": "nsx-op/version",
          "tag": "1.0.0"
        },
        {
          "scope": "nsx-op/namespace_uid",
          "tag": "c54019ac-89cf-4877-a3df-feadbd4049bd"
        },
        {
          "scope": "nsx-op/dns_for",
          "tag": "httproute"
        },
        {
          "scope": "nsx-op/dns_owner_namespace",
          "tag": "test1"
        },
        {
          "scope": "nsx-op/dns_owner_name",
          "tag": "http-3"
        },
        {
          "scope": "nsx-op/dns_gateway_index_list",
          "tag": "eJwqSS0uMdRPTyxJLU-sBAQAAP__IuYFEw"
        },
        {
          "scope": "nsx-op/dns_contributing_owners",
          "tag": "eJzKKCkpKMovLUnVL0ktLjHUB_F1DQ10sIqbYRc2xy5sAQgAAP__TBgiNw"
        }
      ]
    }
  ]
}
```
5. After applying this change, all HTTPRoutes are shown in the scope "dns_contributing_owners" with the plaintext, and the dns_gateway_index_list is also shown as plaintext
```
root@420aec474057587e08ae81c7f6ce8984 [ ~ ]# curl -s -k -u ${cred} https://10.192.49.176/policy/api/v1/orgs/default/projects/project-quality/dns-records | jq '{
  result_count,
  results: [.results[] | {zone_path, record_name, record_type, record_values, tags}]
}'
{
  "result_count": 1,
  "results": [
    {
      "zone_path": "/orgs/default/projects/project-quality/dns-services/default-dns-service/zones/zone1",
      "record_name": "httpbin",
      "record_type": "A",
      "record_values": [
        "192.168.0.7"
      ],
      "tags": [
        {
          "scope": "nsx-op/cluster",
          "tag": "6edf5c92-2967-41a3-9b7c-a3d50fca857b"
        },
        {
          "scope": "nsx-op/version",
          "tag": "1.0.0"
        },
        {
          "scope": "nsx-op/namespace_uid",
          "tag": "c54019ac-89cf-4877-a3df-feadbd4049bd"
        },
        {
          "scope": "nsx-op/dns_for",
          "tag": "httproute"
        },
        {
          "scope": "nsx-op/dns_owner_namespace",
          "tag": "test1"
        },
        {
          "scope": "nsx-op/dns_owner_name",
          "tag": "http-7"
        },
        {
          "scope": "nsx-op/dns_gateway_index_list",
          "tag": "test1/gateway"
        },
        {
          "scope": "nsx-op/dns_contributing_owners",
          "tag": "httproute/test1/http-1,httproute/test1/http-10,httproute/test1/http-2,httproute/test1/http-3,httproute/test1/http-4,httproute/test1/http-5,httproute/test1/http-6,httproute/test1/http-8,httproute/test1/http-9"
        }
      ]
    }
  ]
}
```
6. Add another 10 HTTPRoutes with the same FQDN and referring to the same parent Gateway,
```
root@420aec474057587e08ae81c7f6ce8984 [ ~/wenying ]# kubectl get gateway,httproute -n test1 -owide
NAME                                        CLASS   ADDRESS       PROGRAMMED   AGE
gateway.gateway.networking.k8s.io/gateway   istio   192.168.0.7   True         3h57m

NAME                                          HOSTNAMES                 AGE
httproute.gateway.networking.k8s.io/http-1    ["httpbin.example.com"]   3h57m
httproute.gateway.networking.k8s.io/http-10   ["httpbin.example.com"]   3h57m
httproute.gateway.networking.k8s.io/http-11   ["httpbin.example.com"]   33s
httproute.gateway.networking.k8s.io/http-12   ["httpbin.example.com"]   33s
httproute.gateway.networking.k8s.io/http-13   ["httpbin.example.com"]   33s
httproute.gateway.networking.k8s.io/http-14   ["httpbin.example.com"]   33s
httproute.gateway.networking.k8s.io/http-15   ["httpbin.example.com"]   33s
httproute.gateway.networking.k8s.io/http-16   ["httpbin.example.com"]   32s
httproute.gateway.networking.k8s.io/http-17   ["httpbin.example.com"]   32s
httproute.gateway.networking.k8s.io/http-18   ["httpbin.example.com"]   32s
httproute.gateway.networking.k8s.io/http-19   ["httpbin.example.com"]   32s
httproute.gateway.networking.k8s.io/http-2    ["httpbin.example.com"]   3h57m
httproute.gateway.networking.k8s.io/http-20   ["httpbin.example.com"]   32s
httproute.gateway.networking.k8s.io/http-3    ["httpbin.example.com"]   3h57m
httproute.gateway.networking.k8s.io/http-4    ["httpbin.example.com"]   3h57m
httproute.gateway.networking.k8s.io/http-5    ["httpbin.example.com"]   3h57m
httproute.gateway.networking.k8s.io/http-6    ["httpbin.example.com"]   3h57m
httproute.gateway.networking.k8s.io/http-7    ["httpbin.example.com"]   3h57m
httproute.gateway.networking.k8s.io/http-8    ["httpbin.example.com"]   3h57m
httproute.gateway.networking.k8s.io/http-9    ["httpbin.example.com"]   3h57m

root@420aec474057587e08ae81c7f6ce8984 [ ~/wenying ]# for i in {1..20}; do
  status=$(kubectl get httproute -n test1 "http-$i" -o jsonpath='{.status.parents[*].conditions[?(@.type=="DNSRecordReady")].status}' 2>/dev/null)
  echo "http-$i: ${status:-Not Found / No Status}"
done
http-1: True
http-2: True
http-3: True
http-4: True
http-5: True
http-6: True
http-7: True
http-8: True
http-9: True
http-10: True
http-11: True
http-12: True
http-13: True
http-14: True
http-15: True
http-16: True
http-17: True
http-18: True
http-19: True
http-20: True
```
7. Check the NSX DNSRecord resource, a single resource exists, and tag "dns_contributing_owners_ext" is updated.
```
root@420aec474057587e08ae81c7f6ce8984 [ ~/wenying ]#  curl -s -k -u ${cred} https://10.192.49.176/policy/api/v1/orgs/default/projects/project-quality/dns-records | jq '{
  result_count,
  results: [.results[] | {zone_path, record_name, record_type, record_values, tags}]
}'
{
  "result_count": 1,
  "results": [
    {
      "zone_path": "/orgs/default/projects/project-quality/dns-services/default-dns-service/zones/zone1",
      "record_name": "httpbin",
      "record_type": "A",
      "record_values": [
        "192.168.0.7"
      ],
      "tags": [
        {
          "scope": "nsx-op/cluster",
          "tag": "6edf5c92-2967-41a3-9b7c-a3d50fca857b"
        },
        {
          "scope": "nsx-op/version",
          "tag": "1.0.0"
        },
        {
          "scope": "nsx-op/namespace_uid",
          "tag": "c54019ac-89cf-4877-a3df-feadbd4049bd"
        },
        {
          "scope": "nsx-op/dns_for",
          "tag": "httproute"
        },
        {
          "scope": "nsx-op/dns_owner_namespace",
          "tag": "test1"
        },
        {
          "scope": "nsx-op/dns_owner_name",
          "tag": "http-7"
        },
        {
          "scope": "nsx-op/dns_gateway_index_list",
          "tag": "test1/gateway"
        },
        {
          "scope": "nsx-op/dns_contributing_owners",
          "tag": "httproute/test1/http-1,httproute/test1/http-10,httproute/test1/http-11,httproute/test1/http-12,httproute/test1/http-13,httproute/test1/http-14,httproute/test1/http-15,httproute/test1/http-16,httproute/test1/http-17,httproute/test1/http-18"
        },
        {
          "scope": "nsx-op/dns_contributing_owners_ext",
          "tag": "httproute/test1/http-19,httproute/test1/http-2,httproute/test1/http-20,httproute/test1/http-3,httproute/test1/http-4,httproute/test1/http-5,httproute/test1/http-6,httproute/test1/http-8,httproute/test1/http-9"
        }
      ]
    }
  ]
}
```
8. Delete 6 HTTPRoutes
```
root@420aec474057587e08ae81c7f6ce8984 [ ~/wenying ]# kubectl -n test1 delete httproute http-11 http-12 http-13 http-14 http-15 http-20
httproute.gateway.networking.k8s.io "http-11" deleted from test1 namespace
httproute.gateway.networking.k8s.io "http-12" deleted from test1 namespace
httproute.gateway.networking.k8s.io "http-13" deleted from test1 namespace
httproute.gateway.networking.k8s.io "http-14" deleted from test1 namespace
httproute.gateway.networking.k8s.io "http-15" deleted from test1 namespace
httproute.gateway.networking.k8s.io "http-20" deleted from test1 namespace

root@420aec474057587e08ae81c7f6ce8984 [ ~/wenying ]# for i in {1..20}; do   status=$(kubectl get httproute -n test1 "http-$i" -o jsonpath='{.status.parents[*].conditions[?(@.type=="DNSRecordReady")].
status}' 2>/dev/null);   echo "http-$i: ${status:-Not Found / No Status}"; done
http-1: True
http-2: True
http-3: True
http-4: True
http-5: True
http-6: True
http-7: True
http-8: True
http-9: True
http-10: True
http-11: Not Found / No Status
http-12: Not Found / No Status
http-13: Not Found / No Status
http-14: Not Found / No Status
http-15: Not Found / No Status
http-16: True
http-17: True
http-18: True
http-19: True
http-20: Not Found / No Status
```
9. Check NSX DNSRecord status
```
root@420aec474057587e08ae81c7f6ce8984 [ ~/wenying ]#  curl -s -k -u ${cred} https://10.192.49.176/policy/api/v1/orgs/default/projects/project-quality/dns-records | jq '{
  result_count,
  results: [.results[] | {zone_path, record_name, record_type, record_values, tags, _revision}]
}'
{
  "result_count": 1,
  "results": [
    {
      "zone_path": "/orgs/default/projects/project-quality/dns-services/default-dns-service/zones/zone1",
      "record_name": "httpbin",
      "record_type": "A",
      "record_values": [
        "192.168.0.7"
      ],
      "tags": [
        {
          "scope": "nsx-op/cluster",
          "tag": "6edf5c92-2967-41a3-9b7c-a3d50fca857b"
        },
        {
          "scope": "nsx-op/version",
          "tag": "1.0.0"
        },
        {
          "scope": "nsx-op/namespace_uid",
          "tag": "c54019ac-89cf-4877-a3df-feadbd4049bd"
        },
        {
          "scope": "nsx-op/dns_for",
          "tag": "httproute"
        },
        {
          "scope": "nsx-op/dns_owner_namespace",
          "tag": "test1"
        },
        {
          "scope": "nsx-op/dns_owner_name",
          "tag": "http-7"
        },
        {
          "scope": "nsx-op/dns_gateway_index_list",
          "tag": "test1/gateway"
        },
        {
          "scope": "nsx-op/dns_contributing_owners",
          "tag": "httproute/test1/http-1,httproute/test1/http-10,httproute/test1/http-16,httproute/test1/http-17,httproute/test1/http-18,httproute/test1/http-19,httproute/test1/http-2,httproute/test1/http-3,httproute/test1/http-4,httproute/test1/http-5"
        },
        {
          "scope": "nsx-op/dns_contributing_owners_ext",
          "tag": "httproute/test1/http-6,httproute/test1/http-8,httproute/test1/http-9"
        }
      ],
      "_revision": 91
    }
  ]
}
```
10. Add more HTTPRoute to 30 resources with the same hostname, checking the case about truncating the tag values
```
root@422a2cc5036c07e3ed894483ee25f2fb [ ~/wenying/dns_enhancements ]# for i in {1..30}; do   status=$(kubectl get httproute -n test1 "http-$i" -o jsonpath='{.status.parents[*].conditions[?(@.type=="DNSRecordReady")].status}' 2>/dev/null);   echo "http-$i: ${status:-Not Found / No Status}"; done
http-1: True
http-2: True
http-3: True
http-4: True
http-5: True
http-6: True
http-7: True
http-8: True
http-9: True
http-10: True
http-11: True
http-12: True
http-13: True
http-14: True
http-15: True
http-16: True
http-17: True
http-18: True
http-19: True
http-20: True
http-21: False
http-22: True
http-23: False
http-24: False
http-25: False
http-26: False
http-27: False
http-28: False
http-29: False
http-30: False

# kubectl -n test1 get httproute http-21 -ojsonpath='{.status.parents[*].conditions[?(@.type=="DNSRecordReady")]}'
{"lastTransitionTime":"2026-08-28T04:46:03Z","message":"DNS endpoint validation failed during record build: contributing owners count exceeds maximum tag capacity","observedGeneration":1,"reason":"DNSRecordFailed","status":"False","type":"DNSRecordReady"
```
Check NSX DNSRecord
```
root@422a2cc5036c07e3ed894483ee25f2fb [ ~/wenying/dns_enhancements ]#  curl -s -k -u ${cred} https://192.163.128.16/policy/api/v1/orgs/default/projects/project-quality/dns-records | jq '{
  result_count,
  results: [.results[] | {zone_path, record_name, record_type, record_values, tags}]
}'
{
  "result_count": 1,
  "results": [
    {
      "zone_path": "/orgs/default/projects/project-quality/dns-services/default-dns-service/zones/zone1",
      "record_name": "httpbin",
      "record_type": "A",
      "record_values": [
        "192.168.0.10"
      ],
      "tags": [
        {
          "scope": "nsx-op/cluster",
          "tag": "454d4330-56fb-444f-b6df-7977d426689d"
        },
        {
          "scope": "nsx-op/version",
          "tag": "1.0.0"
        },
        {
          "scope": "nsx-op/namespace_uid",
          "tag": "7e8e3b23-59a3-4f03-b040-9effd03fa916"
        },
        {
          "scope": "nsx-op/dns_for",
          "tag": "httproute"
        },
        {
          "scope": "nsx-op/dns_owner_namespace",
          "tag": "test-1"
        },
        {
          "scope": "nsx-op/dns_owner_name",
          "tag": "http-5"
        },
        {
          "scope": "nsx-op/dns_gateway_index_list",
          "tag": "test-1/gateway"
        },
        {
          "scope": "nsx-op/dns_contributing_owners",
          "tag": "httproute/test-1/http-1,httproute/test-1/http-10,httproute/test-1/http-11,httproute/test-1/http-12,httproute/test-1/http-13,httproute/test-1/http-14,httproute/test-1/http-15,httproute/test-1/http-16,httproute/test-1/http-17,httproute/test-1/http-18"
        },
        {
          "scope": "nsx-op/dns_contributing_owners_ext",
          "tag": "httproute/test-1/http-19,httproute/test-1/http-2,httproute/test-1/http-20,httproute/test-1/http-22,httproute/test-1/http-3,httproute/test-1/http-4,httproute/test-1/http-6,httproute/test-1/http-7,httproute/test-1/http-8,httproute/test-1/http-9"
        }
      ]
    }
  ]
}
```
11. Reduce the number of HTTPRoute resources to 5, checking a single tag is sufficient for all HTTPRoutes per FQDN
```
# for i in {6..30}; do
  kubectl -n test1 delete httproute "http-$i"
done
httproute.gateway.networking.k8s.io "http-6" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-7" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-8" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-9" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-10" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-11" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-12" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-13" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-14" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-15" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-16" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-17" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-18" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-19" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-20" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-21" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-22" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-23" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-24" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-25" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-26" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-27" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-28" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-29" deleted from test-1 namespace
httproute.gateway.networking.k8s.io "http-30" deleted from test-1 namespace

# for i in {1..5}; do   status=$(kubectl get httproute -n test1 "http-$i" -o jsonpath='{.status.parents[*].conditions[?(@.type=="DNSRecordReady")].status}' 2>/dev/null);
   echo "http-$i: ${status:-Not Found / No Status}"; done
http-1: True
http-2: True
http-3: True
http-4: True
http-5: True
```
Check NSX DNSRecord status,
```
#  curl -s -k -u ${cred} https://192.163.128.16/policy/api/v1/orgs/default/projects/project-quality/dns-records | jq '{
  result_count,
  results: [.results[] | {zone_path, record_name, record_type, record_values, tags}]
}'
{
  "result_count": 1,
  "results": [
    {
      "zone_path": "/orgs/default/projects/project-quality/dns-services/default-dns-service/zones/zone1",
      "record_name": "httpbin",
      "record_type": "A",
      "record_values": [
        "192.168.0.10"
      ],
      "tags": [
        {
          "scope": "nsx-op/cluster",
          "tag": "454d4330-56fb-444f-b6df-7977d426689d"
        },
        {
          "scope": "nsx-op/version",
          "tag": "1.0.0"
        },
        {
          "scope": "nsx-op/namespace_uid",
          "tag": "7e8e3b23-59a3-4f03-b040-9effd03fa916"
        },
        {
          "scope": "nsx-op/dns_for",
          "tag": "httproute"
        },
        {
          "scope": "nsx-op/dns_owner_namespace",
          "tag": "test-1"
        },
        {
          "scope": "nsx-op/dns_owner_name",
          "tag": "http-5"
        },
        {
          "scope": "nsx-op/dns_gateway_index_list",
          "tag": "test-1/gateway"
        },
        {
          "scope": "nsx-op/dns_contributing_owners",
          "tag": "httproute/test-1/http-1,httproute/test-1/http-2,httproute/test-1/http-3,httproute/test-1/http-4"
        }
      ]
    }
  ]
}
```

### DIFF
--- a/pkg/nsx/services/dns/builder.go
+++ b/pkg/nsx/services/dns/builder.go
@@ -103,7 +103,7 @@ func (r *EndpointRow) appendRowOwnershipTags(ownerTags []model.Tag) []model.Tag 
 	if r.Endpoint != nil && r.Endpoint.Labels != nil {
 		gwKeys := strings.TrimSpace(r.Endpoint.Labels[EndpointLabelParentGateway])
 		if len(gwKeys) > 0 {
-			gwKey = compressString(gwKeys)
+			gwKey, _ = joinAndPackStrings(strings.Split(gwKeys, ","))
 		}
 	}
 	return appendGatewayAndContributionTags(tags, gwKey, r.contributingOwnerKeys)
--- a/pkg/nsx/services/dns/builder.go
+++ b/pkg/nsx/services/dns/builder.go
@@ -17,7 +17,7 @@ const (
 )
 
 // BuildDnsRecord builds one *model.DnsRecord for row using batchOwner or row.effectiveOwner.
-func (s *DNSRecordService) BuildDnsRecord(batchOwner *ResourceRef, row EndpointRow) *model.DnsRecord {
+func (s *DNSRecordService) BuildDnsRecord(batchOwner *ResourceRef, row EndpointRow) (*model.DnsRecord, error) {
 	owner := batchOwner
 	if row.effectiveOwner != nil {
 		owner = row.effectiveOwner
@@ -54,9 +54,12 @@ func getRecordIDAndPathAndType(recordName, endpointRecordType, zonePath string) 
 	return recID, recordPath, nsxRecordType
 }
 
-func (r *EndpointRow) buildDNSRecord(basicTags []model.Tag) *model.DnsRecord {
+func (r *EndpointRow) buildDNSRecord(basicTags []model.Tag) (*model.DnsRecord, error) {
 	// Append the tags according to the Endpoint labels, e.g., the parent gateway settings for a Route.
-	tags := r.appendRowOwnershipTags(basicTags)
+	tags, err := r.appendRowOwnershipTags(basicTags)
+	if err != nil {
+		return nil, err
+	}
 	recID, path, rt := getRecordIDAndPathAndType(r.nsxRecordName, r.RecordType, r.zonePath)
 	ttl := int64(DefaultRecordTtL)
 	if r.Endpoint.RecordTTL.IsConfigured() {
@@ -76,7 +79,7 @@ func (r *EndpointRow) buildDNSRecord(basicTags []model.Tag) *model.DnsRecord {
 		// Mirror logical FQDN for store indexing / conflict detection; stripped before Policy PATCH (see WrapDnsRecord).
 		Fqdn: common.String(strings.ToLower(r.Endpoint.DNSName)),
 	}
-	return rec
+	return rec, nil
 }
 
 func getNSXDnsRecordType(recType string) string {
@@ -97,13 +100,18 @@ func getNSXDnsRecordType(recType string) string {
 	}
 }
 
-func (r *EndpointRow) appendRowOwnershipTags(ownerTags []model.Tag) []model.Tag {
+func (r *EndpointRow) appendRowOwnershipTags(ownerTags []model.Tag) ([]model.Tag, error) {
 	tags := append([]model.Tag{}, ownerTags...)
 	gwKey := ""
 	if r.Endpoint != nil && r.Endpoint.Labels != nil {
 		gwKeys := strings.TrimSpace(r.Endpoint.Labels[EndpointLabelParentGateway])
 		if len(gwKeys) > 0 {
-			gwKey, _ = joinAndPackStrings(strings.Split(gwKeys, ","))
+			var overflow string
+			var truncated bool
+			gwKey, overflow, truncated = joinAndPackStrings(strings.Split(gwKeys, ","))
+			if truncated || overflow != "" {
+				return nil, fmt.Errorf("gateway index list exceeds maximum tag capacity")
+			}
 		}
 	}
 	return appendGatewayAndContributionTags(tags, gwKey, r.contributingOwnerKeys)
--- a/pkg/nsx/services/dns/contributing.go
+++ b/pkg/nsx/services/dns/contributing.go
@@ -4,10 +4,7 @@
 package dns
 
 import (
-	"bytes"
-	"compress/zlib"
-	"encoding/base64"
-	"io"
+	"slices"
 	"strings"
 
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
@@ -18,7 +15,18 @@ import (
 )
 
 func parseContributingOwnersFromRecord(rec *model.DnsRecord) []string {
-	return parseContributingOwnersTag(decompressContributingTags(rec))
+	tagKeys := parseContributingOwnersTag(decompressContributingTags(rec))
+
+	var allKeys []string
+	allKeys = append(allKeys, tagKeys...)
+
+	additionalTag := firstTagValue(rec.Tags, common.TagScopeDNSRecordAdditionalContributingOwners)
+	if additionalTag != "" {
+		additionalKeys := parseContributingOwnersTag(additionalTag)
+		allKeys = append(allKeys, additionalKeys...)
+	}
+
+	return sortedCopyStrings(allKeys)
 }
 
 func decompressContributingTags(rec *model.DnsRecord) string {
@@ -26,7 +34,7 @@ func decompressContributingTags(rec *model.DnsRecord) string {
 	if encodedContributingKeys == "" {
 		return ""
 	}
-	return decompressString(encodedContributingKeys)
+	return encodedContributingKeys
 }
 
 func parseContributingOwnersTag(raw string) []string {
@@ -45,30 +53,31 @@ func parseContributingOwnersTag(raw string) []string {
 	return out
 }
 
-func formatContributingOwnersTag(keys []string) string {
+func formatContributingOwnersTag(keys []string) (string, string) {
 	if len(keys) == 0 {
-		return ""
+		return "", ""
 	}
-	cp := sortedCopyStrings(keys)
-	plainText := strings.Join(cp, ",")
-	// Compress the consolidated plain text into a tight string under 255 characters.
-	return compressString(plainText)
+	contributionTag, extraTag := joinAndPackStrings(keys)
+	if len(extraTag) > 0 {
+		log.Info("Tag length exceeds dns_contributing_owners limit (255), putting remaining in dns_contributing_owners_ext")
+	}
+	return contributionTag, extraTag
 }
 
 // mergeContributingOwnerKeys returns sorted unique contributing keys (excludes primaryNNKey).
-func mergeContributingOwnerKeys(existing string, add string, primaryNNKey string) string {
-	raw := strings.Join([]string{existing, add}, ",")
+func mergeContributingOwnerKeys(existing []string, add string, primaryNNKey string) []string {
 	seen := sets.New[string]()
-	for _, p := range strings.Split(raw, ",") {
-		k := strings.TrimSpace(p)
-		if k == "" || k == primaryNNKey {
-			continue
-		}
-		if k != "" {
+	for _, k := range existing {
+		k = strings.TrimSpace(k)
+		if k != "" && k != primaryNNKey {
 			seen.Insert(k)
 		}
 	}
-	return formatContributingOwnersTag(seen.UnsortedList())
+	k := strings.TrimSpace(add)
+	if k != "" && k != primaryNNKey {
+		seen.Insert(k)
+	}
+	return sortedCopyStrings(seen.UnsortedList())
 }
 
 func resourceRefFromDNSRecord(rec *model.DnsRecord) (*ResourceRef, bool) {
@@ -111,19 +120,21 @@ func ownerNNIndexKeyForResourceRef(owner *ResourceRef) string {
 	return dnsRecordOwnerKey(createdFor, dnsRecordOwnerNamespacedNameKey(owner.GetNamespace(), owner.GetName()))
 }
 
-func primaryOwnerNNIndexKeyFromRecord(rec *model.DnsRecord) string {
-	return getDNSRecordOwnerNamespacedName(rec)
-}
-
 // appendGatewayAndContributionTags appends the optional GatewayIndexList and ContributingOwners tags
 // onto tags when their values are non-empty, returning the extended slice. The caller is
 // responsible for passing a slice it owns so this function may append to it directly.
-func appendGatewayAndContributionTags(tags []model.Tag, gwKey string, ctag string) []model.Tag {
+func appendGatewayAndContributionTags(tags []model.Tag, gwKey string, contribKeys []string) []model.Tag {
 	if gwKey = strings.TrimSpace(gwKey); gwKey != "" {
 		tags = append(tags, modelTag(common.TagScopeDNSRecordGatewayIndexList, gwKey))
 	}
-	if ctag != "" {
-		tags = append(tags, modelTag(common.TagScopeDNSRecordContributingOwners, ctag))
+	if len(contribKeys) > 0 {
+		plain, overflow := formatContributingOwnersTag(contribKeys)
+		if plain != "" {
+			tags = append(tags, modelTag(common.TagScopeDNSRecordContributingOwners, plain))
+		}
+		if overflow != "" {
+			tags = append(tags, modelTag(common.TagScopeDNSRecordAdditionalContributingOwners, overflow))
+		}
 	}
 	return tags
 }
@@ -134,56 +145,72 @@ func replaceContributingOwnersInTags(tags []model.Tag, newContribKeys []string) 
 		if t.Scope == nil {
 			continue
 		}
-		if *t.Scope != common.TagScopeDNSRecordContributingOwners {
-			out = append(out, t)
+		if *t.Scope == common.TagScopeDNSRecordContributingOwners || *t.Scope == common.TagScopeDNSRecordAdditionalContributingOwners {
 			continue
 		}
-		if len(newContribKeys) > 0 {
-			out = append(out, modelTag(common.TagScopeDNSRecordContributingOwners, formatContributingOwnersTag(newContribKeys)))
+		out = append(out, t)
+	}
+	if len(newContribKeys) > 0 {
+		plain, overflow := formatContributingOwnersTag(newContribKeys)
+		if plain != "" {
+			out = append(out, modelTag(common.TagScopeDNSRecordContributingOwners, plain))
+		}
+		if overflow != "" {
+			out = append(out, modelTag(common.TagScopeDNSRecordAdditionalContributingOwners, overflow))
 		}
 	}
 	return out
 }
 
-// compressString compresses plain-text keys using zlib and encodes the result into RawURLEncoding Base64.
-func compressString(src string) string {
-	if src == "" {
-		return ""
-	}
-	var b bytes.Buffer
-	w := zlib.NewWriter(&b)
-	_, _ = w.Write([]byte(src))
-	_ = w.Close()
-
-	// RawURLEncoding is used to strip trailing padding characters ('='), saving extra bytes.
-	return base64.RawURLEncoding.EncodeToString(b.Bytes())
-}
-
-// decompressString decodes the Base64 payload and inflates the zlib stream back to plain text.
-func decompressString(compressed string) string {
-	if compressed == "" || !strings.HasPrefix(compressed, "eJ") {
-		return compressed
+// joinAndPackStrings sorts and joins src strings into two comma-separated
+// chunks (primary and overflow), each strictly bounded to 255 bytes.
+func joinAndPackStrings(src []string) (string, string) {
+	if len(src) == 0 {
+		return "", ""
 	}
 
-	data, err := base64.RawURLEncoding.DecodeString(compressed)
-	if err != nil {
-		return compressed // Fallback to raw string if Base64 decoding fails.
+	sorted := slices.Clone(src)
+	slices.Sort(sorted)
+
+	var sb, overflowSB strings.Builder
+	sb.Grow(255)
+	overflowSB.Grow(255)
+
+	// push attempts to append an item to the specified builder (truncating single oversized items).
+	// It returns true if the item was successfully appended within the 255-byte limit.
+	push := func(b *strings.Builder, item string) bool {
+		if len(item) > 255 {
+			item = item[:255]
+		}
+		needComma := b.Len() > 0
+		addedLen := len(item)
+		if needComma {
+			addedLen++
+		}
+
+		if b.Len()+addedLen > 255 {
+			return false
+		}
+		if needComma {
+			b.WriteString(",")
+		}
+		b.WriteString(item)
+		return true
 	}
 
-	b := bytes.NewReader(data)
-	r, err := zlib.NewReader(b)
-	if err != nil {
-		return compressed // Fallback to raw string if zlib reader initialization fails.
+	truncated := false
+	for _, item := range sorted {
+		if push(&sb, item) {
+			continue
+		}
+		if !push(&overflowSB, item) {
+			truncated = true
+		}
 	}
-	defer r.Close()
 
-	var out bytes.Buffer
-	maxDecompressedSize := int64(1024 * 1024)
-	limitedReader := io.LimitReader(r, maxDecompressedSize+1)
-	_, err = io.Copy(&out, limitedReader)
-	if err != nil {
-		log.Error(err, "Failed to decompress the string, returning the compressed value", "value", compressed)
-		return compressed
+	if truncated {
+		log.Info("Tag length exceeds the extra tag's limit (255), truncating additional inputs")
 	}
-	return out.String()
+
+	return sb.String(), overflowSB.String()
 }
--- a/pkg/nsx/services/dns/contributing.go
+++ b/pkg/nsx/services/dns/contributing.go
@@ -4,6 +4,7 @@
 package dns
 
 import (
+	"fmt"
 	"slices"
 	"strings"
 
@@ -53,15 +54,18 @@ func parseContributingOwnersTag(raw string) []string {
 	return out
 }
 
-func formatContributingOwnersTag(keys []string) (string, string) {
+func formatContributingOwnersTag(keys []string) (string, string, error) {
 	if len(keys) == 0 {
-		return "", ""
+		return "", "", nil
 	}
-	contributionTag, extraTag := joinAndPackStrings(keys)
+	contributionTag, extraTag, truncated := joinAndPackStrings(keys)
+	if truncated {
+		return "", "", fmt.Errorf("contributing owners count exceeds maximum tag capacity")
+	}
 	if len(extraTag) > 0 {
 		log.Info("Tag length exceeds dns_contributing_owners limit (255), putting remaining in dns_contributing_owners_ext")
 	}
-	return contributionTag, extraTag
+	return contributionTag, extraTag, nil
 }
 
 // mergeContributingOwnerKeys returns sorted unique contributing keys (excludes primaryNNKey).
@@ -123,12 +127,15 @@ func ownerNNIndexKeyForResourceRef(owner *ResourceRef) string {
 // appendGatewayAndContributionTags appends the optional GatewayIndexList and ContributingOwners tags
 // onto tags when their values are non-empty, returning the extended slice. The caller is
 // responsible for passing a slice it owns so this function may append to it directly.
-func appendGatewayAndContributionTags(tags []model.Tag, gwKey string, contribKeys []string) []model.Tag {
+func appendGatewayAndContributionTags(tags []model.Tag, gwKey string, contribKeys []string) ([]model.Tag, error) {
 	if gwKey = strings.TrimSpace(gwKey); gwKey != "" {
 		tags = append(tags, modelTag(common.TagScopeDNSRecordGatewayIndexList, gwKey))
 	}
 	if len(contribKeys) > 0 {
-		plain, overflow := formatContributingOwnersTag(contribKeys)
+		plain, overflow, err := formatContributingOwnersTag(contribKeys)
+		if err != nil {
+			return nil, err
+		}
 		if plain != "" {
 			tags = append(tags, modelTag(common.TagScopeDNSRecordContributingOwners, plain))
 		}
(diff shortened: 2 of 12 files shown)


pkg/nsx/services/dns/recordservice.go
```diff
@@ -72,24 +111,70 @@ type DNSRecordService struct {
...
 	if err != nil {
 		return false, err
 	}
+
 	if len(toUpsert) == 0 && len(toRemove) == 0 {
 		return false, nil
 	}
+
 	toApply, syncErr := s.syncDnsRecordsInNSX(ctx, toUpsert, toRemove)
+	if syncErr != nil {
```
heypnus: If `toApply` and `syncError` are both non-empty, then the local store won't be synced and there will be a mismatch between local store and NSX, right? Maybe we need to apply `toApply` before propagating `syncErr`. Do you know if this is a possible case?

wenyingd: Thanks for the review.

> Do you know if this is a possible case?

In theory, yes, it is possible. Although DNSRecords resources are sync with HAPI, it is still possible that partial of the records are succeeded and others are realized failed.  Updated as suggested.

pkg/nsx/services/dns/contributing.go
```diff
@@ -134,56 +145,72 @@ func replaceContributingOwnersInTags(tags []model.Tag, newContribKeys []string)
...
+	sb.Grow(255)
+	overflowSB.Grow(255)
 
-	data, err := base64.RawURLEncoding.DecodeString(compressed)
-	if err != nil {
-		return compressed // Fallback to raw string if Base64 decoding fails.
+	// push attempts to append an item to the specified builder (truncating single oversized items).
+	// It returns true if the item was successfully appended within the 255-byte limit.
+	push := func(b *strings.Builder, item string) bool {
+		if len(item) > 255 {
```
heypnus, line 192: Could we avoid truncating or dropping owner keys here? These values are used later to look up contributors and decide whether a shared record can be removed. A valid kind/namespace/name key can already be longer than 255 bytes, and any owners beyond the two-tag capacity are currently left out without returning an error. That could make an existing owner disappear from the index. Would it be safer to keep the representation lossless, or return an error when all owners cannot be stored?

pkg/nsx/services/dns/recordservice.go
```diff
@@ -231,7 +343,14 @@ func (s *DNSRecordService) applyDNSUpsertRows(batch *AggregatedDNSEndpoints) ([]
 
 	desiredRecs := make([]*model.DnsRecord, 0, len(batch.Rows))
 	for _, row := range batch.Rows {
-		if rec := s.BuildDnsRecord(batch.Owner, row); rec != nil {
+		// Re-evaluate the row against the current store state while under the local lock
+		// to prevent overwriting existing shared records if the initial validation used a stale cache.
+		freshRow, err := s.validateEndpointRowConflict(row.zonePath, row.Endpoint, row.nsxRecordName, batch.Owner)
```
heypnus, line 437: Could we keep processing the other rows when one FQDN conflicts? Returning here stops the batch before compareRecords, so valid rows are not applied and stale records are not cleaned up. Also, because the validation error now comes back from CreateOrUpdateRecords, the callers treat it as an operational failure and keep retrying it. Maybe we could collect the conflict, continue with the valid rows, and preserve the previous non-retryable validation behavior.

pkg/nsx/services/dns/builder.go
```diff
@@ -103,7 +103,7 @@ func (r *EndpointRow) appendRowOwnershipTags(ownerTags []model.Tag) []model.Tag
 	if r.Endpoint != nil && r.Endpoint.Labels != nil {
 		gwKeys := strings.TrimSpace(r.Endpoint.Labels[EndpointLabelParentGateway])
 		if len(gwKeys) > 0 {
-			gwKey = compressString(gwKeys)
+			gwKey, _ = joinAndPackStrings(strings.Split(gwKeys, ","))
```
heypnus: It looks like the overflow returned by joinAndPackStrings is ignored here. If the parent Gateway list is longer than 255 bytes, those extra Gateway references will not be added to the reverse index, so the GC fallback may miss them. Could we either store the overflow as well or return an error instead of silently dropping it?

pkg/nsx/services/dns/recordservice.go
```diff
@@ -72,16 +111,58 @@ type DNSRecordService struct {
...
+func (s *DNSRecordService) getInvolvedRecordLockKeys(batch *AggregatedDNSEndpoints) (lockKeys []string) {
+	involvedLockKeys := make(map[string]struct{})
+
+	if batch.Owner != nil {
+		for _, row := range batch.Rows {
+			recID, _, recType := getRecordIDAndPathAndType(row.nsxRecordName, row.RecordType, row.zonePath)
+			involvedLockKeys[getLockKey(row.zonePath, recID, recType)] = struct{}{}
+		}
+
+		_, ownedRecs := s.collectRecordsByOwner(batch.Owner.Kind, batch.Owner.GetNamespace(), batch.Owner.GetName())
```
heypnus, line 126: I wonder if we should re-check the owner’s record set after acquiring the locks. We collect ownedRecs before waiting, but applyDNSUpsertRows reads them again afterward. If another reconcile moves the owner from R1 to R2 in between, this call may update R2 while only holding R1’s lock. DeleteRecordByOwnerNN seems to have the same window. An owner-level lock, or retrying when the locked set has changed, might make this safer.

heypnus (commented):
I'm not an expert in this field so the comments are mostly AI-generated. Please check if they are reasonable.

pkg/nsx/services/dns/recordservice.go
wenyingd: It is designed to try to manage the "batch" operations in one transaction, the previous logic is also to return error once the conflict is detected. The current implementation is consistent with the design.

pkg/nsx/services/dns/builder.go
wenyingd: An error is returned when the keys are truncated.

pkg/nsx/services/dns/recordservice.go
wenyingd: Introducing the owner-level lock in the latest change.

pkg/nsx/services/dns/contributing.go
wenyingd: An error is returned when the string is truncated.

TaoZou1:
A quick suggestion regarding the Test Done section:Overflow Cleanup Case: Currently, step 6 deletes 6 HTTPRoutes (leaving 14), so the dns_contributing_owners_ext tag is still present (just shrank). Could you add a quick check showing that deleting further (e.g., down to 5 routes, $<255$ bytes) completely removes/cleans up the ext tag from NSX?Capacity Exceeded (Error Handling) Case: The test done covers the 1-tag to 2-tag transition, but doesn't show what happens when total owners exceed both tags ($>510$ bytes). It would be great to show the error logging/handling when the capacity limit is hit.

TaoZou1 (dismissed):

pkg/nsx/services/dns/contributing.go
heypnus: Thanks. I think there’s still one case where no error is returned. If a single item is longer than 255 bytes, push truncates it to 255 and returns true, so the outer truncated flag remains false. For example, `httproute/ns/<253-character-name>` is 266 bytes. Could you double-check this path?

pkg/nsx/services/dns/recordservice.go
```diff
@@ -156,36 +266,45 @@ func (s *DNSRecordService) validateEndpointRowConflict(zonePath string, ep *extd
...
-			"effectiveOwner", getDNSRecordOwnerNamespacedName(rec), "currentOwner", currentNNKey)
+	}
+
+	if getDNSRecordOwnerNamespacedName(rec) == currentNNKey {
 		row := NewEndpointRow(ep, zonePath, recordName)
-		primaryNN := primaryOwnerNNIndexKeyFromRecord(rec)
-		row.effectiveOwner = effectiveOwner
-		existingContributions := decompressContributingTags(rec)
-		row.contributingOwnerKeys = mergeContributingOwnerKeys(existingContributions, currentNNKey, primaryNN)
+		row.contributingOwnerKeys = parseContributingOwnersFromRecord(rec)
```
heypnus: Should we compare the targets before preserving the contributors here? If A and B initially share target X and primary A later changes to Y, B is still kept as a contributor. If A is deleted afterward, B could be promoted while the record still points to Y, even though B expects X. Should we clear or revalidate the contributors when the targets change?

pkg/nsx/services/dns/contributing.go
wenyingd: I could add a dedicated check on the length of the item directly.

pkg/nsx/services/dns/recordservice.go
wenyingd: Good catch, the race condition is possibly occurred in theory. Clearred the contributors if "target" is updated when proposing the primary.

heypnus: Clearing the contributors avoids the stale-promotion case, but what triggers those owners to reconcile after they are removed from the tags? For example , if 2 LB svc share X, A is the primary, and A changes to Y, B is dropped from the contributor index without receiving an event. If A is deleted before B reconciles again, the record is deleted even though B still exists and still wants X. Could we enqueue the removed contributors, or reject the primary t arget change while contributors still exists?

wenyingd: Thanks for the great catch!

Instead of clearing the contributors (which indeed had the risk of leaving stale records when no controller event re-enqueued those contributors), the latest change has refined the validation rules to handle target and gateway updates seamlessly:
1. **Target IP Changes under the same Parent Gateway**:
   In Gateway API, HTTPRoutes attached to the same parent Gateway share the Gateway's Virtual IP. When the Gateway IP updates, the primary owner updates the target IP of the NSX DNS record while **preserving all existing contributors**. Even if the primary is deleted before a contributor reconciles, the promoted contributor inherits the already-updated Gateway IP.
2. **Parent Gateway Changes**:
   If the primary owner attempts to change its Parent Gateway (e.g., moving from `gw1` to `gw2`) while other contributors are still attached to the DNS record, `validateEndpointRowConflict` will **reject the primary's update with an error**.  In fact, the current implementation would not allows any xxRoute to update its parent Gateway (both the primary and the contributors) if multiple resources are sharing the same DNS Record.
This adopts the second approach you suggested (*"reject the primary target change while contributors still exist"*), preventing primary hijacking and avoiding the need to trigger external resync events across controllers.

pkg/nsx/services/dns/recordservice.go
```diff
@@ -144,7 +255,48 @@ func (s *DNSRecordService) syncDnsRecordsInNSX(ctx context.Context, toUpsert, to
...
+	return kind1 == kind2 || (k8sRouteKindSet.Has(kind1) && k8sRouteKindSet.Has(kind2))
+}
+
+// isRecordSharableWithContributors checks if existing rec is owned by a route kind that supports the contributor mechanism.
+func isRecordSharableWithContributors(rec *model.DnsRecord) bool {
+	owner, ok := resourceRefFromDNSRecord(rec)
+	if !ok {
+		return false
+	}
+	return isRouteKind(owner.Kind)
```
heypnus, line 273: Is restricting contributors to Route kinds intentional? Before this change, two LB Services with the same FQDN and targets could share the record, and the existing test covered that case using Services. This patch changes the test to HTTPRoute and now rejects non-Route owners, while the PR description still mentions shared reconciliation for LoadBalancers. Are we okay with dropping the existing same-target sharing behavior for Services?

wenyingd: Yes, the FQDN is shared only within gateway-api xxRoutes for now, this is consistent with design. For the case to share FQDN among multiple LB Services, we could add the support in the future only when we receive the actual customer requirement.

I would update the descriptions and the commit messages.

heypnus (approved):
Thanks for addressing my comments. LGTM

TaoZou1 (approved):

PR #1522, opened 2026-08-03
Change 500 internal error for standard http response code

seanpang-vmware:
1. return 403 when the failure is because of privilege.
2. return 404 when resource not found in backend logic.

Testing Done:

1. Testing 403 error
SSH'd into the live testbed.
Created a patched version of the role that removed the read permissions for subnets and subnets/status, and applied it.
Ran the test command: kubectl get subnetippools subnet-sample-a -n ipv4-ns-2.
The result was a proper 403 Forbidden error (instead of a 500 internal error)

2. Testing 404 error
Ran the test command: kubectl get subnetdhcpserverstats bogus-identifier -n default
The result was a proper 404 Not Found error (instead of a 500 internal error or a 503 Service Unavailable)


pkg/eas/storage/subnetipppols.go
```diff
@@ -58,6 +69,10 @@ func (s *SubnetIPPoolsStorage) Get(ctx context.Context, namespace, name string)
 	subnets, err := s.nsxClient.SubnetsClient.List(orgID, projectID, vpcID,
 		nil, nil, nil, nil, nil, nil)
 	if err != nil {
+		var unauthorized vapierrors.Unauthorized
```
dantingl: If vpcID or projectID is not existing, error could still be NotFound?

pkg/eas/storage/ipblokusage.go
```diff
@@ -170,6 +190,10 @@ func (s *IPBlockUsageStorage) List(_ context.Context, namespace string) (*easv1a
 		log.Debug("Fetching project IP block usage from NSX", "orgID", orgID, "projectID", pid)
 		nsxList, err := s.nsxClient.ProjectIPBlockUsageClient.List(orgID, pid, nil, nil, nil, nil, nil, nil, nil)
 		if err != nil {
+			var unauthorized vapierrors.Unauthorized
```
dantingl: same here

pkg/eas/storage/subnetdhcpstats.go
```diff
@@ -58,6 +68,10 @@ func (s *SubnetDHCPStatsStorage) Get(ctx context.Context, namespace, name string
 	subnets, err := s.nsxClient.SubnetsClient.List(orgID, projectID, vpcID,
 		nil, nil, nil, nil, nil, nil)
 	if err != nil {
+		var unauthorized vapierrors.Unauthorized
```
dantingl: same here

dantingl:
Is there any way we can check error in a common util function for all APIs?

pkg/eas/storage/subnetipppols.go
seanpang-vmware: Yes, for eas api, user do not need to know the detailed nsx resource related information, so the error exposed should not include these information, a normal 404 error will indicate some resources is missing in the api not only for the input subnet cr.

dantingl (approved):
LGTM

TaoZou1 (approved):

PR #1524, opened 2026-08-06
Support Description, IPBlockNames and labels in Subnet CRD

dantingl:
- Add Description and IPBlockNames fields to SubnetSpec
- Update NSX Subnet service builder to map Description, IPBlockNames, and Kubernetes labels as NSX tags
- Update comparison logic to detect changes to Description and IPBlockNames
- Update MapNSXSubnetToSubnetCR to populate Description and IPBlockNames
- Regenerate CRD manifests, deepcopy, openapi, client interfaces, and API documentation
-  Added E2E cases verifying Subnet creation with description, labels, and VPC private IPBlock.

Testing done:
1. Subnet CRD Description:
    1.a create subnet-test-1 CR with description, checked NSX Subnet description and CR status
    1.b update subnet-test-1 CR with description, checked NSX Subnet description and CR status
    1.c update subnet-test-1 CR with description more than 1024 characters, CR failed to update
    1.d update subnet-test-1 CR without description, checked NSX Subnet description is empty as well
    
2. Subnet CRD labels:
    2.a create subnet-test-2 CR with labels(env: production, tier: backend), checked NSX Subnet tags and CR status
    2.b update subnet-test-2 CR with labels(env: test), checked NSX Subnet tags and CR status
    2.c update subnet-test-2 CR with labels more than 20, NSX threw error 'tags cannot exceed maximum size 26'
    2.d update subnet-test-2 CR without labels, checked NSX Subnet tags which do not have any custom tags
    
3. Subnet CRD with ipBlockNames:
    3.a create subnet-test-3 CR with ipBlockNames ':ipblock-192.168.0.0-netmask-16', checked NSX Subnet's ip_blocks and CR status.
    3.b update subnet-test-3 CR with ipBlockNames 'project-quality-ipblock-10.246.0.0-netmask-16', update failed.
    3.c create subnet-test-4 CR with ipBlockNames ':ipblock-192.168.0.0-netmask-16, project-quality-ipblock-10.246.0.0-netmask-16', NSX threw error 'Mismatch in IP Address Block path'
    3.d create subnet-test-4 CR with ipBlockNames 'project-quality-ipblock-10.246.0.0-netmask-16', check NSX Subnet's ip_blocks and CR status.
    3.e create subnet-test-5 CR with ipBlockNames 'ns-1_dt37d-172_26_0_0_16', checked NSX Subnet's ip_blocks and CR status.
[...]

PR #1525, opened 2026-08-06
Support LB VPC, VIP reuse

TaoZou1:
Introduce several enhancements to the IPAddressAllocation controller to
    support LoadBalancer-specific VPC routing, virtual IP (VIP) reuse, and
    custom IP block allocations.
    Specifically:
    1. Implement VIP reuse via the `nsx.vmware.com/reuse-vip` annotation. When
       specified, the controller retrieves the original allocation and updates
       its tags to point to the new CR's UID, allowing seamless VIP inheritance.
    2. Support LoadBalancer VPC routing via the `nsx.vmware.com/lb` annotation.
       If set to "true", the allocation is routed to the dedicated
       `LoadBalancerVPC` specified in the namespace's VPCNetworkConfiguration
       instead of the default namespace VPC.

   Test done:
    
    | Case ID | Feature | Scenario / Input | Expected Behavior | NSX Extra Resources | User Notification / Status | Result |
| :--- | :--- | :--- | :--- | :---: | :--- | :---: |
| **1.1.1** | LB VPC | `lb: "true"` with configured `loadBalancerVPC` | IP allocated from dedicated LB VPC (`172.26.0.8`) | 1 (intended) | `Ready=True`, `status.allocationIPs` set | **PASS** |
| **1.1.2** | LB VPC | Standard allocation without `lb` annotation | IP allocated from default VPC (`172.26.0.9`) | 1 (intended) | `Ready=True`, `status.allocationIPs` set | **PASS** |
| **1.2.1** | LB VPC | `lb: "true"` with missing `loadBalancerVPC` config | Explicit failure, blocks creation | 0 | `Ready=False`, Warning Event emitted | **PASS** |
| **1.2.2** | LB VPC | `lb: "true"` with malformed `loadBalancerVPC` path | Path parsing error handled cleanly | 0 | `Ready=False`, Warning Event emitted | **PASS** |
| **2.1.1** | VIP Reuse | Same-namespace VIP reuse (`reuse-vip: "tc211-orig"`) | Inherited VIP (`172.26.0.10`), NSX resource tags updated | 0 (reused) | `Ready=True`, `status.allocationIPs` set | **PASS** |
| **2.1.2** | VIP Reuse | Cross-namespace VIP reuse (`reuse-vip: "ns-target/tc212-orig"`) | Cross-namespace VIP inherited (`172.26.0.0`) | 0 (reused) | `Ready=True`, `status.allocationIPs` set | **PASS** |
| **2.1.3** | VIP Reuse | Idempotency check (CR metadata update after reuse) | `needsTagUpdate == false`, no redundant NSX PATCH calls | 0 | `Ready=True`, IP retained | **PASS** |
| **2.1.4** | VIP Reuse | **Operator restart after VIP reuse** | **Local store rebuilt; fallback lookup by `obj.UID` matched seamlessly** | 0 | `Ready=True`, IP retained without errors | **PASS** |
| **2.2.1** | VIP Reuse | Target CR does not exist (`reuse-vip: "ns/nonexistent"`) | Error logged (`CR not found`), auto-requeues via backoff | 0 | `Ready=False`, Warning Event emitted | **PASS** |
| **2.2.3** | VIP Reuse | Whitespace-only format (`reuse-vip: "  "`) | Error logged (`invalid reuse-vip format`) | 0 | `Ready=False`, Warning Event emitted | **PASS** |
| **3.1.2** | Custom IPBlock | Infra-scoped IPBlock (`:ipblock-192.168.0.0-netmask-16`) | Parsed path `/infra/ip-blocks/...`, IP allocated (`192.168.0.9`) | 1 (intended) | `Ready=True`, `status.allocationIPs` set | **PASS** |
| **3.1.3** | Custom IPBlock | Leading/trailing spaces (`  :ipblock-192.168.0.0-netmask-16  `) | Whitespace trimmed cleanly, IP allocated (`192.168.0.10`) | 1 (intended) | `Ready=True`, `status.allocationIPs` set | **PASS** |
| **3.2.1** | Custom IPBlock | Disallowed full policy path (`/infra/ip-blocks/block`) | Rejection with message `full path is not supported` | 0 | `Ready=False`, Warning Event emitted | **PASS** |
| **3.2.2** | Custom IPBlock | Empty block ID / Single colon (`:`) | Rejection with message `IP block ID cannot be empty` | 0 | `Ready=False`, Warning Event emitted | **PASS** |


pkg/nsx/services/ipaddressallocation/ipaddressallocation.go
```diff
@@ -65,7 +67,76 @@ func InitializeIPAddressAllocation(service common.Service, vpcService common.VPC
 }
 
 func (service *IPAddressAllocationService) CreateOrUpdateIPAddressAllocation(obj *v1alpha1.IPAddressAllocation, restoreMode bool) (bool, error) {
-	nsxIPAddressAllocation, err := service.BuildIPAddressAllocation(obj, nil, restoreMode)
+	reuseVIP, hasReuse := obj.Annotations[common.AnnotationIPAllocReuseVIP]
+	if hasReuse && reuseVIP != "" {
+		var targetName, targetNamespace string
+		if strings.Contains(reuseVIP, "/") {
+			parts := strings.Split(reuseVIP, "/")
```
wenqiq: `strings.Split(reuseVIP, "/")` doesn't validate if targetNamespace or targetName are empty (e.g. "ns/" or "a/b/c").
Should use strings.SplitN(reuseVIP, "/", 2) and validate that both targetNamespace and targetName are non-empty here?

pkg/nsx/services/ipaddressallocation/ipaddressallocation.go
```diff
@@ -65,7 +67,76 @@ func InitializeIPAddressAllocation(service common.Service, vpcService common.VPC
...
+				}
+				originalVPCInfo := []common.VPCResourceInfo{vpcResourceInfo}
+
+				if err := service.Apply(&updatedNSXAlloc, originalVPCInfo); err != nil {
+					log.Error(err, "Failed to update tags of original NSX IPAddressAllocation", "ID", *originalNSXAlloc.Id)
+					return false, err
+				}
+			}
+		} else {
+			log.Warn("Original NSX IPAddressAllocation not found in store, cannot reuse VIP", "UID", originalCR.UID)
```
yanjunz97: Do we need to return an error here?

pkg/nsx/services/ipaddressallocation/builder.go
```diff
@@ -68,9 +111,30 @@ func (service *IPAddressAllocationService) BuildIPAddressAllocation(obj metav1.O
 				allocationSize = Int64(int64(o.Spec.AllocationSize))
 			}
 		}
+		if rawBlockName := strings.TrimSpace(o.Spec.IPBlockName); rawBlockName != "" {
+			if strings.HasPrefix(rawBlockName, "/") {
+				log.Error(nil, "Invalid IPBlockName format, only block ID, '<project ID>:<ipBlockID>', or ':ipBlockID' is supported, full path is not supported", "IPBlockName", rawBlockName)
+				return nil, nil, fmt.Errorf("invalid IPBlockName format %s, only block ID, '<project ID>:<ipBlockID>', or ':ipBlockID' is supported, full path is not supported", rawBlockName)
+			} else if strings.Contains(rawBlockName, ":") {
+				parts := strings.SplitN(rawBlockName, ":", 2)
```
wenqiq: If IPBlockName is formatted as "project:" or ":", ipBlockID evaluates to an empty string, generating an invalid NSX path ending in /infra/ip-blocks/.
Add a check to validate that ipBlockID is non-empty after trimming?

pkg/nsx/services/ipaddressallocation/builder.go
```diff
@@ -68,9 +111,30 @@ func (service *IPAddressAllocationService) BuildIPAddressAllocation(obj metav1.O
...
+					if len(VPCInfo) > 0 && VPCInfo[0].OrgID != "" {
+						orgID = VPCInfo[0].OrgID
+					}
+					ipBlock = String(fmt.Sprintf("/orgs/%s/projects/%s/infra/ip-blocks/%s", orgID, projectID, ipBlockID))
+				}
+			} else {
+				ipBlock = String(fmt.Sprintf("/infra/ip-blocks/%s", rawBlockName))
+			}
+		}
 	case *v1alpha1.AddressBinding:
```
wenqiq, line 137: `VPCInfo` is only assigned under case `*v1alpha1.IPAddressAllocation`. For case `*v1alpha1.AddressBinding:` during restore mode, VPCInfo remains nil, causing subsequent `Apply()` calls to fail with `len(VPCInfo) == 0`.
Should also populate VPCInfo for AddressBinding here?

pkg/nsx/services/ipaddressallocation/builder.go
```diff
@@ -32,24 +33,66 @@ func convertIpAddressBlockVisibility(visibility v1alpha1.IPAddressVisibility) v1
...
+
+func (service *IPAddressAllocationService) getVPCInfo(ns string, isLB bool) ([]common.VPCResourceInfo, error) {
+	var VPCInfo []common.VPCResourceInfo
+	if isLB {
+		nc, err := service.VPCService.GetVPCNetworkConfigByNamespace(ns)
+		if err != nil {
+			log.Error(err, "Failed to Get NetworkConfig by Namespace", "Namespace", ns)
+			return nil, err
+		}
+		if nc != nil && nc.Spec.LoadBalancerVPC != "" {
```
yanjunz97, line 55: It seems if LoadBalancerVPC is not specified, the lb annotations is ignored silently. Is this the expected behavior or we shall let user know that annotation does not take effect?

pkg/nsx/services/ipaddressallocation/builder.go
```diff
@@ -32,24 +33,66 @@ func convertIpAddressBlockVisibility(visibility v1alpha1.IPAddressVisibility) v1
...
+	var ipBlock *string
+	var VPCInfo []common.VPCResourceInfo
 
 	switch o := obj.(type) {
 	case *v1alpha1.IPAddressAllocation:
-		if o.Spec.IPAddressType != "" {
-			ipAddressType = controllerscommon.ConvertCRIPAddressTypeToNSX(v1alpha1.IPAddressType(o.Spec.IPAddressType))
-		} else {
-			ipAddressType = controllerscommon.ConvertCRIPAddressTypeToNSX(v1alpha1.IPAddressTypeIPv4)
+		ipAddressType = ipAddressTypeToNSX(o.Spec.IPAddressType)
```
wenqiq, line 91: It seems there is no need to add a new function `ipAddressTypeToNSX` here.

pkg/nsx/services/ipaddressallocation/builder.go
```diff
@@ -32,24 +33,66 @@ func convertIpAddressBlockVisibility(visibility v1alpha1.IPAddressVisibility) v1
...
+}
+
+func (service *IPAddressAllocationService) BuildIPAddressAllocation(obj metav1.Object, subnetPortCR *v1alpha1.SubnetPort, restoreMode bool) (*model.VpcIpAddressAllocation, []common.VPCResourceInfo, error) {
 	ipAddressBlockVisibility := v1alpha1.IPAddressVisibilityPrivate
 	var allocationIps *string
 	var allocationSize *int64
 	var ipAddressType string
 	var ipv6AllocationPrefixLength *int64
+	var ipBlock *string
+	var VPCInfo []common.VPCResourceInfo
```
wenqiq: We had better rename the local variable `VPCInfo` to the `vpcInfo`.

pkg/nsx/services/ipaddressallocation/ipaddressallocation.go
```diff
@@ -65,7 +67,76 @@ func InitializeIPAddressAllocation(service common.Service, vpcService common.VPC
...
+				if err != nil {
+					log.Error(err, "Failed to parse VPC path from original NSX IPAddressAllocation", "Path", *originalNSXAlloc.Path)
+					return false, err
+				}
+				originalVPCInfo := []common.VPCResourceInfo{vpcResourceInfo}
+
+				if err := service.Apply(&updatedNSXAlloc, originalVPCInfo); err != nil {
+					log.Error(err, "Failed to update tags of original NSX IPAddressAllocation", "ID", *originalNSXAlloc.Id)
+					return false, err
+				}
```
yanjunz97: I'm not quite understand the use case here. Just double confirm, as we do not return for update success, does it mean for reuse case, we expect to update the existing NSX IPAddressAllocation and also create a new NSX IPAddressAllocation?

pkg/nsx/services/ipaddressallocation/builder.go
```diff
@@ -68,9 +111,30 @@ func (service *IPAddressAllocationService) BuildIPAddressAllocation(obj metav1.O
 				allocationSize = Int64(int64(o.Spec.AllocationSize))
 			}
 		}
+		if rawBlockName := strings.TrimSpace(o.Spec.IPBlockName); rawBlockName != "" {
+			if strings.HasPrefix(rawBlockName, "/") {
+				log.Error(nil, "Invalid IPBlockName format, only block ID, '<project ID>:<ipBlockID>', or ':ipBlockID' is supported, full path is not supported", "IPBlockName", rawBlockName)
+				return nil, nil, fmt.Errorf("invalid IPBlockName format %s, only block ID, '<project ID>:<ipBlockID>', or ':ipBlockID' is supported, full path is not supported", rawBlockName)
```
yanjunz97: The format seems to be different with ipBlockName in Subnet #1524 
Would you think we shall make them consistent @dantingl 
Not sure if there's concern about exposing project ID in subnet ipBlockName.

pkg/nsx/services/ipaddressallocation/ipaddressallocation.go
```diff
@@ -65,7 +67,76 @@ func InitializeIPAddressAllocation(service common.Service, vpcService common.VPC
 }
 
 func (service *IPAddressAllocationService) CreateOrUpdateIPAddressAllocation(obj *v1alpha1.IPAddressAllocation, restoreMode bool) (bool, error) {
-	nsxIPAddressAllocation, err := service.BuildIPAddressAllocation(obj, nil, restoreMode)
+	reuseVIP, hasReuse := obj.Annotations[common.AnnotationIPAllocReuseVIP]
```
yanjunz97, line 70: I'm curious about if we allow user to remove reuse vip or lb annotation. And if that is allowed, shall we deal with the tags on existing ipaa or ipaa on loadBalancerVPC

pkg/nsx/services/ipaddressallocation/builder.go
dantingl: @TaoZou1 please check Jianjun's comment in the previous CRD definition:
https://github.com/vmware-tanzu/nsx-operator/pull/1468#discussion_r3589977644

pkg/nsx/services/ipaddressallocation/ipaddressallocation.go
TaoZou1: Yes. Fixed

pkg/nsx/services/ipaddressallocation/builder.go
TaoZou1: Yes, Fixed

pkg/nsx/services/ipaddressallocation/builder.go
TaoZou1: not familiar with AddressBinding part

pkg/nsx/services/ipaddressallocation/builder.go
TaoZou1: Done

pkg/nsx/services/ipaddressallocation/builder.go
TaoZou1: easy for UT coverage

pkg/nsx/services/ipaddressallocation/ipaddressallocation.go
TaoZou1: Yes.

pkg/nsx/services/ipaddressallocation/builder.go
TaoZou1: It should return error.

pkg/nsx/services/ipaddressallocation/ipaddressallocation.go
TaoZou1: the CR reuse the existed VIP resource.
it will not create a new one.
	ipAddressAllocationUpdated := common.CompareResource(IpAddressAllocationToComparable(existingIPAddressAllocation),
		IpAddressAllocationToComparable(nsxIPAddressAllocation))

	if !ipAddressAllocationUpdated {
	-> it should go here and update the current CR status

pkg/nsx/services/ipaddressallocation/ipaddressallocation.go
TaoZou1: That's a good point. There is another webhook patch. I prefer to those fields could be only set by the system user.

pkg/nsx/services/ipaddressallocation/ipaddressallocation.go
TaoZou1: you're right,  it should return

pkg/nsx/services/ipaddressallocation/builder.go
TaoZou1: follow the Subnet style

PR #1532, opened 2026-08-13
Implement VPCEndpoint and ServiceEndpoint reconciliation for VPC Private Link

poojav25:
Implements the NSX-side reconciliation for the VPCEndpoint/ServiceEndpoints

1. Wires VpcEndpointClient/VpcServiceEndpointClient (+ their Status sub-resource clients) into pkg/nsx/client.go.
2. Adds pkg/nsx/services/serviceendpoint and pkg/nsx/services/vpcendpoint — build/store/compare/service, modeled on ipaddressallocation. VPCEndpoint.spec.serviceEndpointName is a pre-resolved projectID:vpcID:serviceEndpointID composite string from CCI, so resolving it is pure string parsing (GetVpcServiceEndpointPathFromCCIName) — no k8s lookup, no NSX Search API call. ipAllocationName resolves via the existing same-namespace IPAddressAllocation pattern (copied from StaticRoute).
3. Adds pkg/controllers/serviceendpoint and pkg/controllers/vpcendpoint controllers (Reconcile/handleUpdate/handleDeletion/CollectGarbage/RestoreReconcile), registered in cmd/main.go after IPAddressAllocationService.
4. Adds CEL immutability (self == oldSelf) on ipAllocationName, serviceEndpointName, serviceEndpointIP — NSX documents all three as immutable-after-create; better to reject the edit at the k8s API server than wait for an async NSX rejection.
5. Adds a finalizer (serviceendpoint.nsx.vmware.com/finalizer) to ServiceEndpoint, mirroring the existing Subnet/SubnetConnectionBindingMap pattern — see "Known limitations" below for why this was necessary.
6. Regenerated CRD YAMLs and docs/ref/apis/vpc.md via make manifests generate-api-docs.

**Setup**
```bash 
kubectl apply -f - <<EOF
kind: IPAddressAllocation
metadata:
  name: test-vpcep-ipalloc
  namespace: default
spec:
  ipAddressBlockVisibility: Private
  allocationSize: 1
EOF
kubectl get ipaddressallocation test-vpcep-ipalloc -n default -o yaml
------------------------
status:
  conditions:
  - message: NSX IPAddressAllocation has been successfully created/updated
    reason: IPAddressAllocationReady
    status: "True"
    type: Ready
  allocationIPs: 172.26.0.9
```

**T1 — Create ServiceEndpoint, create VPCEndpoint referencing it across VPCs, verify in NSX**

```bash

kubectl apply -f - <<EOF
kind: ServiceEndpoint
metadata:
  name: test-se-2
  namespace: svc-velero-5i5im
spec:
  serviceEndpointIP: 10.246.0.8
EOF

kubectl apply -f - <<EOF
kind: VPCEndpoint
metadata:
  name: test-vpcep-2
  namespace: default
spec:
  ipAllocationName: test-vpcep-ipalloc
  serviceEndpointName: "project-quality:vmware-system-supervisor-services-vpc_ena6d:test-se_d0bbh"
EOF

---------------
curl -sk -u 'admin:<password>' \
  "https://<nsx-manager>/policy/api/v1/orgs/default/projects/project-quality/vpcs/kube-system_5wsit/vpc-endpoints" 
  
--------------
{
    "results": [
       {
            "vpc_service_endpoint": "/orgs/default/projects/project-quality/vpcs/vmware-system-supervisor-services-vpc_ena6d/vpc-service-endpoints/test-se-2_ailk0",
            "ip_allocation_path": "/orgs/default/projects/project-quality/vpcs/kube-system_5wsit/ip-address-allocations/test-vpcep-ipalloc_qfw3w",
            "resource_type": "VpcEndpoint",
            "id": "test-vpcep-2_2u1jo",
            "display_name": "test-vpcep-2",
            "tags": [
                {"scope": "nsx-op/cluster", "tag": "69ad3e25-f46b-4805-87f2-25162c83e1db"},
                {"scope": "nsx-op/version", "tag": "1.0.0"},
                {"scope": "nsx-op/namespace", "tag": "default"},
                {"scope": "nsx-op/vpcendpoint_name", "tag": "test-vpcep-2"},
                {"scope": "nsx-op/vpcendpoint_uid", "tag": "cdb706db-ad86-452d-af30-930109e0e1e7"},
                {"scope": "nsx-op/namespace_uid", "tag": "7be6f428-7805-4b4f-8e04-1b582dcdfd3d"}
            ],
            "path": "/orgs/default/projects/project-quality/vpcs/kube-system_5wsit/vpc-endpoints/test-vpcep-2_2u1jo",
........
        }
  ]
```
Confirmed the composite-string resolution builds the correct cross-VPC NSX path, no CR lookup needed.

**T3 — CEL immutability: editing serviceEndpointIP after creation is rejected at the k8s API server**

```bash
kubectl patch serviceendpoint test-se -n svc-velero-5i5im --type merge -p '{"spec":{"serviceEndpointIP":"10.246.0.99"}}'

error: serviceendpoints.crd.nsx.vmware.com "test-se" could not be patched: admission webhook ... denied the request:
ServiceEndpoint.crd.nsx.vmware.com "test-se" is invalid: spec.serviceEndpointIP: Invalid value: "string": Value is immutable
```
**T4 —Negative: serviceEndpointName pointing at a non-existent NSX object, rejected synchronously by NSX**
```bash
serviceEndpointName: "project-quality:vmware-system-supervisor-services-vpc_ena6d:does-not-exist"

nsx error code: 661109, message: VpcServiceEndpoint path=[.../vpc-service-endpoints/does-not-exist]
referenced by vpc_service_endpoint not found.
```
**Known limitations discovered during testing**

1. Realization status never confirms success. NSX's status API for these two resource types doesn't reliably return SUCCESS, even when the object was created correctly — so the CR's Ready condition can't be fully trusted yet.


pkg/controllers/vpcendpoint/vpcendpoint_controller.go
```diff
@@ -0,0 +1,221 @@
+/* Copyright © 2026 VMware, Inc. All Rights Reserved.
```
dantingl, line 1: Maybe add some UT cases for both service endpoint and vpc endpoint controllers

pkg/controllers/serviceendpoint/serviceendpoint_controller.go
```diff
@@ -0,0 +1,254 @@
...
+				r.StatusUpdater.DeleteFail(req.NamespacedName, nil, err)
+				return resultRequeue, err
+			}
+			r.StatusUpdater.DeleteSuccess(req.NamespacedName, nil)
+			return common.ResultNormal, nil
+		}
+		return resultRequeue, err
+	}
+	if obj.ObjectMeta.DeletionTimestamp.IsZero() {
+		if !controllerutil.ContainsFinalizer(obj, servicecommon.ServiceEndpointFinalizerName) {
```
dantingl: I remember previously we tried to remove the finalizer on deletion, as it would block namespace deletion if the nsx-operator wasn't running normally. We can rely on GC to clean up objects if the nsx-operator misses any events.
@yanjunz97 correct me if I'm wrong.

pkg/controllers/serviceendpoint/serviceendpoint_controller.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+	r.StatusUpdater.UpdateSuccess(ctx, obj, setReadyStatusTrue)
+	return resultNormal, nil
+}
+
+func (r *ServiceEndpointReconciler) handleDeletion(ctx context.Context, req ctrl.Request, obj *v1alpha1.ServiceEndpoint) (ctrl.Result, error) {
+	r.StatusUpdater.IncreaseDeleteTotal()
+	if err := r.Service.DeleteServiceEndpoint(obj); err != nil {
+		setDeleteFailedStatus(r.Client, ctx, obj, metav1.Now(), err)
+		r.StatusUpdater.DeleteFail(req.NamespacedName, obj, err)
```
dantingl, line 134: I'm thinking maybe we do not need to pass obj to generate K8s events for both DeleteFail and DeleteSuccess, what do you think?

pkg/nsx/services/serviceendpoint/serviceendpoint.go
```diff
@@ -0,0 +1,222 @@
...
+func (service *ServiceEndpointService) DeleteServiceEndpointByNamespacedName(namespace, name string) error {
+	objs := service.ServiceEndpointStore.List()
+	for _, obj := range objs {
+		serviceEndpoint, ok := obj.(*model.VpcServiceEndpoint)
+		if !ok {
+			continue
+		}
+		namespaceMatch, nameMatch := false, false
+		for _, tag := range serviceEndpoint.Tags {
+			if *tag.Scope == common.TagScopeNamespace && *tag.Tag == namespace {
```
dantingl: Do you think we can use namespace and name as a store indexer for lookup

pkg/controllers/vpcendpoint/vpcendpoint_controller.go
poojav25: Added unit tests for both the VPCEndpoint and ServiceEndpoint controllers, covering Reconcile (get errors, not-found/delete, create/update, deletion), garbage collection, and restore.

pkg/controllers/serviceendpoint/serviceendpoint_controller.go
poojav25: Good catch, removed the finalizer. ServiceEndpoint now follows the same pattern as VPCEndpoint — delete is attempted directly, and GC will clean up anything that's missed, so it won't block namespace deletion anymore.

pkg/controllers/serviceendpoint/serviceendpoint_controller.go
poojav25: We do need obj here — it's used by Recorder.Event() to attach the k8s Event to the specific ServiceEndpoint CR, so users can see it via kubectl describe/kubectl get events. If we pass nil, we'd lose that per-object event and only get a log line. Keeping it as is.

pkg/nsx/services/serviceendpoint/serviceendpoint.go
poojav25: Added a namespace+name indexer for the ServiceEndpoint store (same pattern used in subnetipreservation), so lookups by namespace/name are now indexed instead of scanning all objects.

pkg/controllers/vpcendpoint/vpcendpoint_controller.go
```diff
@@ -0,0 +1,221 @@
...
+		return resultRequeue, err
+	}
+	r.StatusUpdater.UpdateSuccess(ctx, obj, setReadyStatusTrue)
+	return resultNormal, nil
+}
+
+func (r *VPCEndpointReconciler) handleDeletion(ctx context.Context, req ctrl.Request, obj *v1alpha1.VPCEndpoint) (ctrl.Result, error) {
+	r.StatusUpdater.IncreaseDeleteTotal()
+	if err := r.Service.DeleteVPCEndpoint(obj); err != nil {
+		r.StatusUpdater.DeleteFail(req.NamespacedName, nil, err)
```
wenqiq: Should also pass `obj` here to align with serviceendpoint controller and other controllers?

pkg/controllers/vpcendpoint/vpcendpoint_controller.go
```diff
@@ -0,0 +1,221 @@
...
+		return resultRequeue, err
+	}
+	r.StatusUpdater.DeleteSuccess(req.NamespacedName, nil)
```
wenqiq: ditto

pkg/nsx/services/vpcendpoint/vpcendpoint.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+	if nsxVPCEndpoint == nil {
+		log.Info("VPCEndpoint not found in store, skip deletion", "obj", obj)
+		return nil
+	}
+	return service.DeleteVPCEndpointByNSXResource(nsxVPCEndpoint)
+}
+
+func (service *VPCEndpointService) DeleteVPCEndpointByNamespacedName(namespace, name string) error {
+	objs := service.VPCEndpointStore.List()
```
wenqiq: `ServiceEndpointStore` uses `indexNamespacedName` to achieve $O(1)$ lookup via `GetByIndex`. `VPCEndpointStore` missed this index, forcing `DeleteVPCEndpointByNamespacedName` to fall back to an $O(N)$ List() iteration. Should we also add the index for `vpcendpoint` in store.go?

pkg/controllers/vpcendpoint/vpcendpoint_controller.go
```diff
@@ -0,0 +1,221 @@
...
+		r.StatusUpdater.UpdateFail(ctx, obj, err, "", setReadyStatusFalse)
+		return resultRequeue, err
+	}
+	r.StatusUpdater.UpdateSuccess(ctx, obj, setReadyStatusTrue)
+	return resultNormal, nil
+}
+
+func (r *VPCEndpointReconciler) handleDeletion(ctx context.Context, req ctrl.Request, obj *v1alpha1.VPCEndpoint) (ctrl.Result, error) {
+	r.StatusUpdater.IncreaseDeleteTotal()
+	if err := r.Service.DeleteVPCEndpoint(obj); err != nil {
```
wenqiq, line 135: Should also add `setDeleteFailedStatus` for `VPCEndpointReconciler` to align with `ServiceEndpointReconciler `?

pkg/nsx/services/serviceendpoint/serviceendpoint.go
```diff
@@ -0,0 +1,212 @@
...
+			return fmt.Errorf("VpcServiceEndpoint %s realization failed", id)
+		default:
+			// keep polling
+			return errNotRealized
+		}
+	})
+}
+
+func (service *ServiceEndpointService) DeleteServiceEndpointByNSXResource(nsxServiceEndpoint *model.VpcServiceEndpoint) error {
+	vpcResourceInfo, err := common.ParseVPCResourcePath(*nsxServiceEndpoint.Path)
```
wenqiq, line 112: If `Path` is nil, it will cause a nil-pointer panic. Suggest to add defensive nil checks here.

pkg/nsx/services/vpcendpoint/vpcendpoint.go
```diff
@@ -0,0 +1,253 @@
...
+			return fmt.Errorf("VpcEndpoint %s realization failed", id)
+		default:
+			// keep polling
+			return errNotRealized
+		}
+	})
+}
+
+func (service *VPCEndpointService) DeleteVPCEndpointByNSXResource(nsxVPCEndpoint *model.VpcEndpoint) error {
+	vpcResourceInfo, err := common.ParseVPCResourcePath(*nsxVPCEndpoint.Path)
```
wenqiq, line 157: ditto

pkg/controllers/serviceendpoint/serviceendpoint_controller.go
dantingl: OK, so you prefer to publish K8s events for these two CR updates

pkg/controllers/serviceendpoint/serviceendpoint_controller.go
```diff
@@ -0,0 +1,239 @@
...
+)
+
+const readyConditionType = "Ready"
+
+// ServiceEndpointReconciler reconciles a ServiceEndpoint object
+type ServiceEndpointReconciler struct {
+	client.Client
+	Scheme        *apimachineryruntime.Scheme
+	Service       *serviceendpoint.ServiceEndpointService
+	VPCService    servicecommon.VPCServiceProvider
```
yanjunz97: Seems this vpc service is not used in the controller? We can remove it

pkg/nsx/services/serviceendpoint/serviceendpoint.go
```diff
@@ -0,0 +1,212 @@
...
+		return err
+	}
+	if existingServiceEndpoint != nil {
+		// Keep the existing NSX id and display_name.
+		nsxServiceEndpoint.Id = String(*existingServiceEndpoint.Id)
+		nsxServiceEndpoint.DisplayName = String(*existingServiceEndpoint.DisplayName)
+		if !common.CompareResource(ServiceEndpointToComparable(existingServiceEndpoint), ServiceEndpointToComparable(nsxServiceEndpoint)) {
+			log.Info("ServiceEndpoint is not changed", "UID", obj.UID)
+			return nil
+		}
```
yanjunz97: Do we support update on service endpoint? I see the only field in spec is immutable, can we always return nil for existingServiceEndpoint, maybe just print some log for this case.

pkg/nsx/services/serviceendpoint/serviceendpoint.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// checkServiceEndpointRealizeState polls until NSX reports SUCCESS or ERROR.
+// TODO: confirm with NSX whether this status API is fully implemented; it hasn't
+// reliably returned SUCCESS in testing.
+func (service *ServiceEndpointService) checkServiceEndpointRealizeState(orgID, projectID, vpcID, id string) error {
+	return retry.OnError(util.NSXTRealizeRetry, func(err error) bool {
+		return errors.Is(err, errNotRealized)
+	}, func() error {
+		status, err := service.NSXClient.VpcServiceEndpointStatusClient.Get(orgID, projectID, vpcID, id)
```
yanjunz97: I think for RealizeState check we do no refer to the status client. You can reuse realizeService.CheckRealizeState, and refer to this

https://github.com/vmware-tanzu/nsx-operator/blob/d17801bc4172e99bc5af30479fc6047a7071f7a4/pkg/nsx/services/staticroute/staticroute.go#L157-L169

pkg/nsx/services/serviceendpoint/builder.go
```diff
@@ -0,0 +1,38 @@
...
+	tags := service.buildServiceEndpointTags(obj)
+	serviceEndpointType := model.VpcServiceEndpoint_SERVICE_ENDPOINT_IP_TYPE_WORKLOAD
+	vpcServiceEndpoint := &model.VpcServiceEndpoint{
+		Id:                    String(service.BuildServiceEndpointID(obj)),
+		DisplayName:           String(service.buildServiceEndpointName(obj)),
+		Tags:                  tags,
+		ServiceEndpointIp:     String(obj.Spec.ServiceEndpointIP),
+		ServiceEndpointIpType: &serviceEndpointType,
+	}
+	return vpcServiceEndpoint, nil
```
yanjunz97: the error is always nil, maybe we can remove it?

pkg/controllers/vpcendpoint/vpcendpoint_controller.go
```diff
@@ -0,0 +1,221 @@
...
+)
+
+const readyConditionType = "Ready"
+
+// VPCEndpointReconciler reconciles a VPCEndpoint object
+type VPCEndpointReconciler struct {
+	client.Client
+	Scheme        *apimachineryruntime.Scheme
+	Service       *vpcendpoint.VPCEndpointService
+	VPCService    servicecommon.VPCServiceProvider
```
yanjunz97: Is this VPCService used in the controller

pkg/controllers/serviceendpoint/serviceendpoint_controller.go
```diff
@@ -0,0 +1,239 @@
...
+		}
+	}
+	if len(errList) > 0 {
+		return fmt.Errorf("errors found in ServiceEndpoint garbage collection: %s", errList)
+	}
+	return nil
+}
+
+// RestoreReconcile re-reconciles Ready CRs missing from the NSX cache.
+func (r *ServiceEndpointReconciler) RestoreReconcile() error {
```
yanjunz97, line 187: Maybe no need to implement the restore reconcile? As we do not need to restore anything from status, we can rely on the normal mode process the ServiceEndpoint re-creation. It won't block the other resources like service creation right?

pkg/controllers/vpcendpoint/vpcendpoint_controller.go
```diff
@@ -0,0 +1,221 @@
...
+		}
+	}
+	if len(errList) > 0 {
+		return fmt.Errorf("errors found in VPCEndpoint garbage collection: %s", errList)
+	}
+	return nil
+}
+
+// RestoreReconcile re-reconciles Ready CRs missing from the NSX cache.
+func (r *VPCEndpointReconciler) RestoreReconcile() error {
```
yanjunz97, line 190: Similar here, i think no need to restore vpcendpoint in restore mode as all the information is from spec and  nothing from status needed to be restored.

pkg/nsx/services/vpcendpoint/builder.go
```diff
@@ -0,0 +1,37 @@
...
+
+func (service *VPCEndpointService) BuildVPCEndpoint(obj *v1alpha1.VPCEndpoint, ipAllocationPath, serviceEndpointPath string) (*model.VpcEndpoint, error) {
+	tags := service.buildVPCEndpointTags(obj)
+	vpcEndpoint := &model.VpcEndpoint{
+		Id:                 String(service.BuildVPCEndpointID(obj)),
+		DisplayName:        String(service.buildVPCEndpointName(obj)),
+		Tags:               tags,
+		IpAllocationPath:   String(ipAllocationPath),
+		VpcServiceEndpoint: String(serviceEndpointPath),
+	}
```
yanjunz97, line 23: As vpcendpoint has dependency on ipallocation, maybe we need to add webhook in ipallocationallocation to block deletion if vpcendpoint referencing it exists.

pkg/nsx/services/vpcendpoint/vpcendpoint.go
```diff
@@ -0,0 +1,253 @@
...
+	existingVPCEndpoint, err := service.indexedVPCEndpoint(obj.UID)
+	if err != nil {
+		log.Error(err, "Failed to get vpcendpoint", "UID", obj.UID)
+		return err
+	}
+	if existingVPCEndpoint != nil {
+		// Keep the existing NSX id and display_name.
+		nsxVPCEndpoint.Id = String(*existingVPCEndpoint.Id)
+		nsxVPCEndpoint.DisplayName = String(*existingVPCEndpoint.DisplayName)
+		if !common.CompareResource(VPCEndpointToComparable(existingVPCEndpoint), VPCEndpointToComparable(nsxVPCEndpoint)) {
```
yanjunz97: Similarly, as we do not support vpcendpoint update, maybe we do not need to compare it?

pkg/controllers/vpcendpoint/vpcendpoint_controller.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+	if obj.ObjectMeta.DeletionTimestamp.IsZero() {
+		return r.handleUpdate(ctx, obj)
+	}
+	return r.handleDeletion(ctx, req, obj)
+}
+
+func (r *VPCEndpointReconciler) handleUpdate(ctx context.Context, obj *v1alpha1.VPCEndpoint) (ctrl.Result, error) {
+	r.StatusUpdater.IncreaseUpdateTotal()
+	if err := r.Service.CreateOrUpdateVPCEndpoint(ctx, obj.Namespace, obj); err != nil {
```
yanjunz97: As we already pass obj, no need to pass obj.Namespace again?

pkg/nsx/services/vpcendpoint/vpcendpoint.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+// checkVPCEndpointRealizeState polls until NSX reports SUCCESS or ERROR.
+// TODO: confirm with NSX whether this status API is fully implemented; it hasn't
+// reliably returned SUCCESS in testing.
+func (service *VPCEndpointService) checkVPCEndpointRealizeState(orgID, projectID, vpcID, id string) error {
+	return retry.OnError(util.NSXTRealizeRetry, func(err error) bool {
+		return errors.Is(err, errNotRealized)
+	}, func() error {
+		status, err := service.NSXClient.VpcEndpointStatusClient.Get(orgID, projectID, vpcID, id)
```
yanjunz97: Similar to the serviceendpoint RealizeState

pkg/controllers/vpcendpoint/vpcendpoint_controller.go
poojav25: Done - handleDeletion now passes obj to both DeleteFail and DeleteSuccess, and added the matching setDeleteFailedStatus call before DeleteFail, mirroring serviceendpoint.

pkg/controllers/vpcendpoint/vpcendpoint_controller.go
poojav25: added

pkg/nsx/services/vpcendpoint/vpcendpoint.go
poojav25: Added indexNamespacedName/indexByNamespacedName to vpcendpoint/store.go and rewired DeleteVPCEndpointByNamespacedName to use GetByIndex, same as ServiceEndpoint.

pkg/nsx/services/serviceendpoint/serviceendpoint.go
poojav25: done

pkg/controllers/vpcendpoint/vpcendpoint_controller.go
poojav25: Dropped the extra namespace parameter from CreateOrUpdateVPCEndpoint; it now reads obj.Namespace internally

pkg/nsx/services/vpcendpoint/vpcendpoint.go
poojav25: updated

pkg/controllers/vpcendpoint/vpcendpoint_controller.go
poojav25: Agreed there's nothing to recover from status. Reduced both to a no-op (return nil) rather than deleting the method outright, since ReconcilerProvider still requires it alongside CollectGarbage/StartController. Removed the now-dead getRestoreList helpers

pkg/controllers/serviceendpoint/serviceendpoint_controller.go
poojav25: Agreed there's nothing to recover from status.

pkg/controllers/vpcendpoint/vpcendpoint_controller.go
poojav25: confirmed, unused. removed it

pkg/nsx/services/serviceendpoint/builder.go
poojav25: updated

pkg/nsx/services/serviceendpoint/serviceendpoint.go
poojav25: Switched to realizestate.InitializeRealizeState(service.Service).CheckRealizeState(...), same as staticroute. Removed the old status-client polling function and the now-unused errNotRealized/retry imports.

pkg/nsx/services/serviceendpoint/serviceendpoint.go
poojav25: Agreed — dropped the CompareResource diff entirely; CreateOrUpdateServiceEndpoint now just logs and returns nil if the CR already exists in NSX. Deleted compare.go since nothing else used it.

pkg/controllers/serviceendpoint/serviceendpoint_controller.go
poojav25: done

pkg/nsx/services/vpcendpoint/vpcendpoint.go
poojav25: Same treatment as ServiceEndpoint — dropped the compare block, skip-with-log if it already exists. Deleted vpcendpoint's compare.go too.

pkg/nsx/services/vpcendpoint/vpcendpoint.go
poojav25: updated

pkg/controllers/vpcendpoint/vpcendpoint_controller.go
poojav25: updated

pkg/nsx/services/vpcendpoint/builder.go
poojav25: added the webhook

pkg/nsx/services/serviceendpoint/serviceendpoint.go
```diff
@@ -0,0 +1,178 @@
...
+// fields are immutable, so once the NSX object exists there is nothing to update.
+func (service *ServiceEndpointService) CreateOrUpdateServiceEndpoint(obj *v1alpha1.ServiceEndpoint) error {
+	existingServiceEndpoint, err := service.indexedServiceEndpoint(obj.UID)
+	if err != nil {
+		log.Error(err, "Failed to get serviceendpoint", "UID", obj.UID)
+		return err
+	}
+	if existingServiceEndpoint != nil {
+		log.Info("ServiceEndpoint already exists, update is not supported, skip", "UID", obj.UID)
+		return nil
```
yanjunz97, line 71: Could you also add a check on if the CR status is correctly updated, refer to this case:

https://github.com/vmware-tanzu/nsx-operator/blob/53921e9900ac82832291c444d18d8d50c37e85a0/pkg/nsx/services/staticroute/staticroute.go#L123-L128

pkg/nsx/services/vpcendpoint/vpcendpoint.go
```diff
@@ -0,0 +1,214 @@
...
+// nothing to update.
+func (service *VPCEndpointService) CreateOrUpdateVPCEndpoint(ctx context.Context, obj *v1alpha1.VPCEndpoint) error {
+	existingVPCEndpoint, err := service.indexedVPCEndpoint(obj.UID)
+	if err != nil {
+		log.Error(err, "Failed to get vpcendpoint", "UID", obj.UID)
+		return err
+	}
+	if existingVPCEndpoint != nil {
+		log.Info("VPCEndpoint already exists, update is not supported, skip", "UID", obj.UID)
+		return nil
```
yanjunz97, line 104: ditto

pkg/nsx/services/serviceendpoint/serviceendpoint.go
```diff
@@ -0,0 +1,178 @@
...
+	nsxServiceEndpointNew, err := service.NSXClient.VpcServiceEndpointClient.Get(orgID, projectID, vpcID, *nsxServiceEndpoint.Id)
+	err = nsxutil.TransNSXApiError(err)
+	if err != nil {
+		return err
+	}
+
+	realizeService := realizestate.InitializeRealizeState(service.Service)
+	if err := realizeService.CheckRealizeState(util.NSXTRealizeRetry, *nsxServiceEndpointNew.Path, []string{}); err != nil {
+		log.Error(err, "Failed to check ServiceEndpoint realization state", "ID", *nsxServiceEndpoint.Id)
+		return err
```
yanjunz97, line 102: I think we should delete nsx resource if we get error from the realizedstate. Otherwise unrealized nsx resource will be left in NSX. Could you refer to this:

https://github.com/vmware-tanzu/nsx-operator/blob/d17801bc4172e99bc5af30479fc6047a7071f7a4/pkg/nsx/services/staticroute/staticroute.go#L157-L169

pkg/nsx/services/vpcendpoint/vpcendpoint.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	nsxVPCEndpointNew, err := service.NSXClient.VpcEndpointClient.Get(orgID, projectID, vpcID, *nsxVPCEndpoint.Id)
+	err = nsxutil.TransNSXApiError(err)
+	if err != nil {
+		return err
+	}
+
+	realizeService := realizestate.InitializeRealizeState(service.Service)
+	if err := realizeService.CheckRealizeState(util.NSXTRealizeRetry, *nsxVPCEndpointNew.Path, []string{}); err != nil {
```
yanjunz97, line 145: ditto

pkg/controllers/serviceendpoint/serviceendpoint_controller.go
dantingl: For DeleteSuccess and DeleteFail in both serviceendpoint_controller and vpcendpoint_controller, some places pass obj, some places pass nil.

PR #1534, opened 2026-08-14
SubnetPort IPv6 enhancement and validation

yanjunz97:
This PR introduces enhancements and validations for IPv6 support for
`SubnetPort` and `AddressBinding`.
It covers the following topics
- AddressBinding is not supported for IPv6
- Not allow update for SubnetPort InterfaceIPType/StaticIPAllocationType
- Return clear error when SubnetPort InterfaceIPType is not a subnetset of the Subnet/SubnetSet IPAddressType

Testing done:
- Created an AddressBinding CR for VM interface with IPv6 single stack. It failed with `address binding is not supported for IPv6`
Observed Subnetport is still realized without error.

```
kubectl get addressbinding -n ns-1  -o yaml addressbinding1
kind: AddressBinding
metadata:
  annotations:
      {"apiVersion":"crd.nsx.vmware.com/v1alpha1","kind":"AddressBinding","metadata":{"annotations":{},"name":"addressbinding1","namespace":"ns-1"},"spec":{"interfaceName":"eth0","vmName":"vm-1"}}
  creationTimestamp: "2026-07-29T04:40:22Z"
  generation: 1
  name: addressbinding1
  namespace: ns-1
  resourceVersion: "27695467"
  uid: 4ccbd95d-c9fb-4998-b3bf-20fc1faa132e
spec:
  interfaceName: eth0
  vmName: vm-1
status:
  conditions:
  - lastTransitionTime: "2026-08-06T07:53:27Z"
    message: 'error occurred while processing the AddressBinding CR. Error: address
      binding is not supported for IPv6'
    reason: AddressBindingNotReady
    status: "False"
    type: Ready
  ipAddress: ""
```

Update InterfactIPType/StaticIPAllocationType on SubnetPort is not supported
```
kubectl apply -f subnetport.yaml -n ns-1           
The SubnetPort "subnetport-1" is invalid: 
* spec.interfaceIPType: Invalid value: "string": Value is immutable
* spec.staticIPAllocationType: Invalid value: "string": Value is immutable
```

Create a IPv4IPv6 SubnetPort on IPv4 Subnet/SubnetSet will fail with message

```yaml
kind: SubnetPort
metadata:
  annotations:
      {"apiVersion":"crd.nsx.vmware.com/v1alpha1","kind":"SubnetPort","metadata":{"annotations":{},"name":"subnetport-6","namespace":"ns-1"},"spec":{"interfaceIPType":"IPv4IPv6","subnet":"subnet-3"}}
  creationTimestamp: "2026-08-21T03:17:41Z"
  generation: 1
  name: subnetport-6
  namespace: ns-1
  resourceVersion: "11570284"
  uid: 5364407d-00e0-4f98-8d54-f362b876c2dd
spec:
  interfaceIPType: IPv4IPv6
  subnet: subnet-3
status:
  attachment: {}
  conditions:
  - lastTransitionTime: "2026-08-21T03:17:41Z"
    message: 'error occurred while processing the SubnetPort CR. Error: Subnet/SubnetSet
      IPAddressType "IPv4" is not a superset of interfaceIPType "IPv4IPv6"'
    reason: SubnetPortNotReady
    status: "False"
    type: Ready
  networkInterfaceConfig: {}
```


pkg/controllers/subnetbinding/subnetbinding_controller.go
```diff
@@ -528,7 +545,12 @@ func (r *Reconciler) validateVpcSubnetsBySubnetSetCR(ctx context.Context, namesp
 	for i := range subnets {
 		subnetPaths[i] = *subnets[i].Path
 	}
-	return subnetPaths, nil
+
+	ipType := subnetSetCR.Spec.IPAddressType
+	if ipType == "" {
+		ipType = v1alpha1.IPAddressTypeIPv4
+	}
+	return subnetPaths, subnetSetCR.Spec.IPAddressType, nil
```
wenqiq: Should return `subnetPaths, ipType, nil` here?

pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -1135,6 +1145,130 @@ func (r *SubnetPortReconciler) updateSubnetStatusOnSubnetPort(subnetPort *v1alph
...
+					hasIPv4 = true
+				} else {
+					hasIPv6 = true
+				}
+			}
+		}
+	}
+
+	isMissing := false
+	switch subnetPort.Spec.StaticIPAllocationType {
```
wenqiq: When `StaticIPAllocationType` is empty "", the switch statement has no case matching "", so `isMissing` remains false. It bypasses IP family realization checks when StaticIPAllocationType is omitted.

pkg/controllers/subnetbinding/subnetbinding_controller.go
yanjunz97: Updated, thanks

pkg/controllers/subnetport/subnetport_controller.go
yanjunz97: StaticIPAllocationType shall be backfilled at this step. Added the default case to return error in this case.

yanjunz97:
This's a discussion regarding BindingMap ipaddresstype validation. We may need to consider more cases for it.
Wenqi will take take of them in PR #1550 . So I remove the bindingmap related changes in this PR.

pkg/apis/vpc/v1alpha1/subnetport_types.go
```diff
@@ -40,6 +43,7 @@ type SubnetPortSpec struct {
 	// will be back-filled based on InterfaceIPType and Subnet configuration.
 	// Supported starting with VCF 9.2.0.
 	// +kubebuilder:validation:Enum=IPv4;IPv6;IPv4IPv6;None
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
```
yanjunz97, line 46: Hi @jianjuns Could you help review if it is fine for us to set StaticIPAllocationType/InterfaceIPType as immutable for SubnetPort CR?
Here's the context:

From supervisor perspective, we do not support workload ipfamily update in 9.2, which means vm operator will not allow user to update the vm ipfamily after the vm is created.

From NSX perspective, NSX API supports SubnetPort.StaticIPAllocationType widen, this means None->IPv4/IPv6/IPv4IPv6 and IPv4/IPv6->IPv4IPv6 is supported, but IPv4IPv6 -> IPv4/IPv6/None or IPv4 <-> IPv6 or IPv4/IPv6 -> None is not supported. By not support, it currently does not report error but the allocated IP will not be released with the update.

The intuitive to set StaticIPAllocationType immutable is to keep consistent with the supervisor side and avoid some unexpected behavior. For example if vm operator/mobility operator accidentally modifies the SubentPort StaticIPAllocationType, it can receive clear error message instead of getting the SubnetPort StaticIPAllocationType updated but not get the actual IP released/assigned.

But as this will make SubnetPort CRD has different behavior as NSX side, would like to see your comment on this.

jianjuns: I think it is fine to set these fields as immutable.

heypnus (dismissed):

pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -164,7 +167,10 @@ func (r *SubnetPortReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			if nsxSubnet != nil && nsxSubnet.IpAddressType != nil {
 				parentIPAddressType = common.ConvertNSXIPAddressTypeToCR(*nsxSubnet.IpAddressType)
 			}
-			interfaceIPType = subnetport.GetDefaultInterfaceIPType(interfaceIPType, parentIPAddressType)
+			interfaceIPType, err = subnetport.GetDefaultInterfaceIPType(interfaceIPType, parentIPAddressType)
```
heypnus, line 170: Could we preserve `subnetPort.Spec.InterfaceIPType` when it is already set, and only derive it from the parent when the spec is empty? Otherwise, an existing IPv6-only port on an `IPv4IPv6` subnet appears to be t rated as IPv4 in the AddressBinding restore path. Would it make sense to add a regression test for this case?

yanjunz97: Good caught, updated, thanks!

wenqiq (approved):
LGTM

heypnus (approved):
LGTM. Thanks for addressing the comments!

PR #1537, opened 2026-08-18
refactor: use standard workqueue and default rate limiting

TaoZou1:
refactor: use standard workqueue and default rate limiting

    This commit refactors the VPCNetworkConfigurationHandler and Kubernetes
    client configuration to use standard libraries and defaults:

    1. Replaced the custom slice-based task queue in VPCNetworkConfigurationHandler
       with client-go's standard workqueue (TypedRateLimitingQueue). This provides
       built-in deduplication and exponential backoff retry mechanisms for failed
       NSX-T API calls, preventing potential memory leaks and ensuring eventual
       consistency.

    2. Removed custom K8S_CLIENT_QPS and K8S_CLIENT_BURST environment variable
       overrides, reverting to client-go's default unlimited rate limiting
       behavior (QPS = -1) to prevent unintended client-side throttling.
       
       
Test done:
test case 1:
run the manager, check the timestamp:
2026-08-19 00:50:26.000 INFO main.go:317 I'm the master now
2026-08-19 00:50:52.000 INFO controller.go:320 Starting workers controller=subnet ...

### DIFF
--- a/pkg/controllers/networkinfo/vpcnetworkconfig_handler.go
+++ b/pkg/controllers/networkinfo/vpcnetworkconfig_handler.go
@@ -3,8 +3,9 @@ package networkinfo
 import (
 	"context"
 	"reflect"
-	"sync"
+	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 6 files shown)


pkg/controllers/networkinfo/vpcnetworkconfig_handler.go
```diff
@@ -29,10 +29,6 @@ const (
 	vpcTaskDelete
 )
 
-const (
-	vpcTaskQueueCapacity = 1000
-)
-
 type vpcTask struct {
 	taskType  vpcTaskType
 	vpcConfig *v1alpha1.VPCNetworkConfiguration
```
yanjunz97: Not related to this change, but I just notice using vpcTask as key may introduce some race conditions.
As vpcConfig is a pointer to VPCNetworkConfiguration with taskType, for the same CR, i.e. the CR with the same Name, it will be considered different key in the work queue.
If it goes through this order:
- update VPCNetworkConfiguration A, (update, A) is put into the queue
- delete VPCNetworkConfiguration A, (delete, A) is put into the queue
- UpdateIPBlocksInfo failed for  (update, A), (update, A) is put into the queue again
- SyncIPBlocksInfo succeeded for (delete, A)
- UpdateIPBlocksInfo added the deleted ipblocks for VPCNetworkConfiguration A
Then the ipblock associated with deleted VPCNetworkConfiguration A will be added to the ipblocksinfo.

I'm thinking about either only queue VPCNetworkConfiguration Name and get CR from k8s client in workerLoop; or get CR from k8s client in UpdateIPBlocksInfo

TaoZou1: Really a good catch! I have updated the key

yanjunz97 (approved):
LGTM

dantingl (approved):

pkg/util/kubernetes.go
```diff
@@ -19,34 +17,12 @@ const (
 	localhostIPv6           = "::1"
 	defaultK8sServicePort   = "6443"
 	K8sServicePortEnv       = "KUBERNETES_SERVICE_PORT"
-	K8sClientQPSEnv         = "K8S_CLIENT_QPS"
-	K8sClientBurstEnv       = "K8S_CLIENT_BURST"
 	K8sClientTimeoutEnv     = "K8S_CLIENT_TIMEOUT"
 	CacheSyncTimeoutEnv     = "CACHE_SYNC_TIMEOUT"
-	DefaultK8sClientQPS     = float32(100)
-	DefaultK8sClientBurst   = 200
```
dantingl, line 27: Could you check if we need to set QPS = -1 to make it unlimited?

pkg/controllers/networkinfo/vpcnetworkconfig_handler.go
```diff
@@ -22,79 +23,74 @@ import (
...
+	return h.ipBlocksInfoService.UpdateIPBlocksInfo(context.Background(), vpcConfig)
 }
 
 func (h *VPCNetworkConfigurationHandler) Create(_ context.Context, e event.CreateEvent, _ workqueue.TypedRateLimitingInterface[reconcile.Request]) {
-	vpcConfigCR := e.Object.(*v1alpha1.VPCNetworkConfiguration)
-	// Enqueue update task to single background worker thread to avoid blocking Informer CacheSync
-	h.enqueueTask(vpcTask{
-		taskType:  vpcTaskUpdate,
-		vpcConfig: vpcConfigCR,
-	})
```
dantingl, line 89: Do you need to cehck if e.Object == nil

pkg/util/kubernetes.go
TaoZou1: no, the lib has set the default value as -1.

pkg/controllers/networkinfo/vpcnetworkconfig_handler.go
TaoZou1: Controller-runtime and client-go informer should guarantee that e.Object will not be nil.

pkg/controllers/networkinfo/vpcnetworkconfig_handler.go
```diff
@@ -22,79 +23,74 @@ import (
...
+			log.Error(err, "Failed to process VPCNetworkConfiguration task, requeuing", "VPCNetworkConfiguration", req.Name)
+			h.queue.AddRateLimited(req)
+		} else {
+			h.queue.Forget(req)
+		}
+		h.queue.Done(req)
+	}
+}
+
+func (h *VPCNetworkConfigurationHandler) processTask(req types.NamespacedName) error {
```
heypnus, line 69: Using the name as the key fixes the stale-object retry from the earlier thread, but I think there’s still an opposite edge case here. If a config is deleted and recreated with the same name before this worker picks up the key, Get() sees the new object and only runs the incremental UpdateIPBlocksInfo(). The delete’s full sync is then skipped, so IP blocks that only belonged to the old object can hang around until the next periodic sync. Could we keep the UID or a fullSync flag in the queued item, so a delete/recreate always triggers SyncIPBlocksInfo()?

TaoZou1: Cool!  That's a edge case. For vpcnetworkconfiguration, it's created by WCP, the name is  'ns name+uuid',
so it's not easy to delete/create the same name CR.

PR #1539, opened 2026-08-20
Leafcert

TaoZou1:
fix(tls): fix WCP leaf cert exact match and noBalancerClient validation

Test Done:
1. set the insecure=false
2. check if the connection has been setup

### DIFF
--- a/pkg/nsx/cluster.go
+++ b/pkg/nsx/cluster.go
@@ -309,7 +309,7 @@ func (cluster *Cluster) createHTTPClient(tr *Transport, timeout time.Duration) *
 
 func (cluster *Cluster) createNoBalancerClient(timeout, idle time.Duration) *http.Client {
 	// #nosec G402: ignore insecure options
-	tlsConfig := tls.Config{InsecureSkipVerify: true}
+	tlsConfig := tls.Config{InsecureSkipVerify: cluster.config.Insecure}
 	transport := &http.Transport{
 		TLSClientConfig: &tlsConfig,
 		IdleConnTimeout: idle * time.Second,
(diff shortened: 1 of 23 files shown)


pkg/nsx/cluster.go
```diff
@@ -241,56 +241,59 @@ func (cluster *Cluster) getCaFile(addr string) string {
...
+				log.Error(err, "Create transport", "get TLS config from cert", cafile)
+				return nil, err
+			}
+		} else {
+			thumbprint := cluster.getThumbprint(addr)
+			tpCount := len(cluster.config.Thumbprint)
+			log.Info("Create Transport", "thumbprint", thumbprint, "tpCount", tpCount)
+			// #nosec G402: ignore insecure options
+			config = &tls.Config{
+				InsecureSkipVerify: true,
```
yanjunz97: Shall we report error or follow golang default verification with `tls.Config{}` if insecure is set to false while ca is not provided to avoid this hard code InsecureSkipVerify?

pkg/nsx/util/utils.go
```diff
@@ -596,8 +596,10 @@ func GetTLSConfigForCert(pemCerts []byte) (*tls.Config, error) {
 		return nil, err
 	}
 
+	log.Info("DEBUG CERT", "IsCA", cert.IsCA, "Subject", cert.Subject.String(), "IPAddresses", cert.IPAddresses, "DNSNames", cert.DNSNames)
```
yanjunz97: Is this debug log? Maybe use debug level or remove it

TaoZou1: it's a info log.
update it to log.Info("Dump CERT",...)

pkg/nsx/cluster.go
TaoZou1: you're right. using the tls.Config{} to align with NCP.

yanjunz97 (approved):
LGTM

PR #1444, opened 2026-06-01
Fix one panic error when sts processDelete

TaoZou1:
Pass nil to processDelete to avoid panic

Test Done:
run ut

PR #1441, opened 2026-05-25
Remove SubnetPort when Pod is in terminating state

TaoZou1:
Previously, the logic ignored Pods in a terminating state to allow for graceful shutdown. However, if a Namespace (NS) or StatefulSet (STS) is deleted and immediately recreated with the same name, the lingering terminating Pods can interfere with garbage collection (GC) or cause resource conflicts.

This change ensures that the SubnetPort is promptly removed when a Pod enters the terminating state, preventing GC failures during rapid delete-and-recreate scenarios.

Test Done:
1. create ns dev
2. create sts sample-sts
3. stop the operator
4. delete the ns dev from VC UI
5. create ns dev and sts again
6. restart the operator
7. check if there is log:
2026-05-27 01:43:44.000 DEBUG statefulset_controller.go:331 Found orphaned subnet port for deleted StatefulSet port=simple-sts-0_n76bx stsUID=610e9b2d-e275-4655-bec2-d0fad89ad3c6
2026-05-27 01:43:45.000 DEBUG statefulset_controller.go:331 Found orphaned subnet port for deleted StatefulSet port=simple-sts-1_n76bx stsUID=610e9b2d-e275-4655-bec2-d0fad89ad3c6

PR #1541, opened 2026-08-25
[cherry-pick]: statefulset feature and fix

TaoZou1:
Cherry-pick 
376b4b16 (HEAD -> tzou-cp-9.1.2-sts, origin/tzou-cp-9.1.2-sts) fix(sts): fix pod termination release semantics and statefulset ordinal range handling
02bfa7bf refactor(nsx): remove ConfigMap vpc_wcp_enhance check for StatefulSet feature
6408ed54 fix(pod): force reuse existing SubnetPort for StatefulSet pods (#1548)
b22f138a Fix one panic error when sts processDelete (#1444)
04b022a9 Remove SubnetPort when Pod is in terminating state (#1441)
3e60e3ec Add StatefulSet pod support for subnet ports (#1383)


pkg/nsx/feature.go
```diff
@@ -0,0 +1,14 @@
+/* Copyright © 2026 VMware, Inc. All Rights Reserved.
```
dantingl, line 1: Should the file be consistent with main branch as feature.go?

TaoZou1: fixed

yanjunz97 (dismissed):
Overall LGTM
I see "remove ConfigMap vpc_wcp_enhance check for StatefulSet feature" is only applied on 912. Just double check, if we would like to remove it on main for now or we will remove it later?

TaoZou1:
> Overall LGTM I see "remove ConfigMap vpc_wcp_enhance check for StatefulSet feature" is only applied on 912. Just double check, if we would like to remove it on main for now or we will remove it later?

main branch still need it now. we'll  remove it later when NSX remove the it first

pkg/controllers/statefulset/statefulset_controller.go
```diff
@@ -0,0 +1,487 @@
...
+		return false
+	},
+	CreateFunc: func(e event.CreateEvent) bool {
+		return false
+	},
+	DeleteFunc: func(e event.DeleteEvent) bool {
+		return true
+	},
+	GenericFunc: func(e event.GenericEvent) bool {
+		return false
```
heypnus, line 497: 🟠 **High**: When a StatefulSet is deleted, Kubernetes sets `metadata.deletionTimestamp`, triggering an update event. Because `PredicateFuncsForStatefulSet.UpdateFunc` only checks for shrinking ordinal ranges, events where `newSts.DeletionTimestamp != nil` are dropped, preventing `r.processDelete` from running during deletion.

```suggestion
UpdateFunc: func(e event.UpdateEvent) bool {
		oldSts, ok1 := e.ObjectOld.(*appsv1.StatefulSet)
		newSts, ok2 := e.ObjectNew.(*appsv1.StatefulSet)
		if !ok1 || !ok2 {
			log.Error(fmt.Errorf("type assertion failed"), "Failed to cast to StatefulSet in update event")
			return false
		}

		if newSts.DeletionTimestamp != nil {
			return true
		}
```

pkg/controllers/statefulset/statefulset_controller.go
```diff
@@ -0,0 +1,487 @@
...
+					log.Info("Pod UID mismatch, deleting subnet port ", "pod", podName, "podUID", pod.UID, "subnetPortUID", podUid)
+				} else if !common.PodIsDeleted(pod) {
+					log.Info("Pod still exists, skipping subnet port deletion", "pod", podName)
+					pendingRunningPod = true
+					continue
+				}
+			}
+		}
+
+		if err := r.SubnetPortService.DeleteSubnetPort(subnetPort); err != nil {
```
heypnus, line 199: 🟠 **High**: If `r.Client.Get` fails with a transient error (other than `NotFound`), the error is ignored and execution proceeds to `DeleteSubnetPort`, which can delete the subnet port of an active, running pod.

pkg/controllers/statefulset/statefulset_controller.go
TaoZou1: for  sts without finalizer, there is no update event when deleting the statefulset.
Fix that in case if the finalizer will be added.

pkg/controllers/statefulset/statefulset_controller.go
TaoZou1: fixed

dantingl (approved):

yanjunz97 (approved):

PR #1543, opened 2026-08-25
[cherry-pick] Retry on empty bindingmap in scaled case (#1457)

yanjunz97:
In scaled case, after HAPI succeeds, NSX Subnet ConnectionBindingMap may not exist in the return of list API. Currently we always read it and save to store. It will hit a nil pointer dereference failure. This PR
- checks if the Subnet ConnectionBindingMap exists and adds a retry logic to get the NSX Subnet ConnectionBindingMap list
- adds a loop to get all the NSX Subnet ConnectionBindingMap through list API

### DIFF
--- a/pkg/nsx/services/subnetbinding/subnetbinding.go
+++ b/pkg/nsx/services/subnetbinding/subnetbinding.go
@@ -2,15 +2,18 @@ package subnetbinding
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


wenyingd (approved):
LGTM

wenqiq (approved):

PR #1545, opened 2026-08-25
fix(mixed-mode): ignore namespace on conflicting T1 and VPC annotations

heypnus:
When a namespace has conflicting T1 and VPC annotations, treat it as
an illegal state, log an error, and ignore it (fail-close).

Testing Done:
1. Automation Tests:
   Pipelines for T1 and VPC passed, details refer to the NCP patch info.

2. Manual Test on Supervisor Testbed:
   Step 1. Enable mixed-mode capability on Supervisor testbed.
   Step 2. Verify startup scan logs with namespace context:
           Started nsx-operator on a testbed with conflicting namespace xxx.
           Verified the error log explicitly records the namespace name:
           > 2026-09-07T02:44:45.890442308Z stdout F 2026-09-07 02:44:45.000 ERROR mixed_mode.go:75 Conflicting network provider annotations (both T1 and VPC are present); namespace will be ignored namespace=xxx
   Step 3. Verify fail-close behavior:
           Created a SecurityPolicy in the conflict namespace and confirmed that the SecurityPolicy was not reconciled.

### DIFF
--- a/pkg/config/mixed_mode.go
+++ b/pkg/config/mixed_mode.go
@@ -50,6 +50,19 @@ func hasT1ActivationAnnotation(annotations map[string]string) bool {
 	return false
 }
 
+// resolveNamespaceProvider determines whether a namespace belongs to T1 and/or VPC
+// provider based on its annotations. If both VPC and T1 annotations are present on
+// the same namespace, this is an illegal state; it logs an error and returns (false, false).
+func resolveNamespaceProvider(annotations map[string]string) (isT1 bool, isVPC bool) {
+	hasVPC := strings.TrimSpace(annotations[VPCNetworkConfigAnnotation]) != ""
+	hasT1 := hasT1ActivationAnnotation(annotations)
+	if hasVPC && hasT1 {
+		log.Error(nil, "Conflicting network provider annotations (both T1 and VPC are present); namespace will be ignored")
+		return false, false
+	}
+	return hasT1, hasVPC
+}
+
 var (
 	capabilitiesGVR = schema.GroupVersionResource{
 		Group:    "iaas.vmware.com",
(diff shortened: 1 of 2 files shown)


pkg/config/mixed_mode.go
```diff
@@ -429,7 +453,8 @@ func IsVPCNamespace(ns *v1.Namespace) bool {
 		return false
 	}
 	if IsPerNamespaceProvidersSupported() {
-		return strings.TrimSpace(ns.Annotations[VPCNetworkConfigAnnotation]) != ""
+		_, isVPC := resolveNamespaceProvider(ns.Annotations)
+		return isVPC
```
lxiaopei, line 483: in main branch, for mixed_mode.go, it has IsT1Namespace func, we should also update IsT1Namespace func to use resolveNamespaceProvider

lxiaopei: also add a unit test for IsT1Namespace with dual annotations

heypnus: Good catch! The issue is because when I was writing this PR 1545, the PR 1413 including the new function IsT1Namespace was not merged. Fixed and added UT, thanks!

pkg/config/mixed_mode.go
```diff
@@ -50,6 +50,24 @@ func hasT1ActivationAnnotation(annotations map[string]string) bool {
...
+func resolveNamespaceProvider(annotations map[string]string) (isT1 bool, isVPC bool) {
+	hasVPC := strings.TrimSpace(annotations[VPCNetworkConfigAnnotation]) != ""
+	hasT1 := hasT1ActivationAnnotation(annotations)
+	if hasVPC && hasT1 {
+		stateMu.RLock()
+		enableVPC := storedEnableVPCNetwork
+		stateMu.RUnlock()
+		if enableVPC {
+			return false, true
+		}
```
TaoZou1, line 79: how about?
if hasVPC && hasT1 {
		stateMu.RLock()
		defer stateMu.RUnlock()
		return !storedEnableVPCNetwork, storedEnableVPCNetwork
	}

heypnus: Good point! done.

dantingl (approved):

lxiaopei (approved):
LGTM

PR #1547, opened 2026-08-26
Add snapshot unit tests framework

yanjunz97:
This PR introduce the snapshot framework and an example on how the snapshot test works with UT.

It utilizes the go-snap but has some differences:
- go-snap does not have a feature based guard, this PR defines a metadata for customization
- go-snap generates the snapshot in each folder where the unit test locates, this PR uses a centralized folder to save all the snapshot files grouped by feature
- go-snap global GC cannot handle the centralized folder correctly, so for feature renaming or deletion, this PR introduces a command `make test-snaps-cleanup` to delete stale feature. This is only required for feature renaming/deletion.

Update the snapshot
```
UPDATE_SNAPS=true go test -gcflags=all=-l ./pkg/controllers/networkinfo/...
```

Update the code and unit test to break the previous design. Run unit test will fail
```
go test -gcflags=all=-l ./pkg/controllers/networkinfo/...                  
--- FAIL: TestNetworkInfoReconciler_UpdateDefaultSubnetSet (0.13s)
    --- FAIL: TestNetworkInfoReconciler_UpdateDefaultSubnetSet/Existing_default_SubnetSet_with_subnetNames_and_IPAddressType_change_(no_delete) (0.02s)
        networkinfo_controller_test.go:2793: 
            	Error Trace:	/Users/zhouya/Documents/go/src/ncp/nsx-operator/pkg/controllers/networkinfo/networkinfo_controller_test.go:2793
            	Error:      	Should be true
            	Test:       	TestNetworkInfoReconciler_UpdateDefaultSubnetSet/Existing_default_SubnetSet_with_subnetNames_and_IPAddressType_change_(no_delete)
        networkinfo_controller_test.go:2794: 
            - Snapshot - 23
            + Received  + 1
            
            @@ -5,28 +5,6 @@
            
                "scenario": "default_subnetset_update"
               },
               "payload": {
            -   "metadata": {
            -    "labels": {
            -     "nsx.vmware.com/default-network": "pod"
            -    },
            -    "name": "pod-default",
            -    "namespace": "ns-1",
            -    "resourceVersion": "MASKED_RESOURCE_VERSION"
            -   },
            -   "spec": {
            -    "accessMode": "Public",
            -    "ipAddressType": "IPv4IPv6",
            -    "ipv4SubnetSize": 32,
            -    "subnetDHCPConfig": {
            -     "dhcpServerAdditionalConfig": {}
            -    },
            -    "subnetDHCPv6Config": {
            -     "dhcpv6ServerAdditionalConfig": {}
            -    },
            -    "subnetNames": [
            -     "subnet-1"
            -    ]
            -   },
            -   "status": {}
            +   "exists": false
               }
              }
            
            at ../../../test/unittests/snapshots/tepless.snap.json:193
            
FAIL
FAIL	github.com/vmware-tanzu/nsx-operator/pkg/controllers/networkinfo	4.607s
FAIL
```

Cleanup stale feature snapshot: `make test-snaps-cleanup`


TaoZou1:
This POC excellently demonstrates the value of snapshot testing for our complex CRDs. However, maintaining a custom snapshot framework (handling concurrency, scrubbing, GC, and CI safety) will introduce unnecessary technical debt. I recommend replacing this custom implementation with a mature, battle-tested library like github.com/gkampitakis/go-snaps. It provides out-of-the-box CI protection, better diff formatting, dynamic scrubbing, and parallel test support without the need for global file locks.

yanjunz97:
> This POC excellently demonstrates the value of snapshot testing for our complex CRDs. However, maintaining a custom snapshot framework (handling concurrency, scrubbing, GC, and CI safety) will introduce unnecessary technical debt. I recommend replacing this custom implementation with a mature, battle-tested library like github.com/gkampitakis/go-snaps. It provides out-of-the-box CI protection, better diff formatting, dynamic scrubbing, and parallel test support without the need for global file locks.

Thanks Tao, I updated the code to use go-snaps. But there's still some gaps between go-snaps lib and what we want, so I keep some of the customization code. The main differences are
- go-snap does not have a feature based guard, I define a metadata for customization
- go-snap generates the snapshot in each folder where the unit test locates, I feel it's annoying and prefer to save them into single folder and grouped by feature
- As we put all the snapshot in the same folder, we cannot use go-snap global GC. It will delete the snap for test not run, which means if I have 2 feature snap, but only run unit test with UPDATE_SNAPS=true for part of the unit test covering only 1 feature, another feature snap will be deleted.  So instead of using the global GC provided by go-snap, I added a command `make test-snaps-cleanup` which shall be ran when a feature is renamed or deleted. The normal updated, e.g. old unit test deleted from a feature can be still handled automatically by go-snap.
- I also keep sanitize utility, it mainly works for k8s resource sanitization to remove those fields may change among tests, i.e. uuid/timestamp/resource version

hack/snaps-cleanup.sh
```diff
@@ -0,0 +1,44 @@
...
+# Find all .snap files in the centralized snapshots directory
+SNAP_FILES=($(find "$SNAPSHOTS_DIR" -maxdepth 1 -name "*.snap" 2>/dev/null || true))
+
+if [ ${#SNAP_FILES[@]} -eq 0 ]; then
+    echo "No snapshot files found."
+    exit 0
+fi
+
+# Find all Go files in the repository, excluding standard non-code or large directories
+GO_FILES=$(find . -type d \( -name .git -o -name .cache -o -name vendor -o -name node_modules -o -name bin -o -name build -o -name snapshots \) -prune -o -type f -name "*.go" -print)
```
heypnus: 🟡 **Medium**: If no Go files are found (for instance, due to an unexpected working directory or `find` failure), `GO_FILES` will be empty. The script will subsequently treat all snapshot files as unreferenced and delete them all. Add a check to abort if `GO_FILES` is empty.

```suggestion
GO_FILES=$(find . -type d \( -name .git -o -name .cache -o -name vendor -o -name node_modules -o -name bin -o -name build -o -name snapshots \) -prune -o -type f -name "*.go" -print)

if [ -z "$GO_FILES" ]; then
    echo "Error: No Go files found to inspect. Aborting snapshot cleanup." >&2
    exit 1
fi
```

hack/snaps-cleanup.sh
```diff
@@ -0,0 +1,44 @@
...
+            found=true
+            break
+        fi
+    done
+    
+    # If the feature name is not referenced in any Go file, delete the snapshot file
+    if [ "$found" = false ]; then
+        echo "Deleting obsolete snapshot file: $snap"
+        rm -f "$snap"
+    fi
```
heypnus: 🟡 **Medium**: Iterating over each Go file in Bash and spawning an individual `grep` process per file results in high overhead (`O(snapshots * go_files)` forks). Additionally, `grep` without `-F` treats feature names as regexes rather than literal strings.

```suggestion
    # Check if the feature name is referenced in any Go file, and delete if obsolete
    if ! echo "$GO_FILES" | xargs grep -q -F "$feature_name" 2>/dev/null; then
        echo "Deleting obsolete snapshot file: $snap"
        rm -f "$snap"
    fi
```

test/unittests/snapshot.go
```diff
@@ -0,0 +1,144 @@
...
+func (g *TestGuard) sanitize(input interface{}) interface{} {
+	bytes, err := json.Marshal(input)
+	if err != nil {
+		return input
+	}
+
+	var data interface{}
+	if err := json.Unmarshal(bytes, &data); err != nil {
+		return input
+	}
```
heypnus, line 114: 🟡 **Medium**: Errors during json.Marshal and json.Unmarshal in sanitize are silently ignored, causing unmasked data to pass through into snapshots. Fail the test explicitly with g.t.Fatalf so serialization failures are caught immediately.

```suggestion
	bytes, err := json.Marshal(input)
	if err != nil {
		g.t.Fatalf("failed to marshal input for sanitization: %v", err)
	}

	var data interface{}
	if err := json.Unmarshal(bytes, &data);	err != nil {
		g.t.Fatalf("failed to unmarshal input for sanitization: %v", err)
	}
```

test/unittests/snapshot.go
```diff
@@ -0,0 +1,144 @@
...
+			// Identify dynamic fields to mask
+			switch strings.ToLower(k) {
+			case "uid":
+				m[k] = "MASKED_UID"
+			case "resourceversion":
+				m[k] = "MASKED_RESOURCE_VERSION"
+			case "creationtimestamp", "timestamp", "deletiontimestamp":
+				if v != nil && v != "" {
+					m[k] = "MASKED_TIMESTAMP"
+				}
```
heypnus, line 153: 🟡 **Medium**: Nil or empty uid and resourceversion fields are unconditionally masked as present. Add a non-empty check (similar to timestamp masking) to avoid masking unassigned or zero-value fields.

```suggestion
			switch strings.ToLower(k) {
			case "uid":
				if v != nil && v != "" {
					m[k] = "MASKED_UID"
				}
			case "resourceversion":
				if v != nil && v != "" {
					m[k] = "MASKED_RESOURCE_VERSION"
				}
			case "creationtimestamp", "timestamp", "deletiontimestamp":
				if v != nil && v != "" {
					m[k] = "MASKED_TIMESTAMP"
				}
```

test/unittests/snapshot.go
```diff
@@ -0,0 +1,144 @@
...
+	default:
+		return val
+	}
+}
+
+// sanitizeFileName replaces invalid file path characters with underscores.
+func sanitizeFileName(name string) string {
+	reg := regexp.MustCompile(`[^a-zA-Z0-9_\-]`)
+	return reg.ReplaceAllString(name, "_")
+}
```
heypnus, line 174: 🔵 **Low**: regexp.MustCompile parses and compiles the regular expression on every call to sanitizeFileName. Move the compiled regex to a package-level variable to avoid redundant compilation.

```suggestion
var sanitizeFileNameRegex = regexp.MustCompile(`[^a-zA-Z0-9_\-]`)

func sanitizeFileName(name string) string {
	return sanitizeFileNameRegex.ReplaceAllString(name, "_")
}
```

TaoZou1:
Category 1: Critical Bugs & Functional Defects
[CRITICAL] hack/snaps-cleanup.sh inadvertently deletes all valid snapshot files

Location: hack/snaps-cleanup.sh:35
Description: The script uses echo "$GO_FILES" | xargs grep -q -F -e "\"$feature_name\"" 2>/dev/null. When the file list is large, or on platforms with tighter environment/argument limits (such as macOS or sandboxed CI environments), xargs fails with errors like sysconf(_SC_ARG_MAX) failed. Because errors are redirected to /dev/null and wrapped in if ! ..., any failure of xargs is treated as "feature name not found in any Go file", causing rm -f "$snap" to immediately execute.
Impact: Running make test-snaps-cleanup or ./hack/snaps-cleanup.sh deletes all existing .snap files in the repository regardless of whether they are actively referenced.
Recommendation: Avoid piping hundreds of filenames through xargs. Instead, use native recursive search (e.g., git grep -q -F "\"$feature_name\"" or grep -rqF --include="*.go" --exclude-dir={.git,.cache,vendor,bin,build,snapshots} "\"$feature_name\"" .). Ensure command execution failures are not blindly swallowed by /dev/null.
[CRITICAL] Runtime panic risk in maskFields due to uncomparable interface types

Location: test/unittests/snapshot.go:124, 128, 132
Description: The code checks if v != nil && v != "" where v is of type interface{}. In Go, comparing an interface{} to a string literal "" when the underlying concrete value is an uncomparable type (such as []interface{} or map[string]interface{}) triggers an immediate runtime panic: panic: runtime error: comparing uncomparable type []interface {}.
Impact: Any nested object where a field named uid or timestamp contains a complex structure or slice will crash test execution.
Recommendation: Safely type-assert to string before comparing:
if s, ok := v.(string); ok && s != "" {
    m[k] = "MASKED_UID"
}
[HIGH] Global state pollution and unexpected file mutation in TestAssertOutput_Success

Location: test/unittests/snapshot_test.go:144
Description: TestAssertOutput_Success sets os.Setenv("UPDATE_SNAPS", "true"). Environment variables are process-wide. When running unit tests with parallel package execution (go test ./... -parallel), this modifies the environment for other concurrently executing test suites, potentially causing them to regenerate/update snapshots instead of validating against them. Furthermore, every regular test run will re-generate my_test_feature.snap on disk.
Recommendation: Test snapshot assertion without mutating the global process environment, or isolate environment overrides inside a mock/in-memory buffer test.
Category 2: Architecture & Framework Integration
[HIGH] Loss of concurrency safety and cross-process race conditions from centralized storage

Location: test/unittests/snapshot.go:67-75, test/unittests/snapshots/
Description: go-snaps by default keeps .snap files adjacent to the test file within the respective package to guarantee parallel safety. Overriding this behavior by forcing all snapshots from different controllers/packages into a single centralized directory (test/unittests/snapshots/) creates concurrency issues:
When running go test ./..., different Go packages run in distinct OS processes.
If two packages share a feature name or concurrently access the centralized snapshots directory, there is no cross-process file-locking mechanism, leading to file write conflicts and corrupted snapshot files.
Recommendation: Adhere to the standard go-snaps convention by storing snapshot files in local __snapshots__ directories adjacent to each test package, or strictly evaluate and benchmark cross-package go test ./... -parallel execution to ensure no lock contention or corruption.
[MEDIUM] Bypassing native go-snaps GC increases maintenance burden

Location: hack/snaps-cleanup.sh, Makefile:96
Description: go-snaps provides built-in snapshot garbage collection (via snaps.Clean(t)) that automatically purges obsolete snapshots during test execution. Centralizing snapshots broke this native mechanism, necessitating a custom Bash script (hack/snaps-cleanup.sh) to detect orphaned files via text searching. This introduces an unnecessary external dependency and cognitive overhead for developers.
Recommendation: Consider leveraging go-snaps native lifecycle hooks (TestMain + snaps.Clean(t)) rather than relying on error-prone shell scripts to clean up snapshots.
Category 3: Edge Cases & Completeness
[MEDIUM] Incomplete dynamic field sanitization for Kubernetes resources

Location: test/unittests/snapshot.go:121-135
Description: maskFields only masks uid, resourceversion, creationtimestamp, timestamp, and deletiontimestamp. In Kubernetes CRD controllers:
status.conditions fields contain lastTransitionTime and lastUpdateTime, which are non-deterministic timestamps.
Metadata frequently includes generation and managedFields.
Impact: Any controller test capturing full CRs with Status.Conditions will produce non-deterministic snapshots that fail across consecutive runs.
Recommendation: Expand the case-insensitive masking list to cover lasttransitiontime, lastupdatetime, and other standard dynamic Kubernetes fields.

yanjunz97:
> [CRITICAL] hack/snaps-cleanup.sh inadvertently deletes all valid snapshot files

Updated the cleanup script

> [CRITICAL] Runtime panic risk in maskFields due to uncomparable interface types

updated the maskFields check

> [HIGH] Global state pollution and unexpected file mutation in TestAssertOutput_Success

Updated the test

> [HIGH] Loss of concurrency safety and cross-process race conditions from centralized storage

Updated the script to add -p 1 for updating snapshot

> [MEDIUM] Bypassing native go-snaps GC increases maintenance burden

go-snaps native lifecycle hooks cannot correctly take care of the centrialized folder, so it is bypassed

>  [MEDIUM] Incomplete dynamic field sanitization for Kubernetes resources

Added more fields for k8s resource

TaoZou1 (dismissed):

test/unittests/snapshot_test.go
```diff
@@ -0,0 +1,191 @@
...
+	defer os.Chdir(oldWd)
+
+	tempDir := t.TempDir()
+	err = os.Chdir(tempDir)
+	require.NoError(t, err)
+
+	// Since tempDir has no go.mod and none of its parents do, findProjectRoot should fail
+	root, err := findProjectRoot()
+	assert.Error(t, err)
+	assert.Empty(t, root)
```
wenqiq: `os.Chdir()` modifies the working directory of the entire operating system process. If other tests or goroutines run concurrently (or if `t.Parallel()` is enabled in the future), any file operations during this window will fail unpredictably.

test/unittests/snapshot.go
```diff
@@ -0,0 +1,174 @@
...
+		return m
+	default:
+		return val
+	}
+}
+
+var sanitizeFileNameRegex = regexp.MustCompile(`[^a-zA-Z0-9_\-]`)
+
+// sanitizeFileName replaces invalid file path characters with underscores.
+func sanitizeFileName(name string) string {
```
wenqiq: `sanitizeFileName()` transforms any non-alphanumeric character (e.g. `/`, `.`) into an underscore `_`. For example, `Feature: "vpc/subnet"` generates `vpc_subnet.snap`. In `snaps-cleanup.sh`, the extracted `feature_name` is `vpc_subnet`. The script searches for `Feature: "vpc_subnet"` in Go code, which does NOT match `Feature: "vpc/subnet"`. The valid snapshot file will be permanently deleted?

test/unittests/snapshot_test.go
yanjunz97: Removed the Chdir part, thanks!

test/unittests/snapshot.go
yanjunz97: Good caught. I fail the feature name with non-alphanumeric character instead of converting them. Thanks!

test/unittests/snapshot_test.go
```diff
@@ -0,0 +1,194 @@
...
+	root, err := findProjectRoot()
+	require.NoError(t, err)
+	assert.NotEmpty(t, root)
+
+	// Verify go.mod exists in the found root
+	_, err = os.Stat(filepath.Join(root, "go.mod"))
+	assert.NoError(t, err)
+}
+
+func TestFindProjectRoottFrom(t *testing.T) {
```
wenqiq: TestFindProjectRoottFrom --> TestFindProjectRootFrom ?

yanjunz97: Updated, thanks!

wenqiq (approved):

TaoZou1 (approved):

PR #1548, opened 2026-08-27
fix(pod): force reuse existing SubnetPort for StatefulSet pods

TaoZou1:
When a StatefulSet pod is recreated, a new Pod UID is assigned. If the original Subnet reaches full capacity, GetSubnetPathForPod failed to recognize the existing SubnetPort because it only queried by Pod UID. This caused a new Subnet to be allocated while builder.go still attempted to reuse the existing SubnetPort ID by stsUID, leading to a Subnet/Port path mismatch and NSX attachment ID conflict error. Fix this issue by:
1. Exporting GetStatefulSetInfo and GetExistingSubnetPortForStatefulSetPod in the subnetport service.
2. Checking for existing StatefulSet SubnetPorts via stsUID + podName in GetSubnetPathForPod, forcing the original Subnet path when found.
3. Adding portID to the reuse log in builder.go for observability.

Test Done:
Cast 1:
1. create rc with instance = 10
2. create the sts pod with instance = 3
3. update the rc with instance = 30
4. delete one of the sts pod
5. check if the pod could be recreated and the subnetport is reused

### DIFF
--- a/pkg/controllers/pod/pod_controller.go
+++ b/pkg/controllers/pod/pod_controller.go
@@ -434,6 +434,19 @@ func (r *PodReconciler) GetSubnetPathForPod(ctx context.Context, pod *v1.Pod) (b
 		log.Debug("NSX SubnetPort had been created, returning the existing NSX Subnet path", "pod.UID", pod.UID, "subnetPath", subnetPath)
 		return true, subnetPath, subnetSetUID, subnetSetLock, interfacetype, "", nil
 	}
+
+	// Check if this is a StatefulSet pod and we can reuse an existing SubnetPort
+	if nsx.StatefulSetPodSubnetPortFeatureEnabled(r.SubnetPortService.NSXClient, r.SubnetPortService.NSXConfig) {
+		stsUID := subnetport.GetStatefulSetUID(pod)
+		if port := r.SubnetPortService.GetExistingSubnetPortForStatefulSetPod(pod.Name, stsUID); port != nil {
+			if port.ParentPath != nil {
+				subnetPath = *port.ParentPath
+			}
+			log.Info("Found existing SubnetPort for StatefulSet pod, forcing old subnet", "podName", pod.Name, "stsUID", stsUID, "subnetPath", subnetPath)
+			return true, subnetPath, subnetSetUID, subnetSetLock, interfacetype, "", nil
+		}
+	}
+
 	log.Info("Got default SubnetSet for Pod, allocating the NSX Subnet", "subnetSet.Name", subnetSet.Name, "subnetSet.UID", subnetSet.UID, "pod.Name", pod.Name, "pod.UID", pod.UID)
 	if r.restoreMode {
 		// For restore case, Pod will be created on the Subnet with matching CIDR
@@ -465,7 +478,7 @@ func (r *PodReconciler) deleteSubnetPortByPodName(ctx context.Context, ns string
 
 	for _, nsxSubnetPort := range nsxSubnetPorts {
 		// Check if this subnet port was created for StatefulSet
-		// Only skip if STS feature is enabled (NSX 9.2.0+)
+		// Only skip if StatefulSet pod SubnetPort feature is enabled
 		if nsx.StatefulSetPodSubnetPortFeatureEnabled(r.SubnetPortService.NSXClient, r.SubnetPortService.NSXConfig) && r.isStatefulSetSubnetPort(nsxSubnetPort) {
 			log.Info("Ignoring subnet port deletion for StatefulSet pod",
 				"pod", name, "statefulset-uid", r.getStsUID(nsxSubnetPort))
--- a/pkg/nsx/services/subnetport/builder.go
+++ b/pkg/nsx/services/subnetport/builder.go
@@ -40,7 +40,7 @@ func (service *SubnetPortService) buildSubnetPort(obj interface{}, nsxSubnet *mo
 	if _, ok := obj.(*corev1.Pod); ok {
 		appId = string(objMeta.UID)
 	}
-	_, stsUID := getStatefulSetInfo(obj)
+	stsUID := GetStatefulSetUID(obj)
 	var externalAddressBinding *model.ExternalAddressBinding
 	var err error
 	var addressBindings []model.PortAddressBindingEntry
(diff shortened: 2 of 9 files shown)


pkg/controllers/pod/pod_controller.go
```diff
@@ -428,6 +428,17 @@ func (r *PodReconciler) GetSubnetPathForPod(ctx context.Context, pod *v1.Pod) (b
 		log.Debug("NSX SubnetPort had been created, returning the existing NSX Subnet path", "pod.UID", pod.UID, "subnetPath", subnetPath)
 		return true, subnetPath, subnetSetUID, subnetSetLock, interfacetype, nil
 	}
+
+	// Check if this is a StatefulSet pod and we can reuse an existing SubnetPort
+	if nsx.StatefulSetPodSubnetPortFeatureEnabled(r.SubnetPortService.NSXClient, r.SubnetPortService.NSXConfig) {
+		stsUID := subnetport.GetStatefulSetUID(pod)
+		if port := r.SubnetPortService.GetExistingSubnetPortForStatefulSetPod(pod.Name, stsUID); port != nil && port.ParentPath != nil {
+			subnetPath = *port.ParentPath
```
lxiaopei: Add check for empty *port.ParentPath

pkg/nsx/services/subnetport/builder.go
```diff
@@ -241,18 +241,32 @@ func (service *SubnetPortService) buildSubnetPort(obj interface{}, nsxSubnet *mo
...
-		return ref.Name, string(ref.UID)
+		return string(ref.UID)
 	}
-	return "", ""
+	return ""
+}
+
+// GetExistingSubnetPortForStatefulSetPod finds an existing SubnetPort for a StatefulSet pod
+func (service *SubnetPortService) GetExistingSubnetPortForStatefulSetPod(podName, stsUID string) *model.VpcSubnetPort {
+	if stsUID == "" {
```
lxiaopei, line 261: if stsUID == "" || service.SubnetPortStore == nil {
    return nil
}
in case of uninitialized service states.

TaoZou1: service.SubnetPortStore can be nil.
it should have value during service init

pkg/controllers/pod/pod_controller.go
TaoZou1: even the *port.ParentPath is ""here, this function can't return false.
else in the builder.go, operator will reuse the subnetport and create it in a new subnet
if it's "", operator will r.SubnetService.GetSubnetByPath(nsxSubnetPath, ...)：
and return 'Failed to get Subnet by path ""'

yanjunz97 (dismissed):

pkg/nsx/services/subnetport/builder.go
```diff
@@ -242,18 +242,32 @@ func (service *SubnetPortService) buildSubnetPort(obj interface{}, nsxSubnet *mo
...
+		return nil
+	}
+	existingPorts := service.SubnetPortStore.GetByIndex(common.TagScopeStatefulSetUID, stsUID)
+	for _, port := range existingPorts {
+		if nsxutil.FindTag(port.Tags, common.TagScopePodName) == podName {
```
dantingl, line 266: Can we use statefulset index tag to find port? is it always the same with podName

pkg/nsx/services/subnetport/builder.go
```diff
@@ -264,18 +278,13 @@ func (service *SubnetPortService) BuildSubnetPortIdAndName(obj *metav1.ObjectMet
 
 	// For StatefulSet pods: check if a SubnetPort with the same StatefulSet UID and pod name exists
 	// Note: STS UID is globally unique, so we only need to check pod name
-	// Only reuse if STS feature is enabled (NSX 9.2.0+ and vpc_wcp_enhance=true)
+	// Only reuse if STS feature is enabled (NSX supports StatefulSet pod SubnetPorts)
```
dantingl: We need to set NSX version to be 9.1.2+

pkg/controllers/pod/pod_controller.go
```diff
@@ -434,6 +434,19 @@ func (r *PodReconciler) GetSubnetPathForPod(ctx context.Context, pod *v1.Pod) (b
...
+
+	// Check if this is a StatefulSet pod and we can reuse an existing SubnetPort
+	if nsx.StatefulSetPodSubnetPortFeatureEnabled(r.SubnetPortService.NSXClient, r.SubnetPortService.NSXConfig) {
+		stsUID := subnetport.GetStatefulSetUID(pod)
+		if port := r.SubnetPortService.GetExistingSubnetPortForStatefulSetPod(pod.Name, stsUID); port != nil {
+			if port.ParentPath != nil {
+				subnetPath = *port.ParentPath
+			}
+			log.Info("Found existing SubnetPort for StatefulSet pod, forcing old subnet", "podName", pod.Name, "stsUID", stsUID, "subnetPath", subnetPath)
+			return true, subnetPath, subnetSetUID, subnetSetLock, interfacetype, "", nil
```
dantingl, line 446: Should line 445-446 put into `if port.ParentPath != nil` condition block?

TaoZou1: It can't be put into condition block.
If it didn't return true here, in the CreateOrUpdateSubnet, it will find the subnetport again and reuse it and create it in a new subnet.
If it return true here, when it returned from this function, it will be failed in function common.IsSharedSubnetPath(ctx, r.Client, nsxSubnetPath, req.Namespace) or r.SubnetService.GetSubnetByPath("", ...)

TaoZou1: no.
It should be put in the current place.
When the function returned,  it will be fail when calling the common.IsSharedSubnetPath(ctx, r.Client, nsxSubnetPath, req.Namespace) or r.SubnetService.GetSubnetByPath("", ...)

If not return true when port.ParentPath is nil, then in the CreateOrUpdateSubnet, it will try to reuse the subnetport in a new subnet which will be failed from nsx side.

pkg/nsx/services/subnetport/builder.go
TaoZou1: Yes, it used the version 9.1.2+.
Updated the comments

pkg/nsx/services/subnetport/builder.go
TaoZou1: use index need to extract the index from podName? 
pod_name tag is the same as podName.

dantingl (dismissed):

yanjunz97 (approved):

dantingl (approved):

PR #1553, opened 2026-09-01
Support reuse-port annotation for cpVM migration and skip vm-default …

seanpang-vmware:
…SubnetSet for kube-system in Supervisor 2.0

- Support reuse-port annotation for cpVM migration

### 1. Test Environment
* **Cluster / Host**: `10.161.60.138`
* **Component**: `nsx-operator` (running in `nsx-ncp` pod, namespace `vmware-system-nsx`)
* **Feature Flags**: `[nsx_v3] vpc_wcp_enhance = True`
---
### 2. Live Cluster Test Matrix
| # | Test Scenario | Annotation / Input | Expected Result | Actual Live Result | Status |
|:--|:---|:---|:---|:---|:--:|
| **1** | **Valid `reuse-port` annotation** | `nsx.vmware.com/reuse-port: "kube-system/vm-44"` | Reuses existing NSX SubnetPort (`vm-44_5rucs`), updates tags/status | SubnetPort reused, tags updated, `Ready: True` | **PASS** |
| **2** | **Non-existent target port** | `nsx.vmware.com/reuse-port: "kube-system/non-existent-vm"` | Fails reconcile, logs error, sets `Ready: False` | `Ready: False`<br>`reused port kube-system/non-existent-vm not found in runtime store` | **PASS** |
| **3** | **Empty annotation value** | `nsx.vmware.com/reuse-port: ""` | Explicitly reports error (does NOT fallback), sets `Ready: False` | `Ready: False`<br>`reused port cannot be empty` | **PASS** |
| **4** | **Invalid format annotation** | `nsx.vmware.com/reuse-port: "invalid-format-without-slash"` | Fails format validation, sets `Ready: False` | `Ready: False`<br>`invalid reuse-port annotation value ..., expected format <namespace>/<name>` | **PASS** |
| **5** | **Baseline CR without annotation** | *None* | Standard allocation via SubnetSet; allocates new IP and ID | Created new port (`test-vm-standard_5rucs`), IP `172.26.0.4/29`, `Ready: True` | **PASS** |
| **6** | **Skip `vm-default` SubnetSet for `kube-system`** | Reconcile `kube-system` namespace | Skips `vm-default` SubnetSet creation in Supervisor 2.0 | Logged: `Skipping default SubnetSet creation for kube-system in Supervisor 2.0` | **PASS** |
---
### 3. Key Verification Highlights
* **Strict Empty Value Validation**: Verified that `nsx.vmware.com/reuse-port: ""` is rejected with `reused port cannot be empty` and does not silently fall back to standard ID generation.
* **Backward Compatibility**: Standard SubnetPort CRDs without annotations continue normal SubnetSet allocation without issues.
* **Supervisor 2.0 Namespace Handling**: Confirmed that `vm-default` SubnetSet creation for `kube-system` is skipped when `vpc_wcp_enhance` is active.

### DIFF
--- a/pkg/controllers/namespace/namespace_controller.go
+++ b/pkg/controllers/namespace/namespace_controller.go
@@ -154,6 +154,10 @@ func (r *NamespaceReconciler) createDefaultSubnetSet(ctx context.Context, ns str
 	defaultSubnetSets := getDefaultSubnetsets(namespaceType)
 	ipFamily := r.NSXConfig.K8sConfig.GetIPAddressType()
 	for name, subnetSetType := range defaultSubnetSets {
+		if ns == "kube-system" && name == types.DefaultVMSubnetSet && r.NSXConfig.NsxConfig.VpcWcpEnhanceEnabled() {
+			log.Info("Skipping default SubnetSet creation for kube-system in Supervisor 2.0")
+			continue
+		}
 		if err := retry.OnError(retry.DefaultRetry, func(err error) bool {
 			return err != nil
 		}, func() error {
--- a/pkg/controllers/subnetport/subnetport_controller.go
+++ b/pkg/controllers/subnetport/subnetport_controller.go
@@ -949,6 +949,26 @@ func (r *SubnetPortReconciler) CheckAndGetSubnetPathForSubnetPort(ctx context.Co
 		existing = true
 		return
 	}
+	if reusePort, ok := subnetPort.Annotations[servicecommon.AnnotationReusePort]; ok {
+		if reusePort == "" {
+			err = fmt.Errorf("reused port cannot be empty")
+			return
+		}
+		parts := strings.Split(reusePort, "/")
+		if len(parts) == 2 && parts[0] != "" && parts[1] != "" {
+			existingPorts := r.SubnetPortService.ListSubnetPortByName(parts[0], parts[1])
+			if len(existingPorts) > 0 && existingPorts[0].ParentPath != nil && len(*existingPorts[0].ParentPath) > 0 {
+				subnetPath = *existingPorts[0].ParentPath
+				existing = true
+				log.Debug("NSX SubnetPort will be reused on the existing NSX Subnet", "subnetPort.UID", subnetPort.UID, "reusePort", reusePort, "subnetPath", subnetPath)
+				return
+			}
+			err = fmt.Errorf("reused port %s not found in runtime store", reusePort)
+			return
+		}
+		err = fmt.Errorf("invalid reuse-port annotation value %s, expected format <namespace>/<name>", reusePort)
+		return
+	}
 	if r.restoreMode {
 		// For restore case, SubnetPort will be created on the Subnet with matching CIDR
 		if subnetPort.Status.NetworkInterfaceConfig.IPAddresses[0].Gateway != "" {
--- a/pkg/nsx/services/subnetport/builder.go
+++ b/pkg/nsx/services/subnetport/builder.go
@@ -177,7 +177,10 @@ func (service *SubnetPortService) buildSubnetPort(obj interface{}, nsxSubnet *mo
 	}
 	namespaceUid := namespace.UID
 
-	nsxSubnetPortID, nsxSubnetPortName := service.BuildSubnetPortIdAndName(objMeta, namespaceUid, stsUID)
+	nsxSubnetPortID, nsxSubnetPortName, err := service.BuildSubnetPortIdAndName(objMeta, namespaceUid, stsUID)
+	if err != nil {
+		return nil, err
+	}
 	nsxSubnetPortPath := fmt.Sprintf("%s/ports/%s", *nsxSubnet.Path, nsxSubnetPortID)
 
 	tags := util.BuildBasicTags(getCluster(service), obj, namespaceUid)
@@ -270,21 +273,45 @@ func (service *SubnetPortService) GetExistingSubnetPortForStatefulSetPod(podName
 	return nil
 }
 
-func (service *SubnetPortService) BuildSubnetPortIdAndName(obj *metav1.ObjectMeta, namespaceUID types.UID, stsUID string) (string, string) {
+func (service *SubnetPortService) BuildSubnetPortIdAndName(obj *metav1.ObjectMeta, namespaceUID types.UID, stsUID string) (string, string, error) {
 	existingSubnetPort, err := service.SubnetPortStore.GetVpcSubnetPortByUID(obj.GetUID())
 	if err == nil && existingSubnetPort != nil {
-		return *existingSubnetPort.Id, *existingSubnetPort.DisplayName
+		return *existingSubnetPort.Id, *existingSubnetPort.DisplayName, nil
+	}
+
+	if reusePort, ok := obj.Annotations[common.AnnotationReusePort]; ok {
+		if reusePort == "" {
+			return "", "", fmt.Errorf("reused port cannot be empty")
+		}
+		parts := strings.Split(reusePort, "/")
+		if len(parts) == 2 && parts[0] != "" && parts[1] != "" {
+			oldNamespace := parts[0]
+			oldName := parts[1]
+			log.Info("Checking reuse-port annotation", "oldNamespace", oldNamespace, "oldName", oldName)
+			existingPorts := service.ListSubnetPortByName(oldNamespace, oldName)
+			if len(existingPorts) > 0 {
+				log.Info("Reusing existing SubnetPort based on annotation",
+					"oldNamespace", oldNamespace, "oldName", oldName, "newNamespace", obj.Namespace, "newName", obj.Name)
+				return *existingPorts[0].Id, *existingPorts[0].DisplayName, nil
+			}
+			return "", "", fmt.Errorf("reused port %s not found in runtime store", reusePort)
+		}
+		return "", "", fmt.Errorf("invalid reuse-port annotation value %s, expected format <namespace>/<name>", reusePort)
 	}
 
 	// For StatefulSet pods: check if a SubnetPort with the same StatefulSet UID and pod name exists
 	// Note: STS UID is globally unique, so we only need to check pod name
 	// Only reuse if StatefulSet pod SubnetPort feature is enabled
-	enableStsFeature := nsx.StatefulSetPodSubnetPortFeatureEnabled(service.NSXClient, service.NSXConfig)
-	if enableStsFeature {
-		if port := service.GetExistingSubnetPortForStatefulSetPod(obj.Name, stsUID); port != nil {
-			log.Info("Reusing existing SubnetPort for StatefulSet pod",
-				"podName", obj.Name, "stsUID", stsUID, "portID", *port.Id)
-			return *port.Id, *port.DisplayName
+	if stsUID != "" && nsx.StatefulSetPodSubnetPortFeatureEnabled(service.NSXClient, service.NSXConfig) {
+		existingPorts := service.SubnetPortStore.GetByIndex(common.TagScopeStatefulSetUID, stsUID)
+		for _, port := range existingPorts {
+			log.Debug("BuildSubnetPortIdAndName", "port", port.Id, "path", port.Path)
+			portName := nsxutil.FindTag(port.Tags, common.TagScopePodName)
+			if portName == obj.Name {
+				log.Info("Reusing existing SubnetPort for StatefulSet pod",
+					"podName", obj.Name, "stsUID", stsUID)
+				return *port.Id, *port.DisplayName, nil
+			}
 		}
 	}
 
@@ -295,7 +322,7 @@ func (service *SubnetPortService) BuildSubnetPortIdAndName(obj *metav1.ObjectMet
 	}
 	return common.BuildUniqueIDWithRandomUUID(objWithNamespaceUID, util.GenerateIDByObject, func(id string) bool {
 		return service.SubnetPortStore.GetByKey(id) != nil
-	}), service.BuildSubnetPortName(obj)
+	}), service.BuildSubnetPortName(obj), nil
 }
 
 func (service *SubnetPortService) BuildSubnetPortName(obj *metav1.ObjectMeta) string {
--- a/pkg/controllers/namespace/namespace_controller_test.go
+++ b/pkg/controllers/namespace/namespace_controller_test.go
@@ -366,6 +366,21 @@ func TestCreateDefaultSubnetSet(t *testing.T) {
 		setupMocks              func(r *NamespaceReconciler) *gomonkey.Patches
 	}{
 		{
+			name:               "Skip case - not create SubnetSet for kube-system when VpcWcpEnhance is true",
+			namespace:          "kube-system",
+			defaultSubnetSize:  24,
+			existingResources:  []client.Object{},
+			expectedError:      false,
+			networkStack:       v1alpha1.FullStackVPC,
+			nameSpaceType:      ctlcommon.SystemNs,
+			expectedSubnetSets: 0,
+			setupMocks: func(r *NamespaceReconciler) *gomonkey.Patches {
+				tr := true
+				r.NSXConfig.NsxConfig.VpcWcpEnhance = &tr
+				return gomonkey.NewPatches()
+			},
+		},
+		{
 			name:              "Skip case - not create SubnetSet for NormalNs",
 			namespace:         "test-ns",
 			defaultSubnetSize: 24,
--- a/pkg/controllers/subnetport/subnetport_controller_test.go
+++ b/pkg/controllers/subnetport/subnetport_controller_test.go
@@ -2152,6 +2152,83 @@ func TestSubnetPortReconciler_CheckAndGetSubnetPathForSubnetPort(t *testing.T) {
 			restoreMode: true,
 			expectedErr: "mocked getSubnetBySubnetPort error",
 		},
+		{
+			name: "ReusePortAnnotationSuccess",
+			prepareFunc: func(t *testing.T, spr *SubnetPortReconciler) *gomonkey.Patches {
+				patches := gomonkey.ApplyFunc((*subnetport.SubnetPortStore).GetVpcSubnetPortByUID,
+					func(s *subnetport.SubnetPortStore, uid types.UID) (*model.VpcSubnetPort, error) {
+						return nil, nil
+					})
+				patches.ApplyFunc((*subnetport.SubnetPortService).ListSubnetPortByName,
+					func(s *subnetport.SubnetPortService, ns string, name string) []*model.VpcSubnetPort {
+						if ns == "kube-system" && name == "old-vm" {
+							return []*model.VpcSubnetPort{
+								{
+									Id:         servicecommon.String("reused-port-id"),
+									ParentPath: servicecommon.String("subnet-path-reused"),
+								},
+							}
+						}
+						return nil
+					})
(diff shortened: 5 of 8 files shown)


pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -949,6 +949,26 @@ func (r *SubnetPortReconciler) CheckAndGetSubnetPathForSubnetPort(ctx context.Co
...
+	if reusePort, ok := subnetPort.Annotations[servicecommon.AnnotationReusePort]; ok {
+		if reusePort == "" {
+			err = fmt.Errorf("reused port cannot be empty")
+			return
+		}
+		parts := strings.Split(reusePort, "/")
+		if len(parts) == 2 && parts[0] != "" && parts[1] != "" {
+			existingPorts := r.SubnetPortService.ListSubnetPortByName(parts[0], parts[1])
+			if len(existingPorts) > 0 && existingPorts[0].ParentPath != nil && len(*existingPorts[0].ParentPath) > 0 {
+				subnetPath = *existingPorts[0].ParentPath
```
TaoZou1: existingPorts[0].ParentPath != nil && len(*existingPorts[0].ParentPath) > 0
you should check it inside {}
since in the builder.go, the reuse logic will not check the ParentPath
so if ParentPath is nil, you should also set existing=true here

pkg/controllers/namespace/namespace_controller.go
```diff
@@ -154,6 +154,10 @@ func (r *NamespaceReconciler) createDefaultSubnetSet(ctx context.Context, ns str
 	defaultSubnetSets := getDefaultSubnetsets(namespaceType)
 	ipFamily := r.NSXConfig.K8sConfig.GetIPAddressType()
 	for name, subnetSetType := range defaultSubnetSets {
+		if ns == "kube-system" && name == types.DefaultVMSubnetSet && r.NSXConfig.NsxConfig.VpcWcpEnhanceEnabled() {
+			log.Info("Skipping default SubnetSet creation for kube-system in Supervisor 2.0")
```
TaoZou1: should you also check the nsx version?
does it only apply to kube-system?
there is other system ns.

pkg/controllers/namespace/namespace_controller.go
```diff
@@ -154,6 +154,10 @@ func (r *NamespaceReconciler) createDefaultSubnetSet(ctx context.Context, ns str
 	defaultSubnetSets := getDefaultSubnetsets(namespaceType)
 	ipFamily := r.NSXConfig.K8sConfig.GetIPAddressType()
 	for name, subnetSetType := range defaultSubnetSets {
+		if ns == "kube-system" && name == types.DefaultVMSubnetSet && r.NSXConfig.NsxConfig.VpcWcpEnhanceEnabled() {
```
dantingl, line 157: There should be a util function to check if the namespace is system namespace

pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -949,6 +949,26 @@ func (r *SubnetPortReconciler) CheckAndGetSubnetPathForSubnetPort(ctx context.Co
...
+				existing = true
+				log.Debug("NSX SubnetPort will be reused on the existing NSX Subnet", "subnetPort.UID", subnetPort.UID, "reusePort", reusePort, "subnetPath", subnetPath)
```
dantingl, line 963: Update existing NSX SubnetPort with the new SubnetPort CR's namespace, namespace uid, CR name and CR uid. Old CR may be deleted after upgrade

pkg/controllers/namespace/namespace_controller.go
```diff
@@ -153,7 +153,15 @@ func getDefaultSubnetsets(namespaceType common.NameSpaceType) map[string]string
 func (r *NamespaceReconciler) createDefaultSubnetSet(ctx context.Context, ns string, nc *v1alpha1.VPCNetworkConfiguration, namespaceType common.NameSpaceType, networkStack v1alpha1.NetworkStackType) error {
 	defaultSubnetSets := getDefaultSubnetsets(namespaceType)
 	ipFamily := r.NSXConfig.K8sConfig.GetIPAddressType()
+	isSystemNs, err := util.IsVPCSystemNamespace(r.Client, ns, nil)
+	if err != nil {
+		log.Error(err, "Failed to check if namespace is system namespace", "Namespace", ns)
+	}
 	for name, subnetSetType := range defaultSubnetSets {
+		if isSystemNs && name == types.DefaultVMSubnetSet && r.NSXConfig.NsxConfig.VpcWcpEnhanceEnabled() {
```
lxiaopei: If util.IsVPCSystemNamespace returns an error , err is logged and discarded. isSystemNs defaults to false.
As a result, if isSystemNs && name == types.DefaultVMSubnetSet ... evaluates to false, and the reconciler proceeds to create the vm-default SubnetSet for the system namespace anyway, violating the intent of this skip logic.

Please return err so the reconciler can retry with exponential backoff.

pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -949,6 +949,28 @@ func (r *SubnetPortReconciler) CheckAndGetSubnetPathForSubnetPort(ctx context.Co
...
 		return
 	}
+	if reusePort, ok := subnetPort.Annotations[servicecommon.AnnotationReusePort]; ok {
+		if reusePort == "" {
+			err = fmt.Errorf("reused port cannot be empty")
+			return
+		}
+		parts := strings.Split(reusePort, "/")
+		if len(parts) == 2 && parts[0] != "" && parts[1] != "" {
+			existingPorts := r.SubnetPortService.ListSubnetPortByName(parts[0], parts[1])
```
lxiaopei: ListSubnetPortByName(parts[0], parts[1]) searches both TagScopeVMNamespace and TagScopeNamespace, matching against both TagScopeSubnetPortCRName and TagScopePodName.

If there are multiple ports matching (e.g., if a Pod and a VM shared the same name in the namespace), it might return the wrong SubnetPort.

pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -949,6 +949,28 @@ func (r *SubnetPortReconciler) CheckAndGetSubnetPathForSubnetPort(ctx context.Co
...
+			if len(existingPorts) > 0 && existingPorts[0].ParentPath != nil && len(*existingPorts[0].ParentPath) > 0 {
+				subnetPath = *existingPorts[0].ParentPath
+				existing = true
+				log.Info("NSX SubnetPort will be reused on the existing NSX Subnet, its tags and display name will be updated with the new SubnetPort CR metadata",
+					"subnetPort.Namespace", subnetPort.Namespace, "subnetPort.Name", subnetPort.Name, "subnetPort.UID", subnetPort.UID,
+					"reusePort", reusePort, "subnetPath", subnetPath, "reusedPortID", *existingPorts[0].Id)
+				return
+			}
+			err = fmt.Errorf("reused port %s not found in runtime store", reusePort)
```
lxiaopei: "runtime store" is an internal implementation detail of nsx-operator, which should not return to end user.

pkg/util/utils.go
```diff
@@ -206,11 +206,21 @@ func IsT1SystemNamespace(c client.Client, ns string, obj *v1.Namespace) (bool, e
...
-	} else if err := c.Get(context.Background(), types.NamespacedName{Namespace: ns, Name: ns}, nsObj); err != nil {
-		return false, client.IgnoreNotFound(err)
+	} else if c != nil {
+		if err := c.Get(context.Background(), types.NamespacedName{Name: ns}, nsObj); err != nil {
+			if err := c.Get(context.Background(), types.NamespacedName{Namespace: ns, Name: ns}, nsObj); err != nil {
+				return false, client.IgnoreNotFound(err)
+			}
+		}
+	}
+	if nsObj.Name == "kube-system" {
```
lxiaopei: this is not needed? since line 209 already check if ns is kube-system.

pkg/util/utils.go
```diff
@@ -206,11 +206,21 @@ func IsT1SystemNamespace(c client.Client, ns string, obj *v1.Namespace) (bool, e
...
+		return true, nil
+	}
 	nsObj := &v1.Namespace{}
 	if obj != nil {
 		nsObj = obj
-	} else if err := c.Get(context.Background(), types.NamespacedName{Namespace: ns, Name: ns}, nsObj); err != nil {
-		return false, client.IgnoreNotFound(err)
+	} else if c != nil {
+		if err := c.Get(context.Background(), types.NamespacedName{Name: ns}, nsObj); err != nil {
+			if err := c.Get(context.Background(), types.NamespacedName{Namespace: ns, Name: ns}, nsObj); err != nil {
```
lxiaopei: consider accepting ctx context.Context as a parameter, instead of use context.Background()

lxiaopei: why here Get Namespace again?

pkg/nsx/services/subnetport/builder.go
```diff
@@ -248,21 +251,45 @@ func (service *SubnetPortService) GetExistingSubnetPortForStatefulSetPod(podName
...
+		parts := strings.Split(reusePort, "/")
+		if len(parts) == 2 && parts[0] != "" && parts[1] != "" {
+			oldNamespace := parts[0]
+			oldName := parts[1]
+			log.Info("Checking reuse-port annotation", "oldNamespace", oldNamespace, "oldName", oldName)
+			existingPorts := service.ListSubnetPortByName(oldNamespace, oldName)
+			if len(existingPorts) > 0 {
+				log.Info("Reusing existing SubnetPort based on annotation",
+					"oldNamespace", oldNamespace, "oldName", oldName, "newNamespace", obj.Namespace, "newName", obj.Name)
+				return *existingPorts[0].Id, service.BuildSubnetPortName(obj), nil
```
lxiaopei: ensure existingPorts[0].Id is not nil

pkg/controllers/namespace/namespace_controller.go
dantingl: I do not think you should skip default SubnetSet creation for system namespace. We still need it for greenfield case

pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -949,6 +949,28 @@ func (r *SubnetPortReconciler) CheckAndGetSubnetPathForSubnetPort(ctx context.Co
 		existing = true
 		return
 	}
+	if reusePort, ok := subnetPort.Annotations[servicecommon.AnnotationReusePort]; ok {
```
dantingl, line 1074: For the original CR, we may also need to check if its port got reused, no need to create new port if there is an event to reconcile the original CR

lxiaopei:
which component will set vpc_wcp_enhance = True in NCP configmap?

pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -502,6 +508,104 @@ func (r *SubnetPortReconciler) deleteSubnetPortByName(ctx context.Context, ns st
...
+	if subnetPort == nil || subnetPort.Status.Attachment.ID == "" {
+		return false
+	}
+	if !util.IsCPVM(subnetPort.Labels) {
+		return false
+	}
+	if r.SubnetPortService == nil || r.SubnetPortService.SubnetPortStore == nil {
+		return false
+	}
+	port := r.SubnetPortService.GetSubnetPortByAttachmentID(subnetPort.Status.Attachment.ID)
```
dantingl, line 523: Why filtering by attachment id, instead of CR uid?

pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -502,6 +508,104 @@ func (r *SubnetPortReconciler) deleteSubnetPortByName(ctx context.Context, ns st
...
+	}
+	portCRUID := nsxutil.FindTag(port.Tags, servicecommon.TagScopeSubnetPortCRUID)
+	if portCRUID != "" && portCRUID != string(subnetPort.UID) {
+		return true
+	}
+	portCRName := nsxutil.FindTag(port.Tags, servicecommon.TagScopeSubnetPortCRName)
+	portNS := nsxutil.FindTag(port.Tags, servicecommon.TagScopeVMNamespace)
+	if portNS == "" {
+		portNS = nsxutil.FindTag(port.Tags, servicecommon.TagScopeNamespace)
+	}
```
dantingl, line 535: Why we need to check CR name and namespace name?

dantingl (commented):
Please also make sure there is no regression for nsx-operator restore logic

dantingl:
> Skip vm-default SubnetSet creation for kube-system in Supervisor 2.0

Do we still have this logic

pkg/controllers/subnetport/subnetport_controller.go
lxiaopei, line 94: it might have a race here.
If the old `kube-system/old-vm` CR was deleted first during a migration, etcd removes it instantly.
  Now imagine the new CR `new-ns/new-vm` with `reuse-port: "kube-system/old-vm"` was just submitted. If the controller reconciles the old CR's deletion *before* the new CR finishes patching the NSX port tags.

pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -733,6 +837,14 @@ func (r *SubnetPortReconciler) CollectGarbage(ctx context.Context) error {
 	var errList []error
 	diffSet := nsxSubnetPortSet.Difference(crSubnetPortIDsSet)
 	for elem := range diffSet {
+		port := r.SubnetPortService.GetSubnetPortByID(elem)
+		if port != nil && r.isCPVMPort(ctx, port) {
```
lxiaopei, line 841: How about upgrade case where SubnetPort was created by the *old* operator version, which **never** added the `LabelCPVM` (`iaas.vmware.com/is-cpvm-subnetport`) tag to NSX, then the old CR in `kube-system` gets deleted during migration. Since the port has no LabelCPVM tag, Garbage collector GC the SubnetPort.

pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -733,6 +837,14 @@ func (r *SubnetPortReconciler) CollectGarbage(ctx context.Context) error {
...
+			if r.isPortReferencedByReusePort(ctx, port) {
```
lxiaopei, line 842: For a batch of 10 stale ports is being collected, `isPortReferencedByReusePort` invokes `r.Client.List(ctx, subnetPortList)` **10 times in a row**, could we List the active SubnetPort CRs  outside the loop in `CollectGarbage`?

pkg/controllers/subnetport/subnetport_controller.go
```diff
@@ -930,6 +1042,16 @@ func (r *SubnetPortReconciler) getSubnetBySubnetPort(subnetPort *v1alpha1.Subnet
 }
 
 func (r *SubnetPortReconciler) CheckAndGetSubnetPathForSubnetPort(ctx context.Context, subnetPort *v1alpha1.SubnetPort) (existing bool, isStale bool, subnetPath string, subnetSetUID *types.UID, subnetSetLock *sync.RWMutex, interfaceType v1alpha1.IPAddressType, staticIPAllocationType v1alpha1.StaticIPAllocationType, err error) {
+	if r.isPortReused(subnetPort) {
```
lxiaopei, line 1045: CheckAndGetSubnetPathForSubnetPort is only invoked when r.isPortReused(subnetPort) is false.
Do not need check again here.

pkg/nsx/services/subnetport/builder.go
```diff
@@ -248,21 +251,58 @@ func (service *SubnetPortService) GetExistingSubnetPortForStatefulSetPod(podName
...
+			log.Info("SubnetPort has reuse-port annotation but is not a cpVM SubnetPort, ignoring reuse-port",
+				"namespace", obj.Namespace, "name", obj.Name)
+		} else {
+			if reusePort == "" {
+				return "", "", fmt.Errorf("reused port cannot be empty")
+			}
+			parts := strings.Split(reusePort, "/")
+			if len(parts) == 2 {
+				oldNamespace := strings.TrimSpace(parts[0])
+				oldName := strings.TrimSpace(parts[1])
```
lxiaopei, line 271: could we put the parse for reusePort in a func? 
the same logic are found in builder.go, subnetport.go, subnetport_controller.go

pkg/nsx/services/subnetport/builder.go
```diff
@@ -248,21 +251,58 @@ func (service *SubnetPortService) GetExistingSubnetPortForStatefulSetPod(podName
...
+				if oldNamespace != "" && oldName != "" {
+					log.Info("Checking reuse-port annotation", "oldNamespace", oldNamespace, "oldName", oldName)
+					existingPorts := service.ListVMSubnetPortByName(oldNamespace, oldName)
+					if len(existingPorts) == 0 {
+						existingPorts = service.ListSubnetPortByName(oldNamespace, oldName)
```
lxiaopei, line 276: I do not get it, ListVMSubnetPortByName will only return SubnetPort for VMs,  ListSubnetPortByName will also return SubnetPort for Pods right？ do we need to use ListSubnetPortByName?

pkg/controllers/subnetport/subnetport_controller.go
seanpang-vmware: this feature is for upgrade scenario, and based on the discussion, even after upgrade, the old CRs will not be deleted, and always there.

pkg/controllers/subnetport/subnetport_controller.go
seanpang-vmware: When a cpVM port is reused by the new SubnetPort CR, the port's tags are updated such that TagScopeSubnetPortCRUID points to the new CR's UID. The old CR (which is reconciling here) only has its attachment ID in its status. If we queried by subnetPort.UID (the old CR's UID), the lookup in the store would return nil because the port is now indexed under the new CR's UID. By looking up by Attachment ID, we find the port and can then compare portCRUID != string(subnetPort.UID) to detect that the port has been taken over by another CR.

pkg/controllers/subnetport/subnetport_controller.go
seanpang-vmware: Streamlined isPortReused so that portCRUID is the primary check (if portCRUID != "" { return portCRUID != string(subnetPort.UID) }). Name and namespace checks are now only a defensive fallback when portCRUID tag is missing.

pkg/controllers/subnetport/subnetport_controller.go
seanpang-vmware: Good catch. In an upgrade scenario, older operator versions did not set LabelCPVM on the NSX port, and the old CR in kube-system gets deleted during migration. To address this, we removed isCPVMPort and now directly check r.isPortReferencedByReusePort(ctx, port) during GC. Since isPortReferencedByReusePort checks all active cpVM SubnetPort CRs in the cluster for matching reuse-port annotations or attachment IDs, the port is reliably preserved regardless of missing tags or deleted old CRs. Added unit test TestSubnetPortReconciler_CollectGarbage_PreservesReusedPort to verify.

pkg/controllers/subnetport/subnetport_controller.go
seanpang-vmware: "Streamlined isPortReused so that portCRUID is the primary check (if portCRUID != "" { return portCRUID != string(subnetPort.UID) }). Name and namespace checks are now only a defensive fallback when portCRUID tag is missing."

PR #1561, opened 2026-09-08
Updating static route CR with ipallocation cause nil pointer error

seanpang-vmware:
During a StaticRoute CR update, if the old CR uses the network field and the new CR uses the ipallocation field, comparing these values via string pointers will cause a nil pointer dereference error. Add protection code to ensure that nil values are handled correctly.

| # | Test Case Description | Scenario / Action | Expected Behavior | Actual Result | Status |
|:---:|---|---|---|---|:---:|
| 1 | Create StaticRoute with static CIDR | `spec.network: 10.100.1.0/24`, `nextHops: [192.168.0.1]` | Route realized on NSX, CR condition `Ready=True` | NSX route created, `Ready=True` | PASS |
| 2 | Add next-hop & Reorder NextHops | Add `192.168.0.2`, then reorder to `[192.168.0.2, 192.168.0.1]` | Reordered hops recognized as equal (order-insensitive), no unnecessary update | Order-insensitive comparison succeeded, `Ready=True` | PASS |
| 3 | Core Fix 1: Switch `network` to `networkIpAllocationName` | Change `spec.network` to `spec.networkIpAllocationName: test-ipalloc-1` | Old has `Network`, new has `nil` `Network`. Safe comparison, no nil dereference panic | Handled `nil` safely, NSX patched with allocation path, `Ready=True` | PASS |
| 4 | Core Fix 2: Switch between `networkIpAllocationName` allocations | Change `networkIpAllocationName` from `test-ipalloc-1` to `test-ipalloc-2` | Both have `nil` `Network`. Safe comparison detects path change without panic | Handled `nil` safely, NSX updated to `test-ipalloc-2`, `Ready=True` | PASS |
| 5 | Core Fix 3: Switch `networkIpAllocationName` back to `network` | Change `networkIpAllocationName: test-ipalloc-2` back to `spec.network: 10.200.2.0/24` | Old has `nil` `Network`, new has `Network`. Safe comparison without panic | Handled `nil` safely, NSX updated to new CIDR, `Ready=True` | PASS |
| 6 | Greenfield StaticRoute with `networkIpAllocationName` | Create new route `test-sr-2` directly with `spec.networkIpAllocationName: test-ipalloc-1` | Path resolved from allocation CR, NSX route created, `Ready=True` | Successfully created, `Ready=True` | PASS |
| 7 | NextHop mutation on IP allocation route | Add hop & reorder hops on `test-sr-2` using `networkIpAllocationName` | Safe comparison with `Network=nil`, order-insensitive hops | Route updated and reordered cleanly, `Ready=True` | PASS |
| 8 | Clean resource deletion & operator stability | Delete all test `StaticRoute` and `IPAddressAllocation` CRs | Resources cleanly deleted from NSX and K8s, 0 operator crashes/restarts | CRs and NSX objects deleted, 0 restarts, 0 panics | PASS |

### DIFF
--- a/pkg/nsx/services/staticroute/compare.go
+++ b/pkg/nsx/services/staticroute/compare.go
@@ -1,28 +1,75 @@
 package staticroute
 
 import (
+	"sort"
+
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
-	"k8s.io/apimachinery/pkg/util/sets"
+
+	"github.com/vmware-tanzu/nsx-operator/pkg/nsx/services/common"
 )
 
-// assume that staticroute doesn't have the same ipaddress, return true if equal
-func (service *StaticRouteService) compareStaticRoute(oldStaticRoute *model.StaticRoutes, newStaticRoute *model.StaticRoutes) bool {
-	if *oldStaticRoute.Network != *newStaticRoute.Network {
-		return false
+type (
+	StaticRoute model.StaticRoutes
+)
+
+type Comparable = common.Comparable
+
+func (sr *StaticRoute) Key() string {
+	if sr != nil && sr.Id != nil {
+		return *sr.Id
 	}
-	oldNextHops := oldStaticRoute.NextHops
-	newNextHops := newStaticRoute.NextHops
-	if len(oldNextHops) != len(newNextHops) {
-		return false
+	return ""
+}
+
+func (sr *StaticRoute) Value() data.DataValue {
+	if sr == nil {
+		return nil
 	}
-	oldHopsSet := sets.NewString()
-	for _, addr := range oldNextHops {
-		oldHopsSet.Insert(*addr.IpAddress)
-	}
-	for _, addr := range newNextHops {
-		if !oldHopsSet.Has(*addr.IpAddress) {
-			return false
+	var nextHops []model.RouterNexthop
+	if len(sr.NextHops) > 0 {
+		nextHops = make([]model.RouterNexthop, len(sr.NextHops))
+		for i, nh := range sr.NextHops {
+			nextHops[i] = model.RouterNexthop{
+				IpAddress: nh.IpAddress,
+			}
 		}
+		sort.Slice(nextHops, func(i, j int) bool {
+			if nextHops[i].IpAddress == nil {
+				return true
+			}
+			if nextHops[j].IpAddress == nil {
+				return false
+			}
+			return *nextHops[i].IpAddress < *nextHops[j].IpAddress
+		})
 	}
-	return true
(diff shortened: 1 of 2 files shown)


pkg/nsx/services/staticroute/compare.go
dantingl: addr.IpAddress may be nil?

pkg/nsx/services/staticroute/compare.go
```diff
@@ -5,9 +5,22 @@ import (
...
+	if a == nil || b == nil {
+		return false
+	}
+	return *a == *b
+}
+
 // assume that staticroute doesn't have the same ipaddress, return true if equal
 func (service *StaticRouteService) compareStaticRoute(oldStaticRoute *model.StaticRoutes, newStaticRoute *model.StaticRoutes) bool {
-	if *oldStaticRoute.Network != *newStaticRoute.Network {
+	if !compareStringPtr(oldStaticRoute.Network, newStaticRoute.Network) {
```
dantingl: Check if newStaticRoute and oldStaticRoute are nil

dantingl:
Commit message: ipallcation -> ipallocation

pkg/nsx/services/staticroute/compare.go
seanpang-vmware: its value could not be nil, if user add nexthop in CR, they could not fill an empty value in ipaddress. the nil error here is because user switch from using network to ipallocation, thus when comparing two CRs, the default value is nil when switching to use the other.

pkg/nsx/services/staticroute/compare.go
seanpang-vmware: the method is for comparing static route CR changes when receive update event, the two CR won't be empty.

below is the invoker's logic, it already handles the situation that the CRs could not be built or could not be found in store.
```
        nsxStaticRoute, err := service.buildStaticRoute(obj, networkIPAllocationPath)
	    if err != nil {
		return err
	}

	existingStaticRoute := service.StaticRouteStore.GetStaticRoutesByCRUID(obj.GetUID())
	if existingStaticRoute != nil {
		// Update the generated NSX static route's id and display_name with the existing configurations.
		nsxStaticRoute.Id = String(*existingStaticRoute.Id)
		nsxStaticRoute.DisplayName = String(*existingStaticRoute.DisplayName)
		if service.compareStaticRoute(existingStaticRoute, nsxStaticRoute) {
```

yanjunz97 (dismissed):
I just notice this staticroute follows a different pattern against other resources to compare the nsx resource.

I think the fix itself shall work. It depends on you if a refactor is preferred.

Ref: https://github.com/vmware-tanzu/nsx-operator/blob/main/pkg/nsx/services/ipaddressallocation/compare.go

yanjunz97 (approved):

TaoZou1 (approved):

PR #1573, opened 2026-09-20
Enhance error message for Edge insufficient capacity case

TaoZou1:
Enhance error message for Edge insufficient capacity case when creating NSX LB resources. Classify LBS capacity and form factor constraint errors using NSX error codes (502102, 502103, 502104, 502105, 10087, 10145, 10149, 23709) and extract both top-level and related error codes from NSX realization alarms. Expose clear, actionable messages on the NamespaceNetworkReady condition.

Test Done:
1. Unit tests:
   - Added and passed unit tests in errors_test.go covering ExtractAllErrorCodes, IsLBSEdgeCapacityError, and GetLBSEdgeCapacityErrorCode.
   - Added and passed unit tests in networkinfo_utils_test.go covering TestGetVPCCreationFailureCondition.
   - Verified integration unit tests in networkinfo_controller_test.go.
2. Live testbed verification :
   - Case 1 (Form Factor constraint / PATCH phase rejection): Set service_size = XLARGE on MEDIUM Edge Node. Observed synchronous HTTP 400 error code 502105 during PATCH / CreateOrUpdateVPC. NamespaceNetworkReady condition displayed: 'Load Balancer Service (size: XLARGE) creation failed: Edge node form factor does not support Load Balancer Service size XLARGE (NSX error code: 502105). Please verify Edge Node sizing in NSX.'
   - Case 2 (Edge Cluster capacity constraint / Realization phase alarm): Set service_size = LARGE on Edge Cluster with insufficient resources. Observed asynchronous realization error with related error code 10087. NamespaceNetworkReady condition displayed: 'Load Balancer Service (size: LARGE) creation failed: Insufficient Edge Cluster capacity on NSX (error code: 10087). Please expand Edge Cluster resources or verify Edge Cluster sizing in NSX.'


pkg/nsx/util/errors.go
```diff
@@ -682,3 +708,147 @@ func IsIPAllocationError(alarm model.PolicyAlarmResource) bool {
...
+	}
+
+	// 3. Structured NsxError / GeneralNsxError
+	if nsxErr, ok := err.(NsxError); ok {
+		if impl, ok := nsxErr.(*GeneralNsxError); ok {
+			addCode(int64(impl.ErrorCode))
+			for _, c := range impl.RelatedErrorCodes {
+				addCode(int64(c))
+			}
+		}
```
heypnus: 🟠 **High**: ExtractAllErrorCodes documents support for wrapped errors, but uses direct type assertions instead of errors.As. Any error wrapped with fmt.Errorf("%w", ...) will bypass the structured error checks; because RealizeStateError.Error() does not output error codes, wrapped realization errors cannot be detected by the fallback regex either. Furthermore, asserting a typed nil pointer causes a nil dereference panic.

TaoZou1: fixed

pkg/nsx/services/realizestate/realize_state.go
```diff
@@ -57,25 +57,34 @@ func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, inte
 			if *result.State == model.GenericPolicyRealizedResource_STATE_ERROR {
 				log.Error(nil, "Found realized state with error", "result", result)
 				var errMsg []string
+				errorCode := 0
```
dantingl, line 74: set errorCode=0 inside the loop?

pkg/nsx/util/errors.go
dantingl: this function is not needed?

pkg/controllers/networkinfo/networkinfo_utils.go
```diff
@@ -377,3 +379,44 @@ func hasVMDefaultSubnets(subnets []v1alpha1.SharedSubnet) bool {
...
+}
+
+// getVPCCreationFailureCondition returns an enhanced NamespaceCondition when VPC or LBS creation fails.
+// If the error is related to Load Balancer Service, it formats a user-friendly and actionable error message
+// instead of exposing the raw NSX error.
+func getVPCCreationFailureCondition(err error, lbsSize string) *v1.NamespaceCondition {
+	if err == nil {
+		return nsMsgVPCIsReady.getNSNetworkCondition()
+	}
+	if msg := enhanceVPCLBSErrorMessage(err, lbsSize); msg != "" {
```
dantingl, line 413: For AVI case, do you know whether NSX could return capacity error code for VPC and SNAT creation?

pkg/nsx/services/realizestate/realize_state.go
TaoZou1: Setting errorCode := 0 here is scoped per error result. We keep it outside the inner for _, alarm := range result.Alarms loop so that the first top-level ErrorCode is captured and preserved (not overwritten/reset across multiple alarms of this result), while any other codes are appended to relatedCodes. Also, since the block returns NewRealizeStateError immediately after the alarms loop, it cleanly exits upon the first failed result.

pkg/nsx/util/errors.go
TaoZou1: Yes, good catch.

pkg/controllers/networkinfo/networkinfo_utils.go
TaoZou1: No Edge LBS Capacity Errors: Since lbProvider == AVILB, createdLBS is nil and no NSX Edge Load Balancer Service is provisioned on NSX. Therefore, NSX will not return any of the LBSEdgeCapacityErrorCodes (e.g. 502102–502105, 10087, 10145, 10149).
VPC / SNAT Capacity Errors: For VPC creation with default SNAT in the AVI case, the bottleneck is external IP capacity rather than Edge sizing. If the external IP block configured on the VPC Connectivity Profile is exhausted, NSX returns IPAM error codes such as 520100 (IpAddressBlocks do not have spare capacity...) or 520015/520054.
Fallback Handling: Since enhanceVPCLBSErrorMessage specifically filters for Edge LBS capacity codes, non-LBS errors (like IPAM exhaustion 520100) return "" and naturally fall back to nsMsgVPCCreateUpdateError, exposing the raw NSX error on the Namespace condition as expected."

pkg/nsx/services/realizestate/realize_state.go
```diff
@@ -57,25 +57,34 @@ func (service *RealizeStateService) CheckRealizeState(backoff wait.Backoff, inte
...
+								relatedCodes = append(relatedCodes, int(*relatedErr.ErrorCode))
 							}
 						}
 					}
 					if nsxutil.IsRetryRealizeError(alarm) {
 						return nsxutil.NewRetryRealizeError(fmt.Sprintf("%s not realized with errors: %s", intentPath, errMsg))
 					}
-					if nsxutil.IsIPAllocationError(alarm) {
-						return nsxutil.NewRealizeStateError(fmt.Sprintf("%s realized with errors: %s", intentPath, errMsg), int(*alarm.ErrorDetails.ErrorCode))
-					}
```
yanjunz97, line 76: Would you know if there is a case like this?
There are many alarms, and IPAllocationError is not the first one. As errorCode only capture the first alarm errorCode, then the IPAllocationError will miss.

yanjunz97 (commented):
Overall LGTM, just a nit

pkg/nsx/services/realizestate/realize_state.go
TaoZou1: Good catch.
I'm not sure if there is such a case.
I notices that the IsIPAllocationError checked all the alarm
even In the vmware-nsxlib it just checkes the first alarm.
I think it's better to check all the error in all the alarms. 
I have updated the code

pkg/nsx/util/errors.go
```diff
@@ -674,11 +698,118 @@ func IsRetryRealizeError(alarm model.PolicyAlarmResource) bool {
...
+		if code > 0 && !seen[code] {
+			seen[code] = true
+			codes = append(codes, code)
+		}
+	}
+
+	// 1. Structured NSXApiError (supports Go 1.13+ error wrapping via errors.As)
+	var apiErr *NSXApiError
+	if errors.As(err, &apiErr) && apiErr != nil {
+		if apiErr.ErrorCode != nil {
```
dantingl, line 739: check apiErr.ApiError? 
apiErr.ErrorCode is apiErr.ApiError.ErrorCode, right?

dantingl (commented):
10087,10145, 10149 are error codes for Edge, do we need to check LB provider

pkg/nsx/util/errors.go
TaoZou1: Good catch! Since *model.ApiError is an embedded pointer in NSXApiError, apiErr.ErrorCode is indeed the promoted field for apiErr.ApiError.ErrorCode. If apiErr.ApiError is nil (e.g. NewNSXApiError(nil, ...)), accessing apiErr.ErrorCode would cause a nil-pointer dereference panic.

Added && apiErr.ApiError != nil guard and added a unit test covering NSXApiError with nil ApiError

dantingl (approved):

PR #1574, opened 2026-09-20
securitypolicy: enforce peer exclusivity for ingress and egress rules via CEL

heypnus:
Add CEL validation rules to SecurityPolicyRule to disallow specifying
'to' or 'destinations' for Ingress rules and 'from' or 'sources' for
Egress rules. This prevents invalid cross-direction configurations from
being admitted and silently dropped by the controller.

The CEL rules use 'self.direction.lowerAscii()' to match 'in'/'ingress'
and 'out'/'egress' in a case-insensitive manner, aligning with the
controller's internal behavior (which converts directions using
'util.ToUpper()').

To prevent Kubernetes API Server's CEL cost estimator from failing
CRD admission with 'estimated rule cost exceeds budget by factor of
more than 100x' (which happens when string functions like lowerAscii()
are evaluated on unbounded strings inside arrays), 'MaxLength=7' is
added to the 'RuleDirection' type definition (since 'Ingress' is the
longest valid value at 7 characters) so that the string length is
statically bounded in the OpenAPI v3 schema.

Testing Done:
1. before the fix, all invalid security policies (T1 egress + from, T1 ingress + to, VPC egress + from, VPC ingress + to) can be created successfully.
2. after the fix, all invalid cases can be denied:
```
root@422e9bca64b679ddbbcc7eb89c66eb28 [ ~/cel-security-policy ]# k apply -f new2_t1_crd
customresourcedefinition.apiextensions.k8s.io/securitypolicies.nsx.vmware.com configured

root@422e9bca64b679ddbbcc7eb89c66eb28 [ ~/cel-security-policy ]# k delete -f t1-egress-from.yaml
securitypolicy.nsx.vmware.com "test-legacy-egress-from" deleted from test-t1-ns namespace
root@422e9bca64b679ddbbcc7eb89c66eb28 [ ~/cel-security-policy ]# k apply -f t1-egress-from.yaml
The SecurityPolicy "test-legacy-egress-from" is invalid: spec.rules[0]: Invalid value: "object": from or sources cannot be set for egress rules

root@422e9bca64b679ddbbcc7eb89c66eb28 [ ~/cel-security-policy ]# k delete -f t1-ingress-to.yaml
securitypolicy.nsx.vmware.com "test-legacy-ingress-to" deleted from test-t1-ns namespace
root@422e9bca64b679ddbbcc7eb89c66eb28 [ ~/cel-security-policy ]# k apply -f t1-ingress-to.yaml
The SecurityPolicy "test-legacy-ingress-to" is invalid: spec.rules[0]: Invalid value: "object": to or destinations cannot be set for ingress rules

root@422e9bca64b679ddbbcc7eb89c66eb28 [ ~/cel-security-policy ]# k apply -f new2_vpc_crd
customresourcedefinition.apiextensions.k8s.io/securitypolicies.crd.nsx.vmware.com configured

root@422e9bca64b679ddbbcc7eb89c66eb28 [ ~/cel-security-policy ]# k delete -f vpc-egress-from.yaml
securitypolicy.crd.nsx.vmware.com "test-secpol-egress-from" deleted from test-vpc-ns namespace
root@422e9bca64b679ddbbcc7eb89c66eb28 [ ~/cel-security-policy ]# k apply -f vpc-egress-from.yaml
The SecurityPolicy "test-secpol-egress-from" is invalid: spec.rules[0]: Invalid value: "object": from or sources cannot be set for egress rules

root@422e9bca64b679ddbbcc7eb89c66eb28 [ ~/cel-security-policy ]# k delete -f vpc-ingress-to.yaml
securitypolicy.crd.nsx.vmware.com "test-secpol-ingress-to" deleted from test-vpc-ns namespace
root@422e9bca64b679ddbbcc7eb89c66eb28 [ ~/cel-security-policy ]# k apply -f vpc-ingress-to.yaml
The SecurityPolicy "test-secpol-ingress-to" is invalid: spec.rules[0]: Invalid value: "object": to or destinations cannot be set for ingress rules
```


timdengyun (dismissed):
LGTM
One question: I recalled that there is no logic to handle validation in such two cases?
If there are existing code to do that, it's good to remove once we have CEL enforcement check.

timdengyun (dismissed):

heypnus:
> LGTM One question: I recalled that there is no logic to handle validation in such two cases? If there are existing code to do that, it's good to remove once we have CEL enforcement check.

We don't have validation in the code. If an ingress rule has `to`, the nsx-op will drop the `to` field silently...

timdengyun (approved):
LGTM, Thanks!

wenqiq (approved):

PR #1577, opened 2026-09-21
fix: prevent slice out-of-bounds panics in GetIPPrefix and NSXCheckVersion

dantingl:
- pkg/util/ip.go: Check slice length before index access in GetIPPrefix to prevent panic on address strings missing CIDR prefix.
- pkg/nsx/cluster.go: Ensure sections slice has minimum length of 3 in featureSupported to avoid panic on product versions with fewer than 3 segments.


pkg/nsx/cluster.go
```diff
@@ -583,7 +583,11 @@ func (nsxVersion *NsxVersion) featureSupported(feature int) bool {
 		// ProductVersion should have at least three sections
 		// each section only have digital value
 		buff := strings.Split(nsxVersion.ProductVersion, ".")
-		sections := make([]int64, len(buff))
+		length := len(buff)
+		if length < 3 {
+			length = 3
+		}
```
TaoZou1, line 589: should we return error/false instead of enlarge the length?

yanjunz97 (dismissed):
LGTM

pkg/nsx/cluster.go
yanjunz97: I feel this determines by if we want to support considering 9.1 as 9.1.0

dantingl: I prefer to add missing segments with 0, the nsxVersion.ProductVersion is just string type without strict format

TaoZou1 (approved):

yanjunz97 (approved):

